PR #21, opened 2020-01-31
Initial support for Nuphar execution provider

KeDengMS:
Steps to try it out:
1. Build onnxruntime with Nuphar:
```
python ./build_perf_tuning.py --onnxruntime_home /path/to/ort --use_nuphar
```
You may specify --llvm_path /path/to/llvm-build/lib/cmake/llvm if llvm is not found. Linux prebuilt could be downloaded [here](http://releases.llvm.org/download.html)

2. build docker image with onnxruntime binaries:
```
cd /OLive_root/docker-images/perf-tuning
docker build -t perf_tuning -f Dockerfile .
```

3. run symbolic shape inference on input model using the image (mounting data which holds the models):
```
 docker run -v ~/data:/data --entrypoint python3.7 -t perf_tuning /perf_tuning/bin/RelWithDebInfo/all_eps/symbolic_shape_infer.py --input /data/path/to/model --output /data/path/to/shape_inferred_model --auto_merge
```

4. run perf_tuning with nuphar:
```
docker run -v ~/data:/data -t perf_tuning -e nuphar --model /data/path/to/shape_inferred_model  --result /data
```

Note that other scripts for model-editing/quantization/AOT are not included yet in this initial step.

### DIFF
--- a/docker-images/perf-tuning/Dockerfile
+++ b/docker-images/perf-tuning/Dockerfile
@@ -3,7 +3,7 @@
 FROM nvidia/cuda:10.0-base
 
 RUN apt-get update && apt-get install -y python3.7 python3.7-dev libgomp1 cuda-cublas-10-0 python3-pip 
-RUN python3.7 -m pip install psutil gputil numpy onnx packaging
+RUN python3.7 -m pip install psutil gputil numpy onnx packaging sympy
 
 RUN mkdir /perf_tuning
 COPY bin /perf_tuning/bin
--- a/docker-images/perf-tuning/Dockerfile
+++ b/docker-images/perf-tuning/Dockerfile
@@ -3,7 +3,7 @@
 FROM nvidia/cuda:10.0-base
 
 RUN apt-get update && apt-get install -y python3.7 python3.7-dev libgomp1 cuda-cublas-10-0 python3-pip 
-RUN python3.7 -m pip install psutil gputil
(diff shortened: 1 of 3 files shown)


docker-images/perf-tuning/Dockerfile
```diff
@@ -3,7 +3,7 @@
 FROM nvidia/cuda:10.0-base
 
 RUN apt-get update && apt-get install -y python3.7 python3.7-dev libgomp1 cuda-cublas-10-0 python3-pip 
-RUN python3.7 -m pip install psutil gputil
+RUN python3.7 -m pip install psutil gputil numpy onnx packaging sympy
```
liuziyue: Why are these modules needed? Are they for symbolic_shape_infer.py?

KeDengMS: Yes this is to simplify the symbolic shape inference to use the same docker, since Nuphar is not in pypi package yet.
```
docker run -v ~/data:/data --entrypoint python3.7 -t perf_tuning /perf_tuning/bin/RelWithDebInfo/all_eps/symbolic_shape_infer.py --input /data/path/to/model --output /data/path/to/shape_inferred_model --auto_merge
```

---
In reply to: [374969815](https://github.com/microsoft/OLive/pull/21#discussion_r374969815) [](ancestors = 374969815)

PR #62, opened 2022-01-14
updated classifiers

anukaal:
added support for python 3.8, 3.9, 3.10


leqiao-1:
Hi @anukaal. Thanks for your PR. 
Now the OLive python package version is related to the customized ONNXRuntime python packages version. And now we only support versions 3.6 - 3.8.

PR #68, opened 2022-02-01
Fixed exception when olive is invoked without parameters

harishsk:
If olive is invoked without any parameters it throws an exception. This PR fixes that.
I also updated .gitignore to include files to ignore during local development.


leqiao-1:
Thanks for your PR.

PR #83, opened 2022-03-29
use `flask.safe_join` instead of `os.path.join`

porcupineyhairs:
I had submitted a security disclosure earlier via huntr.io . Since that has remained pending for quite some time now, I am sending a PR with the proposed fix.

There is a path traversal bug in this project due to the use of `os.path.join` call with untrusted input. This can be fixed by using `flask.safe_join` instead.

### DIFF
--- a/olive/web/backend/app.py
+++ b/olive/web/backend/app.py
@@ -1,7 +1,7 @@
 import os
 import sys
 
-from flask import Flask, jsonify, request, send_file, url_for, send_from_directory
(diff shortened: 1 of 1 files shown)


leqiao-1:
Thanks for your contribution.

JamieSlome:
@leqiao-1 - would it be possible to confirm that this has addressed the following report:

https://www.huntr.dev/bounties/303b6e02-b51f-4f0b-96cf-f7f81da10fff

It is currently private and only accessible to maintainers with repository write permissions 👍

PR #129, opened 2023-03-24
add package publish pipeline

leqiao-1:
add package publish pipeline


.azure_pipelines/package_publish.yaml
```diff
@@ -0,0 +1,20 @@
+trigger: none
```
trajepl, line 1: Emm, should we limited the package publish only to main branch?

trajepl (approved):

jambayk (approved):

PR #131, opened 2023-03-27
Reenable Azure ml test

xiaoyu-work:
Reenable Azure ml test for pipeline.

### DIFF
--- a/examples/bert_ptq_cpu/bert_config.json
+++ b/examples/bert_ptq_cpu/bert_config.json
@@ -21,6 +21,16 @@
                     "conda_file_path": "conda.yaml"
                 }
             }
+        },
+        "docker_system": {
+            "type": "Docker",
+            "config": {
+                "local_docker_config": {
+                    "image_name": "Olive-image",
+                    "build_context_path": "docker",
+                    "dockerfile": "Dockerfile"
+                }
+            }
         }
     },
     "evaluators": {
--- a/examples/bert_ptq_cpu/bert_config.json
+++ b/examples/bert_ptq_cpu/bert_config.json
@@ -26,7 +26,7 @@
             "type": "Docker",
             "config": {
                 "local_docker_config": {
-                    "image_name": "Olive-image",
+                    "image_name": "olive-image",
                     "build_context_path": "docker",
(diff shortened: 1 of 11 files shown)


jambayk:
Since we removed the `bert.py` and `resnet.py` python scripts from the examples, we are not dynamically creating the workspace config jsons before running the tests. I think we need to do this in the test scripts now like in https://github.com/microsoft/Olive/blob/43a2586b24067046ef857213f332b03f47941ea8/test/integ_test/aml_model_test/test_aml_model.py#L53?

or else the tests will fail with workspace config not found.

xiaoyu-work:
> Since we removed the `bert.py` and `resnet.py` python scripts from the examples, we are not dynamically creating the workspace config jsons before running the tests. I think we need to do this in the test scripts now like in
> 
> https://github.com/microsoft/Olive/blob/43a2586b24067046ef857213f332b03f47941ea8/test/integ_test/aml_model_test/test_aml_model.py#L53
> 
> ?
> or else the tests will fail with workspace config not found.

Yeah I just noticed we removed config file here. I'll add this part logic.

jambayk (dismissed):

trajepl (approved):

trajepl:
image
image
If `requirements_file_path` is `None`, the `Path(None)` will throw the exception.

xiaoyu-work:
> image image If `requirements_file_path` is `None`, the `Path(None)` will throw the exception.

Yeah we allowed None value here. I'll add an additional check here.

jambayk:
> image image If `requirements_file_path` is `None`, the `Path(None)` will throw the exception.

This error is actually because of the typing annotation for `requirements_file_path` and not because of the validator. The validator doesn't get called when it is `None`. 
For some reason, the annotation `requirements_file_path: Optional[str] = None` and `requirements_file_path: str = None` behave differently. For the former, it doesn't allow the user to explicitly set `requirements_file_path` to `None` but the latter does. 

image

So, what we want is to remove all `Optional` from the annotations. You can see that none of the other olive code uses `Optional` when annotating optional fields that can take None. It automatically becomes optional when the field is provided a default value.

trajepl (approved):

PR #132, opened 2023-03-28
Support run with no search strategy

jambayk:
This PR adds support for running passes without search strategy. If `search_stategy` is `None`, the passes are run on the input_model and the final output model is evaluated if there is a valid evaluator. The output of the run is the final output model and evaluation results. 

## Tests 
Added tests for the new functionalities introduced. 
Updated relevant pre-existing tests.

### DIFF
--- a/docs/source/overview/options.md
+++ b/docs/source/overview/options.md
@@ -275,6 +275,10 @@ This is a dictionary that contains the information of the engine. The informatio
     - `max_time: [int]` The maximum time of the search in seconds. Only valid for `joint` execution order. By default, there is no
     maximum time.
 
+  If `search_strategy` is `null`, the engine will run the passes in the order they were registered without. Thus, the passes must have empty
+  search spaces. The output of the final pass will be evaluated if there is a valid evaluator. The output of the engine will be the output model
+  of the final pass and its evaluation result.
+
 - `host: [str | Dict]` The host of the engine. It can be a string or a dictionary. If it is a string, it is the name of a system in `systems`.
     If it is a dictionary, it contains the system information. If not specified, it is the local system.
 
@@ -288,6 +292,13 @@ This is a dictionary that contains the information of the engine. The informatio
 
 - `clean_evaluation_cache: [Boolean]` This decides whether to clean the evaluation cache of the engine before running the engine. This is
 `false` by default.
+
+- `output_dir: [str]` The directory to store the output of the engine. If not specified, the output will be stored in the current working
+    directory. For a run with no search, the output is the output model of the final pass and its evaluation result. For a run with search, the
+    output is a json file with the search results.
+
+- `output_name: [str]` The name of the output. This string will be used as the prefix of the output file name. If not specified, there is no
+    prefix.
 
 Please find the detailed config options from following table for each search algorithm:
 
--- a/docs/source/overview/options.md
+++ b/docs/source/overview/options.md
@@ -257,7 +257,7 @@ Please also find the detailed options from following table for each pass:
 
 This is a dictionary that contains the information of the engine. The information of the engine contains following items:
 
-- `search_strategy: [Dict]` The search strategy of the engine. It contains the following items:
+- `search_strategy: [Dict | Boolean | None]` The search strategy of the engine. It contains the following items:
 
     - `execution_order: [str]` The execution order of the optimizations of passes. The options are `pass-by-pass` and `joint`.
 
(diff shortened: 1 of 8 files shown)


olive/cache.py
```diff
@@ -113,3 +114,44 @@ def clean_pass_run_cache(pass_type: str, cache_dir: Union[str, Path] = ".olive-c
     run_jsons = list(run_cache_dir.glob(f"{pass_type}-*.json"))
     for run_json in run_jsons:
         _delete_run(run_json.stem, cache_dir)
+
+
+def save_model(
```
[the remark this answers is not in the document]
jambayk: Since we are saving a model from cache into a user specified location, I think it is still something like `save_model`? It's already under `olive.cache` so I don't think there is too much ambiguity about its use?

jambayk:
> I may have a few concerns to split the case where `search_strategy` is not given by users from the original `engine.run`. There are the reasons:
> 
> 1. If users want to evaluate the models pass-by-pass even if there is no need to search, then current PR would not work.
> 2. You have to maintain two similar code snippets for run_pass -> evaluation -> save.
> 
> How about to set default value for search strategy? That should work when user does not set the strategy in their config and without search_space. image

I had some concerns about this too. However, after discussing with Devang, this case is for when we want to use Olive to only apply passes. One of the requirements of Olive is that it should be the place where users can apply techniques to model instead of calling tools like `quantize_dynamic` themselves. In this user scenario, Olive should take in input model, apply passes and output final model and evaluation. So, this is a distinct scenario from the others. 

1 in your example can be achieved by using a search strategy (perhaps exhaustive search algorithm) + passes with no search space. Once we add the new run results (from my stale PR), it should let users analyze the outputs of different passes. This is not the same scenario as above. 

2 is not a concern since the workflows are different. The `None` search_strategy case in this example can only be used to apply passes and get final model and evaluation. 

Maybe this is over loading the meaning of `search_strategy` and we should enable/disable search with a different flag? That way we can also set a default value for `search_strategy` that will be used if search flag is enabled? The key requirement is that there should be a non-search and search mode for the workflow.

olive/cache.py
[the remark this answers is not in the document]
jambayk: Should I rename it to `dump_model`?

docs/source/overview/options.md
```diff
@@ -275,6 +275,10 @@ This is a dictionary that contains the information of the engine. The informatio
     - `max_time: [int]` The maximum time of the search in seconds. Only valid for `joint` execution order. By default, there is no
     maximum time.
 
+  If `search_strategy` is `null`, the engine will run the passes in the order they were registered without. Thus, the passes must have empty
+  search spaces. The output of the final pass will be evaluated if there is a valid evaluator. The output of the engine will be the output model
+  of the final pass and its evaluation result.
```
devang-ml, line 280: "the passes must have empty search spaces" -- does this mean user should not provide searchable options in the config file?

test/unit_test/test_engine.py
```diff
@@ -44,11 +50,45 @@ def test_register(self):
...
+        assert engine.passes[name]["clean_run_cache"] is False
+
+    def test_register_no_search(self):
+        # setup
+        p = get_onnx_dynamic_quantization_pass(default_to_search=False)
+        name = p.__class__.__name__
+
+        options = {
+            "search_strategy": None,
+        }
```
devang-ml, line 67: - If "search_strategy" is not provided or set to None then run-no-search path.
- If "search_strategy" is True then run the default search strategy
- If "search_strategy" is a dict then validate and run the strategy specified in the dict.

No need to create yet another option for the user to select default search strategy.

olive/engine.py
```diff
@@ -134,38 +135,47 @@ def register(
...
+    def run(self, input_model: OliveModel, verbose: bool = False, output_dir: str = None, output_name: str = None):
         """
         Run all the registered Olive passes on the input model and produce one or more candidate models.
         """
         if not self._initialized:
             self.initialize()
 
+        if self.search_strategy is None:
+            return self._run_no_search(input_model, verbose, output_dir, output_name)
+
```
devang-ml, line 169: Is it possible to avoid separate run function?
One way to look at this path is -- The search space has only single search point for each Pass.

docs/source/overview/options.md
jambayk: Yes. if the provide search able options when search strategy is none, then it errors out when registering the pass.

test/unit_test/test_engine.py
jambayk: This sounds like a good idea. 

what do you think the default search algorithm should be? `exhaustive`?

devang-ml: yes

olive/engine.py
jambayk: Yes, we could combine them. We can create a dummy search strategy to run the passes without search using the pre-existing logic. This also addresses a part of @trajepl 's comment about splitting the logic.

test/unit_test/test_engine.py
jambayk: Also, what should the default value for `search_strategy` be? `True` or `None`?

olive/cache.py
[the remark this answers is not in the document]
trajepl: or `persist_model`?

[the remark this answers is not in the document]
trajepl: Not the blocker for this PR. Feel free to use the one you like. Just leave some comments for better understanding.

trajepl:
> Jambay: 1 in your example can be achieved by using a search strategy (perhaps exhaustive search algorithm) + passes with no search space. Once we add the new run results (from my stale PR), it should let users analyze the outputs of different passes. This is not the same scenario as above.

> Devang: Is it possible to avoid separate run function?
One way to look at this path is -- The search space has only single search point for each Pass.

@jambayk I may not think it is make sense for the new results adding. Both search and no-search, for users(at least me), when I test the models, I always want to compare the model metrics after `olive.engine.run`.  Also it is very easy to add the function to show the best models results even we change the design for SearchResults.

olive/engine.py
jambayk: Done

trajepl (approved):

PR #134, opened 2023-03-29
PassConfigParam: rename `default` -> `default_value`, `default_search` -> `searchable_values`. Introduce `ConditionalDefault`

jambayk:
This PR adds support for setting pass config default values that are conditional on other parameters. It introduces `ConditionalDefault` which is a special type of `Conditional` whose final choices are fixed values instead of `Categoricals`. 

Also adds some special parameter values called `IGNORED` and `INVALID` to better construct search spaces. A search point is invalid if any of the parameters get assigned as `SpecialParamValue.INVALID`. The search algorithms prune these search points during search.  

Renamed `default` -> `default_value` and `default_search` -> `searchable_values` in `PassConfigParam` to make their purposes clear. 
Added comments and documentation for search parameters and `_default_config`.

### DIFF
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -100,13 +100,13 @@ class Pass(AutoConfigClass):
         """
         default_config = self.default_config()
         for key, value in config.items():
-            if value == PassParamDefault.DEFAULT:
-                config[key] = default_config[key].default
+            if value == PassParamDefault.DEFAULT_VALUE:
+                config[key] = default_config[key].default_value
             elif value == PassParamDefault.SEARCHABLE_VALUES:
                 value = default_config[key].searchable_values
                 if value is None:
                     logger.warning(f"Parameter {key} does not have searchable values. Using default value instead.")
-                    value = default_config[key].default
+                    value = default_config[key].default_value
                 config[key] = value
         return config
 
@@ -143,7 +143,7 @@ class Pass(AutoConfigClass):
             if isinstance(value, SearchParameter):
                 # resolve conditional parameters
                 # if categorical with single choice, use that choice directly
-                value = self._resolve_search_parameters(value, fixed_params)
+                value = self._resolve_search_parameter(value, fixed_params)
             if value == SpecialParamValue.INVALID:
                 # TODO: better error message, e.g. what the parent values were, how it was invalid
                 raise ValueError(
@@ -161,7 +161,7 @@ class Pass(AutoConfigClass):
 
         return fixed_params, search_space
 
-    def _resolve_search_parameters(self, param: SearchParameter, fixed_params: Dict[str, Any]) -> Any:
+    def _resolve_search_parameter(self, param: SearchParameter, fixed_params: Dict[str, Any]) -> Any:
         """
         Resolve a search parameter.
         """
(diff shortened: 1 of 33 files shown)


guotuofeng:
>         return config

please add comments to describe the purpose of default/default_search so that we could easily understand the logic.

---
Refers to: olive/passes/olive_pass.py:104 in bf23a59. [](commit_id = bf23a59918cf29aa04b44f7076c0c7ed0243ab4a, deletion_comment = False)

devang-ml:
> > ```
> >     return config
> > ```
> 
> please add comments to describe the purpose of default/default_search so that we could easily understand the logic.
> 
> Refers to: olive/passes/olive_pass.py:104 in [bf23a59](https://github.com/microsoft/Olive/commit/bf23a59918cf29aa04b44f7076c0c7ed0243ab4a). [](commit_id = [bf23a59](https://github.com/microsoft/Olive/commit/bf23a59918cf29aa04b44f7076c0c7ed0243ab4a), deletion_comment = False)

'default' could be renamed as 'default_value' and 'default_search" could be renamed as 'searchable_values" for clarification.

olive/strategy/search_parameter.py
```diff
@@ -137,6 +152,92 @@ def to_json(self):
...
+        """
+        return Categorical([SpecialParamValue.IGNORED])
+
+
+class ConditionalDefault(Conditional):
+    """
+    Parameter with conditional default value
+    """
+
+    def __init__(self, parents: Tuple[str], support: Dict[Tuple[Any], Any], default: Any = SpecialParamValue.INVALID):
```
guotuofeng, line 175: >Any [](http://example.com/codeflow?start=70&length=3)

should the value type be SearchParameter?

olive/passes/olive_pass.py
```diff
@@ -119,38 +126,50 @@ def _init_fixed_and_search_params(
...
+                # if categorical with single choice, use that choice directly
+                value = self._resolve_search_parameters(value, fixed_params)
+            if value == SpecialParamValue.INVALID:
+                # TODO: better error message, e.g. what the parent values were, how it was invalid
+                raise ValueError(
+                    f"Invalid value for parameter '{key}'. Either the parameter or its parents are not fixed."
+                )
             if isinstance(value, SearchParameter):
-                search_space[key] = self._resolve_search_parameter(value, fixed_params)
+                search_space[key] = value
```
guotuofeng, line 146: > search_space[key] = value [](http://example.com/codeflow?start=15&length=26)

should it be put in above if clause?

jambayk: The first one does the simplification of `Categorical`, `Conditional` and `ConditionalDefault`. All of these can became fixed parameters if 1) Categorical has only one option, 2) Conditional has all parents fixed and value is categorical with one option, 3) ConditionalDefault has all parents fixed so its value is also fixed. 

We only add to `search_space` if the above simplified value is still a search parameter. Otherwise, it is a fixed parameter.

guotuofeng (changes requested):
Please add comments or docs to describe the difference between deafult vs default_search

guotuofeng (approved):

PR #151, opened 2023-04-03
Update docs on ONNX FP16 pass and fix defaults

jstoecker:
- Updates docs to include `OnnxFloatToFloat16`
- Both `op_block_list` and `node_block_list` should default to `None`.


jambayk (approved):
Thanks!

PR #154, opened 2023-04-04
olive runtime state center

trajepl:
1. introduce a new class called Footprints which will log all the model_id and related configs during olive.engine.run. This class can serve following model selection/model package/visualization.
2. remove duplicated function from search results.


devang-ml:
> 1. introduce a new class called Footprints which will log all the model_id and related configs during olive.engine.run. This class can serve following model selection/model package/visualization.
> 2. remove duplicated function from search results.

Why is Footprint associated with Engine and not search strategy? Engine is sending signals after each pass run to search strategy anyway and the search strategy is responsible to interpret the signals.

olive/engine/footprint.py
```diff
@@ -0,0 +1,178 @@
...
+logger = logging.getLogger(__name__)
+
+
+class FootprintNodeMetric(ConfigBase):
+    value: dict
+    cmp_direction: dict = {}
+    is_goals_met: bool = False
+
+
+class FootprintNode(ConfigBase):
```
xiaoyu-work, line 22: Can I also have the `inference_config` for a specific Pass output model?

olive/engine/footprint.py
```diff
@@ -0,0 +1,178 @@
...
+    metrics: FootprintNodeMetric = FootprintNodeMetric(value={}, is_goals_met=False)
+
+    date_time: float = datetime.now().timestamp()
+
+    def update(self, **kwargs):
+        for k, v in kwargs.items():
+            setattr(self, k, v)
+
+
+class Footprint:
```
xiaoyu-work, line 39: Can you add some unit tests for this class?

xiaoyu-work:
Can you also update the example tests?

xiaoyu-work:
Engine run will return a `Footprint` object. Is this expected? Does user need to know this `Footprint` class and get their desired result from this `Footprint` object by themselves?

olive/engine/footprint.py
trajepl: You can call the function of get_model_inference_config with given model_id to get the inference_config which is included in model_config.

olive/engine/footprint.py
trajepl: Sure, will add it. Also for the documents for olive footprints, will update it later when all comments are solved.

trajepl:
> > 1. introduce a new class called Footprints which will log all the model_id and related configs during olive.engine.run. This class can serve following model selection/model package/visualization.
> > 2. remove duplicated function from search results.
> 
> Why is Footprint associated with Engine and not search strategy? Engine is sending signals after each pass run to search strategy anyway and the search strategy is responsible to interpret the signals.

I was thinking to use Footprint to log all the runtime model information(model_id/model_config/etc.) during engine.run (pass_run/evaluation no matter what kind of search stretegy olive adapts).  It is not used to handle the signals.

Another concept called search_results is used for search strategy actually. Footprints and search results should not be the same thing.

trajepl:
> Engine run will return a `Footprint` object. Is this expected? Does user need to know this `Footprint` class and get their desired result from this `Footprint` object by themselves?

Will provide docs for that. Also can dump the info into local file for downstream usage. 
Please also share you thoughs. We can discuss that for final output.

xiaoyu-work:
It kinda confused me that `Engine` has an attribute called `footprints` which is a `Footprint` object, while this `Footprint` class also has an attribute called `footprints` which is a dict. Can you somehow rename these to make it clearer? Like `footrpint` in `Engine` and  `footprint_node_dict` in  `Footprint`?

trajepl:
> It kinda confused me that `Engine` has an attribute called `footprints` which is a `Footprint` object, while this `Footprint` class also has an attribute called `footprints` which is a dict. Can you somehow rename these to make it clearer? Like `footrpint` in `Engine` and `footprint_node_dict` in `Footprint`?

Updated it to `nodes`.

trajepl:
> It kinda confused me that `Engine` has an attribute called `footprints` which is a `Footprint` object, while this `Footprint` class also has an attribute called `footprints` which is a dict. Can you somehow rename these to make it clearer? Like `footrpint` in `Engine` and `footprint_node_dict` in `Footprint`?

Done.

olive/engine/footprint.py
```diff
@@ -0,0 +1,178 @@
...
+from datetime import datetime
+from pathlib import Path
+
+from olive.common.config_utils import ConfigBase, config_json_dumps, config_json_loads
+
+logger = logging.getLogger(__name__)
+
+
+class FootprintNodeMetric(ConfigBase):
+    value: dict
```
trajepl, line 17: TODO: should consider the metrics values on different EPs/accelerators for same model.

xiaoyu-work (dismissed):

xiaoyu-work:
Is `get_best_execution` method in `search_strategy.py` still needed?

olive/engine/footprint.py
```diff
@@ -0,0 +1,178 @@
...
+    cmp_direction: dict = {}
+    is_goals_met: bool = False
+
+
+class FootprintNode(ConfigBase):
+    # None for no parent which means current model is the input model
+    parent_model_id: str = None
```
guotuofeng, line 24: >parent_model_id [](http://example.com/codeflow?start=4&length=15)

input_model_id is better or not?

olive/engine/footprint.py
guotuofeng, line 17: >dict [](http://example.com/codeflow?start=11&length=4)

for these dict annotation, would you please help specify the key and value type?

olive/engine/footprint.py
```diff
@@ -0,0 +1,178 @@
...
+class Footprint:
+    """
+    The engine footprint is a class that contains the footprint of the engine runtime.
+    It is used to collect the runtime state of the Olive engine and to organize the state
+    in a way that is easy to visualize and understand for customers.
+    """
+
+    def __init__(
+        self,
+        nodes: OrderedDict = {},
```
guotuofeng, line 48: >{} [](http://example.com/codeflow?start=29&length=2)

please don't use the empty dict or empty list as the default argument. Instead, using None.

[https://stackoverflow.com/questions/366422/how-can-i-avoid-issues-caused-by-pythons-early-bound-default-parameters-e-g-m](https://stackoverflow.com/questions/366422/how-can-i-avoid-issues-caused-by-pythons-early-bound-default-parameters-e-g-m)

olive/engine/footprint.py
```diff
@@ -0,0 +1,178 @@
...
+    def get_pareto_frontier(self):
+        self.mark_pareto_frontier()
+        rls = {k: v for k, v in self.nodes.items() if v.is_pareto_frontier}
+        for _, v in rls.items():
+            logger.info(f"pareto frontier points: {v.model_id} {v.metrics.value}")
+
+        # restructure the pareto frontier points to instance of Footprints node for further analysis
+        return Footprint(nodes=rls, objective_dict=self.objective_dict, is_marked_pareto_frontier=True)
+
+    def _plot_pareto_frontier(self, index=[0, 1]):
```
guotuofeng, line 124: >index=[0, 1] [](http://example.com/codeflow?start=36&length=12)

Not used?

guotuofeng (changes requested):
Please fix the default argument issues

olive/engine/footprint.py
trajepl: input_model_id may be same with the Olive engine input_model_id(torch_model)? I prefer to keep parent.

olive/engine/footprint.py
trajepl: I were thinking to plot pareto frontier with >2D data in the future. Then we can use index to plot any 2 or 3 dimensions.
That says:
if we take accuracy/latency/costs into account in the same time, when visualization, user can use index to chose which 2 or 3 dimension they want to display.

olive/engine/footprint.py
```diff
@@ -0,0 +1,178 @@
...
+        if foot_print_node is not None:
+            _model_id = foot_print_node.model_id
+            self.nodes[_model_id] = foot_print_node
+        elif _model_id in self.nodes:
+            self.nodes[_model_id].update(**kwargs)
+        else:
+            self.nodes[_model_id] = FootprintNode(**kwargs)
+        self.resolve_metrics()
+
+    def get_candidates(self):
```
jambayk, line 99: When we evaluate the baseline, the input model also gets evaluated. So does it also become a candidate?

olive/engine/footprint.py
trajepl: Same structure with the results of metrics evaluation. {metrics_name: value, ....}. 
Will tried to add EPs/accelerators into account when Mike's accelerators changes is merged.

trajepl: Also, already add the comments in code for readability.

olive/engine/footprint.py
```diff
@@ -0,0 +1,189 @@
...
+class FootprintNode(ConfigBase):
+    # None for no parent which means current model is the input model
+    parent_model_id: str = None
+    model_id: str
+    model_config: dict = None
+    from_pass: str = None
+    pass_run_config: dict = None
+    is_pareto_frontier: bool = False
+    # TODO add EP/accelerators for same_model_id metrics
+    metrics: FootprintNodeMetric = FootprintNodeMetric(value={}, cmp_direction={}, is_goals_met=False)
```
jambayk, line 38: I think this has the same potential issue as using {}/[] as default value. Can `metric` also be `None` by default? Looking at the other parts of the code (`resolve_metric`, `get_candidates`), it looks like you already assume the default value of `metric` is `None`.

olive/engine/footprint.py
jambayk: Also, what about the models created by intermediate passes in `pass-by-pass` execution order? They also get evaluated. 

Maybe we need a `is_candiate` flag for models that is true if the pass is the final pass? In the future we could also allow intermediate passes to also be candidates by using a similar flag when registering the pass.

olive/engine/footprint.py
trajepl: I think it is fine for parameter to accept {}, not the default value. But anyway, let me remove this. Thanks. :)

olive/engine/footprint.py
trajepl: Remove the input model from candidates.

Yes, I saw your previous draft PR. There is a flag called is_candidate for final around model. But I am thinking that:
if the intermediate model(in pass-by-pass) is in pareto-frontier, we may should put it in output to let user choose. So I just skip to add is_candidate.

jambayk: That also sounds good to me. Thanks for the clarification. I was wondering about cases where the intermediate passes might not have the model in the desired format (like pytorch model -> pytorch model pass, when we want onnx model) but we don't have such cases currently.

jambayk (approved):

olive/engine/footprint.py
jambayk: I just checked how pydantic models behave with defaults and like you said, it doesn't appear to have the risk of using mutable object as default. Thanks for updating regardless!

olive/engine/footprint.py
guotuofeng: I am thinking here we will hit the mutable default parameter issues?

guotuofeng (approved):

PR #160, opened 2023-04-10
Handle ONNX model with external data for AzureML systems

jambayk:
This PR adds support for ONNX models with external data for AzureML systems (both model evaluation and pass run). 

It also uses a temporary directory for onnx model conversion since the output model path or it's parent could be used by other processes if not provided carefully. 
`all_tensors_to_one_file` default value is now `True` like it is for `onnx.save_model`. This is beneficial since there is only one external data file. Also, for models with unsupported tensor names, the external data files have randomly generated names which is not ideal.

### DIFF
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -87,7 +87,10 @@ class OnnxConversion(Pass):
             "all_tensors_to_one_file": PassConfigParam(
                 type_=bool,
                 default_value=True,
-                description="If true, external data is written to a single file instead of one file per tensor.",
+                description=(
+                    "Effective only if save_as_external_data is True. If true, save all tensors to one external file"
+                    " specified by location. If false, save each tensor to a file named with the tensor name."
+                ),
             ),
         }
 
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -86,7 +86,7 @@ class OnnxConversion(Pass):
             ),
             "all_tensors_to_one_file": PassConfigParam(
                 type_=bool,
-                default_value=False,
+                default_value=True,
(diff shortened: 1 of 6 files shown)


jstoecker (approved):
This is great, thanks!

One small suggestion: "model.onnx" is the default name when `OnnxModel.resolve_path` is called, but it is a parameter. It's possible that a caller gives a different name, in which case the assumption it's always named "model.onnx" will be incorrect. Perhaps remove the parameter (simplest) or embed the name as a member OnnxModel when is_file=false?

jambayk:
> One small suggestion: "model.onnx" is the default name when `OnnxModel.resolve_path` is called, but it is a parameter. It's possible that a caller gives a different name, in which case the assumption it's always named "model.onnx" will be incorrect. Perhaps remove the parameter (simplest) or embed the name as a member OnnxModel when is_file=false?

@jstoecker Do you mean the assumption on line 32 in utils.py? If so, we ensure that it is `model.onnx` when uploading to aml at line 175 in aml_system.py.

jstoecker:
> > One small suggestion: "model.onnx" is the default name when `OnnxModel.resolve_path` is called, but it is a parameter. It's possible that a caller gives a different name, in which case the assumption it's always named "model.onnx" will be incorrect. Perhaps remove the parameter (simplest) or embed the name as a member OnnxModel when is_file=false?
> 
> @jstoecker Do you mean the assumption on line 32 in utils.py? If so, we ensure that it is `model.onnx` when uploading to aml at line 175 in aml_system.py.

You're right. Nevermind!

jstoecker (dismissed):

jstoecker (approved):

jambayk:
@jstoecker Sorry for the bother but could you please approve again. I missed another test. Thanks!

PR #164, opened 2023-04-11
footprint doc & plot pareto frontier

trajepl:
You can also call the following methods to plot the Pareto frontier footprints. Also please make sure you installed `plotly` successfully.
```python
from olive.engine.footprint import Footprint
footprint = Footprint.from_file("footprints.json")
footprint.plot_pareto_frontier()
```


olive/engine/footprint.py
```diff
@@ -99,7 +99,11 @@ def record(self, foot_print_node: FootprintNode = None, **kwargs):
         self.resolve_metrics()
 
     def get_candidates(self):
-        return {k: v for k, v in self.nodes.items() if v.metrics is not None and v.parent_model_id is not None}
+        return {
+            k: v
+            for k, v in self.nodes.items()
+            if v.metrics is not None and len(v.metrics.value) > 0 and v.parent_model_id is not None
```
guotuofeng, line 105: >if v.metrics is not None and len(v.metrics.value) > 0  [](http://example.com/codeflow?start=12&length=54)

if v.metrics and v.metrics.value should be more concise

olive/engine/footprint.py
```diff
@@ -134,15 +138,78 @@ def get_pareto_frontier(self):
...
         # plot pareto frontier
-        # import plotly.graph_objects as go
-        pass
+        try:
+            import pandas as pd
+            import plotly.graph_objects as go
+
+            pd.options.mode.chained_assignment = None
+        except Exception:
+            logger.error("Please make sure you installed pandas and plotly successfully.")
```
guotuofeng, line 169: >error [](http://example.com/codeflow?start=19&length=5)

warning since the error could be ignored?

trajepl: Fixed.

olive/engine/footprint.py
trajepl: done.

olive/engine/footprint.py
```diff
@@ -64,6 +64,13 @@ def __init__(
     def record_objective_dict(self, objective_dict):
         self.objective_dict = objective_dict
 
+    def _is_empty_metric(self, metric: FootprintNodeMetric):
+        if metric is None:
+            return True
+        if metric.value is None or len(metric.value) == 0:
```
guotuofeng, line 70: >metric.value is None or len(metric.value) == 0 [](http://example.com/codeflow?start=11&length=46)

[https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty](https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty)

trajepl: Done.

guotuofeng (approved):

PR #166, opened 2023-04-12
Sample for optimizing Stable Diffusion for DirectML

jstoecker:
Adds a sample that illustrates how to use Olive to convert and optimize Stable Diffusion for ORT+DML. The transformer optimizations require an in-development build of ORT, so this sample will be updated when we have nightly or public builds with the necessary changes. However, I'd like to get eyes on this change sooner than later and avoid missing any breaking changes in Olive itself.

A few additional things to note:
- This sample is unique in that it optimized 5 models, not 1, so there are several config files invoked through stable_diffusion.py
- The helper script also provides a thin wrapper around the `diffusers` library to test inference with the optimized models


PatriceVignola (dismissed):

olive/passes/onnx/conversion.py
```diff
@@ -92,6 +92,11 @@ def _default_config() -> Dict[str, PassConfigParam]:
                     " specified by location. If false, save each tensor to a file named with the tensor name."
                 ),
             ),
+            "external_data_name": PassConfigParam(
+                type_=str,
+                default_value="model.onnx.data",
```
jambayk, line 97: I am thinking about perhaps the default value for this parameter being `None`. If it is `None`, the value used is `{output_model_path.name}.data` like it was before. This is also the value the onnxruntime `save_model_to_file` methods do. 

But this is not required for this PR. We have to update all of the other passes that output onnx models to handle external data correctly.

jambayk (approved):

olive/passes/onnx/conversion.py
jstoecker: That makes sense. I'll do what you suggest.

For context, the reason this parameter is necessary is that the diffusers library expects "weights.pb" and won't copy the external data unless it's named exactly this.

jambayk (approved):

olive/passes/onnx/conversion.py
jambayk: Thanks! I will update the rest of the passes later. We haven't had to deal with large models so far so it hasn't been an issue but our other onnx passes aren't handling the external data issue right now.

PR #167, opened 2023-04-12
add workflow setup

leqiao-1:
add workflow setup
modify olive[openvino] to olive-ai[openvino]


olive/workflows/run/run.py
```diff
@@ -4,14 +4,32 @@
...
+        SystemType.Local: {Device.CPU: ["onnxruntime==1.13.1"], Device.GPU: ["onnxruntime-gpu==1.13.1"]},
+    },
+    "pass": {
+        "OnnxFloatToFloat16": ["onnxconverter-common"],
+        "OrtPerfTuning": ["psutil"],
+        "QuantizationAwareTraining": ["pytorch-lightning"],
+        "OpenVINOConversion": ["openvino==2022.3.0", "openvino-dev[tensorflow,onnx]==2022.3.0"],
+        "OpenVINOQuantization": ["openvino==2022.3.0", "openvino-dev[tensorflow,onnx]==2022.3.0"],
+    },
+}
```
guotuofeng, line 30: could we manage only one extra for these dependencies? from the time being, both here and the setup.py maintain the dependencies.

olive/workflows/run/run.py
```diff
@@ -44,7 +62,37 @@ def automatically_insert_passes(config):
...
+    required_packages = []
+
+    if config.engine.host.type == SystemType.Local:
+        required_packages.extend(
+            DEPENDENCY_MAPPING["device"][config.engine.host.type][config.engine.host.config.device]
+        )
+    else:
+        required_packages.extend(DEPENDENCY_MAPPING["device"][config.engine.host.type])
+
+    for _, pass_config in config.passes.items():
```
jambayk, line 75: The pass dependencies are only required in the local environment if the host for pass is local system. 
You can get the pass' host by something like `pass_config.get("host") or config.engine.host or LocalSystem()` (If the pass doesn't have host option set, it uses engine host. Default engine host is local system.

I think you can install the pass dependencies if type is local, otherwise, just print the dependencies and tell the user to make sure these packages are present in their aml/docker environment.

leqiao-1: change to only install pass related dependencies if it's local target

olive/workflows/run/run.py
leqiao-1: put extra dependencies together in extra_dependencies.json

olive/extra_dependencies.json
```diff
@@ -0,0 +1 @@
+{"azureml": ["azure-ai-ml>=0.1.0b6", "azure-identity"], "docker": ["docker"], "cpu": ["onnxruntime"], "gpu": ["onnxruntime-gpu"], "directml": ["onnxruntime-directml"], "openvino": ["openvino==2022.3.0", "openvino-dev[tensorflow,onnx]==2022.3.0"], "tf": ["tensorflow==1.15.0"]}
```
guotuofeng: format this file and remember to update MANIFEST.in otherwise, this file will not be included in the wheel

olive/workflows/run/run.py
jambayk: In the latest change, I see that you check for the engine host. However, the host for a pass might be different from the engine host if the user provides a separate host for the pass. 

Could you check for the host per pass using the command I shared above? If the local, you can add to the packages list else print for the pass.

olive/extra_dependencies.json
jambayk, line 1: Can you link the sections 
https://github.com/microsoft/Olive/blob/main/README.md?plain=1#L52 
https://github.com/microsoft/Olive/blob/main/docs/source/getstarted/installation.md?plain=1#L65 
to this file and remove the lists from the readme?

olive/model.py
```diff
@@ -165,6 +163,8 @@ def load_model(self) -> onnx.ModelProto:
         return onnx.load(self.model_path)
 
     def prepare_session(self, inference_settings: Dict[str, Any], device: Device):
+        import onnxruntime as ort
```
trajepl, line 166: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/util/lazy_loader.py
https://github.com/python/peps/blob/main/pep-0690.rst

May `lazy load/import` will be more suitable? Look like lazy load can avoid this kinda duplicated code snippet.

olive/workflows/run/run.py
```diff
@@ -44,7 +47,61 @@ def automatically_insert_passes(config):
...
+                __import__(package)
+            except (ImportError):
+                subprocess.check_call(["pip", "install", "{}".format(package)])
+    else:
+        for package in set(DEPENDENCY_MAPPING["device"][config.engine.host.type]):
+            try:
+                __import__(package)
+            except (ImportError):
+                subprocess.check_call(["pip", "install", "{}".format(package)])
+        logger.info(
```
trajepl, line 97: Looks like the info is telling that they should install the packages. But it already finished the installation in the position that you are logging.

setup.py
```diff
@@ -17,22 +18,20 @@ def get_version(rel_path):
...
 
+def get_extra_deps(rel_path):
+    here = os.path.abspath(os.path.dirname(__file__))
+    extra_deps = json.load(open(os.path.join(here, rel_path), "r"))
+    return extra_deps
+
+
 # use techniques described at https://packaging.python.org/en/latest/guides/single-sourcing-package-version/
-# Don't use technique 6 since it need extra dependencies.
+# Don't use technique 6 since it needs extra dependencies.
```
trajepl, line 28: What is techinique 6?

olive/extra_dependencies.json
leqiao-1: added into MANIFEST.in

olive/extra_dependencies.json
leqiao-1: It might be better to keep the install instructions in readme, since it's not specified for workflow run. I added setup usage in tutorial.

olive/model.py
leqiao-1: I will start a new PR for lazy import

olive/workflows/run/run.py
leqiao-1: the packages installed are packages needed for pipelines, like azureml or docker package. The packages in log are required for passes, for example if user wants to run perf tuning on aml, he needs to have psutil in environment config file.

setup.py
leqiao-1: We cannot directly use version variable in __init__.py

olive/workflows/run/run.py
leqiao-1: I made it into two package lists: packages need to be installed locally and packages needed for remote host. For each pass, it will check pass.host, engine.host and local system in order.

examples/directml/squeezenet/README.md
```diff
@@ -8,6 +8,11 @@ Performs optimization pipeline:
 Outputs the best metrics, model, and corresponding Olive config.
 
 ## Optimize SqueezeNet
+First, install required packages according to passes.
```
guotuofeng, line 11: Not sure do we need run setup and optimize in single cmd.

guotuofeng (approved):

guotuofeng (approved):

PR #168, opened 2023-04-12
fix docs bug & try eps first before apply

trajepl:
### Docs fix
image

### Try firstly before apply execution providers
When I tried auto-optimization, especially switch the devices between `cpu` and `gpu`, in most of the cases, not all gpu related eps' environments is ready.
 
image

So how about to filter the invalid execution providers firstly, then tried to apply it to create inference settings?


olive/model.py
```diff
@@ -201,8 +201,22 @@ def prepare_session(self, inference_settings: Dict[str, Any], device: Device):
 
         if len(execution_provider) >= 1 and execution_provider[0] == "DmlExecutionProvider":
             sess_options.enable_mem_pattern = False
+        eps = self.filter_execution_providers(self.model_path, sess_options, execution_provider)
```
guotuofeng, line 204: why do we need this behavior? should not it just fail if user pass invalid EPs?

guotuofeng: the filter EP is not responsibility of the model.py. If we'd like do it, it need to be in another place that invoke model

trajepl: Yes, agree with this.
> the filter EP is not responsibility of the model.py. If we'd like do it, it need to be in another place that invoke model

This PR is draft to bring this idea up for auto optimization.

trajepl: Fixed with today's sync. 
1. When user set the EPs list, just respect their settings. Exit the program even there is the exception.
2. When user does not provider the EPs list, just return the default single EP w.r.t device.(gpu: CudaEP, cpu: CPUEP)

guotuofeng (approved):

PR #173, opened 2023-04-14
Support external data for all onnx passes

jambayk:
Due the limit of protobuf, large onnx models (>2GB) cannot be saved to a single onnx file. The tensors must be saved to an external data file. Our onnx related passes do not handle this case. 

This PR adds options for user to specify whether to save as external data or not. If the model is too large, we override the user's configuration and save as external data regardless. A warning is also generated to inform the user of this. This ensures that the pass doesn't fail when saving the onnx model to file.  

The model save logic is based on a similar implementation by onnxruntime https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/onnx_model.py#L947 but with more control on the name of the external data file.

### DIFF
--- a/olive/passes/onnx/common.py
+++ b/olive/passes/onnx/common.py
@@ -78,16 +78,15 @@ def model_proto_to_file(
         try:
             # save model
             onnx.save_model(model, str(output_path))
-        except ValueError as e:
+        except ValueError:
             # model is too large to save as a single file
-            if "Please use save_as_external_data to save tensors separately from the model file." in str(e):
+            if model.ByteSize() >= onnx.checker.MAXIMUM_PROTOBUF:
                 logger.warning(
                     "Model is too large to save as a single file but 'save_as_external_data' is False. Saving tensors"
                     " as external data regardless."
                 )
                 return model_proto_to_file(model, output_path, True, all_tensors_to_one_file, external_data_name)
-            else:
-                raise e
+            raise
         return False
 
     # location for external data
--- a/olive/passes/onnx/common.py
+++ b/olive/passes/onnx/common.py
@@ -29,7 +29,7 @@ def get_external_data_config():
             default_value=True,
             description=(
                 "Effective only if save_as_external_data is True. If true, save all tensors to one external file"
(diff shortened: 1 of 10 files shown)


olive/passes/onnx/common.py
```diff
@@ -0,0 +1,142 @@
...
+    output_dir = output_path.parent
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    if not save_as_external_data:
+        try:
+            # save model
+            onnx.save_model(model, str(output_path))
+        except ValueError as e:
+            # model is too large to save as a single file
+            if "Please use save_as_external_data to save tensors separately from the model file." in str(e):
```
jstoecker, line 83: With a sufficiently large model I actually get an error like the following, which won't be caught by your `if` check.

Screenshot 2023-04-14 151005

jambayk: Oh, thanks for the heads up. I thought I would catch the error correctly because of this https://github.com/onnx/onnx/blob/main/onnx/__init__.py#L110 

I will test the model save logic with a large model. Do you have one you recommend or I could just create a large dummy model?

jstoecker: You can probably run the SD example. The u-net is the big one:

```
python -m olive.workflows.run --config .\config_unet.json
```

I think this will run correctly even with onnxruntime 1.14. It may fail to evaluate at the end, but it should at least convert successfully.

jambayk: Thanks! I will try this model. I have to see why onnx.save_model doesn't catch the error properly. Looks like it is raising the error from protobuf instead of raising the linked error.

jambayk: Found this issue. `onnx` only added the check in version 1.13.0.

jambayk: @jstoecker I have updated the check and validated it against onnx 1.12.0 also.

jstoecker (approved):
Looks good to me! This is awesome, thanks Jambay.

PR #175, opened 2023-04-17
add dummy dataloader for evaluation

leqiao-1:


trajepl (approved):

olive/evaluator/metric_config.py
```diff
@@ -19,13 +19,16 @@
     "user_script": ConfigParam(type_=Union[Path, str]),
     "data_dir": ConfigParam(type_=Union[Path, str]),
     "batch_size": ConfigParam(type_=int, default_value=1),
+    "input_names": ConfigParam(type_=List, is_object=True),
```
jambayk, line 22: I just saw that this PR has been merged. 
I wanted to ask why these parameters have ‘is_object’ and True? We only use this property if it’s an object/function that cannot be serialized as is imported from user_script. 
I don’t see these parameters being used that way.

leqiao-1: It's an error here. Here is the PR to update arguments: https://github.com/microsoft/Olive/pull/176

PR #181, opened 2023-04-18
dummy inputs to cpu & model to cpu before conversion

trajepl:
To avoid the case where user define the customized evaluation functions and forget to reset the device back to cpu.
Further fix for PR: https://github.com/microsoft/Olive/pull/177

### DIFF
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -120,9 +120,11 @@ class OnnxConversion(Pass):
     def _run_for_config(self, model: PyTorchModel, config: Dict[str, Any], output_model_path: str) -> ONNXModel:
         pytorch_model = model.load_model()
         pytorch_model.eval()
+
+        # TODO: add e2e test for model on cpu but data on gpu; model on gpu but data on cpu
         # put pytorch_model and dummy_inputs at the same device
-        torch_device = pytorch_model.device
-        dummy_inputs = tensor_data_to_device(self._dummy_inputs, torch_device)
+        pytorch_model.to("cpu")
+        dummy_inputs = tensor_data_to_device(self._dummy_inputs, "cpu")
         if isinstance(pytorch_model, torch.jit.RecursiveScriptModule):
             pytorch_model = TraceModelWrapper(pytorch_model)
 
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -9,6 +9,7 @@ from typing import Any, Callable, Dict, List, Union
 import onnx
 import torch
 
+from olive.evaluator.evaluation import tensor_data_to_device
 from olive.model import ONNXModel, PyTorchModel
 from olive.passes import Pass
 from olive.passes.onnx.common import get_external_data_config, model_proto_to_olive_model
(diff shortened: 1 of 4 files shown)


olive/passes/pytorch/qat_utils.py
```diff
@@ -168,7 +169,7 @@ def create_dummy_input(self):
         # dummy inputs
         dummy_inputs = []
         for input_shape, input_type in zip(self.config.input_shapes, input_types):
-            dummy_inputs.append(torch.zeros(input_shape, dtype=input_type))
+            dummy_inputs.append(torch.zeros(input_shape, dtype=input_type, device=Device.CPU))
```
guotuofeng, line 172: better to postpone the data to device when it is used.

olive/passes/onnx/conversion.py
```diff
@@ -119,6 +120,7 @@ def _initialize(self):
     def _run_for_config(self, model: PyTorchModel, config: Dict[str, Any], output_model_path: str) -> ONNXModel:
         pytorch_model = model.load_model()
         pytorch_model.eval()
+        pytorch_model.to(Device.CPU)
```
guotuofeng, line 123: better to add unit test

guotuofeng (changes requested):
Please add unit test for the bug

olive/passes/pytorch/qat_utils.py
trajepl: Done.

olive/passes/onnx/conversion.py
trajepl: Hard to directly test from unit tests. Maybe let us kick off the e2e tests for gpu in next pr when the gpu-cluster is ready.

trajepl:
> Please add test for the bug
> 
> 1. model on cpu but data on gpu
> 2. model on gpu but data on cpu

Will raise it up to team to see what is the available gpu clusters for e2e test in later PR.

olive/passes/onnx/conversion.py
```diff
@@ -119,6 +120,9 @@ def _initialize(self):
     def _run_for_config(self, model: PyTorchModel, config: Dict[str, Any], output_model_path: str) -> ONNXModel:
         pytorch_model = model.load_model()
         pytorch_model.eval()
+        # put pytorch_model and dummy_inputs at the same device
+        torch_device = pytorch_model.device
```
guotuofeng, line 124: seems the pipeline fail here since there is no device attribute on model?

guotuofeng:
> Please add test for the bug
> 
> 1. model on cpu but data on gpu
> 2. model on gpu but data on cpu

Please add TODO for test if we'd like to add test in next PR.

olive/passes/onnx/conversion.py
trajepl: https://stackoverflow.com/questions/58926054/how-to-get-the-device-type-of-a-pytorch-module-conveniently

Look like it is impossible to get device of model, I just put model and data to cpu direcly.

guotuofeng: sure

guotuofeng (approved):

PR #183, opened 2023-04-18
bugfix for latency metrics config in perf tuning

trajepl:
1. When construct the latency metrics configs, it lost several configs like 'user_scirpts'. And the list is hard coded in perf-tuning which can crash again if we change the metrics schema in the future.
2. Enhance the e2e ResNet tests to make it cover the above failed case.
3. typos fix

### DIFF
--- a/olive/evaluator/metric_config.py
+++ b/olive/evaluator/metric_config.py
@@ -53,6 +53,12 @@ def get_user_config_class(metric_type: str):
     return create_config_class(f"{metric_type.title()}UserConfig", default_config, ConfigBase, validators)
 
 
+def get_properties_from_metric_type(metric_type):
+    user_config_class = get_user_config_class(metric_type)
+    # list(user_config_class.schema()["properties"].keys()) # will ignore dataloader_func
+    return list(user_config_class.__fields__)
+
+
 # TODO: automate latency metric config also we standardize accuracy metric config
 class LatencyMetricConfig(ConfigBase):
     warmup_num: int = WARMUP_NUM
--- a/olive/evaluator/metric_config.py
+++ b/olive/evaluator/metric_config.py
@@ -55,7 +55,7 @@ def get_user_config_class(metric_type: str):
 
 def get_properties_from_metric_type(metric_type):
     user_config_class = get_user_config_class(metric_type)
-    # list(user_config_class.schema()["properties"].keys()) # will ignore dataloader_func
+    # avoid to use schema() to get the fields, because it will skip the ones with object type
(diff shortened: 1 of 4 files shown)


jambayk (approved):

olive/passes/onnx/perf_tuning.py
```diff
@@ -42,15 +43,8 @@ def generate_tuning_combos(model, config):
...
-        "data_dir",
-        "dataloader_func",
-        "batch_size",
-        "input_names",
-        "input_shapes",
-        "input_types",
-        "device",
-    ]:
+    # which should be the same as the config in the metric
+    for eval_config in get_properties_from_metric_type(MetricType.LATENCY):
```
jambayk, line 47: Not a breaking issue but can we only add it to `latency_user_config` if it is also in `config`? Right now, this makes the values for all parameters not in `config` as `None` because of the `.get` but this might be problematic in the future if the actual default value is not `None`.

olive/evaluator/metric_config.py
```diff
@@ -53,6 +53,12 @@ def get_user_config_class(metric_type: str):
     return create_config_class(f"{metric_type.title()}UserConfig", default_config, ConfigBase, validators)
 
 
+def get_properties_from_metric_type(metric_type):
+    user_config_class = get_user_config_class(metric_type)
+    # list(user_config_class.schema()["properties"].keys()) # will ignore dataloader_func
```
xiaoyu-work, line 58: Nit: is this comment needed?

trajepl: Thanks. Actually I just do not know why the .schema() method ignored the dataloader_func even it was defined as schema. May some bugs or usage error of pydantic.

trajepl: How about to keep for reminding?

xiaoyu-work: `dataloader_func` is a Callable so it cannot be serialized to json schema. 
Ideally this comment has no relation to the existing codes but I'm okay to keep it if you think it is needed xd.

trajepl: > dataloader_func is a Callable so it cannot be serialized to json schema.
Oh, ic, so the pydantic schema will skip the field which cannot be serialized. I will remove it. Thanks. :)

olive/passes/onnx/perf_tuning.py
trajepl: Good cache. I think it reasonable, will update it.

jambayk (approved):

PR #184, opened 2023-04-18
Add openvino pass unit tests

xiaoyu-work:
### Notes
1. Add OpenVINO pass unit tests.
2. We should add unit tests for other passes and new passes.


olive/passes/openvino/quantization.py
```diff
@@ -27,17 +27,14 @@ def _default_config() -> Dict[str, PassConfigParam]:
                 required=True,
                 description=(
                     "Specific config for openvino.tools.pot.IEEngine. 'engine_config' can be set"
-                    " by passing a dictonary, for example engine_config = {'device': 'CPU'}"
+                    " by passing a dictonary, for example engine_config: {'device': 'CPU'}"
                 ),
             ),
             "dataloader_func": PassConfigParam(
-                type_=Union[Callable, str],
+                type_=str,
```
jambayk, line 34: Is there a reason Callable is not supported anymore?

xiaoyu-work: Ah this was the test change I made before but forgot to change it back. Good catch!

jambayk (approved):

PR #186, opened 2023-04-18
Remove `enum-tools` from docs requirement

jambayk:
`enum-tools` has GPL requirement which causes security compliance issue. So, removing it from docs requirement.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -33,7 +33,6 @@ extensions = [
     "sphinx.ext.napoleon",
     "sphinx_tabs.tabs",
     # "nbsphinx",
-    "enum_tools.autoenum",
     "auto_config_doc",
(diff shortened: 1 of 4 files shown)


docs/source/conf.py
```diff
@@ -33,7 +33,6 @@
     "sphinx.ext.napoleon",
     "sphinx_tabs.tabs",
     # "nbsphinx",
```
xiaoyu-work, line 35: Not related to this PR, but what is `nbsphinx` used for?

xiaoyu-work (approved):

docs/source/conf.py
jambayk: It is an extension that helps render jupyter notebooks as html. We don't have any notebooks in our docs yet so it's commented out.

PR #187, opened 2023-04-19
bugfix: perf_tuning batch size != latency.batch_size

trajepl:
1. Align the batch_size for olive examples, which inconsistency bring gap for perf_tuning metrics and olive evaluations. 

    - [ ] In the future PR, we should directly read batch_size from metrics config.

3. Refine perf-tuning's messy logger info.

### DIFF
--- a/olive/passes/onnx/perf_tuning.py
+++ b/olive/passes/onnx/perf_tuning.py
@@ -56,11 +56,14 @@ def tune_onnx_model(model, config):
 
     tuning_results = []
     for tuning_combo in generate_tuning_combos(model, config):
-        logger.info("Run tuning for: {}".format(tuning_combo))
+        tuning_item = ["provider", "execution_mode", "ort_opt_level", "io_bind"]
+        logger.info("Run tuning for: {}".format(list(zip(tuning_item, tuning_combo))))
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/perf_tuning.py
```diff
@@ -56,11 +56,14 @@ def tune_onnx_model(model, config):
 
     tuning_results = []
     for tuning_combo in generate_tuning_combos(model, config):
-        logger.info("Run tuning for: {}".format(tuning_combo))
+        tuning_item = ["provider", "execution_mode", "ort_opt_level", "io_bind"]
+        logger.info("Run tuning for: {}".format(list(zip(tuning_item, tuning_combo))))
         if tuning_combo[0] == "CPUExecutionProvider" and tuning_combo[3]:
             continue
         tuning_results.extend(threads_num_tuning(model, latency_metric, config, tuning_combo))
-        logger.info("Current tuning results: {}".format(tuning_results[-1]["latency_ms"]))
```
trajepl, line 63: Previous info just record subset of the whole tuning results, which confused me why the best numbers is not in log info.
In `thread_num_tuning`, there will be extra multiple results were pushed into tuning results, but only the last one was logged.

leqiao-1 (approved):

PR #188, opened 2023-04-19
refactor search space generation to support device parameters

guotuofeng:
refactor search space generation to support device parameters

### DIFF
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -160,11 +160,12 @@ class Pass(AutoConfigClass):
         """
         raise NotImplementedError()
 
-    def _resolve_defaults(self, config: Dict[str, Any]) -> Dict[str, Any]:
+    @classmethod
+    def _resolve_defaults(cls, config: Dict[str, Any]) -> Dict[str, Any]:
         """
         Resolve default values.
         """
-        default_config = self.default_config()
+        default_config = cls.default_config()
         for key, value in config.items():
             if value == PassParamDefault.DEFAULT_VALUE:
                 config[key] = default_config[key].default_value
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -2,14 +2,14 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+import inspect
 import logging
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any, Dict, Optional, Tuple, Type, Union
+from typing import Any, Callable, Dict, Optional, Tuple, Type, Union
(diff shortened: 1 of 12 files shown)


devang-ml:
Are you making device an Pass __init__  parameter or _run_for_config parameter?

olive/passes/olive_pass.py
```diff
@@ -315,4 +342,13 @@ def validate_config(cls, v, values):
         return validate_config(v, PassConfigBase, config_class)
 
     def create_pass(self):
-        return Pass.registry[self.type.lower()](self.config, self.disable_search)
+        pass_cls = Pass.registry[self.type.lower()]
+        return create_pass_from_dict(pass_cls, self.config, self.disable_search)
+
+
+def create_pass_from_dict(pass_cls: Type[Pass], config: Dict[str, Any] = None, disable_search=False) -> Pass:
```
jambayk, line 349: can we also accept `str` as type for `pass_cls`? If it is a string, we could do ` Pass.registry[pass_cls.lower()]` to get the actual class. This might make it easier for users since they don't have to import the class.

guotuofeng: I'd like to move the logic to lookup type from string out of this function to make sure it only do one thing.

jambayk: I see. That sounds good to me. Thanks! 

This will only be used by olive workflow or advanced user, so my request is not that important for user experience.

guotuofeng:
> 

I'd like to make the device as generate search space. Once the space is generated, the pass itself should be device agnostic.

guotuofeng:
> Are you making device an Pass **init** parameter or _run_for_config parameter?
I'd like to make the device as generate search space. Once the space is generated, the pass itself should be device agnostic.

olive/passes/olive_pass.py
```diff
@@ -43,25 +43,37 @@ class Pass(AutoConfigClass):
...
-        self._config_class = self.get_config_class(disable_search)
-        self._config = validate_config(config, PassConfigBase, self._config_class)
-        self._config = self._config.dict()
-        self._config = self._resolve_defaults(self._config)
+    @classmethod
+    def __init_subclass__(cls, **kwargs) -> None:
+        """Register the metric."""
+        super().__init_subclass__(**kwargs)
+        if not inspect.isabstract(cls):
+            name = cls.name if cls.name is not None else cls.__name__.lower()
```
jambayk, line 51: `name` was a nullable class attribute for `AutoConfigClass`. We could just use `cls.__name__.lower()` here directly.

jambayk: That is also the actual source of the failure here
image

guotuofeng: yeah, I already realized it and change to use cls.__name__ instead.

trajepl (approved):

PR #190, opened 2023-04-21
add CredScan

leqiao-1:
Since package publish pipeline might be queued seperately, also add CredScan in package publish pipeline


.azure_pipelines/job_templates/olive-build-template.yaml
```diff
@@ -99,6 +99,12 @@ jobs:
       alertWarningLevel: 'High'
     displayName: Component Detection
 
+  - task: CredScan@3
```
jambayk, line 102: not a big issue but can we add the cred scan job under the test_olive job above? instead of test_examples job?

leqiao-1: Moved after test_olive job

jambayk (approved):

PR #191, opened 2023-04-21
Add PythonEnvironmentSystem and MobileNet QNN EP example

jambayk:
This PR adds a new type of system called `PythonEnvironmentSystem`. There are some cases where Olive is not compatible with the python environment where the desired `onnxruntime` package is installed. For instance, the `onnxruntime-qnn` package must be installed on ARM64 python environment but `Olive` is not compatible with ARM64 Windows python. This new environment lets us run Olive on an x86 python environment while doing model inference and evaluation on a different python environment. 

`PythonEnvironmentSystem` currently only supports `accuracy` and `latency` evaluation for `ONNXModel`s. 

We also add an example for preparing a MobileNet model for QNN EP using QDQ quantization.


examples/mobilenet_qnn_qualcomm_npu/README.md
```diff
@@ -0,0 +1,46 @@
...
+### Create an arm64 python environment with onnxruntime-qnn
+Olive is not supported on arm64 version of python so we require the onnxruntime-qnn package to be installed in a separate arm64 environment. This environment can be a [python virtual env](https://docs.python.org/3/library/venv.html).
+
+Set the path to the directory with arm64 python executable as environment variable `QNN_ENV_PATH`
+
+For example if you used a python venv located at `C:\qnn\qnn-venv`, then `QNN_ENV_PATH` is `C:\qnn\qnn-venv\Scripts`
+
+### Prepare workflow config json
+```
+python prepare_config.json
```
nbcsm, line 20: prepare_config.**py**

olive/model.py
```diff
@@ -247,6 +219,45 @@ def get_execution_providers(self, device: Device):
 
         return eps if eps else available_providers
 
+    def get_io_config(self):
+        if self.io_config:
```
trajepl, line 223: Could you help guide me to the usage of io_config or give a short introduction about this new concept in OnnxModel?

olive/systems/python_environment/available_eps.py
```diff
@@ -0,0 +1,17 @@
...
+
+import onnxruntime as ort
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--output_path", type=str, required=True)
+    args = parser.parse_args()
+
+    # get available execution providers
+    available_eps = ort.get_available_providers()
```
trajepl, line 13: What will happed if the setup is not ready for available_providers? In a few cases, the `ort.get_available_providers()` will return tensorrt, cuda and cpu eps, but the local env tensorrt is not ready. When it comes to inferences settings creation, the program crashes.

examples/mobilenet_qnn_qualcomm_npu/README.md
jambayk: Thanks for catching the type. Updated.

olive/model.py
jambayk: Added a docstring for the method.

olive/systems/python_environment/available_eps.py
jambayk: I have updated the logic to check if EP is valid for the model like it is done in `ONNXModel`. Does that solve this issue?

olive/model.py
```diff
@@ -247,6 +219,53 @@ def get_execution_providers(self, device: Device):
 
         return eps if eps else available_providers
 
+    def get_io_config(self):
```
guotuofeng, line 222: >get_io_config [](http://example.com/codeflow?start=8&length=13)

should the logic be part of load_model and let prepre_session use the loaded model?

olive/model.py
```diff
@@ -163,47 +166,16 @@ def load_model(self) -> onnx.ModelProto:
         return onnx.load(self.model_path)
```
guotuofeng, line 166: >return onnx.load(self.model_path) [](http://example.com/codeflow?start=8&length=33)

two things:
1. should we cache the loaded model like PyTorch model does?
2. what about the external data scenarios?

olive/systems/python_environment/python_environment_system.py
```diff
@@ -0,0 +1,261 @@
...
+from olive.model import OliveModel, ONNXModel
+from olive.passes.olive_pass import Pass
+from olive.systems.common import Device, SystemType
+from olive.systems.olive_system import OliveSystem
+from olive.systems.system_config import PythonEnvironmentTargetUserConfig
+
+logger = logging.getLogger(__name__)
+
+
+class PythonEnvironmentSystem(OliveSystem):
```
guotuofeng, line 28: >PythonEnvironmentSystem [](http://example.com/codeflow?start=6&length=23)

should VirtualEnv be clear?

olive/systems/python_environment/python_environment_system.py
```diff
@@ -0,0 +1,261 @@
...
+        self.environ = deepcopy(os.environ)
+        if self.config.environment_variables:
+            self.environ.update(self.config.environment_variables)
+        if self.config.prepend_to_path:
+            self.environ["PATH"] = os.pathsep.join(self.config.prepend_to_path) + os.pathsep + self.environ["PATH"]
+        self.environ["PATH"] = str(self.config.python_environment_path) + os.pathsep + self.environ["PATH"]
+
+        # get available EPs
+        self.available_eps = self.get_available_eps()
+
```
guotuofeng, line 54: could we run the get available eps only when user not provide any EPs to save time?

olive/systems/python_environment/python_environment_system.py
jambayk: Technically this could be a conda environment. Or just any environment that has a python executable.

olive/model.py
jambayk: As discussed offline, `get_io_config` now calls `onnx.load` directly with `use_external_data=False`. 
For `model.load_model`, we can think about this later if needed.

olive/model.py
jambayk: Updated as per our discussion offline

olive/common/ort_inference.py
```diff
@@ -0,0 +1,54 @@
...
+    if isinstance(execution_provider, list):
+        # execution_provider may be a list of tuples where the first item in each tuple is the EP name
+        execution_provider = [i[0] if isinstance(i, tuple) else i for i in execution_provider]
+    elif isinstance(execution_provider, str):
+        execution_provider = [execution_provider]
+
+    for idx, ep in enumerate(execution_provider):
+        if ep == "QNNExecutionProvider":
+            # add backend_path for QNNExecutionProvider
+            execution_provider[idx] = ("QNNExecutionProvider", {"backend_path": "QnnHtp.dll"})
```
guotuofeng, line 46: >execution_provider [](http://example.com/codeflow?start=12&length=18)

do we need break the loop?

jambayk: We could do it. Doesn't really affect anything.

olive/systems/python_environment/python_environment_system.py
jambayk: Done. Lazy loading of get_available_eps. It caches the values when it is run for the first time.

olive/common/ort_inference.py
jambayk: Done.

guotuofeng (dismissed):

guotuofeng (approved):

PR #193, opened 2023-04-23
io bind interface refinement

trajepl:
Previous `io_bind` for perf_tuning must be set as `[True, False]`, then it will be tunned. 
In this PR:
1. we changed the io_bind to bool value where `True`(default) means to tune the best value for io_bind. `False` means not to tune and disable `io_bind` even for CUDAEps.
2. Packaging module bugfix.
3. Typos fix


olive/passes/onnx/perf_tuning.py
```diff
@@ -249,9 +256,9 @@ def _default_config() -> Dict[str, PassConfigParam]:
                 type_=int, default_value=None, description="CPU cores used for thread tuning."
             ),
             "io_bind": PassConfigParam(
-                type_=Union[bool, List[bool]],
-                default_value=False,
-                description="Whether enable IOBinding for ONNX Runtime inference.",
+                type_=bool,
+                default_value=True,
```
leqiao-1, line 260: It might be better to set default value to False. We should only return tuning result with io_binding to the user who actively choice to use io_binding. Else the tuning result with io binding may confuse the user.

trajepl: Thanks for your comment. What I were thinking is to auto-tune io_bind for users. But as you said, most of the users may not have a basic understanding for the usage of io_bind. 

I will update it as false in this PR. 

But if we can pack the io_bind up to our new Packaging module which can hep conduct the executable folders for users?  @xiaoyu-work 

If that is workable, I may suggest to set the default value as True later.

leqiao-1 (approved):

PR #196, opened 2023-04-24
Introducing DistributedOnnxModel to support distributed inferencing

shaahji:
Introducing DistributedOnnxModel to support distributed inferencing

A subclass of OliveModel to manage onnx graphs for multiple ranks (to be used with mpi4py). Pull some of the code from ONNXModel into static/class methods to allow sharing when loading ranked model. Also, implemented special handling for distributed models in pass logic.


devang-ml (approved):

olive/model.py
```diff
@@ -103,6 +110,51 @@ def validate_type(cls, v):
     def create_model(self):
         return REGISTRY[self.type.lower()](**self.config)
 
+def _select_execution_providers(model: OliveModel, inference_settings: Dict[str, Any], device: Device) -> List[str]:
```
trajepl, line 113: eps is only for OnnxModel. It would be better to use `model: ONNXModel`.

olive/model.py
```diff
@@ -58,21 +58,29 @@ def __init__(
...
+    def prepare_session(self, inference_settings: Optional[Dict[str, Any]] = None,
+                        device: Device = Device.CPU, rank: int = None):
         """
         Prepare inference session for Olive model, return in-memory inference session.
         Derived class should implement its specific logic if needed.
         """
         raise NotImplementedError()
 
+    @abstractmethod
+    def get_default_execution_providers(self, device: Device):
```
trajepl, line 78: `execution_providers` is not suitable for torch model. Should be put it under this base class?

trajepl: Also, it leads a failure as there is no implementation in TorchModel for this `abstractmethod`.

olive/model.py
```diff
@@ -444,3 +467,78 @@ def huggingface_model_loader(model_loader):
...
+
+    def load_model(self, rank: int) -> ONNXModel:
+        return ONNXModel(self.model_filepaths[rank], inference_settings=self.inference_settings)
+
+    def prepare_session(self, inference_settings: Optional[Dict[str, Any]] = None,
+                        device: Device = Device.GPU, rank: int = 0):
+        import onnxruntime as ort
+
+        ort_inference_settings = inference_settings or self.inference_settings
+        ort_session_options = _to_ort_session_options(ort_inference_settings)
```
trajepl, line 508: How about make DistributedOnnxModel inherit ONNXModel but no OliveModel?

olive/passes/olive_pass.py
```diff
@@ -271,6 +271,15 @@ def run(self, model: OliveModel, output_model_path: str, point: Optional[Dict[st
             self._initialize()
             self._initialized = True
 
+        if isinstance(model, DistributedOnnxModel):
+            output_filepaths = []
+            for rank in range(0, model.ranks):
```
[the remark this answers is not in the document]
shaahji: Done!

olive/model.py
shaahji: Introduced ONNModelBase that works as a base class for both ONNXModel & DistributedOnnxModel

olive/model.py
shaahji: Reworked logic with the newly introduced ONNXModelBase

olive/model.py
shaahji: That doesn't work well because model_path arg becomes ambiguous. Which model with the collection is the ONNXModel.model_path referring to.  Introduced ONNModelBase that works as a base class for both ONNXModel & DistributedOnnxModel.

devang-ml (approved):

PR #197, opened 2023-04-25
Remove direct use of pass.default_config()

jambayk:
Since, we will be updating `pass.default_config` to take in the accelerator specs to generate the search space, it is not available inside the pass itself or before we know the accelerato specs. 

This PR removes such references by storing the instance of `default_config` that created the specific pass object as a private attribute of the `config_class`. 

Also added TODO for places where we have to update calls to `default_config` by providing the relevant accelerator specs.

### DIFF
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -76,6 +76,7 @@ class Pass(ABC):
         for param, param_config in self._config_class._default_config.items():
             if param_config.is_path:
                 self.path_params.append((param, param_config.required))
+
         self._initialized = False
 
     @classmethod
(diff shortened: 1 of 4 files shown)


olive/passes/olive_pass.py
```diff
@@ -350,29 +349,14 @@ def to_json(self, check_objects: bool = False) -> Dict[str, Any]:
...
+    config: Dict[str, Any] = None
 
     @validator("type")
     def validate_type(cls, v):
         if v.lower() not in Pass.registry:
             raise ValueError(f"Unknown pass type {v}")
         return v
 
-    @validator("config", pre=True, always=True)
-    def validate_config(cls, v, values):
```
guotuofeng, line 362: >validate_config [](http://example.com/codeflow?start=8&length=15)

don't we need validate it anymore?

guotuofeng (approved):

PR #198, opened 2023-04-25
Introduce CompositeOnnxModel

devang-ml:
Introduce CompositeOnnxModel to represent models with encoder and decoder subcomponents as individual OnnxModels.

Separate PRs will introduce support for creating CompositeOnnxModel and converting CompositeOnnxModel into a OnnxModel

### DIFF
--- a/olive/model.py
+++ b/olive/model.py
@@ -78,7 +78,7 @@ class OliveModel(ABC):
         """
         raise NotImplementedError()
 
-    def set_compsite_parent(self, cp):
+    def set_composite_parent(self, cp):
         self.composite_parent = cp
 
     def get_composite_parent(self):
--- a/olive/model.py
+++ b/olive/model.py
@@ -58,6 +58,7 @@ class OliveModel(ABC):
         self.name = name
         self.is_file = is_file
         self.is_aml_model = is_aml_model
+        self.composite_parent = None
 
     @abstractmethod
     def load_model(self, rank: int = None) -> object:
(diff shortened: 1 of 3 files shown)


olive/model.py
```diff
@@ -77,6 +78,12 @@ def prepare_session(
         """
         raise NotImplementedError()
 
+    def set_compsite_parent(self, cp):
```
jambayk, line 81: nit: typo

olive/passes/olive_pass.py
```diff
@@ -336,6 +336,11 @@ def run(self, model: OliveModel, output_model_path: str, point: Optional[Dict[st
             return DistributedOnnxModel(
                 output_filepaths, model.name, version=model.version, inference_settings=model.inference_settings
             )
+        elif isinstance(model, CompositeOnnxModel):
+            components = []
+            for cidx, child in enumerate(model.get_model_components()):
+                components.append(self._run_for_config(child, config, str(f"{output_model_path}_{cidx}")))
```
jambayk, line 342: Not related to this PR but looks like the distributed onnx model implementation missed appending the rank to the `output_model_path`. It is just overwriting the same path with the subsequent ranks.

devang-ml: Yes, good catch. We can address that separately.

olive/model.py
devang-ml: yup. Fixed.

jambayk (dismissed):

jambayk (dismissed):

jambayk (dismissed):

trajepl:
nit: please also add documents to introduce this new changes, separated PR is ok I think.

trajepl (approved):
Merge and resolve conflicts.

jambayk (approved):

PR #199, opened 2023-04-26
Add io_config, dummy_inputs_func and dynamic axes to PytorchModel

jambayk:
Currently, model specific information such as `input_names`, `input_shapes`, etc. are provided as config parameters for passes like `ONNXConversion`. We want to support composite models which are made of multiple models. For such cases, we cannot provide the io information to a pass for all at once. 

So, we make the io information optional init parameters of PytorchModel. The passes will in turn get them from the model itself. This also makes more sense conceptually since these are attributes of the model and not the pass. 

We will make similar changes to the ONNXModel and onnx related passes in the future. Any model specific information should be part of the model and not the pass config.

### DIFF
--- a/examples/bert_ptq_cpu/docker/Dockerfile
+++ b/examples/bert_ptq_cpu/docker/Dockerfile
@@ -6,6 +6,6 @@ RUN pip install azure-ai-ml \
             datasets \
             transformers \
             onnxconverter_common \
-            git+https://github.com/microsoft/Olive.git@myguo/model_type
+            git+https://github.com/microsoft/Olive.git@jambayk/model-info
 
 WORKDIR /olive
--- a/olive/model.py
+++ b/olive/model.py
@@ -349,7 +349,7 @@ class PyTorchModel(OliveModel):
         model_script: Union[str, Path] = None,
         script_dir: Union[str, Path] = None,
         io_config: Union[Dict[str, Any], IOConfig] = None,
-        dummy_input_func: Union[str, Callable] = None,
+        dummy_inputs_func: Union[str, Callable] = None,
         dynamic_axes: Dict[str, Dict[int, str]] = None,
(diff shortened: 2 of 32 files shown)


olive/model.py
```diff
@@ -362,13 +415,49 @@ def load_model(self, rank: int = None) -> torch.nn.Module:
     def prepare_session(self, inference_settings: Dict[str, Any], device: Device, rank: int = None):
         return self.load_model().eval()
 
+    def get_dummy_inputs(self):
```
guotuofeng, line 418: these logic seems not be related to PyTorch model. Instead, it should be part of dataset.

guotuofeng (changes requested):
suggest to move the dummy data generation to dummy dataset like https://github.com/microsoft/Olive/blob/main/olive/evaluator/evaluation.py#L354

olive/model.py
jambayk: It is part of pytorch model. When converting to onnx or during qat, we need dummy input.

jambayk: Can you please unblock this PR? This is urgent for whisper model support.

jambayk: Added TODO as discussed offline.

examples/bert_ptq_cpu/docker/Dockerfile
```diff
@@ -6,6 +6,6 @@ RUN pip install azure-ai-ml \
             datasets \
             transformers \
             onnxconverter_common \
-            git+https://github.com/microsoft/Olive.git
+            git+https://github.com/microsoft/Olive.git@jambayk/model-info
```
trajepl, line 9: Please remember to remove it.

trajepl (approved):

guotuofeng (approved):

PR #205, opened 2023-04-26
Introduce InsertBeamSearch Pass

devang-ml:
`InsertBeamSearchPass` chains two model components (for example, encoder and decoder) together by inserting beam search op in between them.


guotuofeng:
>             is_aml_model=False,

change it to model_storage_kind. Otherwise, it would break.

---
Refers to: olive/model.py:757 in e02b2cf. [](commit_id = e02b2cf8c2afc48f9e2336844378bb30b9821d99, deletion_comment = False)

olive/passes/onnx/insert_beam_search.py
```diff
@@ -0,0 +1,141 @@
...
+            "num_return_sequences",
+            "length_penalty",
+            "repetition_penalty",
+            "",
+            "",
+            "attention_mask",
+        ]
+        beam_outputs = ["sequences"]
+
+        node = helper.make_node("BeamSearch", inputs=beam_inputs, outputs=beam_outputs, name="BeamSearch_zcode")
```
nbcsm, line 49: better to rename value of `name`

devang-ml: done

xiaoyu-work (approved):

PR #206, opened 2023-04-26
Add OrtMixedPrecision pass

xiaoyu-work:
1. Add OrtMixedPrecision pass
2. Add OrtMixedPrecision pass unit tests


olive/passes/onnx/mixed_precision.py
```diff
@@ -0,0 +1,145 @@
...
+    """Convert model to mixed precision."""
+
+    _requires_user_script = True
+
+    @staticmethod
+    def _default_config() -> Dict[str, PassConfigParam]:
+        config = {
+            "op_block_list": PassConfigParam(
+                type_=List[str],
+                default_value=["SimplifiedLayerNormalization", "SkipSimplifiedLayerNormalization", "Relu", "Add"],
```
jambayk, line 22: The default value for gpt2 is different. For the gpt2 case, do we expect the user config to pass the correct one? https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/models/gpt2/gpt2_helper.py#L549

xiaoyu-work: I’m using the default value for whisper model: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/models/whisper/whisper_helper.py#L152. I remember that we had a discussion before that Olive can support different common models, maybe when we have different model cases like whisper gpt2 hf we can have different default values for them.

jambayk: Sounds good to me! We can worry about the other models when it comes up. 

I saw somewhere where the default is something like "auto". if "auto", then we can look at the model type to create the default list. But it is not needed for now.

olive/model.py
```diff
@@ -234,6 +237,52 @@ def prepare_session(self, inference_settings: Dict[str, Any], device: Device, ra
...
+        for graph in self.graphs():
+            for node in graph.node:
+                all_nodes.append(node)
+        return all_nodes
+
+    def graph(self):
+        return self.load_model().graph
+
+    def graphs(self):
+        from onnx import AttributeProto, GraphProto
```
guotuofeng, line 251: >from onnx import AttributeProto, GraphProto [](http://example.com/codeflow?start=8&length=43)

could we move the onnx import to the top of the file it seems the onnx module is already imported.

olive/model.py
```diff
@@ -188,6 +188,8 @@ def __init__(
         model_storage_kind: Union[str, ModelStorageKind] = ModelStorageKind.LocalFile,
         inference_settings: Optional[dict] = None,
     ):
+        from onnx import GraphProto
```
guotuofeng, line 191: >GraphProto [](http://example.com/codeflow?start=25&length=10)

ditto

olive/model.py
```diff
@@ -234,6 +237,52 @@ def prepare_session(self, inference_settings: Dict[str, Any], device: Device, ra
...
+    def graph(self):
+        return self.load_model().graph
+
+    def graphs(self):
+        from onnx import AttributeProto, GraphProto
+
+        if self.all_graphs is not None:
+            return self.all_graphs
+        self.all_graphs = []
+        graph_queue = [self.load_model().graph]
```
guotuofeng, line 256: >self.load_model().graph [](http://example.com/codeflow?start=23&length=23)

better to use the self.graph()?

olive/model.py
```diff
@@ -234,6 +237,52 @@ def prepare_session(self, inference_settings: Dict[str, Any], device: Device, ra
...
 
+    def nodes(self):
+        all_nodes = []
+        for graph in self.graphs():
+            for node in graph.node:
+                all_nodes.append(node)
+        return all_nodes
+
+    def graph(self):
+        return self.load_model().graph
```
guotuofeng, line 248: >self.load_model().graph [](http://example.com/codeflow?start=15&length=23)

do we need to cache the graph to avoid loading the model multiple times when it is called multiple times.

olive/model.py
xiaoyu-work: Ah I didn't notice that. I'll move it to the top.

olive/model.py
xiaoyu-work: Yeah I'll cache it.

olive/model.py
xiaoyu-work: Yeah good catch! Will update it.

olive/model.py
```diff
@@ -290,6 +294,53 @@ def prepare_session(self, inference_settings: Dict[str, Any], device: Device, ra
 
         return get_ort_inference_session(self.model_path, inference_settings)
 
+    def nodes(self):
+        all_nodes = []
+        for graph in self.get_all_graphs():
+            for node in graph.node:
+                all_nodes.append(node)
+        return all_nodes
```
guotuofeng, line 302: yield better than list?
[https://stackoverflow.com/questions/7883962/where-to-use-yield-in-python-best](https://stackoverflow.com/questions/7883962/where-to-use-yield-in-python-best)
[https://www.techbeamers.com/when-prefer-yield-over-return-python/](https://www.techbeamers.com/when-prefer-yield-over-return-python/)

xiaoyu-work: Yeah since we only have iteration use case, I can make a change on this. This part of code is from onnxruntime https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/onnx_model.py#L72

devang-ml (approved):

PR #209, opened 2023-04-27
Add `hf_config` to `PyTorchModel` and whisper example 

jambayk:
## Describe your changes
Add `hf_config` option to PyTorchModel. With `hf_config`, the pytorch model can be specified through a combination of `(task, model_name` or `(model_class,model_name)`.  
If using `model_class`, user can also choose to use the implementation from onnxruntime if available. 
Currently, only whisper is supported. The onnxruntime implementation is needed for creating a final model with beam search. 

`hf_config` also has the option to specify a list of `components`. If given, the underlying pytorch model is split into its components and each one can be created into its own `PyTorchModel` using `olive_pytorch_model.get_component`. 
If the model has components, the `OnnxConversion` pass converts each component into its own ONNXModel and returns a `CompositeONNXModel`. The `CompositeONNXModel` is eventually converted to a single model by the beam search pass. 

Also added whisper example. It needs the latest onnxruntime code which is not in the latest nightly builds yet. It is a WIP. 


examples/whisper/user_script.py
```diff
@@ -0,0 +1,28 @@
+from onnxruntime.transformers.models.whisper.whisper_decoder import WhisperDecoderInputs
+from onnxruntime.transformers.models.whisper.whisper_encoder_decoder_init import WhisperEncoderDecoderInitInputs
+
+
+def encoder_decoder_init_dummy_inputs(model):
+    model = model.load_model()
```
devang-ml, line 6: Could we move this logic inside PyTorchModel's method to get dummy inputs?

examples/whisper/user_script.py
```diff
@@ -0,0 +1,28 @@
...
+        encode_sequence_length=3000,
+        use_decoder_input_ids=True,
+        device="cpu",
+        use_int32_inputs=True,
+    )
+    return tuple(inputs.to_list())
+
+
+def decoder_dummy_inputs(model):
+    model = model.load_model()
```
devang-ml, line 19: same

olive/hf_utils.py
```diff
@@ -0,0 +1,98 @@
...
+            model = model_class.from_pretrained(name)
+            return model
+        except (OSError, ValueError):
+            continue
+
+    return model
+
+
+class ORTWhisperModel(torch.nn.Module):
+    """ORT implementation of whisper model"""
```
devang-ml, line 36: I think moving this into a separate file under new "models" directory would make sense. We may have to offer custom nn.Modules in future also.

olive/hf_utils.py
```diff
@@ -0,0 +1,98 @@
...
+        except AttributeError:
+            raise AttributeError(f"{model_loader} is not found in transformers")
+    elif not isinstance(model_loader, Callable):
+        raise ValueError("model_loader must be a callable or a string defined in transformers")
+
+    return model_loader.from_pretrained
+
+
+def load_huggingface_model_from_model_class(model_class: str, name: str, use_ort_implementation: bool = False):
+    """
```
devang-ml, line 87: What is the reason to make these methods utility instead of member of PyTorch model class?

examples/whisper/user_script.py
jambayk: This is specific for the whisper model so it is not possible to add it to PyTorch model.

jambayk: In the future, when we have automatic dummy input generation for hf models, this won't be necessary anymore. But that is a big feature implementation and cannot be done in time.

olive/hf_utils.py
jambayk: I think we will have a lot of other helper functions like this as we expand support for huggingface models so I moved them to a separate utility file.

olive/model.py
```diff
@@ -750,11 +811,11 @@ def __init__(
         version: Optional[int] = None,
     ):
         super().__init__(
+            framework=Framework.ONNX,
+            model_file_format=ModelFileFormat.ONNX,
             model_path=None,
             name=name,
             version=version,
-            is_file=False,
-            is_aml_model=False,
```
guotuofeng, line 757: specify the model_storage_kind to folder?

olive/model.py
```diff
@@ -750,11 +811,11 @@ def __init__(
         version: Optional[int] = None,
     ):
         super().__init__(
+            framework=Framework.ONNX,
+            model_file_format=ModelFileFormat.ONNX,
```
guotuofeng, line 815: >model_file_format [](http://example.com/codeflow?start=12&length=17)

I am supposing the aml system doesn't support the composite model. right?

docs/source/overview/options.md
```diff
@@ -40,6 +40,15 @@ case insensitive.
 
     - `model_script: [str]` The name of the script provided by the user to assist with model loading.
 
+    - `hf_config: [Dict]` Instead of `model_path` or `model_loader`, the model can be specified using a dictionary describing a huggingface
+    model. This dictionary specifies the following items:
+
+        - `model_name: [str]`: This the model name of the huggingface model such as `distilbert-base-uncased`.
+
+        - `task: [str]`: This is the task type for the model such as `text-classification`.
```
guotuofeng, line 48: better give a link about the supported task list.

olive/model.py
```diff
@@ -750,11 +811,11 @@ def __init__(
         version: Optional[int] = None,
     ):
         super().__init__(
+            framework=Framework.ONNX,
```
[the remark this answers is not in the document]
jambayk: Done.

olive/model.py
jambayk: AML does support composite model right now but it can be added later. since the composite model is just a list of ONNXModels, we should be able to reuse the logic for ONNXModels.

olive/model.py
jambayk: Done. It doesn't really matter since this is a dummy value. The actual values are the storage_kind of the component.

olive/model.py
jambayk: *does not (sorry about the typo_

olive/hf_utils.py
```diff
@@ -0,0 +1,102 @@
...
+        model,
+        None,
+        model.config,
+        decoder_start_token_id=None,
+    )
+
+    return ORTWhisperModel(encoder_decoder_init, decoder, model.config)
+
+
+MODEL_CLASS_TO_ORT_IMPLEMENTATION = {"WhisperForConditionalGeneration": get_ort_whisper_for_conditional_generation}
```
trajepl, line 71: I would suggest the key should be task type in huggingface pipeline. see `load_huggingface_model_from_task`.
The workflow should be like:
1. provide `task_name` & `model_name`(hf model name)
2. get model loader from `load_huggingface_model_from_task` like `WhisperForConditionalGeneration` used in line 53.
3. use the model loader to load model.
4. For specific model, you can apply different type function for ort loading with different task types.
Currently, the key for this map containers Model name + task type which will increase the map size a lot in the future.

But if you think the changes will increase the dev costs, please add TODO for future refactoring.

trajepl: Oh, I see your implementation for task load.

jambayk: I was thinking about something similar too. `model_class` doesn't seem like a sustainable way of doing it. Especially when we will probably need the hf_config for dataset creation in the future too. 
I will add a TODO for refactoring since I don't think I can do it in time for this PR which is needed for other work items (beam search, etc)

PR #224, opened 2023-05-02
Differentiate SD fusions for DML EP based on version

jstoecker:
## Describe your changes

DML in ORT 1.15.0 supports a subset of the transformer fusions, which will be expanded in ORT 1.15.1 to include MHA. This change makes the fusion options more explicit for DML EP based on ORT version. Some fusion options are supported but disabled anyway since it limits the number of contrib ops that don't provide a direct perf benefit for DML EP.

Note that we still need DML 1.11.0 to be published and used in ORT 1.15.0 before nightly ORT-DML python packages can be used in Olive. I'll convert this draft PR to a full PR once we have a nightly build to consume.


olive/passes/onnx/transformer_optimization.py
```diff
@@ -94,30 +94,62 @@ def sd_model_types():
...
+        dml_future = pass_config["target_provider"] == "directml_future"
+
+        if pass_config["target_provider"] == "directml" or dml_future:
+            # Some of these fusions are disabled because they provide no performance advantage,
+            # and it's preferable to limit ops outside the ONNX domain.
+            fusion_options.enable_gelu = is_ort_1_15_0_or_newer
+            fusion_options.enable_layer_norm = is_ort_1_15_0_or_newer
+            fusion_options.enable_attention = is_ort_1_15_1_or_newer or dml_future
+            fusion_options.use_multi_head_attention = is_ort_1_15_1_or_newer or dml_future
+            fusion_options.enable_skip_layer_norm = is_ort_1_15_0_or_newer
```
fdwr, line 134: Propose disabling since it buys no advantage for DML (per conversation with Pat/Justin) and just complicates mapping to WebNN.

fdwr: Unless Pat says otherwise, I'd apply the same to `BiasSkipLayerNorm` too. I don't even see a registration for it in OperatorRegistration.

olive/passes/onnx/transformer_optimization.py
```diff
@@ -94,30 +94,62 @@ def sd_model_types():
...
+            fusion_options.enable_embed_layer_norm = is_ort_1_15_0_or_newer
+            fusion_options.enable_bias_skip_layer_norm = is_ort_1_15_0_or_newer
+            fusion_options.enable_bias_gelu = is_ort_1_15_0_or_newer
```
fdwr, line 137: BiasGelu is an odd one, and quite specific, but it does add a *small* perf benefit for DML (using the ADD1 plus fused activation GELU). So I guess I'd keep it. I'll decompose it in my convertor to WebNN.

olive/passes/onnx/transformer_optimization.py
fdwr: For `EmbedNormalization` (nice ASCII art in the ORT EP btw), I wonder if this is another case where *if* the model already has one, then we need a functional decomposition for it (just so ORT doesn't fall back to CPU because CPU registered the contrib op even though DML was registered as a higher priority EP), but otherwise we don't really get benefit from it 🤔...

olive/passes/onnx/transformer_optimization.py
```diff
@@ -94,30 +94,62 @@ def sd_model_types():
...
+            fusion_options.enable_gelu_approximation = False
+            fusion_options.enable_qordered_matmul = is_ort_1_15_0_or_newer
```
fdwr, line 139: `QOrderedMatMul` I don't see any registrations for it (in the non-Pat branches anyway), and so propose setting false. I wouldn't substitute to any operators that we don't actually have DML registrations for in OperatorRegistration.cpp, because otherwise we *could* get *worse* perf if the CPU registered that contrib op but DML did not.

Also as a general compat/interchange principle, I'd favor emitting as close to standard ONNX as possible, *except* in cases where we know perf is improved by it. The higher the contrib op ratio in a model, the less we can rationalize about expected behavior (the contrib ops are not as well documented, do not have the same breadth/history of testing, and are not opset versioned), and the fewer user tools which can operate with them.

fdwr (commented):
Thanks for updating/sharing the CR and conversing with me about WebNN today.

jambayk (approved):
Thanks for the changes. 
This is not part of the PR but I am wondering about what the best way to address the ort version specific settings would be. The passes run on the `host` which is not always the same as the `target`. So, there could be a mismatch between the version used for during optimization and that during deployment. 
I think we will have to consider it when Mike updates the system to include EP and hardware information. @guotuofeng

examples/directml/stable_diffusion/stable_diffusion.py
```diff
@@ -109,13 +130,17 @@ def optimize(model_name: str, unoptimized_model_dir: Path, optimized_model_dir:
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--optimize", action="store_true", help="Runs the optimization step")
+    parser.add_argument("--optimize_provider", type=str, default="directml", help="EP target for inference")
```
fdwr, line 133: (minor naming) `--optimization_target`? `--target`? `--target-backend`? `--target-provider`? I get that "EP" is used in the context of ORT, but I wonder if "target" or "backend" make more sense generally, and the verb "optimize_provider" sounds like it's optimizing a *provider*, rather than optimizing a *model* for a given target backend.

olive/passes/onnx/transformer_optimization.py
```diff
@@ -94,30 +94,62 @@ def sd_model_types():
...
     @staticmethod
-    def _set_sd_fusion_options(run_config: Dict[str, Any], use_float16: bool):
+    def _set_sd_fusion_options(run_config: Dict[str, Any], pass_config: Dict[str, Any]):
         """Configures fusion options for stable diffusion models"""
         import onnxruntime as ort
 
-        input_model_type = run_config["model_type"]
+        ort_version = version.parse(ort.__version__)
+        is_ort_1_13_or_older = ort_version < version.parse("1.14.0")
+        # is_ort_1_14 = ort_version >= version.parse("1.14.0") and ort_version < version.parse("1.15.0")
```
fdwr, line 103: stale deletable comment?

olive/passes/onnx/transformer_optimization.py
```diff
@@ -94,30 +94,62 @@ def sd_model_types():
...
-            fusion_options.enable_packed_qkv = use_float16 and is_ort_115_or_newer
-            fusion_options.enable_bias_add = is_ort_115_or_newer
+        # TODO: remove dml_future when ORT 1.15.1 with future version of DML is released
+        # This "provider" value is simply a way to test in-development changes without having
+        # to bump the ORT version.
+        dml_future = pass_config["target_provider"] == "directml_future"
+
+        if pass_config["target_provider"] == "directml" or dml_future:
+            # Some of these fusions are disabled because they provide no performance advantage,
+            # and it's preferable to limit ops outside the ONNX domain.
```
fdwr, line 129: Thanks.

olive/passes/onnx/transformer_optimization.py
```diff
@@ -94,30 +94,62 @@ def sd_model_types():
...
+            fusion_options.enable_gelu = is_ort_1_15_0_or_newer
+            fusion_options.enable_layer_norm = is_ort_1_15_0_or_newer
+            fusion_options.enable_attention = is_ort_1_15_1_or_newer or dml_future
+            fusion_options.use_multi_head_attention = is_ort_1_15_1_or_newer or dml_future
+            fusion_options.enable_skip_layer_norm = False
+            fusion_options.enable_embed_layer_norm = is_ort_1_15_0_or_newer
+            fusion_options.enable_bias_skip_layer_norm = False
+            fusion_options.enable_bias_gelu = is_ort_1_15_0_or_newer
+            fusion_options.enable_gelu_approximation = False
```
fdwr, line 138: Do you know what "gelu approximation" emits vs normal Gelu? I doubt it's anything that affects me - just curious.

olive/passes/onnx/transformer_optimization.py
```diff
@@ -94,30 +94,62 @@ def sd_model_types():
...
+            fusion_options.enable_group_norm = is_ort_1_15_1_or_newer or dml_future
+            fusion_options.enable_bias_splitgelu = False
+            fusion_options.enable_packed_qkv = pass_config["float16"] and is_ort_1_15_0_or_newer
+            fusion_options.enable_packed_kv = pass_config["float16"] and is_ort_1_15_0_or_newer
+            fusion_options.enable_bias_add = False
+        else:
+            if input_model_type == "unet":
+                fusion_options.enable_packed_kv = pass_config["float16"]
+                fusion_options.enable_packed_qkv = pass_config["float16"] and is_ort_1_15_0_or_newer
+                fusion_options.enable_bias_add = is_ort_1_15_0_or_newer
```
fdwr, line 152: `enable_bias_add` is False above, but true here. 🤔 Confirming that's intended? Do you know if that emits DML ADD1?

fdwr (approved):

olive/passes/onnx/transformer_optimization.py
jstoecker: This is to mirror logic in upstream ORT stable diffusion scripts for non-DML EP.

examples/directml/stable_diffusion/stable_diffusion.py
jstoecker: I intend to remove this parameter when we have ORT 1.15.1, so I'm not too concerned with the name.

olive/passes/onnx/transformer_optimization.py
jstoecker: I believe it's an alternative mathematical formulation (see https://paperswithcode.com/method/gelu) that sacrifices precision for computational simplicity.

olive/passes/onnx/transformer_optimization.py
jstoecker: I'll delete this when I remove directml_future to avoid spinning the CI again.

jstoecker:
> Thanks for the changes. This is not part of the PR but I am wondering about what the best way to address the ort version specific settings would be. The passes run on the `host` which is not always the same as the `target`. So, there could be a mismatch between the versions used during optimization and deployment. I think we will have to consider it when Mike updates the system to include EP and hardware information. @guotuofeng

That's a good point! In this case the ORT version has implications for both host and target, since this pass relies on modules in the ORT package. The available fusions are dependent on the host ORT package, and the target ORT version determines which fusions are supported at runtime.

PR #227, opened 2023-05-02
Retry test setup for example tests

jambayk:
## Describe your changes
Sometimes, the example tests fail because the setups (preparing model, data, config, etc) fail randomly. Retry the test setups to catch such errors. 
Test fixtures are also updated to make them simpler and consistent across examples. 

### DIFF
--- a/examples/test_bert_ptq_cpu.py
+++ b/examples/test_bert_ptq_cpu.py
@@ -10,15 +10,11 @@ from pathlib import Path
 import pytest
 
 
-@pytest.fixture()
-def example_dir():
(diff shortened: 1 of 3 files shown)


xiaoyu-work:
Any reason why the tests can fail? Should we fix the failure reasons?

xiaoyu-work (approved):

jambayk:
> Any reason why the tests can fail? Should we fix the failure reasons?

@xiaoyu-work I am not sure why the setup fails. It happens rarely so it is hard to replicate the failure on my local machine. For the whisper windows test, it appears to be with decoding the stderr output of the subprocess call. But it succeeds when we run the subprocess call again.

PR #238, opened 2023-05-04
Correct logger severity mapping

jambayk:
## Describe your changes
`olive.set_default_logger_severity` takes int from 0-4 for severity level. The corresponding levels for `logging` are 10-50, not 0-40. Update the mapping for this. 

### DIFF
--- a/olive/__init__.py
+++ b/olive/__init__.py
@@ -24,4 +24,4 @@ def set_default_logger_severity(level):
     :param level: 0: DEBUG, 1: INFO, 2: WARNING, 3: ERROR, 4: CRITICAL
     """
     # logging.DEBUG = 10, logging.INFO = 20, logging.WARNING = 30, logging.ERROR = 40, logging.CRITICAL = 50
-    _logger.setLevel((1 + level) * 10)
+    _logger.setLevel((level + 1) * 10)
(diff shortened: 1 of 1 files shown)


olive/__init__.py
```diff
@@ -20,6 +20,8 @@
 def set_default_logger_severity(level):
     """
     Set log level for olive package.
+
     :param level: 0: DEBUG, 1: INFO, 2: WARNING, 3: ERROR, 4: CRITICAL
     """
-    _logger.setLevel(level * 10)
+    # logging.DEBUG = 10, logging.INFO = 20, logging.WARNING = 30, logging.ERROR = 40, logging.CRITICAL = 50
+    _logger.setLevel((level + 1) * 10)
```
guotuofeng, line 27: better to maintain a map between the log level of ort and python?

olive/__init__.py
```diff
@@ -20,6 +20,8 @@
 def set_default_logger_severity(level):
```
guotuofeng, line 20: btw, do we really need both log_severity_level and ort_log_severity_level?

trajepl (approved):

olive/__init__.py
jambayk: we want to default to ERROR for ort but INFO for olive so I think separate levels is helpful. Otherwise, ort makes the logs messy with it's warnings about graphs, etc.

olive/__init__.py
jambayk: That would work too. Might be more clearer too. What do you think @trajepl?

guotuofeng (approved):

trajepl (approved):

PR #239, opened 2023-05-04
Skip docker system test for bert example

xiaoyu-work:
## Describe your changes
Skip docker system test for bert example until bug is fixed: https://github.com/docker/docker-py/issues/3113


jambayk (approved):
Thanks! 

I think if the next docker release is going to take time, we might have to do the temporary solution of pinning `urllib3` and `requests` for `docker` extra to make docker system usable.

PR #243, opened 2023-05-05
add tests

leqiao-1:


olive/passes/pytorch/qat_utils.py
```diff
@@ -4,8 +4,10 @@
 # --------------------------------------------------------------------------
 import copy
 
+import pytorch_lightning
 import torch
 import torch.quantization.quantization_mappings as tqqm
+from packaging import version
```
trajepl, line 10: Sometime, there will be bug for packaing.version.  Suggest to use `from distutils.version import LooseVersion`

`if LooseVersion(torch.__version__) >= LooseVersion("1.12.0"):`

jambayk: image
python gives this warning to me

leqiao-1: Before we met this issue with packaging.version
image
I will have a check.

trajepl: In Python 3.10 and 3.11, distutils will be formally marked as deprecated. Currently it is 3.7/3.8 for olive?

jambayk: I got the warning above in python 3.8 (3.8.15).

jambayk: Looking at the docs, the example version you shared above doesn’t follow the PEP standards. So do we expect to see versions like those in practice? 
The version above looks like one created for a custom build of torch?

jambayk: Actually, it looks like the version comparison result above is expected? https://packaging.pypa.io/en/stable/version.html 

The first version is a prerelease (1.11.0a) so it is earlier than the release version (1.11.0).

if we want to get it true for prerelease versions also, we could use ‘version.parse(“1.11.0a0+git”).release’ >= ‘version.parse(“1.11.0”).release’?

leqiao-1: I think we can keep version.parse as before.

jambayk: Yes, I agree. The issue you mentioned is expected behavior so I don’t think it’s really an issue.

leqiao-1: last commit reverted

jambayk (approved):

PR #246, opened 2023-05-05
Change `read_timeout` for `azure.ai.ml.MLClient`

jambayk:
## Describe your changes
The default `read_timeout` value for `MLClient` is `3000` seconds. However, this value is too large and causes `AzureMLSystem` pass runs and model evaluations to hang for a long time during the job streaming and download steps when there are multiple jobs being created at the same time (like in the CI pipelines). We already have retries set up for this step, so it is better to reduce the read_timeout value for Olive. 

With this change the bert and resnet ci tests now only take ~50 minutes instead of ~2 hours it takes when the hanging happens. 

We also allow user to set their own default values if they find the connections are timing out even with retries. I don't expect this to be needed much but it is safer to allow the user to overwrite instead of hardcoding it to `60`. 

### DIFF
--- a/olive/systems/system_config.py
+++ b/olive/systems/system_config.py
@@ -31,6 +31,10 @@ class AzureMLTargetUserConfig(TargetUserConfig):
     aml_compute: str
     aml_docker_config: AzureMLDockerConfig
     instance_count: int = 1
+    # read timeout in seconds for HTTP requests, user can increase if they find the default value too small.
+    # The default value from azureml sdk is 3000 which is too large and cause the evaluations and pass runs to
+    # hang for a long time
+    read_timeout: int = 60
     is_dev: bool = False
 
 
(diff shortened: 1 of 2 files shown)


olive/systems/system_config.py
```diff
@@ -31,6 +31,10 @@ class AzureMLTargetUserConfig(TargetUserConfig):
     aml_compute: str
     aml_docker_config: AzureMLDockerConfig
     instance_count: int = 1
+    # read timeout in seconds for HTTP requests, user can increase if they find the default value too small.
+    # The default value from azureml sdk is 3000 which is too large and cause the evaluations and pass runs to
+    # sometimes hang for a long time between retries of job stream and download steps.
+    read_timeout: int = 60
```
trajepl, line 37: Please add the description into Olive docs if it is exposed to users.

jambayk: We don’t have explanation of any of the system config parameter. So there’s not where to add them right now.

i think when we use pydantic fields like discussed in the team meeting, we can make the comment here into a description. It will get automatically documented them.

trajepl (approved):

PR #248, opened 2023-05-08
Refactor engine to support hardware accelerators

guotuofeng:
## Describe your changes
make engine register the Pass config instead of Pass instance to support hardware accelerators

### DIFF
--- a/docs/source/tutorials/advanced_users.md
+++ b/docs/source/tutorials/advanced_users.md
@@ -121,7 +121,7 @@ quantization_config = {
     "weight_type" : "QUInt8"
 }
 # search over the values for the other config parameters
-engine.register(OnnxQuantization, quantization_pass, False)
+engine.register(OnnxQuantization, quantization_config, False)
 ```
 
 ### Run the engine
--- a/docs/source/tutorials/advanced_users.md
+++ b/docs/source/tutorials/advanced_users.md
@@ -123,8 +123,6 @@ quantization_config = {
 # search over the values for the other config parameters
 engine.register(OnnxQuantization, quantization_pass, False)
 ```
-```
-```
 
 ### Run the engine
 Finally, run the engine on your input model. The output will be the best set of parameters for the passes and the output
--- a/docs/source/tutorials/advanced_users.md
+++ b/docs/source/tutorials/advanced_users.md
@@ -105,15 +105,13 @@ Passes available in Olive can be found at ...
 
 ```python
 from olive.passes import OnnxConversion, OnnxQuantization
-from olive.passes.olive_pass import create_pass_from_dict
 
 # Onnx conversion pass
 onnx_conversion_config = {
     "target_opset": 13,
 }
-onnx_conversion_pass = create_pass_from_dict(OnnxConversion, onnx_conversion_config)
 # override the default host with pass specific host
-engine.register(onnx_conversion_pass, host=LocalSystem())
+engine.register(OnnxConversion, onnx_conversion_config, False, host=LocalSystem())
 
 # onnx quantization pass
 quantization_config = {
@@ -123,8 +121,9 @@ quantization_config = {
     "weight_type" : "QUInt8"
 }
 # search over the values for the other config parameters
-quantization_pass = create_pass_from_dict(OnnxQuantization, quantization_config)
-engine.register(quantization_pass)
+engine.register(OnnxQuantization, quantization_pass, False)
+```
+```
 ```
 
 ### Run the engine
--- a/examples/test_bert_ptq_cpu.py
+++ b/examples/test_bert_ptq_cpu.py
@@ -19,11 +19,11 @@ def setup():
     os.chdir(cur_dir)
 
 
-def check_output(footprint):
-    footprint = footprint[0]
-    assert footprint.nodes is not None
-    for v in footprint.nodes.values():
-        assert all([value > 0 for value in v.metrics.value.values()])
+def check_output(footprints):
+    for footprint in footprints:
+        assert footprint.nodes is not None
+        for v in footprint.nodes.values():
+            assert all([value > 0 for value in v.metrics.value.values()])
 
 
 # Skip docker_system test until bug is fixed: https://github.com/docker/docker-py/issues/3113
--- a/examples/test_bert_ptq_cpu.py
+++ b/examples/test_bert_ptq_cpu.py
@@ -20,6 +20,7 @@ def setup():
 
 
 def check_output(footprint):
+    footprint = footprint[0]
     assert footprint.nodes is not None
     for v in footprint.nodes.values():
         assert all([value > 0 for value in v.metrics.value.values()])
--- a/examples/test_bert_ptq_cpu.py
+++ b/examples/test_bert_ptq_cpu.py
@@ -20,7 +20,7 @@ def setup():
 
 
 def check_output(footprints):
-    for footprint in footprints:
+    for footprint in footprints.values():
         assert footprint.nodes is not None
         for v in footprint.nodes.values():
             assert all([value > 0 for value in v.metrics.value.values()])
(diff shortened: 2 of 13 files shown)


examples/test_bert_ptq_cpu.py
```diff
@@ -20,6 +20,7 @@ def setup():
 
 
 def check_output(footprint):
+    footprint = footprint[0]
```
trajepl, line 23: Could you help explain how the accelerators limit the checking only on first footprint node?

olive/engine/engine.py
```diff
@@ -224,8 +243,68 @@ def run(
             json.dump(results, open(results_path, "w"), indent=4)
             return results
 
+        # TODO by myguo: replace the following for loop using the accelerator and excution provider list when adding
+        # accelerator support
+        footprints = []
```
trajepl, line 248: How about use `dict[accelerator_name, footprints]'?

examples/test_bert_ptq_cpu.py
guotuofeng: will change to loop

olive/engine/engine.py
guotuofeng: will make the change when I add accelerator.

olive/engine/engine.py
```diff
@@ -128,28 +129,51 @@ def initialize(self):
...
             if clean_run_cache:
-                cache_utils.clean_pass_run_cache(p.__class__.__name__, cache_dir)
-
-        # list of passes starting from the first pass with non-empty search space
-        # These passes will be added to the search space
-        self.pass_search_spaces = []
-        for pass_name in self.pass_order:
-            p = self.passes[pass_name]["pass"]
-            self.pass_search_spaces.append((pass_name, p.search_space()))
+                cache_utils.clean_pass_run_cache(pass_name, cache_dir)
```
jambayk, line 135: `pass_name` is not always the same as the name of the pass class. It can be user provided too. It is always user provided for workflow runs. 
Can you make this to `p.__name__`  instead?

olive/engine/engine.py
```diff
@@ -224,8 +241,68 @@ def run(
             json.dump(results, open(results_path, "w"), indent=4)
             return results
 
+        # TODO by myguo: replace the following for loop using the accelerator and excution provider list when adding
+        # accelerator support
+        footprints = []
+        for i in range(1):
```
xiaoyu-work, line 247: Please correct me if my understand is not correct: 

the overall idea is the engine will run all same passes with different hardware accelerators. 
Do we have (or will have) any pass that requires a specifical hardware?
Will the engine output best models for all different hardware options or  for each hardware option?

olive/engine/engine.py
guotuofeng: so, we'd like to use the pass type name to clean the cache instead of using the user provided pass name. Is that correct?

olive/engine/engine.py
guotuofeng: Yes, the engine would run pass for different hardware accelerators. Next PR will change all olive pass to accept an accelerator spec which is maintained by engine. 
For each hardware and EP, there will be best models. All of these best models will be returned to user.

olive/engine/engine.py
```diff
@@ -94,10 +95,10 @@ def __init__(
...
         # dictionary of passes
+        self.pass_config = OrderedDict()
+
         # {"pass_name": {"pass": pass, "host": host, "evaluator": evaluator, "clean_run_cache": clean_run_cache}}
-        self.passes = {}
-        # list of pass names in the order they were registered
-        self.pass_order = []
+        self.passes = OrderedDict()
 
         self.footprints = Footprint()
```
jambayk, line 103: I think we would need to refresh this or create a new footprint object per hardware spec loop? Otherwise the models for different iterations are recorded into the same footprint object.

guotuofeng: @trajepl

trajepl: Maybe a field(accelerators) in footprints is better? Do you need me to implement it in the following PR?

guotuofeng: change to dict

olive/engine/engine.py
```diff
@@ -202,37 +224,231 @@ def run(
...
+        if search strategy is None, all passes are run in the order they were registered.
+        Save the final model to {output_dir}/{output_name}_model and json file to {output_dir}/{output_name}_model.json
+        Save evaluation results of the final model, if any, to {output_dir}/{output_name}_metrics.json
+        Return {"model": final_model_json, "metrics": evaluation_results}
+
+        if search strategy is not None, run the search strategy to find candidate models.
+        TODO: save the results using updated RunResult
+        """
+
+        prefix_output_name = f"{output_name}_" if output_name is not None else ""
```
trajepl, line 433: We may need to add `acelerator_spec` into `prefix_output_name`. Otherwise, each accelerator_spec will write the same file for output.

olive/engine/engine.py
```diff
@@ -263,82 +478,19 @@ def run(
...
-            if packaging_config:
-                logger.info("Package output model as artifacts")
-                generate_output_artifacts(
-                    packaging_config, self.footprints, self.footprints.get_last_node(), output_dir
-                )
-
-            return output
-
-        self.footprints.to_file(output_dir / f"{prefix_output_name}footprints.json")
+        self.footprints[accelerator_spec].to_file(output_dir / f"{prefix_output_name}footprints.json")
```
trajepl, line 491: Please take `accelerator_spec` into account when generate the `prefix_output_name`.

trajepl (approved):

olive/engine/engine.py
guotuofeng: will do

docs/source/tutorials/advanced_users.md
```diff
@@ -123,8 +121,7 @@ quantization_config = {
     "weight_type" : "QUInt8"
 }
 # search over the values for the other config parameters
-quantization_pass = create_pass_from_dict(OnnxQuantization, quantization_config)
-engine.register(quantization_pass)
+engine.register(OnnxQuantization, quantization_pass, False)
```
jambayk, line 124: I think this is `quantization_config` instead of `quantization_pass` here?

jambayk:
Can you update the paths to the footprints json and model json according in these examples? 
https://github.com/microsoft/Olive/blob/main/examples/directml/stable_diffusion/stable_diffusion.py#L77
https://github.com/microsoft/Olive/blob/main/examples/whisper/test_transcription.py#L47

olive/engine/engine.py
jambayk, line 1: Can you update the paths to the footprints json and model json according in these examples? 
https://github.com/microsoft/Olive/blob/main/examples/directml/stable_diffusion/stable_diffusion.py#L77
https://github.com/microsoft/Olive/blob/main/examples/whisper/test_transcription.py#L47

jambayk, line 1: I think if we use the accelerator spec to only create the search space for a pass, we should not use the `accelerator_spec` when caching pass runs and models. Once the search space is set, the pass run or the output model should be independent of the accelerator specs. 
Otherwise, we won't be able to take advantage of cached runs between different accelerator specs. Even though the pass run config is the same between two accelerator specs loops, they won't reuse the cached run since the id for the cached run is different. 
The model id is also similarly independent of the accelerator spec.

docs/source/tutorials/advanced_users.md
guotuofeng: good catch

olive/engine/engine.py
guotuofeng: done

guotuofeng: thanks for catch, it seems the accelerator spec in the cache path is needless since the model id contains the info

jambayk (approved):

PR #253, opened 2023-05-09
Refactor metric results and signal results

trajepl:
## Describe your changes
1. Define the results format for single metrics output called `MetricResult`
2. Define the results format for olive multi metrics output called `SignalResult`
3. Extend `sub_type` to accept `list` input to gather the results in one evaluation job if possible.
4. Add `sub_type_for_rank` which is used to sort/search strategy and etc., it is align with the value used for optimization, it must be a `str`.


olive/evaluator/metric.py
```diff
@@ -47,11 +50,12 @@ class LatencySubType(str, Enum):
 class Metric(ConfigBase):
     name: str
     type: MetricType
-    sub_type: Union[AccuracySubType, LatencySubType] = None
+    sub_type: Union[List[Union[str, AccuracySubType, LatencySubType]], AccuracySubType, LatencySubType] = None
+    sub_type_for_rank: str = None
```
jambayk, line 54: I think accepting a list is a good idea since we don't want to waste resources running a new metric for each subtype. 

However, I am not fully sure about the `sub_type_for_rank` idea. This looks like a arbitrary design choice. 

Maybe it is simpler, if we also ask for list for `name`, `priority_rank`, and `goal`? 
How is `goal` resolved and applied with `sub_type_for_rank`?

olive/engine/footprint.py
```diff
@@ -68,27 +70,25 @@ def record_objective_dict(self, objective_dict):
         self.objective_dict = objective_dict
 
     def _is_empty_metric(self, metric: FootprintNodeMetric):
-        return not metric or not metric.value
+        return not metric or not metric.value or not metric.value.signal
```
xiaoyu-work, line 73: Quick question: in which case we could get each `not metric or not metric.value or not metric.value.signal`?

olive/evaluator/metric_config.py
```diff
@@ -90,3 +90,13 @@ def check_value(cls, v, values):
         if values["type"] in ["min-improvement", "max-degradation"] and v < 0:
             raise ValueError(f"Value must be positive for type {values['type']}")
         return v
+
+
+class MetricResult(ConfigBase):
+    key_for_rank: str
```
xiaoyu-work, line 96: Can you explain a little bit about `key_for_rank` and `value_for_rank`?

olive/evaluator/metric.py
```diff
@@ -47,11 +50,12 @@ class LatencySubType(str, Enum):
 class Metric(ConfigBase):
     name: str
     type: MetricType
-    sub_type: Union[AccuracySubType, LatencySubType] = None
+    sub_type: Union[List[Union[str, AccuracySubType, LatencySubType]], AccuracySubType, LatencySubType] = None
```
xiaoyu-work, line 53: Do we consider multiple metrics with same type but different sub_types as a same `Metric` object? Should they be different metrics? I agree we could avoid duplicated calculation for same sub type of a same type metric, but have concerns here if they are considered as a same `Metric`.

olive/engine/footprint.py
trajepl: During Olive runtime, only check `not metric` is enough since Olive only record metric if the results of evaluation is not None.

But for some case, when we tried to create instance of FootprintNodeMetric manually, it will failed when:
1. `not metric.value`: metric = FootprintNodeMetric()
2. `not metric.value.signal`: metric = FootprintNodeMetric(signal=SignalResult())

Actually, this line is completed by copilot. 🤣  Do you need me to simplify to `not metric`?

olive/evaluator/metric.py
trajepl: Here is the motivation for this PR:
1. In model training case, the framework(torch/tf) will minimize the model loss(!= metrics). When the model loss is expected, the final metrics are measured. We can treat `sub_type_for_rank`(_maybe no need_) as model loss here, and `sub_type` list as final metrics shown in dashboard.
2. We should ensure that each `Metric` object is optimization direction even it has many sub_type to measure.

trajepl: Also, customer' request is more like:
they know which metrics they want to optimized but at the same time they also want to double check other metrics, just a quick look. Take `latency` as examples, the percentile numbers will help users to check avg numbers is stat-sig or not.

jambayk: I think the idea of only using one sub type for ranking is limiting and only serves the single customer request you describe. 
The point about having the same optimization direction doesn't take into account that the relationship between the subtypes is not always linear. For example, most of the time when precision increases, recall decreases. 

I think the customer request/use case that you describe could be fulfilled if we allows for a value like `-1` subtype priority rank for metric subtypes we want to evaluate but not use for optimization or ranking. 
This meets the above use case while also being more useful. For example, we might want to optimize both precision and recall at the same time. 

Custom metrics can also return multiple metric values that can be ignored or used for optimization and ranking.

olive/evaluator/metric_config.py
xiaoyu-work: Does this mean we only ranked the one subtype? What about the other two? Also, how to cross-rank different metrics with different sub types?

olive/engine/footprint.py
xiaoyu-work: Yeah it would be better to simplify this if there is no use case of other conditions.

olive/evaluator/metric.py
xiaoyu-work: Yeah I agree, metrics like `precision`, `f1`, `recall`, `auc` they are not quite same. If I'm a customer, I do have a use case in my mind that I would want a separate priority rank for each metric to my preference. `sub_type_for_rank` may not be sufficient for this use case.

trajepl: Yeah, indeed, there always be the case about joint optimization like to optimize `precision` and `recall` at the same time. 

I were thinking that in **current main branch**, we keep the `metric` concept as a single optimization goal. If we want to want to make joint metrics optimization, we have to define different metrics like:
`['metric1': {...}, 'metric2': {...}]`. Then olive will run search strategy to jointly optimize `metric1` and `metric2`.

trajepl: Then for your suggestions, the thing would be like following, 
`
[
  'metric1': {
    'sub_types': {
      'precision': {'rank': 1}
      'recall': {'rank': 2}
      f1': {} // <---------- default rank -1 not for optimization.
    }
  }, 
  'metric2': {
   'sub_types': {
     'latency_avg': {'rank': 3}
     'throughput': {'rank': 4}
   }
 }, 
]
`
In this way, we will transfer the `priority_rank` from metric level to `sub_type` level.

If so, really sounds good than single `sub_type_for_rank`, hahahah. 🤣  I were thinking to keep the original `Metric` definition then stuck in `sub_type_for_rank`.

olive/evaluator/metric.py
```diff
@@ -138,5 +76,67 @@ def validate_goal(cls, v, values):
         return v
 
 
+class Metric(ConfigBase):
+    name: str
```
jambayk, line 80: will the identifier for each type-subtype metric be `{name of Metric}-{name of subtype}`?

olive/evaluator/metric.py
```diff
@@ -39,78 +42,13 @@ class LatencySubType(str, Enum):
...
-
-    @validator("user_config", pre=True)
-    def validate_user_config(cls, v, values):
-        if "type" not in values:
-            raise ValueError("Invalid type")
-
-        user_config_class = get_user_config_class(values["type"])
-        return validate_config(v, ConfigBase, user_config_class)
+    # -1 means no priority which will be evaluated only
+    priority_rank: int = -1
```
jambayk, line 49: if the default value is `-1`, I think we need to validate somewhere that atleast one type-subtype has value 1. Or else we won't have any signal for the search algorithms or metrics to rank output models by. 

Maybe we keep the default as before? assumed as 1 if only one type-subtype is present in the entire metric list, else, require user to provide unique rank or -1 for all type-subtypes

olive/evaluator/metric_config.py
```diff
@@ -90,3 +90,26 @@ def check_value(cls, v, values):
...
+    priority_rank: int
+    higher_is_better: bool
+
+
+class MetricResult(ConfigDictBase):
+    __root__: Dict[str, SubTypeMetricResult] = None
+
+
+class SignalResult(ConfigDictBase):
+    __root__: Dict[str, MetricResult] = None
```
jambayk, line 106: I think instead of nesting the final results with `{"type name": {"subtype name": SubTypeMetricResult}}`, it might be better to return a flattened dict with the type and subtype keys flattened into a single identifier. 

Otherwise, the evaluation results being nested makes the evaluation signals unnecessarily complicated. The search results, algorithms, footprints need to parse the results with multiple loops. 

Once they are evaluated, I don't think the hierarchy of the subtypes matter?

olive/evaluator/metric.py
jambayk: maybe we can move the unique identifier `name` into `SubTypeMetric`? The SubTypeMetric with have fields `name` which is a unique id and `type`/`sub_type`.

trajepl: We will keep the hierarchical structure for Metric definition. But flatten the results in Signal results.

olive/evaluator/olive_evaluator.py
```diff
@@ -28,12 +29,12 @@ def get_metric(self, metric_name: str) -> Metric:
...
+    def evaluate(self, model: OliveModel, target: OliveSystem = LocalSystem()) -> MetricResult:
         return target.evaluate_model(model, self.metrics)
 
 
 class OliveEvaluatorConfig(ConfigBase):
-    metrics: List[Metric]
+    metrics: MetricList
 
     @validator("metrics")
     def validate_metrics(cls, v):
```
jambayk, line 40: I just realized that since we are only validating the metrics in this config class, the check is not done if the evaluator is created directed using the `OliveEvaluator` pass. The evaluation tests directly create the evaluator object so they won't have this validation too. 

maybe we can abstract the `validate_metric` function and use it both here and in `OliveEvaluator` init?

olive/strategy/search_strategy.py
jambayk, line 1: I think when we initialize the Search strategy at line 84, we should filter out the metrics in `objective_dict` that have priority rank -1. 
That way the search results and the search algorithms only interact with the relevant metrics while making decisions.

jambayk: The objective dict needs to have the rank information too. We should also order the self.objectives list by the priority rank so that the model sorting follows this order.

trajepl: Fixed.

olive/evaluator/olive_evaluator.py
trajepl: Yes, actually not only for Evaluator. The same problems for Metric, RunConfig. Will sent the seprated PR to fix that.

olive/evaluator/metric_config.py

olive/evaluator/metric.py
trajepl: Still keep the default value as -1 to avoid user to provide rank for those metrics they do not want to compare. The behavior will be like
1. When there is only sub metric, user can provide empty priority rank olive will assign as 1. (done in validator
2. When there is multi sub metrics like bert example, user can just assign the rank value to the metrics they care. For other metrics, -1 will be assign.

olive/evaluator/metric.py
trajepl: Updated.

olive/common/config_utils.py
```diff
@@ -115,6 +115,41 @@ def from_json(cls, json_dict: dict) -> "ConfigBase":
         return cls.parse_raw(json.dumps(json_dict))
 
 
+class ConfigListBase(ConfigBase):
```
guotuofeng, line 118: what's the consideration that we don't use the native list/dict and use the customized list/dict?

olive/engine/engine.py
```diff
@@ -248,7 +235,7 @@ def run(
                 results = self._evaluate_model(input_model, input_model_id, self.evaluator_config, i, verbose)
                 result_name = f"{prefix_output_name}metrics"
                 results_path = output_dir / f"{result_name}.json"
-                json.dump(results, open(results_path, "w"), indent=4)
+                json.dump(results.dict(), open(results_path, "w"), indent=4)
```
guotuofeng, line 238: remember close the file

olive/systems/azureml/aml_evaluation_runner.py
```diff
@@ -54,7 +54,7 @@ def main(raw_args=None):
     metric_result = target.evaluate_model(model, [metric])
 
     # save metric result json
-    json.dump(metric_result, open(Path(common_args.pipeline_output) / "metric_result.json", "w"))
+    json.dump(metric_result.dict(), open(Path(common_args.pipeline_output) / "metric_result.json", "w"))
```
guotuofeng, line 57: close the file

olive/systems/azureml/aml_system.py
```diff
@@ -435,9 +434,9 @@ def evaluate_model(self, model: OliveModel, metrics: List[Metric]) -> Dict[str,
             for metric in metrics:
                 metric_json = output_dir / "named-outputs" / metric.name / "metric_result.json"
                 if metric_json.is_file():
-                    metric_results[metric.name] = json.load(open(metric_json))["result"]
+                    metric_results.update(json.load(open(metric_json))["__root__"])
```
guotuofeng, line 437: ditto

guotuofeng (changes requested):
could we not use the customized list/dic?

guotuofeng (changes requested):
could we not use the customized list/dic?

olive/common/config_utils.py
trajepl: Use to simplify the access on pydantic data. Not aim to customized list/dict.

examples/whisper/whisper_template.json
```diff
@@ -35,7 +35,9 @@
                 {
                     "name": "latency",
                     "type": "latency",
-                    "sub_type": "avg",
+                    "sub_types": [
+                        {"name": "avg","priority_rank": 2}
```
jambayk, line 39: priority rank 1

docs/source/overview/options.md
```diff
@@ -141,19 +141,23 @@ information of the evaluator contains following items:
 
     - `type: [str]` The type of the metric. The supported types are `accuracy`, `latency` and `custom`.
 
-    - `subtype: [str]` The subtype of the metric. Please refer to [AccuracySubtype](accuracy_sub_type) and
-    [LatencySubtype](latency_sub_type) for the supported sub-types. It is `null` for `custom` type.
+    - `subtyps: [List[Dict]]` The subtypes of the metric. Cannot be null or empty. Each subtype is a dictionary that contains following items:
```
jambayk, line 144: typo

docs/source/overview/options.md
```diff
@@ -141,19 +141,23 @@ information of the evaluator contains following items:
...
 
-    - `higher_is_better: [Boolean]` True if the metric is better when it is higher. It is `true` for `accuracy` type and `false` for `latency` type.
+        - `name: str` The name of the subtype. Please refer to [AccuracySubtype](accuracy_sub_type) and [LatencySubtype](latency_sub_type)
```
jambayk, line 146: I think we need a name for the custom metrics too? It cannot be null?

olive/engine/engine.py
```diff
@@ -249,7 +236,7 @@ def run(
                 result_name = f"{prefix_output_name}metrics"
                 results_path = output_dir / f"{result_name}.json"
                 with open(results_path, "w") as f:
-                    json.dump(results, f, indent=4)
+                    f.write(results.json())
```
guotuofeng, line 239: I like this since it would remove the "__root__", right?

trajepl: Yes~

docs/source/overview/options.md
trajepl: Updated.

olive/evaluator/olive_evaluator.py
trajepl: https://github.com/microsoft/Olive/pull/256 Looks like it was fixed in this PR already.

.gitignore
```diff
@@ -155,3 +155,4 @@ cython_debug/
 
 # AzureML
 .azureml
+.*.py
```
guotuofeng, line 158: why ignore all python files?

docs/source/overview/options.md
```diff
@@ -152,19 +152,23 @@ information of the evaluator contains following items:
...
-    - `subtype: [str]` The subtype of the metric. Please refer to [AccuracySubtype](accuracy_sub_type) and
-    [LatencySubtype](latency_sub_type) for the supported sub-types. It is `null` for `custom` type.
+    - `subtypes: [List[Dict]]` The subtypes of the metric. Cannot be null or empty. Each subtype is a dictionary that contains following items:
 
-    - `higher_is_better: [Boolean]` True if the metric is better when it is higher. It is `true` for `accuracy` type and `false` for `latency` type.
+        - `name: str` The name of the subtype. Please refer to [AccuracySubtype](accuracy_sub_type) and [LatencySubtype](latency_sub_type)
+        for the supported sub-types. For `custom` type, if the result of the evaluation is a dictionary, the name of the subtype should be the key of the dictionary. Otherwise, the name of the subtype could be any unique string user gives.
 
-    - `goal: [Dict]` The goal of the metric. It is a dictionary that contains following items:
+        - `priority_rank: [int]` The priority rank of the subtype. The subtype with the highest priority rank will be used to evaluate the. Note that it should be unique among all subtypes in the metric.
```
guotuofeng, line 160: the term priority is similar with rank, could we use one?

examples/test_resnet_ptq_cpu.py
```diff
@@ -30,7 +30,7 @@ def check_output(footprints):
     for footprint in footprints.values():
         assert footprint.nodes is not None
         for v in footprint.nodes.values():
-            assert all([value > 0 for value in v.metrics.value.values()])
+            assert all([value.value > 0 for value in v.metrics.value.values()])
```
guotuofeng, line 33: so many value, metrics.value.values().value. Is it a little bit wired?

docs/source/overview/options.md
trajepl: Updated to priority.

olive/evaluator/metric_config.py
```diff
@@ -90,3 +90,43 @@ def check_value(cls, v, values):
...
+
+def joint_metric_key(metric_name, sub_type_name):
+    return f"{metric_name}{MetricResult.delimiter}{sub_type_name}"
+
+
+def flatten_metric_sub_type(metric_dict: Dict[str, Dict]):
+    flatten_results = {}
+    for metric_name, metric_res in metric_dict.items():
+        for sub_type_name, sub_type_res in metric_res.items():
+            key = f"{metric_name}{MetricResult.delimiter}{sub_type_name}"
```
guotuofeng, line 126: joint_metric_key?

.gitignore
trajepl: Debug purpose, for .xxx.py. Removed.

examples/test_resnet_ptq_cpu.py
trajepl: Updated to increase readability.

olive/evaluator/olive_evaluator.py
```diff
@@ -237,9 +264,13 @@ def _evaluate_onnx_accuracy(
     def _evaluate_onnx_latency(
         self, model: OliveModel, metric: Metric, dataloader: Dataset, device: Device = Device.CPU, post_func=None
     ) -> Dict[str, Any]:
-        warmup_num = metric.metric_config.warmup_num
-        repeat_test_num = metric.metric_config.repeat_test_num
-        sleep_num = metric.metric_config.sleep_num
+        warmup_num, repeat_test_num, sleep_num = None, None, None
```
guotuofeng, line 267: lot of redundancy code. Could we remove them by a method or function?

test/unit_test/model/test_pytorch_model.py
```diff
@@ -8,13 +8,15 @@
...
 import pandas as pd
+import pytest
 import transformers
 from azureml.evaluate import mlflow as aml_mlflow
 
 from olive.model import PyTorchModel
 
 
 class TestPyTorchMLflowModel(unittest.TestCase):
+    @pytest.fixture(autouse=True)
```
guotuofeng, line 19: suggest to make the changes unrelated to metrics to a separated PR.

docs/source/overview/options.md
```diff
@@ -152,19 +152,23 @@ information of the evaluator contains following items:
...
-    - `subtype: [str]` The subtype of the metric. Please refer to [AccuracySubtype](accuracy_sub_type) and
-    [LatencySubtype](latency_sub_type) for the supported sub-types. It is `null` for `custom` type.
+    - `subtypes: [List[Dict]]` The subtypes of the metric. Cannot be null or empty. Each subtype is a dictionary that contains following items:
 
-    - `higher_is_better: [Boolean]` True if the metric is better when it is higher. It is `true` for `accuracy` type and `false` for `latency` type.
+        - `name: str` The name of the subtype. Please refer to [AccuracySubtype](accuracy_sub_type) and [LatencySubtype](latency_sub_type)
+        for the supported sub-types. For `custom` type, if the result of the evaluation is a dictionary, the name of the subtype should be the key of the dictionary. Otherwise, the name of the subtype could be any unique string user gives.
 
-    - `goal: [Dict]` The goal of the metric. It is a dictionary that contains following items:
+        - `priority: [int]` The priority rank of the subtype. The subtype with the highest priority rank will be used to evaluate the. Note that it should be unique among all subtypes in the metric.
```
guotuofeng, line 160: rename the term

docs/source/tutorials/configure_metrics.rst
```diff
@@ -193,6 +193,6 @@ If you have multiple metrics to evaluate, you can configure them in the followin
             ]
         }
 
-You need to specify :code:`"priority_rank": <rank>` for the metrics if you have multiple metrics.
-Olive will use the priority_ranks of the metrics to determine the best model.
-If you only have one metric, you can omit :code:`"priority_rank": 1`.
+You need to specify :code:`"priority": <rank>` for the metrics if you have multiple metrics.
```
guotuofeng, line 196: need we change it to priority?

trajepl: Good catch, let me check.

trajepl: Updated.

olive/evaluator/olive_evaluator.py

olive/evaluator/metric.py
```diff
@@ -39,78 +42,13 @@ class LatencySubType(str, Enum):
...
-# so that we get a single metric config for a single type
-# This way, the user can return multiple metrics at once
-class Metric(ConfigBase):
-    name: str
-    type: MetricType
-    sub_type: Union[AccuracySubType, LatencySubType] = None
-    higher_is_better: bool = True
-    priority_rank: int = 1
-    goal: MetricGoal = None
+class MetricSubType(ConfigBase):
```
guotuofeng, line 45: SubMetric?

olive/systems/python_environment/python_environment_system.py
```diff
@@ -145,9 +146,13 @@ def evaluate_latency(self, model: ONNXModel, metric: Metric) -> float:
...
         dataloader, _, _ = OliveEvaluator.get_user_config(metric)
-        warmup_num = metric.metric_config.warmup_num
-        repeat_test_num = metric.metric_config.repeat_test_num
-        sleep_num = metric.metric_config.sleep_num
+        warmup_num, repeat_test_num, sleep_num = None, None, None
+        for sub_type in metric.sub_types:
+            if sub_type.metric_config:
+                warmup_num = sub_type.metric_config.warmup_num
+                repeat_test_num = sub_type.metric_config.repeat_test_num
+                sleep_num = sub_type.metric_config.sleep_num
```
guotuofeng, line 154: redundancy code?

olive/evaluator/olive_evaluator.py
```diff
@@ -148,48 +165,68 @@ def get_user_config(metric: Metric):
...
+        metric_res = {}
+        for sub_type in metric.sub_types:
+            metric_res[sub_type.name] = SubTypeMetricResult(
+                value=latency_metrics[sub_type.name],
+                priority=sub_type.priority,
+                higher_is_better=sub_type.higher_is_better,
+            )
+        return MetricResult.parse_obj(metric_res)
+
+    def _get_latency_config_from_metric(self, metric: Metric):
```
guotuofeng, line 221: make it method of Metric so all other place could reuse it?

trajepl: Agree.~

jambayk (commented):
Please also update the directml example configs.

trajepl:
> Please also update the directml example configs.

Updated.

docs/source/overview/options.md
```diff
@@ -197,8 +201,12 @@ information of the evaluator contains following items:
...
                 "name": "accuracy",
                 "type": "accuracy",
-                "sub_type": "accuracy_score",
-                "priority_rank": 1,
+                "sub_types": [
+                    {"name": "accuracy_score", "priority": 1, "goal": {"type": "max-degradation", "value": 0.01}},
+                    {"name": "f1_score"},
+                    {"name": "auc", "metric_config": {"reorder": true}}
+                ],
+                "priority": 1,
```
jambayk, line 209: I don't think this should be here?

olive/engine/footprint.py
```diff
@@ -74,29 +75,29 @@ def record_objective_dict(self, objective_dict):
...
-                if metric_name in self.objective_dict:
-                    cmp_direction = 1 if self.objective_dict[metric_name]["higher_is_better"] else -1
-                    self.nodes[k].metrics.cmp_direction[metric_name] = cmp_direction
-                    _goal = self.objective_dict[metric_name]["goal"]
-                    if _goal is None:
-                        is_goals_met.append(True)
-                    else:
-                        is_goals_met.append(v.metrics.value[metric_name] * cmp_direction >= _goal)
+                if metric_name not in self.objective_dict:
+                    logger.debug("There is no goal set for metric: {metric_name}.")
```
jambayk, line 90: nit: I think this message should be that it does not have priority rank?

olive/evaluator/olive_evaluator.py
```diff
@@ -477,9 +511,14 @@ def _evaluate_accuracy(
 
     def _evaluate_latency(
         self, model: PyTorchModel, metric: Metric, dataloader: Dataset, device: Device = Device.CPU, post_func=None
-    ) -> Dict[str, Any]:
-        warmup_num = metric.metric_config.warmup_num
-        repeat_test_num = metric.metric_config.repeat_test_num
+    ) -> MetricResult:
+        warmup_num, repeat_test_num = None, None
```
guotuofeng, line 515: get_latency_config_from_metric?

olive/evaluator/olive_evaluator.py
```diff
@@ -590,16 +622,32 @@ class OliveEvaluatorConfig(ConfigBase):
...
+        for metric in v:
+            for sub_type in metric.sub_types:
+                sub_type_names.add(joint_metric_key(metric.name, sub_type.name))
+                if sub_type.priority != -1:
+                    sub_type_with_rank.add(sub_type.name)
+                    rank_set.add(sub_type.priority)
+
+        if not rank_set and len(sub_type_names) == 1:
+            logger.debug(
+                """No priority is specified, but only one sub type
```
guotuofeng, line 641: will the multiple line string also be prepend with the leading whitespace in the multiple line string here?

olive/engine/engine.py
```diff
@@ -497,40 +488,60 @@ def resolve_goals(
...
+        baseline = MetricResult()
+        for goal in goals.values():
+            _evaluated = False
+            for sub_goal in goal.values():
+                if not sub_goal:
+                    break
+                if sub_goal.type != "threshold":
+                    assert self.evaluator_config is not None, "Default evaluator must be provided to resolve goals"
+                    if verbose:
+                        logger.info("Computing baseline for metrics ...")
```
guotuofeng, line 510: could we use logger.debug without checking the verbose flag?

trajepl: Good suggestion. How about to fix in separated PR. The verbose flag is added in the first version of olive by Jambay.

olive/evaluator/olive_evaluator.py
trajepl: Fixed.

olive/engine/engine.py
guotuofeng: sure, for existing code, I suggest to fix in another PR, I put comments here only for new added code.

olive/engine/engine.py
```diff
@@ -476,12 +459,20 @@ def resolve_objectives(
...
         goals = self.resolve_goals(input_model, input_model_id, metrics, accelerator_spec, verbose)
-        objective_dict = {
-            metric.name: {"higher_is_better": metric.higher_is_better, "goal": goals.get(metric.name)}
-            for metric in metrics
-        }
+        objective_dict = {}
+        for metric in metrics:
+            for sub_type in metric.sub_types:
+                if sub_type.priority <= 0:
+                    continue
```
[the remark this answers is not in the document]
trajepl: No need. We should not optimize the metrics with priority <= 0 which are evaluated for references.

olive/evaluator/olive_evaluator.py
trajepl: Yes, updated.

olive/engine/engine.py
trajepl: removed.

docs/source/overview/options.md
```diff
@@ -152,19 +152,23 @@ information of the evaluator contains following items:
 
     - `type: [str]` The type of the metric. The supported types are `accuracy`, `latency` and `custom`.
 
-    - `subtype: [str]` The subtype of the metric. Please refer to [AccuracySubtype](accuracy_sub_type) and
-    [LatencySubtype](latency_sub_type) for the supported sub-types. It is `null` for `custom` type.
+    - `subtypes: [List[Dict]]` The subtypes of the metric. Cannot be null or empty. Each subtype is a dictionary that contains following items:
 
-    - `higher_is_better: [Boolean]` True if the metric is better when it is higher. It is `true` for `accuracy` type and `false` for `latency` type.
+        - `name: str` The name of the subtype. Please refer to [AccuracySubtype](accuracy_sub_type) and [LatencySubtype](latency_sub_type)
+        for the supported sub-types. For `custom` type, if the result of the evaluation is a dictionary, the name of the subtype should be the key of the dictionary. Otherwise, the name of the subtype could be any unique string user gives.
```
guotuofeng, line 158: do we need unify to sub-type or subtype?

docs/source/overview/options.md
```diff
@@ -152,19 +152,23 @@ information of the evaluator contains following items:
...
 
-    - `goal: [Dict]` The goal of the metric. It is a dictionary that contains following items:
+        - `priority: [int]` The priority of the subtype. The subtype with the highest priority will be used to evaluate the. Note that it should be unique among all subtypes in the metric.
```
guotuofeng, line 160: missing something?

olive/evaluator/olive_evaluator.py
```diff
@@ -94,13 +105,28 @@ def _evaluate_custom(
...
-        return eval_func(model, metric.user_config.data_dir, metric.user_config.batch_size, device)
+        raw_res = eval_func(model, metric.user_config.data_dir, metric.user_config.batch_size, device)
+        metric_res = {}
+        for sub_type in metric.sub_types:
+            if isinstance(raw_res, Number):
+                assert len(metric.sub_types) == 1, "Only one sub type is allowed for single value custom metric"
+                metric_res[sub_type.name] = SubMetricResult(
+                    value=raw_res, priority=sub_type.priority, higher_is_better=sub_type.higher_is_better
+                )
+            elif isinstance(raw_res, Dict):
```
guotuofeng, line 120: dict?

trajepl: Should be Dict from typing. `dict` is available after python3.9. 
https://stackoverflow.com/questions/37087457/difference-between-defining-typing-dict-and-dict

trajepl: Updated to dict.

docs/source/overview/options.md
```diff
@@ -209,8 +216,11 @@ information of the evaluator contains following items:
             {
                 "name": "latency",
                 "type": "latency",
-                "sub_type": "avg",
-                "priority_rank": 2,
+                "sub_types": [
+                    {"name": "avg", "goal": {"type": "percent-min-improvement", "value": 20}},
```
xiaoyu-work, line 220: Do we need `"priority": 2,` for second metric?

docs/source/overview/quicktour.md
```diff
@@ -121,7 +121,11 @@ returns metrics values for each output model.
             {
                 "name": "latency",
                 "type": "latency",
-                "sub_type": "avg",
+                "sub_types": [
+                    {"name": "avg", "priority": 2, "goal": {"type": "percent-min-improvement", "value": 20}},
```
xiaoyu-work, line 125: Do we need "priority": 2, for only one metric?

docs/source/overview/quicktour.md
```diff
@@ -217,7 +221,11 @@ python -m olive.workflows.run --config config.json
                 {
                     "name": "latency",
                     "type": "latency",
-                    "sub_type": "avg",
+                    "sub_types": [
+                        {"name": "avg", "priority": 2, "goal": {"type": "percent-min-improvement", "value": 20}},
```
xiaoyu-work, line 225: Same as above

docs/source/tutorials/advanced_users.md
```diff
@@ -59,7 +59,11 @@ from olive.evaluator.olive_evaluator import OliveEvaluatorConfig
 latency_metric = Metric(
     name="latency",
     type=MetricType.LATENCY,
-    sub_type=LatencySubType.AVG,
+        sub_types=[{
+        "name": LatencySubType.AVG,
+        "priority": 2,
```
xiaoyu-work, line 64: Same as above

docs/source/tutorials/configure_metrics.rst
```diff
@@ -110,35 +118,40 @@ sepcify its name in :code:`"evaluate_func"` field, and Olive will call your func
             {
                 "name": "accuracy",
                 "type": "custom",
+                sub_types: [
```
xiaoyu-work, line 121: `sub_types` -> `"sub_types"`

docs/source/tutorials/configure_metrics.rst
```diff
@@ -63,34 +69,36 @@ Latency Metric
             {
                 "name": "latency",
                 "type": "latency",
-                "sub_type": "avg",
+                "sub_types": [
+                    {"name": "avg", "priority": 2, "goal": {"type": "percent-min-improvement", "value": 20}}
```
xiaoyu-work, line 73: Can we remove `priority` here in example? Or set it to 1? 2 is confusing me here.

examples/bert_inc_ptq_cpu/bert_inc_static_config.json
```diff
@@ -36,8 +35,8 @@
                 {
                     "name": "latency",
                     "type": "latency",
-                    "sub_type": "avg",
-                    "priority_rank": 2,
+                    "sub_types": [{"name": "avg", "priority": 2}],
+                    "priority": 2,
```
xiaoyu-work, line 39: Can we remove this as it is not needed anymore.

olive/engine/engine.py
```diff
@@ -879,7 +890,7 @@ def _evaluate_model(
         return signal
 
     def _get_top_ranked_nodes(self, metrics: List[Metric], footprint: Footprint, k: int) -> List[FootprintNode]:
-        metric_priority = [metric.name for metric in sorted(metrics, key=lambda x: x.priority_rank)]
+        metric_priority = [metric.name for metric in sorted(metrics, key=lambda x: x.priority)]
```
xiaoyu-work, line 893: There is no `priority` attribute for Metric, should we use the subtype's priority instead here?

test/unit_test/systems/azureml/output_metrics/pipeline_output/named-outputs/accuracy/metric_result.json
```diff
@@ -1 +1,7 @@
-{"result": 0.99618}
+{
+    "accuracy-accuracy_score": {"value": 0.99618, "priority": 1, "higher_is_better": true},
+    "accuracy-f1_score": {"value": 0.99618, "priority": 1, "higher_is_better": true},
```
xiaoyu-work, line 3: Question: will all subtype metrics have `priority 1`?

test/unit_test/systems/docker/output_local_path/eval_res.json
```diff
@@ -1 +1,7 @@
-{"accuracy": 0.99618}
+{
+    "accuracy-accuracy_score": {"value": 0.99618, "priority": 1, "higher_is_better": true},
+    "accuracy-f1_score": {"value": 0.99618, "priority": 1, "higher_is_better": true},
```
xiaoyu-work, line 3: Same as above

xiaoyu-work:
Overall looks good to me. Thanks!

guotuofeng (approved):

test/unit_test/systems/azureml/output_metrics/pipeline_output/named-outputs/accuracy/metric_result.json
trajepl: Fixed. Only one metric should have `1` as top priority.

olive/engine/engine.py
trajepl: Updated to use sorted objective_dict.

guotuofeng (approved):

PR #258, opened 2023-05-12
Add support for HF AML model

xiaoyu-work:
## Describe your changes
* Add support for HF AML model
* Add bert HF AML model example

### DIFF
--- a/olive/model.py
+++ b/olive/model.py
@@ -302,6 +302,7 @@ class ONNXModel(ONNXModelBase):
         model_storage_kind: Union[str, ModelStorageKind] = ModelStorageKind.LocalFile,
         inference_settings: Optional[dict] = None,
         use_ort_extensions: bool = False,
+        hf_config: Union[Dict[str, Any], HFConfig] = None,
     ):
         super().__init__(
             model_path=model_path,
@@ -314,6 +315,8 @@ class ONNXModel(ONNXModelBase):
         self.io_config = None
         self.graph = None
         self.all_graphs: Optional[List[GraphProto]] = None
+        # huggingface config
+        self.hf_config = validate_config(hf_config, HFConfig) if hf_config else None
 
     @staticmethod
     def resolve_path(file_or_dir_path: str, model_filename: str = "model.onnx") -> str:
@@ -400,7 +403,11 @@ class ONNXModel(ONNXModelBase):
     def to_json(self, check_object: bool = False):
         config = super().to_json(check_object)
         config["config"].update(
-            {"inference_settings": self.inference_settings, "use_ort_extensions": self.use_ort_extensions}
+            {
+                "inference_settings": self.inference_settings,
+                "use_ort_extensions": self.use_ort_extensions,
+                "hf_config": self.hf_config,
+            }
         )
         return serialize_to_json(config, check_object)
 
--- a/olive/model.py
+++ b/olive/model.py
@@ -207,12 +207,12 @@ class HFComponent(ConfigBase):
 
 
 class HFConfig(ConfigBase):
+    load_model_from_hub: bool = True
     model_name: str = None
     task: str = None
     # TODO: remove model_class and only use task
     model_class: str = None
     use_ort_implementation: bool = False
-    load_model_from_hub: bool = True
     components: List[HFComponent] = None
     dataset: Dict[str, Any] = None
 
@@ -223,6 +223,12 @@ class HFConfig(ConfigBase):
 
         if not v and not values["task"]:
             raise ValueError("Either task or model_class must be specified")
+        return v
+
+    @validator("model_name", always=True)
+    def validate_model_name(cls, v, values):
+        if values["load_model_from_hub"] and not v:
+            raise ValueError("Model_name needs to be specified when load_model_from_hub is True")
         return v
 
 
@@ -544,11 +550,12 @@ class PyTorchModel(OliveModel):
             user_module_loader = UserModuleLoader(self.model_script, self.script_dir)
             model = user_module_loader.call_object(self.model_loader, self.model_path)
         elif self.hf_config is not None:
+            input_model = self.hf_config.model_name if self.hf_config.load_model_from_hf_hub else self.model_path
             if self.hf_config.task:
-                model = load_huggingface_model_from_task(self.hf_config.task, self.hf_config.model_name)
+                model = load_huggingface_model_from_task(self.hf_config.task, input_model)
             else:
                 model = load_huggingface_model_from_model_class(
-                    self.hf_config.model_class, self.hf_config.model_name, self.hf_config.use_ort_implementation
+                    self.hf_config.model_class, input_model, self.hf_config.use_ort_implementation
                 )
         else:
             if self.model_file_format == ModelFileFormat.PYTORCH_ENTIRE_MODEL:
--- a/olive/model.py
+++ b/olive/model.py
@@ -207,7 +207,7 @@ class HFComponent(ConfigBase):
 
 
 class HFConfig(ConfigBase):
-    model_name: str
+    model_name: str = None
     task: str = None
     # TODO: remove model_class and only use task
     model_class: str = None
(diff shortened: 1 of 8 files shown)


xiaoyu-work:
Question: is `HFConfig` useful for HF onnx model? How many (and which) attributes in `HFConfig` can be used on HF onnx model?  (like `model_name` will be used to fetch tokenizer from HF model hub so `model_name` is needed)

olive/model.py
```diff
@@ -207,11 +207,12 @@ class HFComponent(ConfigBase):
 
 
 class HFConfig(ConfigBase):
-    model_name: str
+    model_name: str = None
     task: str = None
     # TODO: remove model_class and only use task
     model_class: str = None
     use_ort_implementation: bool = False
+    load_model_from_hub: bool = True
```
jambayk, line 215: I think a validator, that checks that `model_name` is not `None` if `load_model_from_hub` is `True` would be helpful.

olive/model.py
jambayk, line 1: At line 547, I think we would use `self.hf_config.model_name` if `self.hf_config.load_model_from_hub = True` and `self.model_path` if `self.hf_config.load_model_from_hub = False`? 

This way, we could support not only aml models, but also locally saved hf models? For local runs, I guess the user could provide the local path as `self.hf_config.model_name` and it would still work. But uploading to aml system would not work. If the provided it as `model_name` and set `self.hf_config.load_model_from_hub = False` then, the aml upload could work too.

olive/systems/azureml/aml_evaluation_runner.py
```diff
@@ -42,12 +42,22 @@ def main(raw_args=None):
...
     model_config = get_model_config(common_args)
-    model = ModelConfig.from_json(model_config).create_model()
-
     # load metric
     metric_config = json.load(open(metric_args.metric_config))
     metric = create_metric(metric_config, metric_args)
 
+    # Replace hf config model_name with input model path to load model from input model path
+    hf_model_name = None
+    if model_config["config"].get("hf_config") and not model_config["config"]["hf_config"]["load_model_from_hub"]:
```
jambayk, line 51: If we did that loading from `model_path` in `PytorchModel` that I suggested in the other comment, I think this step could be skipped?

olive/model.py
xiaoyu-work: Sounds a good idea. Will update this.

olive/model.py
xiaoyu-work: Will add.

olive/systems/azureml/aml_pass_runner.py
```diff
@@ -100,8 +91,8 @@ def main(raw_args=None):
     model_json = output_model.to_json()
 
     # Replace output model HF config with input model HF config
-    if hf_config:
-        model_json["config"]["hf_config"] = hf_config
+    if input_model_config["config"].get("hf_config"):
+        model_json["config"]["hf_config"] = input_model_config["config"]["hf_config"]
```
jambayk, line 95: We don't use snpe or openvino models on aml so this will not fail but this assumes all models have `hf_config`. 

Is this part required? I think the hf_config needs to be provided to the output model by the pass that created it. Otherwise we have mismatch in the output model's config between aml system and local system. Here the system is autoinserting hf_config to all output models.

olive/systems/azureml/aml_evaluation_runner.py
```diff
@@ -42,12 +42,21 @@ def main(raw_args=None):
 
     # load model
     model_config = get_model_config(common_args)
-    model = ModelConfig.from_json(model_config).create_model()
-
     # load metric
     metric_config = json.load(open(metric_args.metric_config))
     metric = create_metric(metric_config, metric_args)
 
+    # HF model tokenizer will be loaded from HF model hub
```
jambayk, line 49: I think this is already done by the run config validator https://github.com/microsoft/Olive/blob/main/olive/workflows/run/config.py#LL67C1-L67C1? Or do we expect the metric data config to be changed somewhere else?

olive/systems/azureml/aml_pass_runner.py
xiaoyu-work: I assume all previous model which has `hf_config` is a hf model so the next output model is also a hf model. This is correct?

jambayk: Currently, the hf configs are propagated only from pytorch model to composite onnx model by the onnx conversion pass. 

None of the other passes or models use it.

olive/systems/azureml/aml_evaluation_runner.py
xiaoyu-work: This part logic in config.py is running when initiating Olive from config file. This does not apply to runtime when we got a new model to evaluate it.

olive/systems/azureml/aml_pass_runner.py
jambayk: But I think propagating to all output models sounds fine too. We would need to update the output of all the passes to do this though.

xiaoyu-work: If we can do this when running pass that would be great, then we can remove this logic here, otherwise the output hf onnx model will not have the hf config data.

olive/systems/azureml/aml_evaluation_runner.py
jambayk: Once the metric is set, I don't think the `model_name` changes?  All models come from the same source model, so I think the `model_name` is the same. Otherwise we would have different data for different models.

olive/systems/azureml/aml_pass_runner.py
jambayk: Maybe we can do it in a different PR so that your PR doesn't need to worry about it?
I also think it is better for the passes to do it if we want all models to have hf_config. This way the behavior is consistent for all systems since the pass inserts it.

olive/systems/azureml/aml_evaluation_runner.py
jambayk: I am not fully familiar with how the new `data_config` behaves so you can ignore this XD

xiaoyu-work: It sounds reasonable. I was having an error before that the metric does not have such data for outputed onnx model but I can't remember the exact scenario. Let me test it here.

olive/systems/azureml/aml_pass_runner.py
xiaoyu-work: Yeah let's do this in a separate PR. Thanks!

olive/systems/azureml/aml_evaluation_runner.py
xiaoyu-work: Seems it works. Maybe my previous case happens when the tokenizer was loaded from local files.

olive/systems/azureml/aml_evaluation_runner.py
```diff
@@ -42,12 +42,12 @@ def main(raw_args=None):
 
     # load model
     model_config = get_model_config(common_args)
-    model = ModelConfig.from_json(model_config).create_model()
-
     # load metric
     metric_config = json.load(open(metric_args.metric_config))
     metric = create_metric(metric_config, metric_args)
 
+    model = ModelConfig.from_json(model_config).create_model()
```
jambayk, line 49: nit: is this just a leftover from undone changes?

olive/systems/azureml/aml_evaluation_runner.py
jambayk: Thanks for checking!

olive/systems/azureml/aml_evaluation_runner.py
xiaoyu-work: Ah yeah I moved this line from before metric to after metric since I was going to update hfconfig for model config and metric config both.

olive/systems/azureml/aml_pass_runner.py
```diff
@@ -89,6 +89,11 @@ def main(raw_args=None):
 
     # save model json
     model_json = output_model.to_json()
+
+    # Replace output model HF config with input model HF config
```
jambayk, line 93: Since, we are planning to make the passes autoinsert the hf_config, are we still keeping this section?

jambayk: Or we remove this when that PR happens?

xiaoyu-work: This is needed for aml run for now. We should remove this after we make the changes on passes

jambayk (approved):

PR #261, opened 2023-05-15
Support input aml model run locally

xiaoyu-work:
## Describe your changes

1. Add AML model download logic.
2. Since we have more use cases for AML client other than AMLSystem. Extract AML client creation logic out to a standalone logic.
3. Support input aml model run locally.

Notes:
Since OliveModel class is getting more and more AML model related attribute, we may need a more structured AML model class and local model class. 


examples/bert_hf_cpu_aml/bert_config.json
```diff
@@ -1,11 +1,17 @@
...
+        "resource_group": "<place_holder>",
+        "workspace_name": "<place_holder>"
+    },
     "input_model":{
         "type": "PyTorchModel",
         "config": {
             "name": "bert-hf",
             "version": "3",
             "model_storage_kind": "azureml",
+            "aml_storage_name": "bert_hf",
```
trajepl, line 14: How about move aml_client info from engine entry to input model's property, Like:
image

olive/engine/engine.py
```diff
@@ -207,6 +208,7 @@ def run(
         self,
         input_model: OliveModel,
         packaging_config: Optional[PackagingConfig] = None,
+        aml_client: Optional[AzureMLClient] = None,
```
guotuofeng, line 211: do we really need to make the aml_client concept known to engine? 

Ideally, only model.py should download the model when the model is loaded.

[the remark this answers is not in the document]
xiaoyu-work: Yeah there are 2 options here: either put aml_client to model or engine. For me, Engine owns this makes more sense than model. Another use case for Engine will be that Engine will be responsible for packaging output models to AML in the near future. Then Engine will still need AML client.

xiaoyu-work: Another use case for aml_client for Engine will be AML packaging. Engine will need to have aml_client to package out models to AML with AML_client. This part of logic is not sent as a PR for now but will be needed in the near future.

devang-ml (commented):
We need to support packaging output models from Footprints to AML.

olive/engine/engine.py
guotuofeng: could we make the aml_client part of the packaging logic instead of engine?

olive/workflows/run/config.py
```diff
@@ -136,6 +138,10 @@ def _resolve_config_str(v, values, alias, component_name="systems"):
     components = values[component_name] or {}
     if sub_component not in components:
         raise ValueError(f"{alias} {sub_component} not found in systems")
+    if component_name == "systems" and components[sub_component].type == "AzureML":
```
jambayk, line 141: can you move this auto insert to `_resolve_system` at line `149`? 

I think it can be something like 
```
def _resolve_system(v, values, system_alias, component_name="systems"):
    v = _resolve_config_str(v, values, system_alias, component_name=component_name) 
    v[system_alias] = validate_config(v[system_alias], SystemConfig)
    if v[system_alias].type == "AzureML":
         .... auto insert code .... 
``` 

Otherwise, it will only be auto inserted if the system is referenced by name. We also accept system config dicts directly.

xiaoyu-work: Sure will do. This default `component_name="systems"` confused me a lot. If there is no specific use case here, I'll remove this.

jambayk: Yes, it will always be "systems" for `_resolve_system`, so it is redundant. 

`_resolve_config_str` also has direct references to `system` which should actually be `component_name`. Looks like the PR that generalized the `resolve_config_str` code missed them.

xiaoyu-work: Yeah I mean the default value in `_resolve_config_str` for system. We also use it for data_config right now. I've removed all system keyword.

jambayk: Thanks! it looks good to me now

jambayk:
One of the ci jobs cancelled by itself. I think you will need to push an empty commit to retrigger the ci pipelines.

olive/azureml/azureml_client.py
```diff
@@ -0,0 +1,89 @@
...
+                raise ValueError(f"aml_config_path {v} does not exist")
+            if not Path(v).is_file():
+                raise ValueError(f"aml_config_path {v} is not a file")
+        return v
+
+    def create_client(self):
+        return AzureMLClient(**self.dict())
+
+
+class AzureMLClient(object):
```
guotuofeng, line 45: do we really need AzureMLClien? or we can directly return MLClient in create_client?

olive/engine/engine.py
```diff
@@ -668,9 +676,15 @@ def _run_passes(
                 model.model_storage_kind == ModelStorageKind.AzureMLModel
                 and not self.host_for_pass(pass_id).system_type == SystemType.AzureML
             ):
-                error_msg = "Azure ML model only supports AzureMLSystem for Olive Pass"
-                logger.error(error_msg)
-                raise Exception(error_msg)
+                model_download_path = self._model_cache_path / "azureml_input_model"
+                model_path = model.download_from_azureml(self.azureml_client, model_download_path)
```
guotuofeng, line 680: >model [](http://example.com/codeflow?start=29&length=5)

what about the input model is an onnx model?

olive/model.py
```diff
@@ -104,6 +107,25 @@ def prepare_session(
         """
         raise NotImplementedError()
 
+    def download_from_azureml(self, azureml_client: AzureMLClient, download_path: Union[Path, str]):
```
guotuofeng, line 110: >azureml_client [](http://example.com/codeflow?start=36&length=14)

use MLClient?

olive/workflows/run/run.py
```diff
@@ -124,6 +124,10 @@ def run(config: Union[str, Path, dict], setup: bool = False):
     # input model
     input_model = config.input_model.create_model()
 
+    # Azure ML Client
+    if config.azureml_client:
+        config.engine.azureml_client = config.azureml_client
```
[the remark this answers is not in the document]
guotuofeng: shouldn't the config happens in config json?

olive/model.py
xiaoyu-work: AzureMLClient is a wrapper here for MLClient. Otherwise if we specify MLClient type hint we will need to import MLClient to model.py

olive/engine/engine.py
```diff
@@ -55,6 +57,7 @@ def __init__(
         host: Optional[OliveSystem] = None,
         target: Optional[OliveSystem] = None,
         evaluator_config: Optional[OliveEvaluatorConfig] = None,
+        azureml_client: Optional[AzureMLClientConfig] = None,
```
guotuofeng, line 60: >azureml_client [](http://example.com/codeflow?start=8&length=14)

seems useless?

xiaoyu-work: Oh yeah good catch! Will remove this.

xiaoyu-work: Yeah good catch! Will remove this.

olive/model.py
guotuofeng: We could use "MLClient" as the type hints.

xiaoyu-work: Yeah that's what I mean. MLClient is from azure ai package "from azure.ai.ml import MLClient", we don't want to import this in model module level.

olive/workflows/run/run.py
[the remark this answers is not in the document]
guotuofeng: could not the following works?

"azureml_client": {
    "subscription_id": "<place_holder>",
    "resource_group": "<place_holder>",
    "workspace_name": "<place_holder>"
},
"engine": {
        "search_strategy": {
            "execution_order": "joint",
            "search_algorithm": "random",
            "search_algorithm_config": {
                "num_samples": 3,
                "seed": 0
            }
        },
        "evaluator": "common_evaluator",
        "host": "local_system",
        "target": "local_system",
        "azureml_client": "azureml_client"
        "clean_cache": true,
        "cache_dir": "cache"
    }

olive/model.py
guotuofeng: we need not import the MLClient. We could either use string or just leave the type hints empty here

olive/workflows/run/run.py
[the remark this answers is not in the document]
guotuofeng: Please consider that case that there are multiple workspaces in one olive run in future since we all not sure the assumption that only use one workspace is right.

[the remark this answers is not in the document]
xiaoyu-work: I read that aml documentation. So the question here is, will Olive handle the sharing operation? Like if the user wants to use the model registered in another workspace, should they complete the sharing operation already?

olive/model.py
xiaoyu-work: I don't like a placeholder string or empty type hint here. Any reason that model download should not know AzureMLClient?

guotuofeng: Because model will have knowledge about the implementation detail of AzureMLClient. This is a design flaw we need avoid.

olive/workflows/run/run.py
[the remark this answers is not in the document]
guotuofeng: not just about share model. Please consider the case: model is registered in workspace A. but compute target is in workspace B. 

Since no one know whether our current assumption is correct or not, we'd better to keep that in mind when we design/implement to consider the extensibility.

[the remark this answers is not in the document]
xiaoyu-work: I just realize that the assets sharing is between workspace and registry but not workspace and workspace. So in this case there will only be one workspace in total.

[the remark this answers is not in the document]
xiaoyu-work: We can add support registry support into AMLClient in the future.

olive/model.py
guotuofeng: [TYPE_CHECKING](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) should be the thing your are looking for.

olive/engine/engine.py
```diff
@@ -668,9 +675,19 @@ def _run_passes(
...
                 and not self.host_for_pass(pass_id).system_type == SystemType.AzureML
             ):
-                error_msg = "Azure ML model only supports AzureMLSystem for Olive Pass"
-                logger.error(error_msg)
-                raise Exception(error_msg)
+                if not self.azureml_client_config:
+                    raise ValueError("AzureML client config is required to download the model from AzureML storage")
+                model_download_path = self._model_cache_path / "azureml_input_model"
+                model_path = model.download_from_azureml(
+                    self.azureml_client_config.create_client(), model_download_path
```
guotuofeng, line 682: better to move out of the for loop?

xiaoyu-work: Since each pass and evaluator can have its own host and target, I was thinking one case that all previous passes and evaluations are running on AML but one of the middle are running locally. This is to handle that case so I put this in the loop.

guotuofeng (dismissed):

olive/engine/engine.py
xiaoyu-work: Although currently, we will need to downlaod the model from the AML after the first pass run on AML, this is a gap from AML side but not our side. We will support this logic in the future once AML fixes this.

olive/model.py
```diff
@@ -10,14 +10,16 @@
...
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Union
 
 import onnx
 import torch
 import yaml
 from onnx import AttributeProto, GraphProto
 from pydantic import validator
 
+if TYPE_CHECKING:
```
trajepl, line 21: Nice~!

olive/systems/azureml/aml_system.py
```diff
@@ -33,17 +33,17 @@ class AzureMLSystem(OliveSystem):
 
     def __init__(
         self,
-        aml_config_path: str,
+        azureml_client_config: AzureMLClientConfig,
```
trajepl, line 36: Should we also keep the legacy `aml_config_path` before the releasement?

trajepl (dismissed):

olive/systems/azureml/aml_system.py
xiaoyu-work: Even we keep this, aml system wont't work with aml_config_path as we changed the mlclient creation logic inside. Do we want to keep this?

guotuofeng (dismissed):

trajepl (approved):

PR #274, opened 2023-05-17
fix azureml-evaluate-mlflow version

leqiao-1:
## Describe your changes
fix azureml-evaluate-mlflow version
azureml-evaluate-mlflow changed model_loader flavor name from `hftransformers` to `hftransformersv2` when updating to 0.8.0. Current ci test may fail when loading pytorch mlflow model. This is a workaround to make sure the test won't fail recently.

### DIFF
--- a/test/requirements-test.txt
+++ b/test/requirements-test.txt
@@ -14,5 +14,5 @@ psutil
 pytorch_lightning
 torchvision
 datasets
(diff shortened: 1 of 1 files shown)


trajepl:
Which kind of error would happen if we do not pin the version? Could you help update it in PR description? May help better tracking further.

PR #275, opened 2023-05-17
Support for evaluating distributed onnx models

shaahji:
Support for evaluating distributed onnx models

### DIFF
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -5,7 +5,7 @@
 import logging
 import time
 from abc import ABC, abstractmethod
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Tuple
 
 import numpy as np
 import torch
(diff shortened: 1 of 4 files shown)


olive/evaluator/olive_evaluator.py
```diff
@@ -275,6 +275,172 @@ def _evaluate_latency(
...
+
+        os.environ["OMPI_COMM_WORLD_RANK"] = str(local_rank)
+        os.environ["OMPI_COMM_WORLD_SIZE"] = str(world_size)
+
+        from mpi4py import MPI
+
+        local_rank = MPI.COMM_WORLD.Get_rank()
+
+        inference_settings["execution_provider"] = ["CUDAExecutionProvider", "CPUExecutionProvider"]
+        inference_settings["provider_options"] = [{"device_id": str(local_rank)}, {}]
```
devang-ml, line 297: The EP should be selected based on the accelerator_spec provided by the Engine to the evaluate_method. Right now it does not, so please add a FIXME note here.

shaahji: Do we have the accelerator_spec functionality in the current implementation or is this something of a TODO?

shaahji: Done!

PR #277, opened 2023-05-17
Auto remove unit test resulting files

trajepl:
## Describe your changes
When I run unit test locally, always generate many temp files in root folder. It often breaks my git track for changed files.

This PR is used to move those temp files into temp folder and remove them when test is done.

### DIFF
--- a/olive/passes/pytorch/quantization_aware_training.py
+++ b/olive/passes/pytorch/quantization_aware_training.py
@@ -86,6 +86,7 @@ class QuantizationAwareTraining(Pass):
             ),
             "gpus": PassConfigParam(type_=int, description="Number of GPUs to use."),
             "seed": PassConfigParam(type_=int, default_value=None, description="Random seed for training."),
+            "checkpoint_path": PassConfigParam(type_=str, description="Path to save checkpoints."),
         }
 
     def _run_for_config(self, model: PyTorchModel, config: Dict[str, Any], output_model_path: str) -> PyTorchModel:
--- a/olive/passes/pytorch/quantization_aware_training.py
+++ b/olive/passes/pytorch/quantization_aware_training.py
@@ -86,7 +86,7 @@ class QuantizationAwareTraining(Pass):
             ),
             "gpus": PassConfigParam(type_=int, description="Number of GPUs to use."),
             "seed": PassConfigParam(type_=int, default_value=None, description="Random seed for training."),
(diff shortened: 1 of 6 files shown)


test/unit_test/engine/test_engine.py
```diff
@@ -212,10 +214,13 @@ def test_pass_exception(self, caplog):
             model = PyTorchModel(model_loader=pytorch_model_loader, model_path=None)
 
             # execute
-            engine.run(model)
+            output_dir = Path(tempfile.TemporaryDirectory().name)
```
jambayk, line 217: looks like this is not actually using a temporary directory? You get a tempfile name but the actual tempfile gets deleted right away since there is no context manager or there is no reference to the temporary directory object 

It is not an issue since `output_dir` gets cleaned up manually. But will it reach the clean up step at 223 if the assert step fails?

jambayk: if you keep a reference to the actual object 
```
temp_dir = tempfile.TemporaryDirectory()
output_dir = Path(tempfile.name)
``` 

It will clean up by itself when it goes out of scope. Or you can manually clean up at the end to be safe `temp_dir.cleanup()`

trajepl: Good catche. Updated.

olive/passes/pytorch/quantization_aware_training.py
```diff
@@ -86,6 +86,7 @@ def _default_config() -> Dict[str, PassConfigParam]:
             ),
             "gpus": PassConfigParam(type_=int, description="Number of GPUs to use."),
             "seed": PassConfigParam(type_=int, default_value=None, description="Random seed for training."),
+            "checkpoint_path": PassConfigParam(type_=str, description="Path to save checkpoints."),
```
jambayk, line 89: if this is None, can the trainer handle it?

trajepl: Yes. The trainer will call os.getcwd to set as current folder. 
Do you suggest to set it as None by default?

trajepl: Updated.

jambayk: I see. thanks for the info. 
I just wanted to make sure the trainer can handle None. The default value here is already None since the default for `default_value` is None

jambayk (approved):

PR #255, opened 2023-05-10
Enable hardware accelerator in Olive

guotuofeng:
## Describe your changes

Enable hardware accelerator for Olive

## Checklist before requesting a review
- [x] Add unit tests for this change.
- [x] Make sure all tests can pass.
- [x] Update documents if necessary.
- [x] Format your code by running `pre-commit run --all-files`

PR #284, opened 2023-05-19
Update minimum ORT version for DML attention optimizations

jstoecker:
## Describe your changes

Currently, the transformer optimizer will only perform certain optimizations if the ORT version is 1.15.1. However, the dependencies are now *expected* to be available in ORT version 1.15.0. I will leave this PR as a draft until the ORT 1.15.0 release candidate is available and verified.

- [X] Update documents if necessary.
- [X] Format your code by running `pre-commit run --all-files`


examples/directml/stable_diffusion/config_safety_checker.json
```diff
@@ -56,7 +56,27 @@
                 "float16": true,
                 "use_gpu": true,
                 "keep_io_types": false,
-                "target_provider": "directml_future"
+                "optimization_options": {
```
PatriceVignola, line 59: Whatever is in the `examples` folder is not part of the olive package, right? Do users really have to manually set which fusions they want and which ones they don't want? It feels like we should be deciding which fusions are appropriate for DML.

jstoecker: Correct, examples are not published as part of olive-ai package.

I agree with you that Olive should ideally decide fusions in a more automatic fashion, but the logic I added to OrtTransformerOptimizer is a bit dubious long term: it relies on a special "target_provider" parameter that will be obsolete with some upcoming changes (e.g. see #255). I want to revisit this in the future when accelerators are integrated.

PatriceVignola (approved):

examples/directml/stable_diffusion/stable_diffusion.py
```diff
@@ -179,7 +183,7 @@ def optimize(
         olive_run(olive_config)
 
         footprints_file_path = (
-            Path(__file__).resolve().parent / "footprints" / f"{submodel_name}_cpu-cpu_footprints.json"
+            Path(__file__).resolve().parent / "footprints" / f"{submodel_name}_gpu-dml_footprints.json"
```
guotuofeng, line 186: >gpu-dml_footprints.json" [](http://example.com/codeflow?start=79&length=24)

I could not verify the stable diffusion scenario. 

Does this change works for dml after the accelerator feature is in main?

jstoecker: It does! I really need to write a test for these scripts soon. Optimization shouldn't really require DML EP or a GPU, but the latency evaluator _does_ require DML. I'll try to find a way to have a test that skips evaluation but otherwise produces valid models for DML EP.

PR #285, opened 2023-05-19
Unify model path configurations using `ResourcePath`

jambayk:
## Describe your changes
This PR introduces a new class called `olive.resource_path.ResourcePath` that is able to handle many different types of model sources, both local and remote. 
Currently, the model path/source is specified through multiple parameters `model_path`, `name`, `version`, `model_storage_kind`. However, there is no systematic way in which these parameters interact. They were added as new features were implemented. 

Using `ResourcePath`, we can unify all of these into a single class that can be used by all model classes. 
The supported resource types are local file, local folder, string name, azureml model, azureml datastore and azureml job outputs. We can add other sources easily. 

We also remove `name` from `OliveModel` since it doesn't have any use and might be a point of confusion. 


test/unit_test/systems/azureml/output_pass/pipeline_output/named-outputs/pipeline_output/output_model.onnx
guotuofeng, line 1: could we not check-in binary files?

jambayk: Done. I also deleted the `.json` file in the same folder. The files are created during test time now.

olive/engine/engine.py
```diff
@@ -758,15 +764,19 @@ def _run_pass(
         # new model id
         input_model_number = input_model_id.split("_")[0]
         output_model_id = f"{self._get_new_model_number()}_{pass_name}-{input_model_number}-{hash_dict(pass_config)}"
-        output_model_path = str(self._model_cache_path / f"{output_model_id}")
+        output_model_path = self._model_cache_path / f"{output_model_id}" / "output_model"
```
guotuofeng, line 767: the output_model_id need to be modified to add the accelerator suffix if the pass is not accelerator agnostic. 
https://github.com/microsoft/Olive/pull/286. 

How do we handle this scenario if the output model name changes from model id to hardcode output_model?

olive/resource_path.py
```diff
@@ -0,0 +1,387 @@
...
+        return self.value
+
+
+LOCAL_RESOURCE_TYPES = [ResourceType.LocalFile, ResourceType.LocalFolder, ResourceType.StringName]
+AZUREML_RESOURCE_TYPES = [ResourceType.AzureMLModel, ResourceType.AzureMLDatastore, ResourceType.AzureMLJobOutput]
+
+_resource_config_registry: Dict[ResourceType, Type[ConfigBase]] = {}
+
+
+class ResourceConfig(ConfigBase):
```
guotuofeng, line 41: could we unify the hf_config by using the ResourceConfig?

olive/resource_path.py
```diff
@@ -0,0 +1,387 @@
...
+        if not inspect.isabstract(cls) and cls._type is not None:
+            _resource_config_registry[cls._type] = cls
+
+    @abstractmethod
+    def get_path(self) -> str:
+        """Return the resource path as a string."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: bool = False) -> str:
```
guotuofeng, line 57: is download better?

guotuofeng, line 57: force better?

olive/resource_path.py
```diff
@@ -0,0 +1,387 @@
...
+                    ml_client.models.download,
+                    [self.name],
+                    {"version": self.version, "download_path": temp_dir},
+                    max_tries=3,
+                    delay=5,
+                    exceptions=ServiceResponseError,
+                )
+                new_path.parent.mkdir(parents=True, exist_ok=True)
+                shutil.move(temp_dir / self.name / model_path.name, new_path)
+        except Exception as e:
```
guotuofeng, line 282: the try/catch logging is needless since the caller will catch the exception and handle it

olive/systems/azureml/aml_system.py
```diff
@@ -136,47 +144,59 @@ def _create_model_args(self, model_json: dict, tmp_dir: Path):
...
+            model_resource_path = ResourcePath.create_resource_path(model_json["config"]["model_path"])
+            asset_type = RESOURCE_TYPE_TO_ASSET_TYPE[model_resource_path.type]
+
+            if model_resource_path.type in AZUREML_RESOURCE_TYPES:
+                # ensure that the model is in the same workspace as the system
+                model_workspace_config = model_resource_path.config.azureml_client.get_workspace_config()
+                system_workspace_config = self.azureml_client_config.get_workspace_config()
+                for key in model_workspace_config:
+                    if model_workspace_config[key] != system_workspace_config[key]:
+                        raise Exception(
```
guotuofeng, line 157: better not raise Exception

olive/workflows/snpe/convertquantize/convertquantize.py
```diff
@@ -44,10 +44,10 @@ def convertquantize(
...
     if model_file.suffix == ".onnx":
         logger.info("Loading model...")
-        model = ONNXModel(model_file, name)
+        model = ONNXModel(model_file)
     elif model_file.suffix == ".pb":
         logger.info("Loading model...")
-        model = TensorFlowModel(model_file, name)
+        model = TensorFlowModel(model_file)
     else:
         raise Exception(f"Unsupported model format: {model_file.suffix}")
```
guotuofeng, line 52: raise ValueError?

test/integ_test/aml_model_test/test_aml_model.py
```diff
@@ -61,7 +49,38 @@ def test_aml_model():
...
+            "output_names": ["output"],
+        },
+    )
+    return pytorch_model
+
+
+def get_olive_workspace_config():
     subscription_id = os.environ.get("WORKSPACE_SUBSCRIPTION_ID")
     if subscription_id is None:
         raise Exception("Please set the environment variable WORKSPACE_SUBSCRIPTION_ID")
```
guotuofeng, line 86: ditto

test/integ_test/evaluator/azureml_eval/utils.py
```diff
@@ -128,7 +125,7 @@ def get_aml_target():
     )
 
 
-def generate_olive_workspace_config(workspace_config_path):
+def get_olive_workspace_config():
     subscription_id = os.environ.get("WORKSPACE_SUBSCRIPTION_ID")
     if subscription_id is None:
         raise Exception("Please set the environment variable WORKSPACE_SUBSCRIPTION_ID")
```
guotuofeng, line 131: not raise Exception

olive/resource_path.py
jambayk: `hf_config` is not only used to load models but also hold information about the hf configuration of the original model. It is also inherited by output models. So, it is not the same as resource path. 

You can see here that the model_path can be used to load the model even when there is hf config https://github.com/microsoft/Olive/blob/main/olive/model.py#L568/ 

I think we can make changes in a separate PR if we want to rethink this part.

olive/resource_path.py
jambayk: I think `overwrite` describes the behavior better?

olive/resource_path.py
jambayk: I added the try except for logging the error but I can remove it too. Do you think the error logging is not needed?

olive/systems/azureml/aml_system.py
jambayk: What error should I raise? ValueError? We need them to be in the same workspace for this to work.

olive/resource_path.py
guotuofeng: If the exception is not rethrown, the catch and log is fine. I prefer remove it because the caller will catch the rethown exception and do error handling. From this point of view, it seems the logging is duplicated?

olive/systems/azureml/aml_system.py
guotuofeng: You can find the appropriated exception at https://docs.python.org/3/library/exceptions.html#exception-hierarchy.

Some best practice is https://stackoverflow.com/questions/2052390/manually-raising-throwing-an-exception-in-python.

olive/engine/engine.py
guotuofeng: How could my PR for reruse run logic looks like by including your PR?

olive/resource_path.py
guotuofeng: We can reconsider whether hf_config could be aligned with the resource path concept in future.

olive/engine/engine.py
jambayk: There is no effect as we discussed offline. Resolving this conversation.

olive/systems/azureml/aml_system.py
jambayk: Updated to `ValueError` since it depends on the input model.

olive/workflows/snpe/convertquantize/convertquantize.py
jambayk: Will address in the future when we refactor the other exceptions in Olive.

test/integ_test/aml_model_test/test_aml_model.py
jambayk: Will address in the future when we refactor the other exceptions in Olive.

test/integ_test/evaluator/azureml_eval/utils.py
jambayk: Will address in the future when we refactor the other exceptions in Olive.

olive/resource_path.py
jambayk: Done. I removed the try/except part since the logging is redundant as you mentioned.

olive/resource_path.py
```diff
@@ -0,0 +1,379 @@
...
+        dir_path = Path(dir_path).resolve()
+        dir_path.mkdir(parents=True, exist_ok=True)
+
+        # path to save the resource to
+        if name:
+            new_path_name = Path(name).with_suffix(self.path.suffix).name
+        else:
+            new_path_name = self.path.name
+        new_path = dir_path / new_path_name
+        _overwrite_helper(new_path, overwrite)
```
trajepl, line 158: If overwrite is false and the file is already dumped to local file system, should be raise the exception to stop olive?
How about to catch the exception from this method?

olive/resource_path.py
```diff
@@ -0,0 +1,379 @@
...
+    )
+
+
+class AzureMLModelConfig(ResourceConfig):
+    """Config for an AzureML model."""
+
+    _type = ResourceType.AzureMLModel
+    azureml_client: AzureMLClientConfig = Field(..., description="AzureML client config.")
+    name: str = Field(..., description="Name of the model.")
+    version: Union[int, str] = Field(..., description="Version of the model.")
```
trajepl, line 240: Double check: will the version be required for AML model? What if None?

olive/resource_path.py
```diff
@@ -0,0 +1,379 @@
...
+            ml_client.models.create_or_update,
+            [
+                Model(
+                    path=self.get_path(),
+                    name="olive-backend-model",
+                    description="Model created by Olive backend. Ignore this model.",
+                    type=AssetTypes.CUSTOM_MODEL,
+                )
+            ],
+            max_tries=3,
```
trajepl, line 323: How about to extract the `max_tries` or `delay` to default configs? If there is a few network issue, use can adjust the numbers.

olive/systems/azureml/aml_system.py
```diff
@@ -21,13 +21,23 @@
...
-from olive.model import ModelConfig, ModelStorageKind, OliveModel, ONNXModel
+from olive.model import ModelConfig, OliveModel
 from olive.passes.olive_pass import Pass
+from olive.resource_path import AZUREML_RESOURCE_TYPES, AzureMLModelConfig, ResourcePath, ResourceType
 from olive.systems.common import AzureMLDockerConfig, SystemType
 from olive.systems.olive_system import OliveSystem
 
 logger = logging.getLogger(__name__)
 
+RESOURCE_TYPE_TO_ASSET_TYPE = {
```
trajepl, line 32: How about maintain this map under `resource_type.py`?

olive/systems/azureml/aml_system.py
```diff
@@ -138,47 +146,59 @@ def _create_model_args(self, model_json: dict, tmp_dir: Path):
...
+                    ml_client.models.create_or_update,
+                    [
+                        Model(
+                            path=model_resource_path.get_path(),
+                            name="olive-backend-model",
+                            description="Model created by Olive backend. Ignore this model.",
+                            type=AssetTypes.CUSTOM_MODEL,
+                        )
+                    ],
+                    max_tries=3,
```
trajepl, line 181: Ditto.

olive/systems/azureml/aml_system.py
```diff
@@ -21,13 +21,23 @@
...
+from olive.resource_path import AZUREML_RESOURCE_TYPES, AzureMLModelConfig, ResourcePath, ResourceType
 from olive.systems.common import AzureMLDockerConfig, SystemType
 from olive.systems.olive_system import OliveSystem
 
 logger = logging.getLogger(__name__)
 
+RESOURCE_TYPE_TO_ASSET_TYPE = {
+    ResourceType.LocalFile: AssetTypes.URI_FILE,
+    ResourceType.LocalFolder: AssetTypes.URI_FOLDER,
+    ResourceType.StringName: AssetTypes.URI_FILE,  # this is a place holder since we won't upload a string as input
```
trajepl, line 35: Could we set the value of `StringName` as None to stand for place holder? Or just remove this line?

olive/workflows/run/config.py
```diff
@@ -65,8 +66,33 @@ class RunConfig(ConfigBase):
...
+        input_model_path = v["config"].get("model_path")
+        # if not a dict, return the original value
+        if not input_model_path or not isinstance(input_model_path, dict):
+            return v
+
+        resource_path_type = input_model_path.get("type")
+        # insert aml_client if resource_path_type is azureml and not already set
+        if resource_path_type in AZUREML_RESOURCE_TYPES:
+            model_aml_client = input_model_path.get("config", {}).get("azureml_client")
+            print("here", model_aml_client)
```
trajepl, line 83: Remove debug line.

olive/workflows/run/config.py
```diff
@@ -65,8 +66,33 @@ class RunConfig(ConfigBase):
...
+            if not model_aml_client:
+                if "azureml_client" not in values:
+                    raise ValueError("azureml_client is required for azureml resource path")
+                v["config"]["model_path"]["config"]["azureml_client"] = values["azureml_client"]
+        print(v)
```
trajepl, line 88: Ditto.

docs/source/api/models.rst
```diff
@@ -10,12 +10,12 @@ ONNX Model
 
 .. _composite_onnx_model:
 CompositeOnnxModel Model
-----------
+--------------------------
```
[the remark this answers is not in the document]
jambayk: It is not an error but sphinx gives warning if the underlines are shorter than the text. You can see the rule here https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#headings

examples/bert_inc_ptq_cpu/bert_inc_config.json
```diff
@@ -3,7 +3,6 @@
     "input_model":{
         "type": "PyTorchModel",
         "config": {
-            "model_path": null,
```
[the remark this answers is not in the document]
jambayk: Yes, the `model_path` is optional or can be `None` in general. If it is required for a specific type like ONNXModel, then it's class init has it as required.

olive/resource_path.py
```diff
@@ -0,0 +1,379 @@
...
+        return self.value
+
+
+LOCAL_RESOURCE_TYPES = [ResourceType.LocalFile, ResourceType.LocalFolder, ResourceType.StringName]
+AZUREML_RESOURCE_TYPES = [ResourceType.AzureMLModel, ResourceType.AzureMLDatastore, ResourceType.AzureMLJobOutput]
+
+_resource_config_registry: Dict[ResourceType, Type[ConfigBase]] = {}
+
+
+class ResourceConfig(ConfigBase):
```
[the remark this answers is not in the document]
jambayk: This is a good idea. We already have autoconfigclass and I replicated the behavior with a different code style. I will change the `ResourcePath` class to `AutoConfigClass`.

olive/resource_path.py
jambayk: I think it is better to fail here? We cannot guarantee that the file present already in the save_dir is the same as the file we want to download. 
The olive engine will never encounter this failure since it handles model download using cache and also sets `overwrite` to `True` when this method is used.

olive/resource_path.py
jambayk: Yes version is required. I think that is safer since if we default to something like `latest`, we cannot guarantee caching. The `latest` version might change.

olive/systems/azureml/aml_system.py
jambayk: I kept this here since the azureml packages are optional and we cannot have this mapping in the `resource_paths.py` file. It is also only relevant to the azureml system so I think it is better for the azureml system to maintain this mapping?

olive/systems/azureml/aml_system.py
jambayk: In methods like `create_model_inputs`, it uses the mapped values so `None` is not accepted.

olive/resource_path.py
jambayk: Done. Exposed in `AzureMLClientConfig`

olive/engine/engine.py
```diff
@@ -810,15 +816,19 @@ def _run_pass(
...
+        output_model_path = str(output_model_path)
 
         # prune if invalid search_point
         if not p.validate_search_point(pass_search_point) and not self.no_search:
             output_model = PRUNED_CONFIG
         else:
             # run pass
+            host = self.host_for_pass(pass_id)
+            if host.system_type != SystemType.AzureML:
+                input_model = self._prepare_non_local_model(input_model)
```
guotuofeng, line 830: need the input_model_id be updated?

olive/engine/engine.py
```diff
@@ -608,7 +614,7 @@ def _get_new_model_number(self):
         while True:
             new_model_number = self._new_model_number
             self._new_model_number += 1
-            if list(self._model_cache_path.glob(f"{new_model_number}_*.json")) == []:
```
guotuofeng, line 611: what's the consideration to remove the file extension? would it introduce any side effect?

olive/engine/engine.py
jambayk: The input model_id is the same. What this method does is to download the remote resource (or load from cache) and update the model object's `local_resource_path` attribute. this attribute is used to get `model.model_path`

olive/systems/azureml/aml_pass_runner.py
```diff
@@ -103,12 +102,24 @@ def main(raw_args=None):
     # this is to handle passes like OrtPerfTuning that use the same model file as input
     model_json["same_model_path_as_input"] = False
     if model_json["config"]["model_path"] is not None:
-        model_path = str(Path(model_json["config"]["model_path"]).resolve())
-        if model_path == input_model_path:
+        # create a resource path from the model path
+        model_resource_path = ResourcePath.create_resource_path(model_json["config"]["model_path"])
```
guotuofeng, line 106: do we need assert the model_resource_path is local resource?

jambayk: It gets checked indirectly when the pass, model or evaluator calls `model.model_path` https://github.com/microsoft/Olive/blob/jambayk/model-source/olive/model.py#L67

olive/systems/azureml/aml_pass_runner.py
```diff
@@ -103,12 +102,24 @@ def main(raw_args=None):
...
+        # string representation of the model path
+        model_path_str = model_resource_path.get_path()
+        if model_path_str == input_model_path:
+            # if the model path is the same as the input model path, set model_path to None
+            # and set same_model_path_as_input to True
             model_json["config"]["model_path"] = None
             model_json["same_model_path_as_input"] = True
-        else:
-            model_json["config"]["model_path"] = str(Path(model_path).relative_to(Path(common_args.pipeline_output)))
+        elif model_resource_path.type in [ResourceType.LocalFile, ResourceType.LocalFolder]:
```
guotuofeng, line 114: I am supposing the model_resource_path should be always local path here?

jambayk: It can also be a `ResourceType.StringName`

docs/source/api/models.rst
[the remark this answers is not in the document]
trajepl: Thanks!

olive/systems/azureml/aml_pass_runner.py
jambayk: Currently, all our passes create model with local files/folders so the StringName case doesn't happen. But I think it is still better to cover all cases?

olive/systems/azureml/aml_system.py
```diff
@@ -323,40 +343,23 @@ def _load_model(self, input_model, output_model_path, pipeline_output_path):
...
-                if Path(output_model_path).suffix != Path(downloaded_model_path).suffix:
-                    output_model_path += Path(downloaded_model_path).suffix
-                if downloaded_model_path.is_file():
-                    # model is a file
-                    shutil.copy(downloaded_model_path, output_model_path)
-                else:
-                    # model is a directory
-                    shutil.copytree(downloaded_model_path, output_model_path, dirs_exist_ok=True)
-            model_path = output_model_path
+                model_path = model_json["config"]["model_path"]
```
guotuofeng, line 362: when the branch could be reached? I am supposing the local branch should always be hit.

jambayk: This is the same case as the one described in the previous comment https://github.com/microsoft/Olive/pull/285#discussion_r1203259172 

We don't have any passes currently that generate such models, but the model path could also be a string name. So this case is for completeness.

olive/resource_path.py
jambayk: I originally created the `_overwrite_helper` function to abstract common code that is used by all resource path types. so this logic would otherwise been directly inside the `save_to_dir` function. 

I can rename the `_overwrite_helper` function if that helps? With regards to the behavior, I think we should fail if overwrite=False and a similarly named file/folder already exists.

olive/systems/azureml/aml_system.py
trajepl: Asking this is because, we clone all ResourceType and set the mapped aml asset types. When there is any other similar requests, we may need to maintain that again in other system.

But currently it looks ok, as only aml is using that and we need to resolve the issue of packaging dependency. Thanks.

jambayk: Sounds good to me. 

Originally, I was thinking about putting the responsibility of creating aml input and args on the resource_paths so that we can add more resource types without having to touch the aml system. Maybe we can see if that's a better idea in the future?

olive/systems/azureml/aml_system.py
trajepl: Ic, you may do not want to add too many condition when create the model input. 

Then it is fine.

olive/systems/azureml/aml_system.py
trajepl: Yes. Sounds good.

olive/resource_path.py
trajepl: I do not touch too much detailed code. But what if if path exists, we force the overwrite to true?

jambayk: if `overwrite` is true and file/folder exists, we delete the existing resource and write the new one to the same path. `_overwrite_helper` does the deletion. the new files are saved by the code after the `_overwrite_helper` is called.

olive/systems/azureml/aml_system.py
guotuofeng: I think the output model should always be local file or local folder. right?

olive/systems/azureml/aml_pass_runner.py
guotuofeng: I could not imagine which case could generate the output model with only string names.

olive/systems/azureml/aml_pass_runner.py
jambayk: Done

olive/systems/azureml/aml_pass_runner.py
jambayk: Done. asserts that it is local as discussed offline

olive/systems/azureml/aml_system.py
jambayk: Done. asserts that it is local as discussed offline

olive/resource_path.py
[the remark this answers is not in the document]
jambayk: Done.

olive/cache.py
```diff
@@ -109,10 +110,49 @@ def clean_pass_run_cache(pass_type: str, cache_dir: Union[str, Path] = ".olive-c
         _delete_run(run_json.stem, cache_dir)
 
 
+def get_non_local_resource(resource_path: ResourcePath, cache_dir: Union[str, Path] = ".olive-cache"):
```
guotuofeng, line 113: add unit test for the get_non_local_resource?

jambayk: Done

olive/engine/engine.py
```diff
@@ -138,16 +138,16 @@ def initialize(self):
         """
         Initialize engine state. This should be done before running the registered passes.
         """
-        cache_dir = self._config.cache_dir
+        self.cache_dir = self._config.cache_dir
```
guotuofeng, line 141: do we need make cache_dir as part of the engine attribute? it seems the save_model still use self._config.cache_dir?

olive/engine/engine.py
```diff
@@ -810,15 +820,19 @@ def _run_pass(
             accelerator_spec,
         ]
         output_model_id = "-".join(map(str, output_model_id_parts))
-        output_model_path = str(self._model_cache_path / f"{output_model_id}")
+        output_model_path = self._model_cache_path / f"{output_model_id}" / "output_model"
+        output_model_path.parent.mkdir(parents=True, exist_ok=True)
+        output_model_path = str(output_model_path)
```
guotuofeng, line 825: do we need the conversion?

olive/model.py
```diff
@@ -709,13 +724,11 @@ def to_json(self, check_object: bool = False):
 
 
 class OptimumModel(OliveModel):
-    def __init__(self, model_path: str, name: str, model_components: List[str]):
+    def __init__(self, model_path: str, model_components: List[str]):
```
guotuofeng, line 727: we only support string path here. right?

olive/engine/engine.py
jambayk: Oh yeah, I forgot about `self._config.cache_dir`. I will revert this change and use `self._config.cache_dir` in the `cache_utils.get_non_local_resource` call.

olive/engine/engine.py
jambayk: Conversion to string? Yes, since the `pass._run_for_config` signature has the expected type as string. The pass implementations also assume it is provided as string`.

olive/model.py
jambayk: good catch! This was added in the middle of my PR so I missed updating here. Done

olive/engine/engine.py
jambayk: Done

olive/engine/engine.py
guotuofeng: I assume pass the Path will still work for all passes

guotuofeng: It doesn't matter. it is not big deal

guotuofeng (approved):

guotuofeng (approved):

trajepl:
The failure was caused by timeout when run component detection.  @jambayk Please rebase main branch with latest changes which Lei fixed.

PR #286, opened 2023-05-22
Allow pass run could be reused cross different hardware

guotuofeng:
## Describe your changes
Allow pass run could be reused cross different hardware

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -702,14 +702,13 @@ class Engine:
         pass_name: int,
         input_model_number: str,
         pass_config: dict,
-        is_accelerator_agonistic: bool,
         accelerator_spec: AcceleratorSpec,
     ):
         """
         Get the path to the run json.
         """
         pass_config_hash = hash_dict(pass_config)
-        if is_accelerator_agonistic:
+        if not accelerator_spec:
             run_json_path = self._run_cache_path / f"{pass_name}-{input_model_number}-{pass_config_hash}.json"
         else:
             run_json_path = (
@@ -723,7 +722,6 @@ class Engine:
         pass_config: dict,
         input_model_id: str,
         output_model_id: str,
-        is_accelerator_agonistic: bool,
         accelerator_spec: AcceleratorSpec,
     ):
         """
@@ -736,9 +734,7 @@ class Engine:
             "output_model_id": output_model_id,
         }
         input_model_number = input_model_id.split("_")[0]
-        run_json_path = self.get_run_json_path(
-            pass_name, input_model_number, pass_config, is_accelerator_agonistic, accelerator_spec
-        )
+        run_json_path = self.get_run_json_path(pass_name, input_model_number, pass_config, accelerator_spec)
         try:
             with open(run_json_path, "w") as f:
                 json.dump(run_json, f, indent=4)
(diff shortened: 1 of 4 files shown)


trajepl (approved):

olive/passes/olive_pass.py
```diff
@@ -74,6 +74,8 @@ def __init__(
         config_class, default_config = self.get_config_class(accelerator_spec, disable_search)
 
         self._accelerator_spec = accelerator_spec
+        self._is_accelerator_agnostic = True
```
jambayk, line 77: I think we could make this a class attribute like 
image

Otherwise, I don't see where the pass developer sets this property without changing `__init__`. Or do we expect the value for this property to change based on the pass run config? I think it is simpler if the pass just has a class attribute for this?

olive/engine/engine.py
```diff
@@ -848,11 +877,20 @@ def _run_pass(
                 if self.no_search:
                     raise  # rethrow the exception if no search is performed
 
+        # Do we need get the output_model_id from the output_model.model_path like the following?
+        # if output_model != PRUNED_CONFIG and output_model.model_path:
+        #     # get the updated output_model_id from the output model path
+        #     output_model_id = Path(output_model.model_path).name
+        if not p.is_accelerator_agnostic:
```
jambayk, line 884: This line should go between 860 and 861.

jambayk: Otherwise outputs from the same not accelerator agnostic pass will overwrite each other's outputs.

olive/passes/olive_pass.py
guotuofeng: I deliberately not make it class attribute is because for some passes, it could/could not depend hardware by different EP. 
For example, the quantization could depend hardware if the EP is DML. it should be independent on hardware if the EP is CPU. In this way, the CPU quantization could be reused if the EP is CPU or other EP that doesn't impact the quantization behavior. The dml quantization should only be reused if the EP is DML. 

How do you think?

jambayk (dismissed):

olive/engine/engine.py
```diff
@@ -710,19 +734,21 @@ def _cache_run(self, pass_name: int, pass_config: dict, input_model_id: str, out
...
             logger.error(f"Failed to cache run: {e}")
 
-    def _load_run(self, input_model_id: str, pass_name: int, pass_config: dict):
+    def _load_run(self, input_model_id: str, pass_name: int, pass_config: dict, accelerator_spec: AcceleratorSpec):
         """
         Load the run from the cache directory.
         """
         input_model_number = input_model_id.split("_")[0]
-        run_json_path = self.get_run_json_path(pass_name, input_model_number, pass_config)
+        run_json_path = self.get_run_json_path(pass_name, input_model_number, pass_config, accelerator_spec)
```
jambayk, line 749: Actually, I think this logic might not work. 
For example, we ran with accelerator_spec `A` which returns `is_accelerator_agnostic=True` and then `B` which returns `False`. The second run for `B` will reuse the run for `A` even though that is not correct. 

Since we now have a pass method for `is_accelerator_agnostic`, we can directly provide the correct value for `accelerator_spec` for this method at line `823`

jambayk (approved):

PR #291, opened 2023-05-22
Add dolly-v2-7b example

PatriceVignola:


olive/passes/onnx/optimum_conversion.py
```diff
@@ -0,0 +1,52 @@
...
+from typing import Any, Dict, Union
+
+from olive.model import CompositeOnnxModel, ONNXModel, OptimumModel
+from olive.hardware.accelerator import AcceleratorSpec
+from olive.passes import Pass
+from olive.passes.onnx.common import get_external_data_config
+from olive.passes.pass_config import PassConfigParam
+from pathlib import Path
+
+from optimum.exporters.onnx import main_export as export_optimum_model
```
guotuofeng, line 14: delay the import so that the optimum could be an extra for Olive?

olive/passes/onnx/optimum_merging.py
```diff
@@ -0,0 +1,82 @@
...
+from olive.passes.onnx.common import get_external_data_config
+from olive.passes.pass_config import PassConfigParam
+from pathlib import Path
+
+from optimum.onnx import merge_decoders
+import onnx
+import onnxruntime
+import os
+from onnx import ModelProto
+
```
guotuofeng, line 19: I guess you need run pre-commit run --all-files

olive/passes/onnx/optimum_merging.py
```diff
@@ -0,0 +1,82 @@
...
+from olive.hardware.accelerator import AcceleratorSpec
+from olive.model import CompositeOnnxModel, ONNXModel
+from olive.passes import Pass
+from olive.passes.onnx.common import get_external_data_config
+from olive.passes.pass_config import PassConfigParam
+from pathlib import Path
+
+from optimum.onnx import merge_decoders
+import onnx
```
guotuofeng, line 15: better not import the optimum at top level since it would make it to be requirement of Olive instead of an extra

olive/passes/onnx/optimum_merging.py
```diff
@@ -0,0 +1,82 @@
...
+            all_tensors_to_one_file=True,
+            location="decoder_model_merged.onnx.data",
+        )
+
+        # Doing a dry run of ORT allows us to remove the initializers that were orphaned by the merging step
+        sess_options = onnxruntime.SessionOptions()
+        sess_options.optimized_model_filepath = output_model_path
+
+        execution_provider = config["execution_provider"]
+        onnxruntime.InferenceSession(output_model_path, sess_options, providers=[execution_provider])
```
guotuofeng, line 80: >InferenceSession [](http://example.com/codeflow?start=20&length=16)

do we need specify the graph_optimization_level?

guotuofeng:
Does the following documentation need to be updated?

https://github.com/microsoft/Olive/blob/main/docs/source/examples.md
https://github.com/microsoft/Olive/blob/main/docs/source/overview/options.md
https://github.com/microsoft/Olive/blob/main/docs/source/api/passes.rst

guotuofeng:
better add readme.md to the example?

olive/passes/onnx/transformer_optimization.py
```diff
@@ -93,7 +94,7 @@ def _run_for_config(self, model: ONNXModel, config: Dict[str, Any], output_model
         for key in get_external_data_config():
             del run_config[key]
 
-        output_model_path = ONNXModel.resolve_path(output_model_path)
+        output_model_path = ONNXModel.resolve_path(os.path.join(output_model_path, os.path.basename(model.model_path)))
```
jambayk, line 97: Is there a reason for this change? Ideally, we don't want to manipulate the output path much other than adding a suffix or creating a file in the directory. Since it might affect the caching of the model.

olive/passes/onnx/optimum_merging.py
```diff
@@ -0,0 +1,85 @@
...
+
+        onnx.save(
+            merged_model,
+            output_model_path,
+            save_as_external_data=True,
+            all_tensors_to_one_file=True,
+            location="decoder_model_merged.onnx.data",
+        )
+
+        # Doing a dry run of ORT allows us to remove the initializers that were orphaned by the merging step
```
jambayk, line 77: Does this actually do anything? The model is already saved to disk so does it persist the orphaned initializers removal past the current inference session? Maybe the model should be saved again after the inference session is created so that the model on disk has the changes?

olive/passes/onnx/transformer_optimization.py
PatriceVignola: The default resulting model name of `resolve_path` is simply `model.onnx`, no matter what the original model is named.

It works for pipelines with a single model, but for pipelines with multiple models (e.g. `decoder_model.onnx` and `decoder_with_past_model.onnx`), it makes it very hard to debug intermediate steps, or even use the result of intermediate steps for something else. All this change does is name the optimized model the same way that the unoptimized version is named.

olive/passes/onnx/optimum_merging.py
PatriceVignola: It does. Setting `optimized_model_filepath` automatically overwrites the model on disk after applying the optimizations.

olive/extra_dependencies.json
```diff
@@ -24,5 +24,8 @@
     ],
     "inc": [
         "neural-compressor"
+    ],
+    "optimum": [
```
jambayk, line 28: Please also add the pass dependencies here https://github.com/microsoft/Olive/blob/main/olive/workflows/run/run.py#L64

olive/passes/onnx/transformer_optimization.py
jambayk: I see. Thanks for the clarification. For the other pass where we have composite models, we create a subdirectory for each component so they are in different folders. 
I think this change is okay for now since the file is still in the parent directory of the `output_model_path`. I will revisit this later to standardize how the paths are manipulted.

guotuofeng: I agree with @PatriceVignola. The output model path might be modified based on the arguments. For example, the output model path might be appended with hardware info if the pass decide it is hardware dependent like https://github.com/microsoft/Olive/pull/286/files#diff-89a2a3c72dc004cdf3c6603d382f434ec3f7527fc9065a7f3c420ceaf3a7cbf9R110-R111

olive/passes/onnx/optimum_merging.py
jambayk: Oh I see. I wasn't aware about this functionality. Does it handle the external data correctly too while doing the overwrite?

guotuofeng: @jambayk , this line would generate the optimized model. For example, the constants in model will be transformed as initializers.

olive/model.py
```diff
@@ -708,6 +708,24 @@ def to_json(self, check_object: bool = False):
...
+            framework=Framework.PYTORCH,
+        )
+        self.model_components = model_components
+
+    def load_model(self, rank: int = None):
+        raise NotImplementedError()
+
+    def prepare_session(self, inference_settings: Dict[str, Any], device: Device, rank: int = None):
+        raise NotImplementedError()
+
```
jambayk, line 727:
```
        config = {
            "type": self.__class__.__name__,
            "config": {"name": self.name, "model_path": self.model_path, "model_components": self.model_components
        }
        return serialize_to_json(config, check_object)
```

guotuofeng (approved):

PR #292, opened 2023-05-23
Bert example bugfix for f1_score

trajepl:
## Describe your changes
For bert example, when we apply `f1_score`, olive got the unmatched number with official report as the `task` is not assigned to `binary` for bert-mnli task.

https://huggingface.co/Intel/bert-base-uncased-mrpc

image

This PR added the `task` parameters into accuracy base class which can help get correct numbers.

image

### DIFF
--- a/docs/source/overview/options.md
+++ b/docs/source/overview/options.md
@@ -197,7 +197,7 @@ information of the evaluator contains following items:
                 "type": "accuracy",
                 "sub_types": [
                     {"name": "accuracy_score", "priority": 1, "goal": {"type": "max-degradation", "value": 0.01}},
-                    {"name": "f1_score"},
+                    {"name": "f1_score", "metric_config": {"task": "binary"}},
                     {"name": "auc", "metric_config": {"reorder": true}}
                 ],
                 "user_config":{
--- a/docs/source/overview/options.md
+++ b/docs/source/overview/options.md
@@ -197,7 +197,7 @@ information of the evaluator contains following items:
                 "type": "accuracy",
                 "sub_types": [
                     {"name": "accuracy_score", "priority": 1, "goal": {"type": "max-degradation", "value": 0.01}},
-                    {"name": "f1_score", "metric_config": {"task": "binary"}},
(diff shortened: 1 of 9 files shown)


docs/source/overview/options.md
```diff
@@ -197,7 +197,7 @@ information of the evaluator contains following items:
                 "type": "accuracy",
                 "sub_types": [
                     {"name": "accuracy_score", "priority": 1, "goal": {"type": "max-degradation", "value": 0.01}},
-                    {"name": "f1_score"},
+                    {"name": "f1_score", "metric_config": {"task": "binary"}},
```
guotuofeng, line 200: it seems the old version doesn't have the parameter

trajepl: image
Oh, yes. there is no `task` in v0.8.0 of torhmetircs. Do you know why we limit the version range as `>=0.7.0, <0.10.1`"? Can we update it?

Or I can fix the bert bug with other configs like `multiclass=2`

trajepl: Updated.

trajepl: Updated

guotuofeng: the version is pinned by Jambey, he has more context.

trajepl: Ic, anyhow I updated it with another parameter which exited in `>=0.7.0, <0.10.1`.

jambayk: We pinned the version some time last year when the new version that came out was incompatible with our evaluation code. 
there was a big change in the api. I think they introduced the tasks concept and it broke our code. 

I think we can revisit the versions if needed. Maybe check the torch vision version, map the config params and call the relevant torch metric functions for the version?

trajepl: Thanks. It is fine now. We change the logic for metric default config. Currently, we use inspect to generate the metric config with the origin default configs.  
Then no matter what kinda version user choose, just provide right parameters would be ok.

guotuofeng (approved):

guotuofeng (approved):

PR #302, opened 2023-05-25
Integrate Intel® Neural Compressor accuracy aware tuning

yuwenzho:
## Describe your changes

Integrate Intel® Neural Compressor accuracy aware tuning into Olive.


examples/bert_inc_ptq_cpu/README.md
```diff
@@ -14,7 +14,23 @@ python -m pip install -r requirements.txt
...
+Accuracy aware tuning is one of unique features provided by Intel® Neural Compressor quantization. This feature can be used to solve accuracy loss pain points brought by applying low precision quantization and other lossy optimization methods. Intel® Neural Compressor also support to quantize all quantizable ops without accuracy tuning, user can decide whether to tune the model accuracy or not. Please check the [doc](https://github.com/intel/neural-compressor/blob/master/docs/source/quantization.md) for more details.
+
+User can decide to tune the model accuracy by setting 'evaluate_func'. If not set, accuracy of the model will not be tuned.
+
+```json
+"quantization": {
+    "type": "IncQuantization",
+    "config": {
+        "user_script": "user_script.py",
+        "evaluate_func": "eval_accuracy",
```
trajepl, line 28: Does that say if user choose the build-in accuracy metrics in the part of `olive.evaluator` , they still need to provide `evaluate_func` here? 
Also the the `accuracy_criterion` looks a bit duplicated with olive's metric goal. 

If we can leverage the olive.evaluator metrics here Instead of use `evaluate_func`?

yuwenzho:
Hi, I refined INC `eval_func` and `accuracy_criterion`. Now they are set according to Olive accuracy metric, rather than being additionally set by the user.

olive/passes/onnx/inc_quantization.py
```diff
@@ -208,9 +264,148 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
         config.update(get_external_data_config())
         return config
 
+    def _set_eval_func(self, accuracy_metric, sub_type):
+        # set eval_func for INC according to Olive accuracy metric
+        def eval_func(model):
+            # eval_func for Intel® Neural Compressor quantization take model as input,
```
jambayk, line 270: I think this can be simplified by creating a temporary `ONNXModel` and calling the evaluator model. This way, we won't need need to create a custom inference session or custom metric that takes in an onnx model proto. 

Please refer to this commit for a sample implementation https://github.com/microsoft/Olive/commit/3f2b6fe3081691b9dee97a8b12f773df68d310b6 

Please feel free to cherry pick the commit if it meets your requirements.

ftian1: thanks, we will follow up your suggestion on that.

yuwenzho:
@jambayk I refined evaluation function based on your commit. Thanks for your help!

trajepl (approved):

trajepl (dismissed):

trajepl (approved):

olive/passes/olive_pass.py
```diff
@@ -314,6 +315,9 @@ def _resolve_config(
             config = cls._validate_user_script(config, user_module_loader, default_config)
         return config
 
+    def set_evaluator(self, evaluator):
```
guotuofeng, line 318: what's the type of the evaluator?

olive/engine/engine.py
```diff
@@ -816,6 +816,10 @@ def _run_pass(
         pass_config = p.config_at_search_point(pass_search_point)
         pass_config = p.serialize_config(pass_config)
 
+        # set evaluator to pass
```
guotuofeng, line 819: the evaluator info will be dropped when running in AzureML case. 

@jambayk , do we need support it?

guotuofeng (changes requested):
the evaluator info set by manually in engine level would be dropped when running in AzureML.

olive/engine/engine.py
jambayk: Yes, I think we will need to support it. However, I am not sure what the best way to do this is. 

I think it might be better to insert the evaluator config into the pass config like we do for `data_config` instead of setting a model attribute to the pass object. What do you think @trajepl?

trajepl: Agree with Jambay‘s idea. Something like the flag of `_requires_user_script` in Pass's attribute, then we can replace the evaluator_config for specific pass.

But think it a bit more[rough idea], we have many components like `user_script/script_dir`, `data_config`, `evaluator`, `aml_client` and etc. are required by engine/pass/system etc.  Maybe we should propose the component like central component in olive?
It may not change the user-facing config. But will give a clear definition and unified way for engine/pass/system to pull the corresponding configs from this central component.

guotuofeng: does this mean we need allow pass refer to the evaluation config defined in "evaluator" section? or we need copy the config and put them as plain text in pass config?

trajepl: Currently, the latter one which copy the config. Because Olive need to ensure pass run can be directly started even there is no outside evaluation config.

guotuofeng: I suppose reference to evaluator defined in evaluators section should be able to work and it is more concise than the latter solution. Are there any blockers?

trajepl: If you mean the reference like olive system, I think we are at the same line. I am just thinking you may mean the reference to the config object.

Finally, the run_config.parser will replace the reference automatically with the full config.

guotuofeng: Yeah, I mean the behavior like system target or host.

guotuofeng: @yuwenzho , please ping @trajepl if you need help on how to do the change.

ftian1: sure, we will do that after the Chinese Dragon Boat Festival. thanks for your support.

yuwenzho:
@trajepl  I optimized the evaluator, but I'm not sure if it meets your expectations. Could you please review it for me?

olive/passes/olive_pass.py
```diff
@@ -48,6 +50,8 @@ class Pass(ABC):
     _requires_user_script: bool = False
     # True if pass configuration requires data configuration which will leverage data container for pass execution
     _requires_data_config: bool = False
+    # True if pass configuration requires evaluation configuration
```
guotuofeng, line 53: better not use a class level field for eval customization since only inc_quantization only use it from the time being. 

The other two _requires_xxx is user by all passes.

olive/passes/pass_config.py
```diff
@@ -93,6 +94,19 @@ def get_data_config(required: Optional[bool] = False):
     return data_config
 
 
+def get_eval_config(required: Optional[bool] = False):
+    eval_config = {
+        "evaluator": PassConfigParam(
```
guotuofeng, line 99: @jambayk , @trajepl , how should we differentiate the evaluator with https://microsoft.github.io/Olive/overview/options.html#passes-information if we make it as part of default config?

olive/passes/onnx/inc_quantization.py
```diff
@@ -201,9 +257,140 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
...
+            )
+
+        _evaluator_config = deepcopy(run_config)
+
+        accuracy_criterion = AccuracyCriterion()
+        tuning_criterion = TuningCriterion()
+        eval_func = None
+        accuracy_metric = None
+
+        if self._eval_config is not None:
```
guotuofeng, line 346: >_eval_config [](http://example.com/codeflow?start=16&length=12)

it seems the config only use metrics definition. @jambayk , @trajepl should we use `metrics` instead `evaluator` here to avoid confusion?

examples/bert/README.md
```diff
@@ -26,8 +26,39 @@ This workflow performs BERT optimization on CPU with Intel® Neural Compressor q
...
+                ]
+            }
+        ]
+    }
+},
+"passes": {
+    "quantization": {
+        "type": "IncQuantization",
+        "config": {
+                "evaluator": "common_evaluator"
```
guotuofeng, line 52: >"evaluator": "common_evaluator" [](http://example.com/codeflow?start=16&length=31)

this design seems have conflicts with [https://microsoft.github.io/Olive/overview/options.html#passes-information](https://microsoft.github.io/Olive/overview/options.html#passes-information)

guotuofeng (changes requested):
it seems the evaluator only use accuracy metrics definition. 

@jambayk , @trajepl one better approach is we could use metrics instead of evaluator in pass config so that we can differentiate our original pass evaluator with the customized metrics definition here. 

Please let me know if you have any better idea.

olive/passes/pass_config.py
trajepl: Yes, otherwise, the config for one passes will be like:
```json
"passes": {
    "quantization": {
        "type": "IncQuantization",
        "config": {
                "evaluator": "common_evaluator" <---------------
            }

         "evaluator": "common_evaluator" <---------------
    }
}
```

The two same evaluators look a bit confusing.

trajepl: After offline syncing, @ftian1 , we suggest to implement it as Mike's suggestion to directly **pass the metric you needed with config way**. It will be like:

```json
"passes": {
    "quantization": {
        "type": "IncQuantization",
        "config": {
            "metric": {
                    "name": "accuracy",
                    "type": "accuracy",
                    "sub_types": [
                        {"name": "accuracy_score", "priority": 1}
                    ],
                    "user_config":{
                        "post_processing_func": "post_process",
                        "dataloader_func": "create_dataloader",
                        "batch_size": 1
                    }
            }
        }
         "evaluator": "common_evaluator" 
    }
}
```

trajepl: Then we can remove the flag of `_requires_eval_config` as it is not common flag for all passes like Mike's comment.

yuwenzho:
`metric` is now set to plain text in the pass config. Please help review. Thank you!

olive/passes/onnx/inc_quantization.py
```diff
@@ -95,6 +98,22 @@
...
+    "tuning_criterion": PassConfigParam(
+        type_=dict,
+        default_value={},
+        description="""
+            Instance of TuningCriterion class. In this class you can set strategy, strategy_kwargs,
+            timeout, max_trials and objective.
+        """,
+    ),
+    "metric": PassConfigParam(
+        type_=Metric,
```
trajepl, line 110: How about set as `type_=Optional[Metric], default_value=None`. Otherwise the type(dict) of default_value looks a bit inconsistent with `Metric`.

trajepl (approved):

trajepl:
> `metric` is now set to plain text in the pass config. Please help review. Thank you!

Thanks for the quick changes! It looks good to me basically. Please fix the comments, then I will try to trigger the CI pipeline.

yuwenzho:
> > `metric` is now set to plain text in the pass config. Please help review. Thank you!
> 
> Thanks for the quick changes! It looks good to me basically. Please fix the comments, then I will try to trigger the CI pipeline.
 
I fixed the comment. Thank you all for your help！

examples/bert/README.md
```diff
@@ -26,8 +26,32 @@ This workflow performs BERT optimization on CPU with Intel® Neural Compressor q
...
+
+User can decide to tune the model accuracy by setting accuracy `metric` in Intel® Neural Compressor quantization pass config. If not set, accuracy of the model will not be tuned.
+
+```json
+"passes": {
+    "quantization": {
+        "type": "IncQuantization",
+        "config": {
+            "metric": {
+                "name": "accuracy",
```
guotuofeng, line 41: >name [](http://example.com/codeflow?start=17&length=4)

@trajepl  do we need use type or name here?

olive/passes/onnx/inc_quantization.py
```diff
@@ -201,9 +264,140 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
...
+        _inc_quantization_config = deepcopy(run_config)
+
+        accuracy_criterion = AccuracyCriterion()
+        tuning_criterion = TuningCriterion()
+        eval_func = None
+        accuracy_metric = None
+
+        if len(_inc_quantization_config["metric"]) != 0:
+            if "name" in _inc_quantization_config["metric"]:
+                metric_name = _inc_quantization_config["metric"]["name"]
```
guotuofeng, line 355: >name [](http://example.com/codeflow?start=66&length=4)

should use type instead of name here?

trajepl: Yes. You are right.

examples/bert/README.md
trajepl: The evaluation function in `custom` can be `latency` or something else. How about to ignore the `accuracy` check? Clear docs may be enough.

guotuofeng (dismissed):

guotuofeng:
It seems there is bug in the condition check for len(metric)

guotuofeng (approved):

PR #305, opened 2023-05-26
Update whisper ort dependency to 1.15.0

jambayk:
## Describe your changes
Change whisper dependency from ort nightly to 1.15.0 stable release. 

### DIFF
--- a/examples/whisper/requirements.txt
+++ b/examples/whisper/requirements.txt
@@ -1,5 +1,4 @@
-coloredlogs
-flatbuffers
 onnx==1.13.1
+onnxruntime==1.15.0
 torch>=1.13.1
 transformers>=4.23.1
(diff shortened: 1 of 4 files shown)


examples/whisper/requirements.txt
```diff
@@ -1,5 +1,4 @@
-coloredlogs
-flatbuffers
 onnx==1.13.1
+onnxruntime==1.15.0
```
guotuofeng, line 2: need we pin the version or just use the latest ort version?

guotuofeng (approved):

examples/whisper/requirements.txt
jambayk: We need to pin it right now. It needs at least 1.15.0 but there are some new changes in ort main branch that are not compatible with what we have right now. 

I am planning to make the required changes to support versions > 1.15.0 (latest nightly) in a follow up PR.

PR #311, opened 2023-05-31
Restructure example tests and bert/resnet examples

jambayk:
## Describe your changes
The different bert and resnet examples are now consolidated under single `bert` and `resnet` directories. The config files have been simplified to remove unused sections. The separate `static` and `dynamic` configs have also been removed. 

All example tests are now under an `examples/test` subdirectory. Common code has been abstracted to `utils.py`.


examples/bert/bert_auto.json
```diff
@@ -13,15 +13,22 @@
...
         "common_evaluator": {
             "metrics":[
                 {
                     "name": "accuracy",
                     "type": "accuracy",
-                    "sub_types": [{"name": "accuracy_score", "priority": 1}],
+                    "sub_types": [
+                        {"name": "accuracy_score", "priority": 1}
+                    ],
                     "user_config":{
```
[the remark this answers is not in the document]
jambayk: sub_types is required. This structure was introduced in this PR https://github.com/microsoft/Olive/commit/7f69c3dfa5bd81cdbc892a331a8e7ca3094b6f67.

examples/bert/bert_auto.json
```diff
@@ -13,15 +13,22 @@
         }
     },
     "systems": {
-        "local_system": {"type": "LocalSystem"}
+        "local_system": {
+            "type": "LocalSystem",
```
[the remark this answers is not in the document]
jambayk: That is true. The engine defaults to local system with cpu. Should I remove it from all the bert and resnet configs? I kept them so that the user knows how to change the system if needed.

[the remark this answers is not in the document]
devang-ml: Yes, please remove them.

I'm going to update the quick tour to make Olive intro simpler for users.

examples/bert/bert_auto.json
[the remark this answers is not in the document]
devang-ml: ok.

examples/bert/bert_auto.json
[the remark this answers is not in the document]
jambayk: Done. I removed the local system from all the resnet and bert configs. Also updated the default system to initialize with CPU.

olive/engine/engine.py
```diff
@@ -65,15 +65,15 @@ def __init__(
...
+            self.host = LocalSystem([Device.CPU])
 
         # engine target
         if target is not None:
             self.target = target
         elif self._config.target is not None:
             self.target = self._config.target.create_system()
         else:
-            self.target = LocalSystem()
+            self.target = LocalSystem([Device.CPU])
```
guotuofeng, line 76: I suppose this is needless since the default CPU is already handled.

jambayk: I wanted the default to be clear since the default value for the LocalSystem accelerator is None and it generates this warning https://github.com/microsoft/Olive/blob/main/olive/engine/engine.py#L96.

olive/engine/engine.py
```diff
@@ -65,15 +65,15 @@ def __init__(
         elif self._config.host is not None:
             self.host = self._config.host.create_system()
         else:
-            self.host = LocalSystem()
+            self.host = LocalSystem([Device.CPU])
```
guotuofeng, line 68: host itself doesn't use the accelerators actually, so it could be safely removed.

olive/common/ort_inference.py
```diff
@@ -52,7 +52,9 @@ def get_ort_inference_session(
         if ep == "QNNExecutionProvider":
             # add backend_path for QNNExecutionProvider
             execution_provider[idx] = ("QNNExecutionProvider", {"backend_path": "QnnHtp.dll"})
-            break
+        if isinstance(ep, list) and len(ep) == 2 and isinstance(ep[1], dict):
```
guotuofeng, line 55: >if isinstance(ep, list) and len(ep) == 2 and isinstance(ep[1], dict) [](http://example.com/codeflow?start=8&length=68)

what's the purpose of this line? are there any examples or which part could trigger it?

jambayk: When we load a cached model that has a tuned inference setting, the tuple is loaded as a list like `["CPUExecutionProvider", {}]`. The ort runtime session generates a warning saying the format is not right. So, I added this line to correctly parse it to a tuple.

jambayk: You can see the tuple being created by the perf tuner here https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/perf_tuning.py#L110

olive/engine/engine.py
guotuofeng: make sense then

olive/common/ort_inference.py
jambayk: I can also remove this change from this PR since it is unrelated to it. We can revisit this issue later. I also don't think we should be ignoring the second element in tuples here https://github.com/microsoft/Olive/blob/main/olive/common/ort_inference.py#L47.

jambayk: Removed it and created a work item to track it.

guotuofeng (approved):

PR #319, opened 2023-06-05
Add hf model custom components support

xiaoyu-work:
## Describe your changes

1. Add hf model custom components support. Support custom components definition in HFConfig.
2. Update Whisper example to new structure.
3. Remove Whisper model specific code in Olive.


olive/model/hf/hf_model.py
```diff
@@ -0,0 +1,98 @@
...
+        super().__init__()
+        self.encoder = encoder
+        self.decoder = decoder
+        self.encoder_init = encoder_init
+        self.decoder_init = decoder_init
+        self.config = config
+
+
+class ORTWhisperModel(torch.nn.Module):
+    """ORT implementation of whisper model"""
```
devang-ml, line 71: This should be outside the Olive, in the customer code.

xiaoyu-work: I'll remove this once I refactor whisper example.

olive/model/hf/hf_model.py
```diff
@@ -0,0 +1,51 @@
...
+    @validator("model_class", always=True)
+    def task_or_model_class_required(cls, v, values):
+        if values["model_name"]:
+            if not v and not values.get("task", None):
+                raise ValueError("Either task or model_class must be specified")
+            return v
+
+
+class HFModelBase(torch.nn.Module):
+    """Base class for all HF models"""
```
devang-ml, line 41: This is not a HFModelBase, it is just a PyTorchNetwork

We have used the OliveModelBase and its derived classes to handle ONNX, PT etc., models and their weights on the disk.

examples/whisper/user_script.py
```diff
@@ -5,8 +5,147 @@
 from pathlib import Path
 
 import numpy as np
-from onnxruntime.transformers.models.whisper.whisper_decoder import WhisperDecoderInputs
-from onnxruntime.transformers.models.whisper.whisper_encoder_decoder_init import WhisperEncoderDecoderInitInputs
+from onnxruntime.transformers.models.t5.past_helper import PastKeyValuesHelper
+from onnxruntime.transformers.models.whisper.whisper_decoder import WhisperDecoder, WhisperDecoderInputs
+from onnxruntime.transformers.models.whisper.whisper_encoder_decoder_init import (
```
devang-ml, line 10: Could we just copy WhisperDecoder etc., here in the example folder ?

devang-ml (commented):
Also add another whisper example that uses standard HF whisper model components only.

olive/model/hf/hf_model.py
devang-ml: Other alternative is to rename OliveModelBase as OliveModelWrapper (and its derived classes accordingly) and use PyTorchModel here instead of HFModelBase.

docs/source/overview/options.md
```diff
@@ -72,9 +72,20 @@ case insensitive.
         - `task: [str]`: This is the task type for the model such as `text-classification`. The complete list of supported task can be found
         at [huggingface-tasks](https://huggingface.co/docs/transformers/v4.28.1/en/main_classes/pipelines#transformers.pipeline.task).
 
+        - `feature: [str]`: The ONNX export features. This is only needed for HuggingFace hub model. Default to `default`. You can find more info at [Export to ONNX](https://huggingface.co/docs/transformers/serialization)
+
         - `model_class: [str]`: Instead of the `task`, the class of the model can be provided as well. Such as `DistilBertForSequenceClassification`
 
-        - `model_config: [str]`: The config of the model can be provided as well. Such as `WhisperConfig`. See
+        - `use_custom_implementation: [bool]`: If components is implemented by the user. If `False`, Olive will load model from Huggingface hub from `model_name` and `task` or `model_class`. If `True`, user will need to provide detailed components implementation.
+
```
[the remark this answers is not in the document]
xiaoyu-work: Will remove this.

olive/model/hf/hf_model.py
xiaoyu-work: Let me revisit this if we actually need a model class for HF model in Olive. If only conversion, we may not need a model class here.

examples/whisper/user_script.py
xiaoyu-work: I thought that would be duplicated code from onnxruntime. It will be too much to copy all codes from onnxruntime repo.

olive/model/hf/hf_model.py
guotuofeng: I agree with @devang-ml , the HFModelBase is confusing.

examples/whisper/user_script.py
devang-ml: Neither onnxruntime nor Olive are supposed to be the vendor of the WhisperDecoder etc., components. If we copy this here in the example folder then it becomes clear to the user that this is how they can provide their own custom components. 

If we can establish this then eventually we can remove the WhisperDecoder etc., from the onnxruntime repository.

olive/model/hf/hf_model.py
xiaoyu-work: Removed HFModelBase class.

examples/whisper/user_script.py
xiaoyu-work: Added custom components implement.

olive/model/model.py
```diff
@@ -9,29 +9,29 @@
 from abc import ABC, abstractmethod
```
guotuofeng, line 9: nit: this file could be be merged with __init__.py

xiaoyu-work: Sounds good. Will do.

devang-ml (approved):

PR #329, opened 2023-06-07
Aml_client.models.download -> aml fsspec.download

trajepl:
## Describe your changes
1. Replace the `aml_client.models.dwonload` with azureml fsspec for Datastore resource download
2. Add the interface to let user directly input datastore url without aml client preparation.

### DIFF
--- a/olive/resource_path.py
+++ b/olive/resource_path.py
@@ -3,12 +3,13 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 import logging
+import re
 import shutil
 import tempfile
 from abc import abstractmethod
 from enum import Enum
 from pathlib import Path
-from typing import Any, Callable, Dict, Union
+from typing import Any, Callable, Dict, Optional, Union
 
 from pydantic import Field, validator
 
@@ -26,6 +27,7 @@ class ResourceType(str, Enum):
     StringName = "string_name"
     AzureMLModel = "azureml_model"
     AzureMLDatastore = "azureml_datastore"
+    AzureMLModelDatastore = "azureml_model_datastore"
     AzureMLJobOutput = "azureml_job_output"
 
     def __str__(self) -> str:
@@ -316,9 +318,7 @@ class AzureMLModel(ResourcePath):
         return str(new_path)
 
 
-class AzureMLDatastore(ResourcePath):
-    """AzureML datastore resource path"""
-
+class AzureMLDataStore(ResourcePath):
     name = ResourceType.AzureMLDatastore
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


olive/resource_path.py
```diff
@@ -316,64 +317,126 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
...
 
     def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: bool = False) -> str:
         # there is no direct way to download a file from a datastore
         # so we will use a workaround to download the file by creating a aml model
         # that references the file and downloading the model
-        from azure.ai.ml.constants import AssetTypes
-        from azure.ai.ml.entities import Model
-        from azure.core.exceptions import ServiceResponseError
+        try:
+            from azureml.fsspec import AzureMachineLearningFileSystem
```
guotuofeng, line 392: need we add an item in the extra?

olive/resource_path.py
```diff
@@ -316,64 +317,126 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
...
+            )
+        return self.config.datastore_url
+
+    def is_file(self, fsspec) -> bool:
+        return fsspec.info(self.get_relative_path()).get("type") == "file"
+
+    def get_relative_path(self) -> str:
+        if not self.config.relative_path:
+            assert self.config.datastore_url
+            self.config.relative_path = re.split("/datastores/.*/paths/", self.config.datastore_url)[-1]
```
jambayk, line 372: I think we should not change the value of the `self.config` after validation is done. This causes the hash of the resource path to change depending on which methods were called and which config fields were changed. 

Also, if we convert this resource path and load this again, we will hit warning at line 328 which was not originally present. since we added relative_path to the original config.

jambayk: Also, based on the priority of `datastore_url`, I think we should only return the original `relative_path` if `datastore_url` is None? Since it ``relative_path` is ignored if `datastore_url` is None

olive/resource_path.py
```diff
@@ -316,64 +317,126 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
...
+        return self.config.relative_path
+
+    def get_aml_client_config(self) -> AzureMLClientConfig:
+        if not self.config.azureml_client:
```
jambayk, line 376: same comments as above for `relative_path`.

olive/resource_path.py
jambayk: Yes, I think we can add `azureml-fsspec` to the pre-existing `azureml` extra?

olive/resource_path.py
```diff
@@ -316,64 +317,126 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
...
+                },
+                max_tries=self.config.azureml_client.max_operation_retries,
+                delay=self.config.azureml_client.operation_retry_interval,
+            )
+            downloaded_resource = Path(temp_dir) / relative_path.name
+            if downloaded_resource.is_file():
+                # only if the resource is a existed file we will move it to the new path
+                shutil.move(downloaded_resource, new_path)
+            else:
+                shutil.copytree(temp_dir, new_path)
```
jambayk, line 437: not a big issue but would shutil.move work for this part too? If so, I think that might be better/faseter since shutil.move  will be just a rename. the temp_dir is in the same partition as `dir_path` so there is no copy done in move. 

something like
```
new_path.parent.mkdir(parents=True, exist_ok=True)
shutil.move(downloaded_resource  is is_file else temp_dir, new_path)
``` 
?

olive/resource_path.py
guotuofeng: the azureml-fsspec would introduce azureml-dataset which depends on azureml sdk1.

trajepl: image
No dependencies on azureml-core. The dependent package size is about 260Mb.

olive/resource_path.py
```diff
@@ -316,64 +317,126 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
...
+                f"{_get_azureml_resource_prefix(workspace_config)}"
+                f"/datastores/{self.config.datastore_name}/paths/{self.config.relative_path}"
+            )
+        return self.config.datastore_url
+
+    def is_file(self, fsspec) -> bool:
+        return fsspec.info(self.get_relative_path()).get("type") == "file"
+
+    def get_relative_path(self) -> str:
+        if not self.config.relative_path:
```
jambayk, line 372: based on the highest priority of `datastore_url`, I think this condition should be `if self.config.datastore_url`?

olive/resource_path.py
```diff
@@ -316,64 +317,126 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
...
+            assert self.config.datastore_url
+            return re.split("/datastores/.*/paths/", self.config.datastore_url)[-1]
+        return self.config.relative_path
+
+    def get_aml_client_config(self) -> AzureMLClientConfig:
+        if not self.config.azureml_client:
```
jambayk, line 378: same comment as above for condition.

olive/resource_path.py
trajepl: Fixed.

olive/resource_path.py
trajepl: Fixed.

jambayk (approved):
Thanks!

PR #330, opened 2023-06-07
📝 Bert gpu cuda/trt example

trajepl:


examples/bert/bert_cuda.json
```diff
@@ -0,0 +1,103 @@
+{
```
EmmaNingMS, line 1: Suggest renaming the title of this config file to: bert_cuda_gpu.json. Same to the trt config file.

examples/bert/bert_trt.json
```diff
@@ -0,0 +1,104 @@
...
+        }
+    },
+    "passes": {
+        "conversion": {
+            "type": "OnnxConversion",
+            "config": {
+                "target_opset": 13
+            }
+        },
+        "transformers_optimization": {
```
EmmaNingMS, line 61: TRT has its own graph optimization logic. So we only need to feed in the original ONNX graph.

EmmaNingMS (changes requested):

examples/bert/bert_cuda.json
trajepl: Updated.

EmmaNingMS (approved):

PR #331, opened 2023-06-07
🐛 Fix: cannot get accelerator spec info in perf tuning

trajepl:

### DIFF
--- a/olive/passes/onnx/perf_tuning.py
+++ b/olive/passes/onnx/perf_tuning.py
@@ -332,11 +332,11 @@ class OrtPerfTuning(Pass):
         }
 
     def _run_for_config(self, model: ONNXModel, config: Dict[str, Any], output_model_path: str) -> ONNXModel:
-        if "providers_list" not in config:
+        if not config.get("providers_list"):
             # add the provider to the config if user doesn't provide the execution providers
             config["providers_list"] = [self._accelerator_spec.execution_provider]
 
-        if "device" not in config:
+        if not config.get("device"):
             config["device"] = self._accelerator_spec.accelerator_type
 
         config = self._config_class(**config)


olive/passes/onnx/perf_tuning.py
```diff
@@ -332,11 +332,11 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
         }
 
     def _run_for_config(self, model: ONNXModel, config: Dict[str, Any], output_model_path: str) -> ONNXModel:
-        if "providers_list" not in config:
+        if not config.get("providers_list"):
             # add the provider to the config if user doesn't provide the execution providers
             config["providers_list"] = [self._accelerator_spec.execution_provider]
```
devang-ml, line 337: User should not have need to provide separate providers_list to ORTPerfTuning pass. We should remove that ability.

The ORTPerfTuning (and other places) should only get EP info from AcceleratorSpec. AcceleratorSpec should be the only place where the user can specify EP(s)

guotuofeng: I think the providers_list is used for standalone purpose. for example, the ORTPerfTuning could be used without engine.

guotuofeng:
what's the bug? could its description be more specific?

trajepl:
> what's the bug? could its description be more specific?

`providers_list` and `device` are not required and the default value are None and `cpu`. It means that the check of `providers_list not in config` is always `False`.

```python
if "providers_list" not in config: # <--- the config always have the key of providers_list which is optional in default_config, default value is None
    # add the provider to the config if user doesn't provide the execution providers
    config["providers_list"] = [self._accelerator_spec.execution_provider]

if "device" not in config:
    config["device"] = self._accelerator_spec.accelerator_type
```

You can see that even I give the TRT ep in engine, olive is still use cpu to run perf tuning.
image

PR #333, opened 2023-06-09
Add RawDataContainer, SNPE eval and pass accept common dataloader, Migrate snpe examples to workflow jsons

jambayk:
## Describe your changes
This PR introduces a new data container called `RawDataContainer` where the data is saved in directory as numpy binary files. 
See the doc for `RawDataset` for more details. 
SNPE quantization and evaluation converts the generic dataloader to snpe dataloader so they are not limited snpe data loaders anymore.

In the future, we can expand the dataset to support diferent types of data (.npy, .npz, pickle, etc), 

The snpe examples have been changed to workflow config jsons that use the raw data config. 
SNPE evaluation and quantization now accept generic dataloaders such as torch dataloader. They no longer require users to provide SNPE dataloaders.

### DIFF
--- a/olive/data/component/dataloader.py
+++ b/olive/data/component/dataloader.py
@@ -3,16 +3,9 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 
-import logging
-import tempfile
-from pathlib import Path
-
-import numpy as np
 from torch.utils.data import DataLoader
 
 from olive.data.registry import Registry
-
-logger = logging.getLogger(__name__)
 
 
 @Registry.register_dataloader()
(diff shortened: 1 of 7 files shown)


olive/engine/engine.py
```diff
@@ -289,7 +289,7 @@ def run(
                     result_name = f"{prefix_output_name}metrics"
                     results_path = output_dir / f"{result_name}.json"
                     with open(results_path, "w") as f:
-                        f.write(results.json())
+                        json.dump(results.to_json(), f, indent=4)
```
guotuofeng, line 292: >to_json [](http://example.com/codeflow?start=42&length=7)

what's the difference between to_json and json()?

olive/data/component/dataloader.py
```diff
@@ -48,3 +55,120 @@ def rewind(self):
...
     return _CalibrationDataReader(dataloader, **kwargs)
+
+
+@Registry.register_dataloader()
+def default_snpe_dataloader(_dataset, snpe_model, batch_size=None):
+    from olive.snpe.data_loader import SNPEProcessedDataLoader
+    from olive.snpe.utils.input_list import create_input_list
+
+    class _SNPEProcessedDataLoader(SNPEProcessedDataLoader):
+        def __init__(self, dataset, snpe_model, batch_size=None):
```
guotuofeng, line 66: >snpe_model [](http://example.com/codeflow?start=36&length=10)

seems we only need the io_config?

olive/engine/engine.py
jambayk: the .json() returns a json string (it is a built-in method of pydantic BaseModel). `to_json` is a method of `ConfigBase` that returns a json as python dictionary.

olive/data/container/data_container.py
```diff
@@ -65,6 +65,18 @@ def create_dataloader(self):
         pre_process_dataset = self.pre_process(dataset)
         return self.dataloader(pre_process_dataset)
 
+    def create_snpe_dataloader(self, model, evaluation=True):
```
guotuofeng, line 68: >create_snpe_dataloader [](http://example.com/codeflow?start=8&length=22)

do we need add new signature?

jambayk: I just discussed this with Jiapeng. The snpe dataloader is used only for snpe model evaluation or quantization, so it is better kept as a separate method. Unless, we want to add different condition to the `create_dataloader` and `create_calibration_dataloader` methods.

olive/data/component/dataloader.py
jambayk: yes, it only needs the `input_names` and `input_shapes`.

olive/evaluator/olive_evaluator.py
```diff
@@ -156,22 +156,23 @@ def get_user_config(metric: Metric):
         eval_func = user_module.load_object(evaluate_func)
```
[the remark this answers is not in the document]
jambayk: It depends on the model and data. 

for the user_config part, it is used to create dummy data. 
for snpe_model, we need to get it from the model itself since we don't know the model's io config before hand. SNPE permutes the input shapes from `batchXchannelXwidthXheight` to `batchXwidthXheightXchannel` so we need both the original shapes (from data) and the model shape.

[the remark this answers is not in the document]
jambayk: for onnx model, we cannot use it to create dummy data so the latency metric user config might have io config inside them.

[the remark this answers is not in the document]
jambayk: I think we could revisit these later if needed.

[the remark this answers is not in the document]
jambayk: @jiapeng I think there are some plans to maybe remove `user_config` method of creating dataloaders in the future? Maybe we can review this then. Since the user_config `input_names`, `output_names` way of creating dummy data is already covered by the `DummyDataContainer`.

olive/engine/engine.py
guotuofeng: could I suppose the json.load(json()) == to_json()?

olive/data/container/data_container.py
guotuofeng: Is it possible there would be lot of customized create_xx_dataloader in future? that's my concerns.

olive/engine/engine.py
jambayk: Yes, it only pretty prints the json file. the contents are the same.

olive/data/container/data_container.py
jambayk: That is a valid concern. I added this directly to the container since `SNPEModel` is an olive model type and it also needs snpe dataloader. I think for other passes, they would need to do the conversion from olive dataloader to their specific type in the pass itself. 

I could also do the same. remove this method and instead get olive dataloader. And the convert it to snpe dataloader in the snpe evaluation and quantization pass.

olive/data/component/dataloader.py
guotuofeng: if so, should we pass io_config here instead of the mdoel itself?

olive/data/container/data_container.py
trajepl: How about this? @jambayk 
You can inherit the RawDataContainer and change the dataloader to snpe_dataloader. In this way, we can keep the base class as simple as possible. Then the usage in SNPE case will be clearer.
```python
@Registry.register(DataContainerType.DATA_CONTAINER)
class SNPEDataContainer(RawDataContainer):
default_components_type: ClassVar[dict] = {
    DataComponentType.LOAD_DATASET.value: "raw_dataset"
    DataComponentType.DATALOADER.value: "default_snpe_dataloader"
}
```

trajepl: Or we can create snpe template in `https://github.com/microsoft/Olive/pull/333/files/f9b7c7eac227bfa2b9f6f05ff80516184b472c33#diff-bfa09a24344a0657b60de3f87a8b40be86b65f9e34ab3ac259f0c631ba9d6c3c` based on RawDataContainer.

olive/evaluator/olive_evaluator.py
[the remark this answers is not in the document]
trajepl: Sounds good to me currently. Thanks.

In the future, maybe we can consider to let io_config spread to onnx model as well?
 Or another method for dynamic axis, we can assign the value if user does not provide it. Just make sure it is consist during the comparation in perf_tuning?

olive/data/component/dataloader.py
jambayk: Done.

olive/evaluator/olive_evaluator.py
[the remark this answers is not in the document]
jambayk: Auto assigning would work too but I don't know if the perf tuning results would be as useful as using user provided shapes. Since we want to tune for data the user expects to use the model irl.

olive/data/container/data_container.py
trajepl: A bit differences with offline discussion where I were thinking to replace the default dataloader to `default_snpe_dataloader` in `RawDataContainer`. Then the `RawDataContainer` only serves SNPE. I agreed that is not the way to go.

[1] The new suggestion is that we keep the `RawDataContainer` and to produce the `SNPEDataContainer`.  Then in SNPE config, we can directly use `SNPEDataContainer`. But for onnx/torch model, we can use `RawDataContainer`. **That is determined by users.**

[2] For this, it sounds good to me as well. 
> If people are worried about the new create_snpe_dataloader method, I think a better solution is to not add the method. The evaluation and quantization code will then do the following:
datacontainer.create_dataloader -> regular dataloader -> Convert regular dataloader to snpe dataloader


For me, I have no strong tendency on method [1] or [2]. Both of them can resolve Mike's concern.

trajepl: Besides, quick question about this: I do not get your points. Is there any necessary case where we must call snpe_dataloader independently? From my understanding, all the method in data components can be call independently.
>  The snpe dataloader code has no dependency on the RawDataContainer. It can work with any datacontainer.

olive/evaluator/olive_evaluator.py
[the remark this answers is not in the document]
trajepl: Good catch. The real case will be more flexible. The auto io_config assigning will work for perf_tuning accurately only when user know we will use the io_config in input model for perf tuning as well.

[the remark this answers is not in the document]
jambayk: Okay. I will close this conversation for now since it is related to a different feature.

olive/data/container/data_container.py
jambayk: actually, pushed the changes for [2] directly. Please let me know if that addresses the concerns.

olive/data/template.py
```diff
@@ -55,3 +55,33 @@ def huggingface_data_config_template(model_name, task, **kwargs) -> DataConfig:
             **kwargs,
         },
     )
+
+
+def raw_data_config_template(
```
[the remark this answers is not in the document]
trajepl: It is fine to keep it for me. I created the template module for:
1. demonstrate popular data config.
2. simplify the call action.

trajepl (approved):

PR #335, opened 2023-06-09
cleanup deadcode

guotuofeng:

### DIFF
--- a/olive/workflows/run/config.py
+++ b/olive/workflows/run/config.py
@@ -190,7 +190,10 @@ def _resolve_evaluator(v, values):
         return v
 
     evaluator = v.get("evaluator")
-    if not isinstance(evaluator, str):
+    if isinstance(evaluator, dict):
+        for idx, metric in enumerate(evaluator.get("metrics", [])):
+            evaluator["metrics"][idx] = _resolve_data_config(metric, values, "data_config")
+    elif not isinstance(evaluator, str):
         return v
 
     # resolve evaluator name to evaluators member config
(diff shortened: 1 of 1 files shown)


olive/workflows/run/config.py
```diff
@@ -125,7 +124,6 @@ def validate_evaluation_only(cls, v):
 
     @validator("passes", pre=True, each_item=True)
     def validate_pass_host_evaluator(cls, v, values):
-        v = _resolve_system(v, values, "host")
```
jambayk, line 128: I don't think this is deadcode, we use the `pass_config.host` here https://github.com/microsoft/Olive/blob/main/olive/workflows/run/run.py#L153. Users are able to specify pass specific hosts if they want.

olive/workflows/run/config.py
```diff
@@ -192,12 +190,7 @@ def _resolve_evaluator(v, values):
         return v
 
     evaluator = v.get("evaluator")
-    if isinstance(evaluator, dict):
-        v["evaluator"] = _resolve_system(evaluator, values, "target")
-        for metrics in v["evaluator"].get("metrics", []):
```
jambayk, line 197: LIne 195 is not used but I think 197-... is required. The evaluator for passes can be specified as dict. They might have `data_config` that need to be resolved.

jambayk: Maybe it can be changed to be similar to line 110 under `resolve_evaluators` 
```
if isinstance(evaluator_dict): 
   for idx, metric in enumerate(v["evaluator].get("metrics", [])):
            v["evaluator"]["metrics"][idx] = _resolve_data_config(metric, values, "data_config")
``` 

This way it is clear that the original `v` is being updated.

olive/workflows/run/config.py
guotuofeng: I could not find the the host for pass is configured in either olive_pass.py or in pass_config.py. where is the schema defined?

trajepl (approved):

jambayk (approved):

PR #352, opened 2023-06-14
☁️ AML Dataset

trajepl:
## Describe your changes
Add AML dataset support for `data_dir`, `user_script` and `script_dir` for evaluation and pass running.
1. In local system, olive will download aml dataset to local, then start olive optimization.
2. In aml system, olive can directly mount the data to evaluation/pass_run aml job with the way of `Input(type=AssetTypes.URI_FOLDER, path="azureml://xxx/xxx"`
3. Remove AzureCliCredential as we update the pipeline with SP credential.
4. Remove AzureCLI


olive/resource_path.py
```diff
@@ -490,8 +535,25 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
...
         return str(new_path)
 
 
-OLIVE_RESOURCE_ANNOTATIONS = Optional[Union[Path, str, ResourcePath, ResourcePathConfig]]
+OLIVE_RESOURCE_ANNOTATIONS = Optional[Union[str, Path, ResourcePath, ResourcePathConfig]]
+OLIVE_DOWNLOAD_SUCCESSFUL_FLAG = ".olive_resource_flag"
+LOCAL_CACHE_DIR = "./cache/aml_datastore_data"
+
+
+def get_local_path(resource_path: OLIVE_RESOURCE_ANNOTATIONS) -> str:
```
jambayk, line 547: Instead of doing all of the flag manipulation, I think it might be easier to do cache the remote resource like in cache https://github.com/microsoft/Olive/blob/main/olive/cache.py#L113

olive/common/config_utils.py
```diff
@@ -191,6 +192,16 @@ def validate_object(v, values, field):
     return v
 
 
+def validate_resource_path(v, values, field):
+    from olive.resource_path import create_resource_path
+
+    try:
+        v = create_resource_path(v)
+    except (ValueError, ValidationError) as e:
```
guotuofeng, line 200: ValidationError is also ValueError

olive/common/user_module_loader.py
```diff
@@ -16,11 +16,11 @@ class UserModuleLoader:
     Only used for objects that are not json serializable.
     """
 
-    def __init__(self, user_script: Optional[Union[Path, str]], script_dir: Optional[Union[Path, str]] = None):
-        self.user_script = user_script
-        self.script_dir = script_dir
+    def __init__(self, user_script: OLIVE_RESOURCE_ANNOTATIONS, script_dir: OLIVE_RESOURCE_ANNOTATIONS = None):
```
guotuofeng, line 19: what's the scenario that we need the user_script is in AzureML? I am supposing it should be only local file.

olive/passes/olive_pass.py
```diff
@@ -80,7 +81,7 @@ def __init__(
         if self._requires_user_script:
             self._user_module_loader = UserModuleLoader(self._config["user_script"], self._config["script_dir"])
         if self._requires_data_config:
-            data_config = self._config["data_config"] or {}
+            data_config = self._config.get("data_config") or {}
```
guotuofeng, line 84: get("data_config", {})

trajepl: `data_config` is in self._config but the value is None. So the logic here is like:
```python
if 'data_config' not in self._config or self._config['data_config'] is None:
    data_config = {}
```

olive/resource_path.py
trajepl: Coool, updated.

olive/common/user_module_loader.py
trajepl: It is modified to align the dependencies interface. Then Olive can consume all the path dependencies with ResourcePath. Also if user have to mound their code which is hosted in AML, they can directly pass the url.

examples/resnet/conda.yaml
```diff
@@ -5,6 +5,7 @@ dependencies:
   - python=3.8.13
   - pip=20.2
   - pip:
+      - azureml-fsspec
```
jambayk, line 8: nit: this is probably not required? Since it is on the aml side there is no use for fsspec.

examples/resnet/resnet_ptq_cpu.json
```diff
@@ -46,7 +46,14 @@
                     ],
                     "user_config":{
                         "user_script": "user_script.py",
-                        "data_dir": "data",
+                        "data_dir": {
+                            "type": "azureml_datastore",
+                            "config": {
+                                "azureml_client": "azureml_client",
```
jambayk, line 52: do we need the user to provide this parameter? For the model path and azureml system, we auto-insert the `azureml_client` if it is not present in the resource config.

olive/cache.py
```diff
@@ -100,8 +99,6 @@ def clean_pass_run_cache(pass_type: str, cache_dir: Union[str, Path] = ".olive-c
 
     This function deletes all runs for a given pass type as well as all child models and evaluations.
     """
-    assert pass_type.lower() in PASS_REGISTRY, f"Invalid pass type {pass_type}"
```
jambayk, line 103: why is this deleted?

olive/cache.py
```diff
@@ -148,6 +145,21 @@ def get_non_local_resource(resource_path: ResourcePath, cache_dir: Union[str, Pa
...
+    If the resource path is a local resource, the path is returned.
+    If the resource path is an AzureML resource, the resource is downloaded to the cache and the path is returned.
+    """
+    if resource_path is None:
+        return None
+
+    if resource_path.is_local_resource() or resource_path.is_string_name():
+        return resource_path.get_path()
+    elif resource_path.is_azureml_resource():
+        return get_non_local_resource(resource_path).get_path()
```
jambayk, line 160: This is missing the cache_dir argument. 

We might have to add `.olive_cache` to gitignore since this cache is different from the run cache directory `cache`. Maybe we can find a way to propagate the cache from the run to other parts? Like setting an environment variable like `OLIVE_CACHE_DIR`? And make the default value for cache_dir be `.olive_cache` only if `OLIVE_CACHE_DIR` is not set.

examples/resnet/conda.yaml
trajepl: Oh, sry. We may need that still. As in aml system, we call `fsspec.info` to check if the input is file or folder.

jambayk: Doesn't this change to local file/folder when seen from the aml_pass_runner or aml_evaluation_runner? 

The check is only done by the aml system running on the local environment. but this conda yaml is for the azureml job environment.

examples/resnet/resnet_ptq_cpu.json
```diff
@@ -46,7 +46,14 @@
                     ],
                     "user_config":{
                         "user_script": "user_script.py",
-                        "data_dir": "data",
+                        "data_dir": {
```
jambayk, line 49: should we create two versions of this json (one with local data and another with aml data)? So that we have a resnet example that works for users with no access to the datastore.

trajepl: Agree. Will do it.

olive/evaluator/metric.py
```diff
@@ -202,3 +203,11 @@ def get_latency_config_from_metric(metric: Metric):
             sleep_num = sub_type.metric_config.sleep_num
             break
     return warmup_num, repeat_test_num, sleep_num
+
+
+def localize_metrics_user_config(metrics: List[Metric]) -> List[Metric]:
```
guotuofeng, line 208: >localize [](http://example.com/codeflow?start=4&length=8)

use another term. 

Basically, localize refer to make english to chinese, which is confusing when user refer to globalization/localization.

olive/passes/olive_pass.py
guotuofeng: then the original line seems ok. we don't need change it. right?

olive/cache.py
```diff
@@ -148,6 +145,21 @@ def get_non_local_resource(resource_path: ResourcePath, cache_dir: Union[str, Pa
     return local_resource_path
 
 
+def get_local_path(resource_path: ResourcePath, cache_dir: Union[str, Path] = ".olive-cache"):
```
guotuofeng, line 148: >cache_dir [](http://example.com/codeflow?start=48&length=9)

this parameter seems useless? if so, we could make the function as method of ResourcePath?

guotuofeng (changes requested):
it seems get_local_path should be method of ResourcePath?

examples/resnet/conda.yaml
trajepl: Ic

guotuofeng:
> def get_non_local_resource(resource_path: ResourcePath, cache_dir: Union[str, Path] = ".olive-cache"):

the name is confusing. how about get_resource_local_path?

---
Refers to: olive/cache.py:110 in 3cb8a2b. [](commit_id = 3cb8a2b10584a3b7359eb4b9247d9d3f1f1f5fd1, deletion_comment = False)

olive/passes/olive_pass.py
```diff
@@ -80,7 +81,7 @@ def __init__(
         if self._requires_user_script:
```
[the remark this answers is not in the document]
guotuofeng: Yeah, @jambay is right, we need download the resource only when needed to be run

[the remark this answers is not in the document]
trajepl: Seems, it is not a good idea to make `user_script` and `script_dir` accepting azureml resource. It will bring more code structure changes. Let us only accept local file/folder now.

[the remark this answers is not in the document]
trajepl: Based current design, the UserModuleLoad is used to check the function name right or not in `user_script`. 
It is not good idea to make it lazy I think. Let us say user may want to know the olive config is right or not before any pass run/evaluation.

olive/cache.py
trajepl: As the `get_non_local_resouce`(need to import resource_path) is under `cache.py`, and we also need to implement `get_local_path`(need to rely on methods in cache) which is **used broadly in Pass related files**. 

Also as the registry need to import all the detailed passes, it brings the circular imports issues. 

Then I simplify remove this check as it involves all the passes related python module.

Would live to add it back if we have more good solutions.

trajepl:
> download_resource

Updated.

olive/evaluator/metric.py
trajepl: Remove it as we already have the cache.

olive/passes/pass_config.py
```diff
@@ -61,7 +59,7 @@ def get_user_script_config(
 ) -> Dict[str, PassConfigParam]:
     type_ = str
     if allow_path:
-        type_ = Union[Path, str]
+        type_ = OLIVE_RESOURCE_ANNOTATIONS
```
guotuofeng, line 62: >OLIVE_RESOURCE_ANNOTATIONS [](http://example.com/codeflow?start=16&length=26)

revert to Union[Path, str]?

olive/evaluator/metric_config.py
```diff
@@ -2,22 +2,22 @@
...
 SLEEP_NUM = 0
 
-
+user_path_config = ["script_dir", "data_dir", "user_script"]
 _common_user_config = {
-    "script_dir": ConfigParam(type_=Union[Path, str]),
-    "user_script": ConfigParam(type_=Union[Path, str]),
-    "data_dir": ConfigParam(type_=Union[Path, str]),
+    "script_dir": ConfigParam(type_=OLIVE_RESOURCE_ANNOTATIONS, is_path=True),
+    "user_script": ConfigParam(type_=OLIVE_RESOURCE_ANNOTATIONS, is_path=True),
```
guotuofeng, line 19: >user_script [](http://example.com/codeflow?start=5&length=11)

do we need to make the user_script and script_dir consistent with olive pass one?

olive/common/user_module_loader.py
```diff
@@ -16,11 +16,21 @@ class UserModuleLoader:
     Only used for objects that are not json serializable.
     """
 
-    def __init__(self, user_script: Optional[Union[Path, str]], script_dir: Optional[Union[Path, str]] = None):
-        self.user_script = user_script
-        self.script_dir = script_dir
-        if self.user_script:
-            self.user_module = import_user_module(user_script, script_dir)
+    def __init__(
+        self, user_script: Optional[OLIVE_RESOURCE_ANNOTATIONS], script_dir: Optional[OLIVE_RESOURCE_ANNOTATIONS] = None
```
guotuofeng, line 20: >OLIVE_RESOURCE_ANNOTATIONS [](http://example.com/codeflow?start=36&length=26)

I suppose this piece of code is forget to revert?

olive/systems/azureml/aml_system.py
```diff
@@ -130,79 +146,79 @@ def run_pass(
...
+            model_workspace_config = model_resource_path.get_aml_client_config().get_workspace_config()
+            system_workspace_config = self.azureml_client_config.get_workspace_config()
+            for key in model_workspace_config:
+                if model_workspace_config[key] != system_workspace_config[key]:
+                    logger.warning(
+                        f"Model workspace {model_workspace_config} is different from system workspace"
+                        f" {system_workspace_config}. Olive will download the model to local storage, then upload it to"
+                        "the system workspace."
+                    )
+                    return Input(type=asset_type, path=get_local_path(model_resource_path))
```
guotuofeng, line 175: >get_local_path [](http://example.com/codeflow?start=55&length=14)

for azureml scenario, do we need download the model locally and upload it again for pipelien jobs?

olive/systems/azureml/aml_system.py
```diff
@@ -130,79 +146,79 @@ def run_pass(
...
+            logger.debug(f"Creating aml model for job output {model_resource_path}")
+            aml_model = retry_func(
+                ml_client.models.create_or_update,
+                [
+                    Model(
+                        path=model_resource_path.get_path(),
+                        name="olive-backend-model",
+                        description="Model created by Olive backend. Ignore this model.",
+                        type=AssetTypes.CUSTOM_MODEL,
                     )
```
guotuofeng, line 192: why do we need create the model?

olive/passes/pass_config.py
trajepl: Emm, it is still the resouce_path. But we filter the case of aml resouce out for script_dir and user_scriptl

olive/systems/azureml/aml_system.py
```diff
@@ -130,79 +146,79 @@ def run_pass(
...
 
-            # we keep the model path as a string in the config file
-            if model_resource_path.type != ResourceType.StringName:
-                model_path = Input(type=asset_type, path=model_resource_path.get_path())
-                model_json["config"]["model_path"] = None
+        return None
+
+    def _create_model_args(self, model_json: dict, tmp_dir: Path):
+        args = {}
+        for item in ["model_script", "script_dir", "model_path"]:
```
guotuofeng, line 215: >"model_script", "script_dir" [](http://example.com/codeflow?start=21&length=28)

seems we need not handle the path operation for model_script and script_dir?

olive/systems/docker/utils.py
```diff
@@ -58,23 +59,23 @@ def create_metric_volumes_list(metrics: List[Metric], container_root_path: Path,
     for metric in metrics:
         metric_path = container_root_path / "metrics" / metric.name
         if metric.user_config.user_script:
-            user_script_path = str(Path(metric.user_config.user_script).resolve())
-            user_script_name = Path(metric.user_config.user_script).name
+            user_script_path = get_local_path(metric.user_config.user_script)
```
guotuofeng, line 62: >get_local_path [](http://example.com/codeflow?start=31&length=14)

no need call get_local_path for user_script and script_dir. Only need for data_dir

olive/passes/pass_config.py
guotuofeng: where do we filter them? I suggest to revert it to Union[Path, str] to keep consistence since all other places about script_dir and user_script is just local path.

the same with aml_system.py(model_script, script_dir).

olive/common/config_utils.py
```diff
@@ -159,6 +159,7 @@ class ConfigParam(ConfigBase):
     required: bool = False
     default_value: Any = None
     is_object: bool = False
+    is_path: bool = False
```
guotuofeng, line 162: >is_path: bool = False [](http://example.com/codeflow?start=4&length=21)

does the movement impact the generated document? I see the subclass has document but the base class doesn't have.

olive/evaluator/metric_config.py

olive/passes/pass_config.py
trajepl: Updated to revert back to path|str for all user_script and script_dir

olive/cache.py
jambayk: I see. the lazy import make sense for now. Thanks!

olive/systems/azureml/aml_system.py
jambayk: This is a workaround to use job outputs as aml inputs. 
image

This comment describes why we do it. I have asked the aml sdk team for help since aml jobs are supposed to be supported paths https://learn.microsoft.com/en-us/azure/machine-learning/how-to-manage-models?view=azureml-api-2&tabs=cli%2Cuse-local#supported-paths but it doesn't work as expected. They haven't replied yet so we are using this as a workaround for now.

olive/common/config_utils.py
trajepl: image
I think it is fine. The description all set in `PassConfigParam`

olive/azureml/azureml_client.py
```diff
@@ -98,22 +98,17 @@ def _get_credentials(self):
         2. DefaultAzureCredential
         3. InteractiveBrowserCredential
         """
-        from azure.identity import AzureCliCredential, DefaultAzureCredential, InteractiveBrowserCredential
+        from azure.identity import DefaultAzureCredential, InteractiveBrowserCredential
 
         logger.debug("Getting credentials for MLClient")
         try:
-            credential = AzureCliCredential()
+            credential = DefaultAzureCredential()
```
guotuofeng, line 105: nit: if we'd like to by pass managed identity, we could pass argument to Default one by disable it.

olive/systems/azureml/aml_system.py
guotuofeng: @jambayk , need we add TODO here?

olive/workflows/run/config.py
```diff
@@ -143,6 +158,18 @@ def validate_pass_search(cls, v, values):
...
+            data_dir_name = "data_dir"
+            for config_key, config_value in v["config"].items():
+                if config_key == data_dir_name and isinstance(config_value, dict):
+                    rp_type = config_value.get("type")
+                    if rp_type in AZUREML_RESOURCE_TYPES:
+                        rp_aml_client = config_value.get("config", {}).get("azureml_client")
+                        if not rp_aml_client:
+                            if "azureml_client" not in values:
+                                raise ValueError("azureml_client is required for azureml resource path in config")
+                            config_value["config"]["azureml_client"] = values["azureml_client"]
```
guotuofeng, line 171: it seems the logic is exactly same with above one? could we reuse it by using a method?

olive/azureml/azureml_client.py
trajepl: Sounds good to me. Will add it in a separated PR.

olive/workflows/run/config.py
trajepl: Updated.

guotuofeng (approved):

PR #358, opened 2023-06-16
enable linkcheck in doc build

guotuofeng:
## Describe your changes
enable linkcheck in doc build

### DIFF
--- a/.azure_pipelines/job_templates/olive-build-doc-template.yaml
+++ b/.azure_pipelines/job_templates/olive-build-doc-template.yaml
@@ -37,6 +37,7 @@ jobs:
           cd docs
           python -m pip install -r requirements.txt
           make html
+          make linkcheck
         windows: False
 
     - task:  PublishPipelineArtifact@1
--- a/docs/README.md
+++ b/docs/README.md
@@ -22,6 +22,7 @@ At `docs`:
 
 ```bash
 make html
+make linkcheck
 ```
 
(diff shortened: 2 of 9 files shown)


jambayk (commented):
How do we handle cases where some new files are being created in the same PR as the one adding new links? In this case the links to the new files won't exist at the time of the PR. 

I think it might be better to create a new pipeline that is triggered after a PR merges and only runs `make linkcheck`. This way we get a status of the links but it doesn't block the PR validation pipelines or the doc publish pipeline. 

So, it will be similar to this pipeline https://github.com/microsoft/Olive/blob/main/.azure_pipelines/build-doc.yaml.

guotuofeng:
> How do we handle cases where some new files are being created in the same PR as the one adding new links? In this case the links to the new files won't exist at the time of the PR.
> 
> I think it might be better to create a new pipeline that is triggered after a PR merges and only runs `make linkcheck`. This way we get a status of the links but it doesn't block the PR validation pipelines or the doc publish pipeline.
> 
> So, it will be similar to this pipeline https://github.com/microsoft/Olive/blob/main/.azure_pipelines/build-doc.yaml but only for linkcheck

It seems there are not ideal solution for your mentioned scenario. Another solution is to add the file firstly in a PR, then update the link in following PR, which is same with your suggestion in theory.

trajepl (approved):

PR #360, opened 2023-06-16
Update Vitis AI Quantization Pass

sheng-xiao:
New Features:
1. Support power-of-two quantization with QOP format.
2. Support more quantization configurations in power-of-two quantization in both QDQ and QOP format.
    a) Support signed and unsigned configurations.
    b) Support symmetry and asymmetry configurations.
    c) Support per-tensor and per-channel configurations.
3. Support onnx models larger than 2GB.
4. Support using CUDAExecutionProvider for calibration in quantization.

Bugfixes:
1. Fix the OverflowError caused by the "Inf" scale.
2. Fix some RuntimeErrors caused by quant_preprocessing.


olive/passes/onnx/vitis_ai/refine.py
```diff
@@ -8,10 +8,9 @@
 
 from olive.passes.onnx.vitis_ai.quant_utils import pos2scale, scale2pos
 
-logger = logging.getLogger(__name__)
-
 refine_op_type = ["DequantizeLinear", "QuantizeLinear"]
 postfix = "_Output"
+logging.basicConfig(level=logging.INFO)
```
jambayk, line 13: can you please use the `logger = logging.getLogger(__name__)` along with `logger.info`, `logger.debug`? This makes the logs from all olive code under the `olive` namespace. 

`logging.basicConfig(level=logging.INFO)` is not recommended for python libraries since it changes the root logger. The root logger should only be changed by the user themselves.

jambayk (changes requested):

olive/passes/onnx/vitis_ai/refine.py
guotuofeng: Jambay is right, there should only one global logging.basicConfig call in one python process.

olive/passes/onnx/vitis_ai_quantization.py
```diff
@@ -129,39 +140,66 @@
...
     ),
+    "need_layer_fusing": PassConfigParam(
+        type_=bool,
+        default_value=False,
+        searchable_values=Boolean(),
+        description="""
+            Perform layer fusion for conv-relu type operations
+        """,
+    ),
+    "execution_providers": PassConfigParam(
```
guotuofeng, line 156: the accelerator_spec on the run has the execution provider info. Is this one needed?

olive/passes/onnx/vitis_ai_quantization.py
```diff
@@ -189,7 +227,7 @@ class VitisAIQuantization(Pass):
 
     def _initialize(self):
         super()._initialize()
-        self.tmp_dir = tempfile.TemporaryDirectory()
+        self.tmp_dir = tempfile.TemporaryDirectory(prefix="olive_vaiq_tmp")
 
     @staticmethod
     def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
```
guotuofeng, line 233: this parameter contains the ep info. when running, this parameter is Pass.accelerator_spec.

olive/passes/onnx/vitis_ai_quantization.py
```diff
@@ -285,30 +321,57 @@ def _run_for_config(self, model: ONNXModel, config: Dict[str, Any], output_model
...
+        tmp_model_path = str(tmp_dir_path / Path(output_model_path).name)
 
         # get the dataloader
-        if config["dataloader_func"]:
+        if self._user_module_loader.user_module:
             dataloader = self._user_module_loader.call_object(
-                config["dataloader_func"],
-                config["data_dir"],
-                config["batch_size"],
+                self._fixed_params["dataloader_func"],
```
guotuofeng, line 336: @jambayk , I suppose you did change the fixed_param to config. right?

PR #364, opened 2023-06-19
Update Vitis-AI quantization pass

sheng-xiao:
New Features:
1. Support power-of-two quantization with QOP format.
2. Support more quantization configurations in power-of-two quantization in both QDQ and QOP format.
    a) Support signed and unsigned configurations.
    b) Support symmetry and asymmetry configurations.
    c) Support per-tensor and per-channel configurations.
3. Support onnx models larger than 2GB.
4. Support using CUDAExecutionProvider for calibration in quantization.
Bugfixes:
1. Fix the OverflowError caused by the "Inf" scale.
2. Fix some RuntimeErrors caused by quant_preprocessing.


sheng-xiao:
Hi @guotuofeng , I accidentally closed the previous PR: https://github.com/microsoft/Olive/pull/360, please help to review this new PR. I've solved the issues in the previous PR and merged the latest codes in the PR. Thanks very much.

olive/passes/onnx/vitis_ai_quantization.py
```diff
@@ -213,24 +242,20 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
         # exposed extra options config
         config.update(deepcopy(_exposed_extra_options_config))
         config.update(deepcopy(_extra_options_config))
+
+        # external data config
+        config.update(get_external_data_config())
+        config["execution_providers"] = PassConfigParam(
```
guotuofeng, line 248: It seems the execution provider param is needless? what's the consideration that allow user to override the execution providers passed by engine?

olive/passes/onnx/vitis_ai_quantization.py
```diff
@@ -6,28 +6,29 @@
...
 from olive.strategy.search_parameter import Boolean, Categorical, Conditional
 
 logger = logging.getLogger(__name__)
 
-# common config for vai_q_onnx quantization
+# common config for Vitis-AI quantization
 vai_q_onnx_quantization_config = {
     "data_dir": PassConfigParam(
-        type_=OLIVE_RESOURCE_ANNOTATIONS,
+        type_=Union[Path, str],
```
trajepl, line 31: Please rebase the code from main branch where we pushed a new changes which support AML dataset path for `data_dir`.

olive/passes/onnx/vitis_ai_quantization.py
```diff
@@ -288,29 +319,58 @@ def _run_for_config(self, model: ONNXModel, config: Dict[str, Any], output_model
...
+        tmp_dir_path = Path(tmp_dir.name)
+        tmp_model_path = str(tmp_dir_path / Path(output_model_path).name)
+
         # get the dataloader
+        # TODO: only use data config
         if config["dataloader_func"]:
             dataloader = self._user_module_loader.call_object(
                 config["dataloader_func"],
-                get_local_path(config["data_dir"]),
+                config["data_dir"],
```
trajepl, line 336: Ditto.

trajepl (dismissed):

olive/passes/onnx/vitis_ai_quantization.py
sheng-xiao: We want to use the execution providers to run the calibration in this quantization pass. For example, user can run quantize calibration on GPU to speed up.

guotuofeng: the execution provider associated with local target will be automatically populated to each pass as accelerator_spec. If you want to override the ep, you can use the execution_providers for engine like https://github.com/microsoft/Olive/blob/main/examples/directml/stable_diffusion/config_vae_decoder.json#LL89C1-L90C1

So, it seems this parameter could be removed.

sheng-xiao: Yes, we agree with you. We are directly using the Pass.accelerator.execution_provider in the codes. The 'config' here is a local variable and will not override the ep.

guotuofeng: The issue here is that the execution_providers is exposed to the user by _default_config. As the result, the generated document will includes the execution_providers as a parameter. Instead, we could remove the line here and explicitly pass it when invoking quantize_static in line 347

sheng-xiao: The problem here is that we cannot fetch "accelerator_spec.execution_provider" in the "_run_for_config" function in line 347, so we used a local variable to pass the value. Is there any other way to pass the ep value in "_run_for_config"?

jambayk: The same `accelerator_spec` is actually saved in the pass as `self._accelerator_spec` so you can access it in the pass itself. 

If you are doing this, please update this method to return `False`. https://github.com/microsoft/Olive/blob/main/olive/passes/olive_pass.py#L104

guotuofeng: @jambayk is right.

guotuofeng (approved):

PR #371, opened 2023-06-21
📈 Make data_config optional for Metric

trajepl:
## Describe your changes
Make data_config optional for Metrics. Then we can remove the pass specific changes for io_config. It will be helpful for the case to ignore the io_config in evaluator configs.


olive/evaluator/olive_evaluator.py
```diff
@@ -146,6 +147,27 @@ def evaluate(
                 raise TypeError(f"{metric.type} is not a supported metric type")
         return flatten_metric_result(metrics_res)
 
+    @staticmethod
+    def enhance_user_config_with_model_config(metric: Metric, model: OliveModel):
```
guotuofeng, line 151: >enhance_user_config_with_model_config [](http://example.com/codeflow?start=8&length=37)

fill_metric_user_config_with_model_io?

olive/snpe/data_loader.py
```diff
@@ -163,15 +164,15 @@ def load_data(self) -> Tuple[str, str, np.ndarray]:
 class SNPERandomDataLoader(SNPEDataLoader):
     def __init__(
         self,
-        io_config: dict,
+        io_config: IOConfig,
```
guotuofeng, line 167: >IOConfig [](http://example.com/codeflow?start=19&length=8)

better to keep original dict to keep the logic simpler instead of accept IOConfig and do conversion using dict()

olive/snpe/snpe.py
```diff
@@ -53,7 +54,7 @@ def get_throughput_options(self):
 
 
 class SNPEInferenceSession:
-    def __init__(self, model_path: str, io_config: dict, session_options: SNPESessionOptions = None):
+    def __init__(self, model_path: str, io_config: IOConfig, session_options: SNPESessionOptions = None):
```
guotuofeng, line 57: >IOConfig [](http://example.com/codeflow?start=51&length=8)

ditto here so to avoid import IOConfig

olive/model/__init__.py
```diff
@@ -454,9 +458,9 @@ def tensor_dtype_to_np_dtype(tensor_type):
                 io_config[f"{prefix}_shapes"].append(shape)
 
         # save io_config
-        self.io_config = io_config
+        self.io_config = IOConfig.parse_obj(io_config)
```
guotuofeng, line 461: > IOConfig.parse_obj(io_config) [](http://example.com/codeflow?start=24&length=30)

return dict instead of IOConfig will be much simpler. It is a standard behavior without needing know the definition.

guotuofeng (changes requested):
we need keep the original io_config as built-in dict instead of our customized IOConfig. In this way, the logic will be much clear than the IOConfig one. At the same time, the IOConfig introduce dependency on IOConfig definition.

trajepl:
Close this PR to split several changes.

olive/evaluator/olive_evaluator.py
```diff
@@ -128,6 +129,10 @@ def evaluate(
     ) -> MetricResult:
         metrics_res = {}
         for metric in metrics:
+            # use model io_config if user does not specify input_names and input_shapes
+            # only do this if data_config or dataloader is not provided
+            # priority: dataloader_func > data_config > user_config.input_names/input_shapes > model io_config
```
guotuofeng, line 134: @jambayk , please make sure the comment is right since we change the priority order you previous added.

olive/evaluator/olive_evaluator.py
```diff
@@ -146,6 +151,28 @@ def evaluate(
...
+        if not is_io_config_static(io_config):
+            logger.debug(
+                "Model input shapes are not static. Cannot use inferred input shapes for creating dummy data. This will"
+                " cause an error when creating dummy data for tuning."
+            )
+        if io_config and not metric.user_config.input_names and not metric.user_config.input_shapes:
+            # input_types is optional
+            metric.user_config.input_names = io_config["input_names"]
+            metric.user_config.input_shapes = io_config["input_shapes"]
+            metric.user_config.input_types = io_config.get("input_types")
```
trajepl, line 173: As input_types is optional, default value is float32, we use `get` here to handle the case of `KeyNotExists`

olive/model/__init__.py
```diff
@@ -488,7 +492,7 @@ def __init__(
         super().__init__(framework=Framework.PYTORCH, model_file_format=model_file_format, model_path=model_path)
 
         # io config for conversion to onnx
-        self.io_config = validate_config(io_config, IOConfig) if io_config else None
+        self.io_config = io_config if io_config and validate_config(io_config, IOConfig) else None
```
jambayk, line 495: if we want the dict, can you do `validate_config(io_config, IOConfig).dict() if io_config else None`? 

The io config validation does some conversions such as making `dynamic_axes` keys into int from string.

olive/model/model_config.py
```diff
@@ -12,10 +12,10 @@
 class IOConfig(ConfigBase):
     # TODO remove input names, shapes and types, turn to use olive dataset config.
     input_names: List[str]
-    input_shapes: List[List[int]] = None
+    input_shapes: List[List[Union[int, str]]] = None
```
jambayk, line 15: Why is `str` added? For the IOConfig use case, it will always need to be `int`.

olive/evaluator/olive_evaluator.py
jambayk: Yes, this new priority looks good to me.

olive/evaluator/olive_evaluator.py
```diff
@@ -128,6 +129,10 @@ def evaluate(
     ) -> MetricResult:
         metrics_res = {}
         for metric in metrics:
+            # use model io_config if user does not specify input_names and input_shapes
+            # only do this if data_config or dataloader is not provided
+            # priority: dataloader_func > data_config > user_config.input_names/input_shapes > model io_config
+            metric = OliveEvaluator.fill_metric_user_config_with_model_io(metric, model)
```
jambayk, line 135: I think it would be better and safer if we do not change the original object. We reuse the evaluator and metric objects for multiple models so I don't think the evaluation of one model should change the metric object. 

In general, I think we should not change any of our shared objects or object created from config jsons after config validation. It affects hashing+caching as well as sharing the object.

olive/evaluator/olive_evaluator.py
guotuofeng: if the input_types doesn't exist, should we ignore the assignment or assign None?

trajepl: The logic is like:
1. if user provide 'input_types' we use that.
2. if not, use `float32` as default.

olive/model/model_config.py
trajepl: To suppot the case where io_config generated from OnnxModel.

jambayk: I see. Thanks!

olive/evaluator/olive_evaluator.py
guotuofeng: I suppose io_config.get("input_types") would return None if no input_types?

olive/model/model_config.py
trajepl: In previous commit, I unified the `get_io_config` to return `IOConfig` which met failure for onnx model. But it is fine now, I change the `IOConfig` to `dict`.

olive/model/__init__.py
trajepl: Updated.

olive/evaluator/olive_evaluator.py

olive/evaluator/olive_evaluator.py
```diff
@@ -154,6 +155,9 @@ def evaluate(
     @staticmethod
     def fill_metric_user_config_with_model_io(metric: Metric, model: OliveModel):
```
guotuofeng, line 156: the method name should be updated?

olive/evaluator/olive_evaluator.py
trajepl: Yes, see this under `dataset.py`
image

olive/evaluator/olive_evaluator.py
trajepl: I prefer to `generate` as it won't change the original metrics.

jambayk:
Thanks! Everything LGTM. Please feel free to merge once Mike approves.

olive/evaluator/olive_evaluator.py
guotuofeng: better to add comments here for the None assignment.

olive/data/component/load_dataset.py
```diff
@@ -38,8 +38,8 @@ def huggingface_dataset(data_name=None, subset=None, split="validation", **kwarg
 
 
 @Registry.register_dataset()
-def dummy_dataset(input_names, input_shapes, input_types):
-    return DummyDataset(input_names, input_shapes, input_types)
+def dummy_dataset(input_shapes, input_names=None, input_types=None):
```
guotuofeng, line 41: >dummy_dataset(input_shapes, input_names=None, input_types=None) [](http://example.com/codeflow?start=4&length=63)

would the parameter order change impact existing code?

test/unit_test/systems/python_environment/test_python_environment_system.py
```diff
@@ -100,7 +100,7 @@ def test_evaluate_accuracy(self, mock_compute_accuracy):
         expected_call = mock_compute_accuracy.mock_calls[0]
         # python environment call
         actual_call = mock_compute_accuracy.mock_calls[1]
-        assert actual_call.args[0] == expected_call.args[0]
```
guotuofeng, line 103: why expected_call.args[0] is different with metric?

trajepl: It is changed with the reason of (`1actual_call.args[0]` is metric config):
1. `actual_call` directly call `compute_accuracy` which won't be generated config with model's io_config.
2. `exptected_call` call `evaluate_model` as the accuracy method for local system is private. But `evaluate_model` will generate a metric config with model's io_config. It made `actual_call.args[0] != expected_call.args[0]`

trajepl: I changed the actual_call with `evalaute_model` to align the tests. 

Also @jambayk  Could you help double check? If we can make sure the `compute_accuracy` won't be call beyond the `evalatue_model`, then it is fine. Otherwise, we may need to update metric io config in the level of `compute_latency/accuracy`

test/unit_test/systems/python_environment/test_python_environment_system.py
```diff
@@ -91,7 +91,7 @@ def test_evaluate_accuracy(self, mock_compute_accuracy):
         expected_res = local_system.evaluate_model(model, [metric], DEFAULT_CPU_ACCELERATOR)["accuracy-accuracy_score"]
 
         # execute
-        actual_res = self.system.evaluate_accuracy(model, metric)[AccuracySubType.ACCURACY_SCORE]
+        actual_res = self.system.evaluate_model(model, [metric], DEFAULT_CPU_ACCELERATOR)["accuracy-accuracy_score"]
```
guotuofeng, line 94: @jambayk , I will let you decide the change. the change LGTM except the test case change.

guotuofeng (commented):
I will let @jambayk confirm the change of test

test/unit_test/systems/python_environment/test_python_environment_system.py
jambayk: Yes, it looks good to me. thanks!

jambayk (approved):

PR #372, opened 2023-06-21
🐞Fix the execution_provider load issue

trajepl:
## Describe your changes
JSON cannot persist the Tuple for python. 

It will cause that the load_model cannot load the correct execution providers.

image

### DIFF
--- a/olive/common/ort_inference.py
+++ b/olive/common/ort_inference.py
@@ -43,8 +43,8 @@ def get_ort_inference_session(
             sess_options.add_session_config_entry(key, value)
 
     if isinstance(execution_provider, list):
-        # execution_provider may be a list of tuples where the first item in each tuple is the EP name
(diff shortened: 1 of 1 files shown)


olive/common/ort_inference.py
```diff
@@ -43,8 +43,8 @@ def get_ort_inference_session(
             sess_options.add_session_config_entry(key, value)
 
     if isinstance(execution_provider, list):
-        # execution_provider may be a list of tuples where the first item in each tuple is the EP name
-        execution_provider = [i[0] if isinstance(i, tuple) else i for i in execution_provider]
+        # execution_provider may be a list of tuples/lists where the first item in each tuple is the EP name
+        execution_provider = [i[0] if isinstance(i, tuple) or isinstance(i, list) else i for i in execution_provider]
```
guotuofeng, line 47: isinstance(i, (tuple, list)) should be concise

trajepl: Updated.

PR #380, opened 2023-06-26
Make Olive compatible with ORT 1.16.0 (nightly)

jambayk:
## Describe your changes
ORT main branch recently removed some functions in https://github.com/microsoft/onnxruntime/commit/89f8f20a61ceac0276c42d894c38400d86b175d0. 

For vitis ai code, the import of these functions is made lazy (only imported where needed). vitis ai code needs to be updated later to actually be compatible with ORT 1.16.0. Since this version of ORT is not released yet, it is not urgent. 

For onnx quantization passes, `optimize_model` config parameter is removed in ORT and we do the same. Using this option was already not recommended so I think it might be okay to remove it from Olive onnx quantization passes and always set it to `False` in older versions of ORT. As described in the code comments, we already have `quant_preprocess` which does the same optimization step as part of preprocessing. 
The commit above also says "This PR also remove the optimization from quantization. User needs to call pre-process to optimize the model". 

Whisper example requirements updated to `onnx==1.14.0` since the workflow fails for ORT nightly otherwise. ORT versions 1.15.0,1.15.1 are now compatible with onnx 1.14.0 because of https://github.com/microsoft/Olive/commit/a531f77b87738c0916e08d123da74f30532a5a26.

### DIFF
--- a/olive/passes/onnx/vitis_ai/calibrate.py
+++ b/olive/passes/onnx/vitis_ai/calibrate.py
@@ -19,7 +19,7 @@ from onnxruntime.quantization.calibrate import (
     CalibrationDataReader,
     MinMaxCalibrater,
 )
-from onnxruntime.quantization.quant_utils import QuantType, clone_model_with_shape_infer
+from onnxruntime.quantization.quant_utils import QuantType
 
 from olive.passes.onnx.vitis_ai.quant_utils import PowerOfTwoMethod, quantize_data_pof2s
 
@@ -59,6 +59,10 @@ class PowOfTwoCalibrater(CalibraterBase):
         Make all quantization_candidates op type nodes as part of the graph output.
         :return: augmented ONNX model
         """
+        # not supported in ORT >= 1.16.0
+        # TODO: Update code to support different versions of ORT
+        from onnxruntime.quantization.quant_utils import clone_model_with_shape_infer
+
         model = clone_model_with_shape_infer(self.model)
 
         self.tensors_to_calibrate, value_infos = self.select_tensors_to_calibrate(model)
@@ -77,7 +81,6 @@ class PowOfTwoCalibrater(CalibraterBase):
         self.intermediate_outputs = []
 
     def collect_data(self, data_reader: CalibrationDataReader):
-
         while True:
             inputs = data_reader.get_next()
             if not inputs:
@@ -134,7 +137,6 @@ class PowOfTwoCollector(CalibrationDataCollector):
         )
 
     def collect(self, name_to_arr):
-
         self.name_to_arr = name_to_arr
 
         return
@@ -170,7 +172,6 @@ def create_calibrator_power_of_two(
     execution_providers=["CPUExecutionProvider"],
     extra_options={},
 ):
-
     calibrator = None
 
     # default settings for min-max algorithm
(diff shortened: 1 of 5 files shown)


olive/passes/onnx/vitis_ai/calibrate.py
```diff
@@ -59,6 +59,10 @@ def augment_graph(self):
         Make all quantization_candidates op type nodes as part of the graph output.
         :return: augmented ONNX model
         """
+        # not supported in ORT >= 1.16.0
+        # TODO: Update code to support different versions of ORT
+        from onnxruntime.quantization.quant_utils import clone_model_with_shape_infer
```
guotuofeng, line 64: should we add try/except for ImportError? if the exception happens, it seems we can directly use the model without clone? @sheng-xiao

jambayk: Maybe we can do it in a separate PR? I have created a TODO here for tracking.

olive/passes/onnx/vitis_ai/quantize.py
```diff
@@ -166,6 +166,10 @@ def quantize_static(
     if not op_types_to_quantize or len(op_types_to_quantize) == 0:
         op_types_to_quantize = list(QLinearOpsRegistry.keys())
 
+    # not supported in ORT >= 1.16.0
+    # TODO: Update code to support different versions of ORT
+    from onnxruntime.quantization.quant_utils import load_model
+
     model = load_model(Path(model_input), optimize_model)
```
guotuofeng, line 173: @sheng-xiao , it seems the parameter is already removed in 1.16, should we remove it so that we could use load_model_with_shape_infer?

olive/passes/onnx/vitis_ai/calibrate.py
guotuofeng: sure

guotuofeng (approved):

olive/passes/onnx/vitis_ai/calibrate.py
jambayk: Thanks!

olive/passes/onnx/vitis_ai/quantize.py
jambayk: This is the commit of interest FYI https://github.com/microsoft/onnxruntime/commit/89f8f20a61ceac0276c42d894c38400d86b175d0

`load_model_with_shape_infer` was changed to add a step `add_infer_metadata(model)` which was previously under `load_model`. If this is important, we might have to use `load_model` for versions < 1.16.0 and `load_model_with_shape_infer` for >=? 

same for `clone_model_with_shape_infer` which doesn't exist in 1.16.0

jambayk: Created an issue to track it https://github.com/microsoft/Olive/issues/382

sheng-xiao: Thanks for your update on making the import lazily. We tested for the publicly released ORT 1.14/1.15 only. We are now working on updating Vitis-AI pass to be compatible with the nightly ORT version. Further discussion will be at https://github.com/microsoft/Olive/issues/382

PR #381, opened 2023-06-26
Support data_config in OpenVINOQuantization quantization pass

jambayk:
## Describe your changes
Add support for `data_config` in `OpenVINOQuantization` pass. If data_config is given, the pass converts it into the correct `openvino.tools.pot.api.DataLoader` format.

### DIFF
--- a/olive/passes/openvino/quantization.py
+++ b/olive/passes/openvino/quantization.py
@@ -115,10 +115,9 @@ class OpenVINOQuantization(Pass):
 
         class _OVDataloader(DataLoader):
             def __init__(self, dataloader):
-                self.dataloader = dataloader
                 self.data = []
                 self.labels = []
-                for data, label in self.dataloader:
+                for data, label in dataloader:
                     if isinstance(data, dict):
                         data = {k: np.array(v) for k, v in data.items()}
                     elif isinstance(data, tuple):
--- a/olive/passes/openvino/quantization.py
+++ b/olive/passes/openvino/quantization.py
@@ -5,8 +5,9 @@
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Union
 
+import numpy as np
+
 from olive.cache import get_local_path
-from olive.common.user_module_loader import UserModuleLoader
 from olive.hardware.accelerator import AcceleratorSpec
 from olive.model import OpenVINOModel
 from olive.passes import Pass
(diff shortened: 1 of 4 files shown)


olive/passes/openvino/quantization.py
```diff
@@ -97,3 +103,38 @@ def _run_for_config(self, model: OpenVINOModel, config: Dict[str, Any], output_m
...
+        Create an openvino.tools.pot.api.DataLoader instance from a common dataloader.
+        """
+        try:
+            from openvino.tools.pot.api import DataLoader
+        except ImportError:
+            raise ImportError("Please install olive-ai[openvino] to use OpenVINO pass")
+
+        class _OVDataloader(DataLoader):
+            def __init__(self, dataloader):
+                self.dataloader = dataloader
```
guotuofeng, line 118: >self.dataloader [](http://example.com/codeflow?start=16&length=15)

it seems the data member is not needed since all the data is already transformed to data/label?

olive/passes/openvino/quantization.py
```diff
@@ -97,3 +103,38 @@ def _run_for_config(self, model: OpenVINOModel, config: Dict[str, Any], output_m
         openvino_model = OpenVINOModel(model_path)
 
         return openvino_model
+
+    def _create_dataloader(self, common_dataloader):
+        """
+        Create an openvino.tools.pot.api.DataLoader instance from a common dataloader.
```
guotuofeng, line 109: >dataloader [](http://example.com/codeflow?start=75&length=10)

torch Dataloader?

jambayk: It is a torch dataloader in most of our use cases and examples. But I don't think it will be the same since the `dataloader` component in `olive.data` is not specified to return a torch dataloader. 
For example, the dummy datacontainer dataloader is not torch dataloader. 
@trajepl please correct me if I am wrong.

olive/passes/openvino/quantization.py
jambayk: Removed.

olive/passes/openvino/quantization.py
trajepl: It is right. The `dataloader` in `olive.data`

But somehow the `Dataloader` in OpenVINO is more like the `Dataset` in torch.

guotuofeng (approved):

PR #384, opened 2023-06-27
🐎Raise the failure in time after docker.container.run failed

trajepl:
## Describe your changes
Sometimes, the CI pipeline failed with Connection error for docker tests, but the raised error is shown as None metric results. 
That is caused by:
docker.container.run failed, but olive continue run to get the empty results
MetricResult accept empty value which will put off the failure again.
image

### DIFF
--- a/olive/common/config_utils.py
+++ b/olive/common/config_utils.py
@@ -129,7 +129,7 @@ class ConfigListBase(ConfigBase):
 
 
 class ConfigDictBase(ConfigBase):
-    __root__: Dict[str, Any] = None
+    __root__: Dict[str, Any]
 
     def __iter__(self):
         return iter(self.__root__)
--- a/olive/common/config_utils.py
+++ b/olive/common/config_utils.py
@@ -116,7 +116,7 @@ class ConfigBase(BaseModel):
 
 
 class ConfigListBase(ConfigBase):
-    __root__: List[Any] = None
+    __root__: List[Any]
 
     def __iter__(self):
         return iter(self.__root__)
--- a/olive/systems/docker/docker_system.py
+++ b/olive/systems/docker/docker_system.py
@@ -157,7 +157,11 @@ class DockerSystem(OliveSystem):
             )
             for line in container.logs(stream=True):
                 logger.info(line.strip().decode())
-            exit_code = container.wait()
+            exit_code = container.wait()["StatusCode"]
+            if exit_code != 0:
+                raise docker.errors.ContainerError(
+                    container, exit_code, eval_command, self.image, "Docker container evaluation failed"
+                )
             logger.debug("Docker container evaluation completed successfully")
         finally:
             # clean up dev mount regardless of whether the run was successful or not
@@ -171,12 +175,6 @@ class DockerSystem(OliveSystem):
                     volumes=[dev_mount_str, clean_up_mount_str],
                 )
                 logger.debug("Dev mount cleaned up successfully")
-
-        exit_code = exit_code["StatusCode"]
-        if exit_code != 0:
-            raise docker.errors.ContainerError(
-                container, exit_code, eval_command, self.image, "Docker container evaluation failed"
-            )
 
         metric_json = Path(output_local_path) / f"{eval_output_name}"
         return metric_json
--- a/olive/systems/docker/docker_system.py
+++ b/olive/systems/docker/docker_system.py
@@ -171,11 +171,11 @@ class DockerSystem(OliveSystem):
                     volumes=[dev_mount_str, clean_up_mount_str],
                 )
 
-        if docker.version_info[0] >= 3:
-            exit_code = exit_code["StatusCode"]
-
+        exit_code = exit_code["StatusCode"]
         if exit_code != 0:
-            raise RuntimeError(f"Container exited with code {exit_code}")
+            raise docker.errors.ContainerError(
+                container, exit_code, eval_command, self.image, "Docker container evaluation failed"
+            )
 
         metric_json = Path(output_local_path) / f"{eval_output_name}"
         return metric_json
--- a/olive/systems/docker/docker_system.py
+++ b/olive/systems/docker/docker_system.py
@@ -156,7 +156,8 @@ class DockerSystem(OliveSystem):
                 image=self.image, command=eval_command, volumes=volumes_list, detach=True, **run_command
             )
             for line in container.logs(stream=True):
-                print(line.strip().decode())
+                logger.info(line.strip().decode())
+            exit_code = container.wait()
             logger.debug("Docker container evaluation completed successfully")
         finally:
             # clean up dev mount regardless of whether the run was successful or not
@@ -169,7 +170,12 @@ class DockerSystem(OliveSystem):
                     command=f"python {clean_up_mount_path} --dev_mount_path {dev_mount_path}",
                     volumes=[dev_mount_str, clean_up_mount_str],
                 )
-                logger.debug("Dev mount cleaned up successfully")
+
+        if docker.version_info[0] >= 3:
+            exit_code = exit_code["StatusCode"]
+
+        if exit_code != 0:
+            raise RuntimeError(f"Container exited with code {exit_code}")
 
         metric_json = Path(output_local_path) / f"{eval_output_name}"
         return metric_json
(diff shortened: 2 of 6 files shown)


test/integ_test/evaluator/docker_eval/test_docker_evaluation.py
```diff
@@ -56,3 +59,17 @@ def test_evaluate_model(self, model_cls, model_config, metric, expected_res):
         for sub_type in metric.sub_types:
             joint_key = joint_metric_key(metric.name, sub_type.name)
             assert actual_res[joint_key].value >= expected_res
+
+    @patch("olive.systems.docker.docker_system.docker_utils.create_evaluate_command")
```
trajepl, line 63: I used the `patch` in integration test. Please let me know if you insist to keep no any patch in integration tests.

 https://stackoverflow.com/questions/52107522/is-it-in-considered-a-good-practice-to-mock-in-integration-test

xiaoyu-work: Yeah I don't quite like mock in integ test. This test is an unit test rather an integ test. What is the purpose of this test?

guotuofeng: whenever we use patch, it is something unit test.

olive/common/config_utils.py
```diff
@@ -129,7 +129,7 @@ def __len__(self):
 
 
 class ConfigDictBase(ConfigBase):
-    __root__: Dict[str, Any] = None
```
guotuofeng, line 132: should the ConfigListBase.__root__ be non-None?

olive/systems/docker/docker_system.py
```diff
@@ -169,7 +170,12 @@ def _run_container(self, tempdir, model: OliveModel, metrics: List[Metric], cont
                     command=f"python {clean_up_mount_path} --dev_mount_path {dev_mount_path}",
                     volumes=[dev_mount_str, clean_up_mount_str],
                 )
-                logger.debug("Dev mount cleaned up successfully")
+
+        if docker.version_info[0] >= 3:
+            exit_code = exit_code["StatusCode"]
+
+        if exit_code != 0:
+            raise RuntimeError(f"Container exited with code {exit_code}")
```
guotuofeng, line 178: raise ContainerError?

olive/systems/docker/docker_system.py
```diff
@@ -169,7 +170,12 @@ def _run_container(self, tempdir, model: OliveModel, metrics: List[Metric], cont
                     command=f"python {clean_up_mount_path} --dev_mount_path {dev_mount_path}",
                     volumes=[dev_mount_str, clean_up_mount_str],
                 )
-                logger.debug("Dev mount cleaned up successfully")
+
+        if docker.version_info[0] >= 3:
```
guotuofeng, line 174: why need this check?

test/integ_test/evaluator/docker_eval/test_docker_evaluation.py
guotuofeng: https://en.wikipedia.org/wiki/Mock_object

image

trajepl: Removed.

olive/systems/docker/docker_system.py
trajepl: Updated with docker.errors.ContainerError 
image

olive/systems/docker/docker_system.py
trajepl: For docker which version less that 3, the exit_code is integer not dict. 
But I double checked the docker version in Olive, which should be the latest 6.x.x where the exit_code is dictionary. 

Removed this check.

olive/common/config_utils.py
trajepl: I think it is fine. It cannot be None but can be `{}` since the typing of `__root__` is Dict.
We should avoid to assign None to a Dict type, right?

trajepl: Sorry, updated the List to None as well.

olive/systems/docker/docker_system.py
```diff
@@ -171,5 +172,11 @@ def _run_container(self, tempdir, model: OliveModel, metrics: List[Metric], cont
                 )
                 logger.debug("Dev mount cleaned up successfully")
 
+        exit_code = exit_code["StatusCode"]
+        if exit_code != 0:
+            raise docker.errors.ContainerError(
+                container, exit_code, eval_command, self.image, "Docker container evaluation failed"
+            )
+
```
guotuofeng, line 180: better to move this part into the try clause

olive/systems/docker/docker_system.py
```diff
@@ -171,5 +172,11 @@ def _run_container(self, tempdir, model: OliveModel, metrics: List[Metric], cont
                 )
                 logger.debug("Dev mount cleaned up successfully")
```
guotuofeng, line 173: @xiaoyu-work , why we need start another docker to clean up the mount path? for the mount path in docker container, the container is already removed. for the host path, it is temp folder which will be automatically removed?

olive/systems/docker/docker_system.py
```diff
@@ -156,7 +156,8 @@ def _run_container(self, tempdir, model: OliveModel, metrics: List[Metric], cont
                 image=self.image, command=eval_command, volumes=volumes_list, detach=True, **run_command
             )
             for line in container.logs(stream=True):
-                print(line.strip().decode())
+                logger.info(line.strip().decode())
+            exit_code = container.wait()
```
guotuofeng, line 160: >container [](http://example.com/codeflow?start=24&length=9)

do we need call container.remove()?

olive/systems/docker/docker_system.py
xiaoyu-work: I didn't add this. Seems @jambayk added this for cleaning dev mount https://github.com/microsoft/Olive/pull/210. I have a same question here. Can we just remove the container?

guotuofeng: @jambayk

guotuofeng: @trajepl , it seems [dev_mount_cleanup.py](https://github.com/microsoft/Olive/pull/210/files#diff-b6ef65dfa0431903cc41c3a9a6ecc495b211f1844bb33a6c83e9919085c26abc) could be removed and so did create_dev_cleanup_mount

olive/systems/docker/docker_system.py
trajepl: Done.

guotuofeng:
>                 )

add try/except here to avoid the cleanup?

---
Refers to: olive/systems/docker/docker_system.py:176 in 6052d74. [](commit_id = 6052d74ce14941ea8672951c9a503aacf3284f74, deletion_comment = False)

olive/systems/docker/docker_system.py
```diff
@@ -156,7 +156,12 @@ def _run_container(self, tempdir, model: OliveModel, metrics: List[Metric], cont
                 image=self.image, command=eval_command, volumes=volumes_list, detach=True, **run_command
             )
             for line in container.logs(stream=True):
-                print(line.strip().decode())
+                logger.info(line.strip().decode())
+            exit_code = container.wait()["StatusCode"]
```
guotuofeng, line 160: >container [](http://example.com/codeflow?start=24&length=9)

container.remove()?

olive/systems/docker/docker_system.py
trajepl: This part from #210 is to restart the container with `root` user to remove the generated files in container with `root` permission

This image shows the host case which will create the new `pycache` with local users.
image

But when it comes to container, Olive run container with `root` users who will generated the file in mounted path which cannot be removed when the context of temp directory ends.

trajepl: Run docker with `--user $(id -u):$(id -g)` would help. But there is another issue when the docker is started with host user but the user is not in sudo group, huggingface model download will fail as they will cache the model under `/.cache/xxx` which is the root path. 

I will start a new PR if there is another elegant way.

olive/systems/docker/docker_system.py
trajepl: ignore the remove in case user wants to debug

guotuofeng (approved):

olive/systems/docker/docker_system.py
jambayk: I added the dev mount clean up because of the reason Jiapeng mentioned above. Even thought we mount a temporary directory, it cannot automatically clean up the pycache which is created with root permissions by the container.

PR #387, opened 2023-06-29
MoE Experts Distributor

shaahji:
MoE Experts Distributor

Introducing a pass to distribute experts in onnx model by splitting the graph equally, making each individual such graph capable of running on its own GPU.

NOTE: The implementation requires a custom onnxruntime build with "--use_mpi" flag enabled.

### DIFF
--- a/olive/passes/onnx/moe_experts_distributor.py
+++ b/olive/passes/onnx/moe_experts_distributor.py
@@ -1,0 +1,275 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+import json
+import pprint
+from pathlib import Path
+from typing import Any, Callable, Dict, List
+
+import numpy
+import onnx
+from google.protobuf.json_format import MessageToDict
+from google.protobuf.message import Message
+from onnxruntime.transformers.onnx_model import OnnxModel
+from pydantic import validator
+
+from olive.hardware.accelerator import AcceleratorSpec
+from olive.model import DistributedOnnxModel, ONNXModel
+from olive.passes import Pass
+from olive.passes.onnx.common import get_external_data_config
+from olive.passes.pass_config import PassConfigParam
+
+_domain = "com.microsoft"
+_json_separators = (",", ": ")
+_expert_pattern = [
+    "Slice",  # 14
+    "Concat",  # 13
+    "MatMul",  # 12
+    "Cast",  # 11
+    "Relu",  # 10
+    "MatMul",  # 9
+    "Cast",  # 8
+    "Slice",  # 7
+    "Mul",  # 6
+    "Div",  # 5
+    "Add",  # 4
+    "Gather",  # 3
+    "Shape",  # 2
+    "Reshape",  # 1
+    "Reshape",  # 0
+]
+
+
+def _dump_graph(node: Message, filepath: str):
+    with open(filepath, "wt") as strm:
+        json.dump(MessageToDict(node), fp=strm, indent=2, sort_keys=True, separators=_json_separators)
+        strm.flush()
+
+
+def _iterate_experts(model: OnnxModel, producers: Dict[str, Message]):
+    for node in model.get_nodes_by_op_type(_expert_pattern[0]):
+        path = model.match_parent_path(node, _expert_pattern[1:], output_name_to_node=producers)
+        if path:
+            path.reverse()
+            path.append(node)
+            yield path
+
+
+def _create_node_name(model: OnnxModel, op_type: str, prefix_a: str, prefix_b: str):
+    prefix = ""
+    last_slash = -1
+    for i in range(0, min(len(prefix_a), len(prefix_b))):
+        if prefix_a[i] == prefix_b[i]:
+            prefix += prefix_a[i]
+            if prefix_a[i] == "/":
+                last_slash = i
+        else:
+            break
+
+    if last_slash > 0:
+        prefix = prefix[: last_slash + 1]
+
+    if not prefix.endswith("/"):
+        prefix += "/"
+    prefix += op_type
+
+    return model.create_node_name(op_type, prefix)
+
+
+def _create_ranked_model(
+    nodes: Dict[str, Message],
+    producers: Dict[str, Message],
+    consumers: Dict[str, Message],
+    experts: List[List[str]],
+    world_size: int,
+    num_experts: int,
+    rank: int,
+):
+    experts_per_rank = int(num_experts / world_size)
+    start_index = rank * experts_per_rank
+    end_index = start_index + experts_per_rank
+
+    for expert in experts:
+        div_node = nodes[expert[5]]
+        concat_node = nodes[expert[13]]
+        reshape_node = nodes[expert[1]]
+
+        div_node_oname = div_node.output[0]
+        reshape_node_oname = reshape_node.output[0]
+
+        mul_nodes = [n for n in consumers[div_node_oname] if n.op_type == "Mul"]
+        matmul_nodes = [producers[iname] for iname in concat_node.input if producers[iname].op_type == "MatMul"]
+        slice_nodes = [n for n in consumers[reshape_node_oname] if n.op_type == "Slice"]
+
+        slice_nodes[start_index].input[1] = slice_nodes[0].input[1]
+
+        for i in range(num_experts):
+            if (i < start_index) or (i >= end_index):
+                mul_nodes[i].input.remove(div_node_oname)
+                concat_node.input.remove(matmul_nodes[i].output[0])
+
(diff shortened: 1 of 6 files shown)


olive/passes/onnx/moe_experts_distributor.py
```diff
@@ -0,0 +1,258 @@
...
+                _replace_constant_value(producers[concat_iname_0], int(num_experts / experts_per_rank))
+
+                concat_iname_1 = parent.input[1]
+                _replace_constant_value(producers[concat_iname_1], experts_per_rank)
+
+                break
+
+
+def run(world_size: int, input_filepath: str, output_dirpath: str, debug: bool = False):
+    basename = os.path.splitext(os.path.basename(input_filepath))[0]
```
jambayk, line 185: We prefer to use `pathlib.Path` instead of `os.path` to do path manipulations. Could you use that instead?

I think this becomes much more simplified too. 
line 185: `basename = Path(input_filepath).stem`
line 193: `dump_graph(model.model, str(Path(output_dirpath) / f"{basename}.graph"))`

similarly for the other instances.

shaahji: Done!

olive/passes/onnx/moe_experts_distributor.py
```diff
@@ -0,0 +1,259 @@
...
+    basename = Path(input_filepath).stem
+    output_dirpath = Path(output_dirpath)
+
+    model = OnnxModel(onnx.load_model(input_filepath))
+    producers = model.output_name_to_node()
+    consumers = model.input_name_to_nodes()
+
+    if debug:
+        OnnxModel.graph_topological_sort(model.model.graph)
+        _dump_graph(model.model, str(output_dirpath / f"{basename}.graph"))
```
jambayk, line 194: is there a reason you implemented `_dump_graph` instead of saving the model using `onnx.save_model`?

jambayk: or directly call `model.save_model_to_file` which also does the topological sort inside it? https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/onnx_model.py#L1047

shaahji: `_dump_graph` is in text format (basically, json) not the usual binary onnx format.

jambayk: Oh I see. Does it make it more readable?

shaahji: It's for debugging purposes and that's why wrapped with `debug` flag. It's simpler to compare to json objects than visually comparing two graphs in tool like netron. It has been handy.

jambayk: That sounds good to me. Was wondering what the differences were. Thanks!

examples/switch/model_4n_2l_8e.onnx
jambayk, line 1: Is there a way to get this model from an open source on the fly or create it? We try to not check in binary files into the repo if possible.

shaahji: Unfortunately not. The model is generated by the moe_examples repository and the requires custom onnxruntime build. 

Everything with distributed inferencing is largely work in progress. These issues will continue until onnxruntime publishes distributed variants of nightly builds or releases the same. At the moment, the feature is not publicly advertised.

devang-ml: The core of the example model is MoE layer that is using ORT MoE implementation.  ORT MoE is in https://github.com/pytorch/ort/tree/main/ort_moe . 

There are other MoE implementations available. In theory the moe_experts_dstributor.py should work for any MoE model. It may need update in pattern matching MoE layer and Experts.

We should just write a note in README saying that the input for this example should use ORT MoE implementation of MoE layer instead of adding a .onnx file in the repo.

olive/passes/onnx/moe_experts_distributor.py
```diff
@@ -0,0 +1,259 @@
...
+
+    experts = []
+    num_experts = None
+    for expert in _iterate_experts(model, producers):
+        if debug:
+            pprint.pprint([n.name for n in expert])
+
+        div_node = expert[5]
+        div_node_oname = div_node.output[0]
```
devang-ml, line 203: What is so special about 5th expert ?

shaahji: It's not the 5th expert. It's the 5th node of every expert which happens to be the DIV node. Indexing is based on what was matched to identify the experts.

olive/passes/onnx/moe_experts_distributor.py
```diff
@@ -0,0 +1,259 @@
...
+
+def _fix_shapes(
+    model: OnnxModel,
+    nodes: Dict[str, Message],
+    producers: Dict[str, Message],
+    consumers: Dict[str, Message],
+    experts: List[List[str]],
+    world_size: int,
+    num_experts: int,
+    rank: int,
```
trajepl, line 151: `rank` is not used in this function, do we  need to keep it here still? Maybe any other case will use `rank` in the future?

olive/passes/onnx/moe_experts_distributor.py
```diff
@@ -0,0 +1,259 @@
...
+):
+    experts_per_rank = int(num_experts / world_size)
```
trajepl, line 153: nit: int(a/b) -> a // b

olive/passes/onnx/moe_experts_distributor.py
```diff
@@ -0,0 +1,259 @@
...
+    """
+
+    @staticmethod
+    def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        config = {
+            "world_size": PassConfigParam(
+                type_=int,
+                default=2,
+                required=True,
+                description=("Number of GPU nodes to distribute the model for. Must be greater than 1."),
```
trajepl, line 251: Could we add validation to raise the unavailable config value? Refer to https://github.com/microsoft/Olive/blob/b11692168ae025d3544dff0348a109dea3a7d24e/olive/passes/snpe/conversion.py#L87

olive/passes/onnx/moe_experts_distributor.py
```diff
@@ -0,0 +1,259 @@
...
+
+    if debug:
+        OnnxModel.graph_topological_sort(model.model.graph)
+        _dump_graph(model.model, str(output_dirpath / f"{basename}.graph"))
+
+    experts = []
+    num_experts = None
+    for expert in _iterate_experts(model, producers):
+        if debug:
+            pprint.pprint([n.name for n in expert])
```
trajepl, line 200: How about to use olive.logger?

olive/passes/onnx/moe_experts_distributor.py
trajepl, line 251: Also quick question: does the ORT support multi-nodes inferencing for MoE model? If it supports, how we specify the hostfile/master ip/etc.?

Let us say, we have 8 machines, each machine with 4 GPUs. Then the world_size should be 32 if we use all of the GPUs. But if we just want run MoE model with partial GPUs, how we specify the GPUs we want to use? 

Not aim to block this PR, just wonder to figure out: is there any limitation to run MoE or other models which need to be served with multi-nodes/gpus.

shaahji: > Also quick question: does the ORT support multi-nodes inferencing for MoE model?

By multi-node if you mean, Multi GPU, then yes. ORT supports multi-gpu thru' a constant parameter `group_size` in the model itself. As for multi-machine, I don't suppose this is supported yet.

In short, ORT can use multiple GPUs on the same machine but there is support for gpus spread across multiple machine is possibly work in progress.

olive/passes/onnx/moe_experts_distributor.py
shaahji: This is meant for dev debugging only. No point polluting the user log output. To a user this wouldn't make much sense unless he/she is educated with the details of the implementation.

Note - This is _debug_ not _verbose_.

olive/passes/onnx/moe_experts_distributor.py
shaahji: Done!

olive/passes/onnx/moe_experts_distributor.py
shaahji: Done!

olive/passes/onnx/moe_experts_distributor.py
shaahji: Removed!

examples/switch/model_4n_2l_8e.onnx
shaahji: Added README.md

devang-ml (approved):

PR #388, opened 2023-06-29
Updated custom metric logic

xiaoyu-work:
## Describe your changes

1. Added `metric_func` config to custom metric. Olive will run the inference for custom eval func for user. User doesn't need to do inference by themselves.
2. Updated and fixed evaluator unit tests.

### DIFF
--- a/docs/source/tutorials/configure_metrics.rst
+++ b/docs/source/tutorials/configure_metrics.rst
@@ -165,6 +165,14 @@ In your :code:`"user_script.py"`, you need to define a function that takes in an
             # evaluate model
             # return metric value
 
+Alternatively, if you only need Olive run the inference and you will calculate the metric by yourself, you can speficy :code:`"metric_func": "None"` in the metric configuration.
+Olive will run the inference with you model with the data you provided, and return the inference results to you. You can then calculate the metric by yourself::
+
+        def metric_func(preds, targets):
+            # calculate metric
+            # return metric value
+
+If you provide both :code:`"evaluate_func"` and :code:`"metric_func"`, Olive will call :code:`"evaluate_func"` only.
 
 Multi Metrics configuration
 ----------------------------
--- a/examples/resnet/user_script.py
+++ b/examples/resnet/user_script.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 import torch
+import torchmetrics
 from onnxruntime.quantization.calibrate import CalibrationDataReader
 from pytorch_lightning import LightningDataModule, LightningModule
 from torch.utils.data import DataLoader, Dataset
@@ -10,7 +11,6 @@ from torchvision import transforms
 from torchvision.datasets import CIFAR10
 
 from olive.constants import Framework
-from olive.evaluator.accuracy import AccuracyScore
 from olive.model import OliveModel
 
 # -------------------------------------------------------------------------
@@ -137,7 +137,11 @@ def eval_accuracy(model: OliveModel, data_dir, batch_size, device, execution_pro
             preds.extend(outputs.tolist())
             target.extend(labels.data.tolist())
 
-    return AccuracyScore().measure(preds, target)
+    preds_tensor = torch.tensor(preds, dtype=torch.int)
+    target_tensor = torch.tensor(target, dtype=torch.int)
+    accuracy = torchmetrics.Accuracy()
+    result = accuracy(preds_tensor, target_tensor)
+    return result.item()
 
 
 # -------------------------------------------------------------------------
--- a/olive/evaluator/metric_config.py
+++ b/olive/evaluator/metric_config.py
@@ -18,7 +18,9 @@ user_path_config = ["data_dir"]
 _common_user_config = {
     "script_dir": ConfigParam(type_=Union[Path, str]),
     "user_script": ConfigParam(type_=Union[Path, str]),
+    "inference_settings": ConfigParam(type_=dict),
     "data_dir": ConfigParam(type_=OLIVE_RESOURCE_ANNOTATIONS, is_path=True),
+    "dataloader_func": ConfigParam(type_=Union[Callable, str], is_object=True),
     "batch_size": ConfigParam(type_=int, default_value=1),
     "input_names": ConfigParam(type_=List),
     "input_shapes": ConfigParam(type_=List),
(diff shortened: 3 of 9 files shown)


olive/evaluator/metric_config.py
```diff
@@ -29,17 +31,19 @@
...
         "post_processing_func": ConfigParam(type_=Union[Callable, str], is_object=True),
-        "inference_settings": ConfigParam(type_=dict),
     },
     "custom": {
-        "evaluate_func": ConfigParam(type_=Union[Callable, str], required=True, is_object=True),
+        "evaluate_func": ConfigParam(type_=Union[Callable, str], required=False, is_object=True),
+        "evaluate_func_pytorch": ConfigParam(type_=Union[Callable, str], required=False, is_object=True),
+        "evaluate_func_onnx": ConfigParam(type_=Union[Callable, str], required=False, is_object=True),
+        "evaluate_func_snpe": ConfigParam(type_=Union[Callable, str], required=False, is_object=True),
+        "evaluate_func_openvino": ConfigParam(type_=Union[Callable, str], required=False, is_object=True),
```
guotuofeng, line 44: I suppose the evaluate_func should be enough for the functionality. what's the consideration that expose those function?

guotuofeng (changes requested):
better not expose these suffix eval functions

olive/evaluator/metric_config.py
xiaoyu-work: In some cases, user needs to evaluate different type models. If we only have one custom eval func, it is not ideal for the user to evaluate all kinds of models in the same func.

examples/resnet/user_script.py
```diff
@@ -103,41 +101,8 @@ def resnet_calibration_reader(data_dir, batch_size=16):
...
-        for input_data, labels in dataloader:
-            if isinstance(input_data, dict):
-                result = sess(**input_data)
-            else:
-                result = sess(input_data)
-            outputs = post_process(result)
-            preds.extend(outputs.tolist())
-            target.extend(labels.data.tolist())
-
-    return AccuracyScore().measure(preds, target)
```
guotuofeng, line 140: we need keep these logic since user would have different logic generate the preds.

olive/evaluator/metric_config.py
guotuofeng: could you please clarify? I still think we don't need them.

examples/resnet/user_script.py
xiaoyu-work: Yeah this is what I mentioned in my PR description. I'm not sure what is the other use case for user for their own eval func. I only expose loaded model for the custom eval func now. I doubt if they insist to inference by themselves. If there is any different case from our inference.

olive/evaluator/metric_config.py
xiaoyu-work: You can reference current `eval_accuracy` custom eval func in resnet example. There is an IF statement to distinguish different model type and run the inference: onnx model or pytorch model. So the concerns here are:
1. If there are multiple model types in a same workflow, but we only have one custom eval func for them, all different types model evaluation logic will be put in a same evaluation func.
2. As for custom eval function, I prefer to offer user a loaded model as an argument in the custom eval func. Instead of pass an additional arg like model type, having separate eval func for different model type will be cleaner.

guotuofeng: that's is user logic, user should have the full control about how to do the evaluation/inference from their point of view. 
Basically, we should provide the functionality support to allow user have full control of get the predicts/target from raw data.

xiaoyu-work: I mean, giving them control of getting the predicts/target from raw data and having these 4 additional eval func are different topics. Like we discussed, I will add the capability of inference by user for this custom eval func back.

olive/evaluator/olive_evaluator.py
```diff
@@ -61,6 +61,18 @@ def get_inference_settings(self, metric: Metric) -> Dict[str, Any]:
             else None
         )
 
+    @abstractmethod
```
guotuofeng, line 64: It seems we can remove this method?

olive/systems/python_environment/python_environment_system.py
```diff
@@ -100,7 +100,7 @@ def evaluate_accuracy(self, model: ONNXModel, metric: Metric) -> float:
         """
         Evaluate the accuracy of the model.
         """
-        dataloader, post_func, _ = OliveEvaluator.get_user_config(metric)
+        dataloader, _, post_func = OliveEvaluator.get_user_config(metric)
```
guotuofeng, line 103: good catch

olive/evaluator/olive_evaluator.py
xiaoyu-work: What do you mean by removing this?

olive/evaluator/olive_evaluator.py
```diff
@@ -296,6 +319,18 @@ def _evaluate_onnx_accuracy(
...
+    def _evaluate_onnx_accuracy(
+        self,
+        model: ONNXModel,
+        metric: Metric,
+        dataloader: Dataset,
+        post_func=None,
+        device: Device = Device.CPU,
+        execution_providers: Union[str, List[str]] = None,
+    ) -> MetricResult:
+        preds, targets = self._inference(model, metric, dataloader, post_func, device, execution_providers)
```
guotuofeng, line 333: do we need enable io_bind for GPU?

olive/evaluator/olive_evaluator.py
```diff
@@ -95,16 +107,19 @@ def _evaluate_custom(
...
-        # return eval_func(model, metric, dataloader, device, post_func)
-        raw_res = eval_func(
-            model,
-            get_local_path(metric.user_config.data_dir),
-            metric.user_config.batch_size,
-            device,
-            execution_providers,
-        )
+        raw_res = None
+        if metric.user_config.metric_func:
```
guotuofeng, line 111: I suppose the priority of evaluation function should be higher than that of metric function

xiaoyu-work: Updated.

docs/source/tutorials/configure_metrics.rst
```diff
@@ -165,6 +165,14 @@ In your :code:`"user_script.py"`, you need to define a function that takes in an
...
             # return metric value
 
+Alternatively, if you only need Olive run the inference and you will calculate the metric by yourself, you can speficy :code:`"metric_func": "None"` in the metric configuration.
+Olive will run the inference with you model with the data you provided, and return the inference results to you. You can then calculate the metric by yourself::
+
+        def metric_func(preds, targets):
+            # calculate metric
+            # return metric value
+
+If you provide both :code:`"evaluate_func"` and :code:`"metric_func"`, Olive will call :code:`"metric_func"` only.
```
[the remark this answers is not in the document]
xiaoyu-work: Updated.

guotuofeng (approved):

PR #389, opened 2023-06-30
add Olive exception skeleton

guotuofeng:
## Describe your changes
Sometimes, the quantization would fail https://dev.azure.com/aiinfra/Model%20optimization%20Toolkit/_build/results?buildId=321080&view=logs&j=83e9455c-f768-5dd6-ffb8-4d16293cb8a9&t=8c02a2da-1821-5532-3cad-c631fe1ed2f8 

Olive should have the ability to handle this kind of error not caused by Olive bug. when this happens, the quantization should be ignored. 

As the result, we introduce OliveException system to allow Olive throw the ignorable exception in pass run

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -18,6 +18,7 @@ from olive.engine.packaging.packaging_config import PackagingConfig
 from olive.engine.packaging.packaging_generator import generate_output_artifacts
 from olive.evaluator.metric import Metric, MetricResult, joint_metric_key
 from olive.evaluator.olive_evaluator import OliveEvaluatorConfig
+from olive.exception import OliveException
 from olive.hardware import AcceleratorLookup, AcceleratorSpec, Device
 from olive.model import ModelConfig, OliveModel
 from olive.passes.olive_pass import Pass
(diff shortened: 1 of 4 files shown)


olive/engine/engine.py
```diff
@@ -817,6 +822,9 @@ def _run_passes(
             else:
                 signal = self._evaluate_model(model, model_id, evaluator_config, accelerator_spec)
             logger.debug(f"Signal: {signal}")
+        else:
+            signal = None
+            logger.warning("Skipping evaluation as model was pruned")
```
trajepl, line 827: As the `signal` will be set anytime, maybe we can remove the Line 851.

olive/engine/engine.py
```diff
@@ -883,6 +891,9 @@ def _run_pass(
                 input_model = self._prepare_non_local_model(input_model)
             try:
                 output_model = host.run_pass(p, input_model, output_model_path, pass_search_point)
+            except OliveException as e:
```
trajepl, line 894: Better to OlivePassException? As we aim to catch the exception from pass run.

olive/exception/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+
+class OlivePassException(OliveException):
+    """
+    Base class for Olive pass exceptions.
+    """
+
+    pass
+
+
+class OliveEvaluationException(OliveException):
```
trajepl, line 21: Will we need the evaluation exception class? As there is no any `prune` mechanism in evaluation part.

guotuofeng: this is used to raise if there are some failure when evaluation model.

jambayk (approved):

PR #390, opened 2023-06-30
⛏️Pin protobuf to 3.20.3

trajepl:
## Describe your changes

With onnx is released to version 1.14.0, the required protobuf is set as >3.20.2. But for `transformers`, I met the error of protobuf version. It failed when run following importing:
`from transformers import AutoModelForMaskedLM, AutoTokenizer, PreTrainedModel, PreTrainedTokenizerBase`

image

So I tried to pin the protobuf version as 3.20.3 to meet the requirements from `onnx` and `transformers`

LInks:
1. onnx protobuf limitation: https://github.com/onnx/onnx/blob/rel-1.14.0/requirements.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ onnx
 optuna
 pandas
 pydantic
-protobuf !=3.20.0, !=3.20.1
+protobuf==3.20.3
 pyyaml
 torch
 torchmetrics>=0.7.0, <0.10.1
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -3,7 +3,7 @@ onnx
 optuna
 pandas
 pydantic<2.0.0
-protobuf !=3.20.0, !=3.20.1
+protobuf==3.20.3
```
guotuofeng, line 6: could we pin using <=?

trajepl: Updated to <4.0.0

guotuofeng (approved):

PR #396, opened 2023-07-04
🈹Remove EP validation from olive model

trajepl:

### DIFF
--- a/olive/model/__init__.py
+++ b/olive/model/__init__.py
@@ -194,35 +194,6 @@ class ONNXModelBase(OliveModel):
         self.inference_settings = inference_settings
         self.use_ort_extensions = use_ort_extensions
 
-    def _is_valid_ep(self, filepath: str, ep: str = None):
-        # TODO: should be remove if future accelerators is implemented
-        # It should be a bug for onnxruntime where the execution provider is not be fallback.
-        import onnxruntime as ort
-
-        try:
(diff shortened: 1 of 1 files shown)


olive/model/__init__.py
```diff
@@ -194,35 +194,6 @@ def __init__(
...
-    def _is_valid_ep(self, filepath: str, ep: str = None):
-        # TODO: should be remove if future accelerators is implemented
-        # It should be a bug for onnxruntime where the execution provider is not be fallback.
-        import onnxruntime as ort
-
-        try:
-            sess_options = ort.SessionOptions()
-            if self.use_ort_extensions:
-                # register custom ops for onnxruntime-extensions
-                from onnxruntime_extensions import get_library_path
```
trajepl, line 206: This function filter the case in which we cannot create ort inference sessions in runtime.  I supposed we can replace it with `AcceleratorLookup.get_execution_providers_for_device(device)` where we can filter the EP with device and `ort.get_available_providers`.

please help double check.

guotuofeng: even if the EPs are filtered with available EPs, it is possible that fail when create inferencesession since the get available EP doesn't need the dependencies.

guotuofeng (changes requested):
we need reconsider the remove since the is_vald_ep is not same with the filter logic.

PR #401, opened 2023-07-06
🤗NER task support in Huggingface dataset

trajepl:
## Describe your changes
1. Add support to NER task dataset.
2. Attach the test config file for NER model

### DIFF
--- a/olive/data/component/post_process_data.py
+++ b/olive/data/component/post_process_data.py
@@ -30,3 +30,22 @@ def text_classification_post_process(_output_data, **kwargs):
         _, preds = torch.max(_output_data, dim=1)
 
     return preds
+
+
+@Registry.register_post_process()
+def ner_post_process(_output_data, **kwargs):
+    """Post-process data for NER task.
+
+    Args:
+        data (object): Model output to be post-processed.
+        **kwargs: Additional arguments.
+
+    Returns:
+        object: Post-processed data.
+    """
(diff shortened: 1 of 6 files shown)


lainey1570 (approved):
The PR looks good and it would meet my requirements for `ner`.

olive/data/component/post_process_data.py
```diff
@@ -30,3 +30,22 @@ def text_classification_post_process(_output_data, **kwargs):
         _, preds = torch.max(_output_data, dim=1)
 
     return preds
+
+
+@Registry.register_post_process()
+def ner_post_process(_output_data, **kwargs):
```
xiaoyu-work, line 36: Any documentations need to be updated about how to use this?

trajepl: https://github.com/microsoft/Olive/pull/401/files#diff-fce744f2268555668fc1868242a7833b6691a78bc883d9dbfc50fcc23f70a0b9

Here I add the template config file for `ner` task. This function will not be exposed to user directly now. When user set the task as ner, the corresponding pre_process, post_process will be picked.

trajepl: https://github.com/microsoft/Olive/blob/main/docs/architecture.md#data-container Here is the doc to use that function internally now.

lainey1570 (approved):

PR #402, opened 2023-07-06
Fix metric output bug

xiaoyu-work:
## Describe your changes

Fix metric output bug

### DIFF
--- a/olive/engine/packaging/packaging_generator.py
+++ b/olive/engine/packaging/packaging_generator.py
@@ -34,11 +34,8 @@ def generate_output_artifacts(
     if sum([len(f.nodes) if f.nodes else 0 for f in pf_footprints.values()]) == 0:
         logger.warning("No model is selected. Skip packaging output artifacts.")
         return
-    try:
-        if packaging_config.type == PackagingType.Zipfile:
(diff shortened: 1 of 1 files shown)


olive/engine/packaging/packaging_generator.py
```diff
@@ -34,8 +34,11 @@ def generate_output_artifacts(
     if sum([len(f.nodes) if f.nodes else 0 for f in pf_footprints.values()]) == 0:
         logger.warning("No model is selected. Skip packaging output artifacts.")
         return
-    if packaging_config.type == PackagingType.Zipfile:
-        _generate_zipfile_output(packaging_config, footprints, pf_footprints, output_dir)
+    try:
+        if packaging_config.type == PackagingType.Zipfile:
+            _generate_zipfile_output(packaging_config, footprints, pf_footprints, output_dir)
+    except Exception as e:
```
trajepl, line 40: Seems we should avoid to catch generic `Exception` from previous discussion about to unify the exception in Olive.  Not sure in this case, do we know the detailed exception if possible?

xiaoyu-work: Good catch! This shouldn't be added here. Let me remove this.

trajepl (approved):

PR #403, opened 2023-07-07
➕ Opt in input model evaluation

trajepl:
## Describe your changes
Currently, if you want to get the evaluation result from input_model, they need to set:
1. `evaluation_only`: in this case, olive only evaluate input model without passes run.
2. or to set metric `goal` to tell the metric tolerance based on the baseline.

Several customers would love to compare the input model metrics with output model's metrics by default. So in this PR, we make input model is evaluated as long as there is the given metrics. 
If user do not need to evaluate input model, they can manually set `evalute_input_model` as `False`.

With the PR, the default behaviors. is like:
1. evaluate input model with given metrics.
2. if user gives the metric goal, olive read the result in step 1 from cache.
3. then olive will run search or no_search passes w.r.t given metrics.

   Replace  `evaluation_only` with `evaluate_input_model` and set False by default.

### DIFF
--- a/.azure_pipelines/job_templates/olive-build-doc-template.yaml
+++ b/.azure_pipelines/job_templates/olive-build-doc-template.yaml
@@ -18,7 +18,7 @@ jobs:
         versionSpec: 3.8
       displayName: Use Python 3.8
 
-    # checout release branch if doc_version provided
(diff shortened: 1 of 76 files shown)


guotuofeng:
could use only do evaluation on input model without optimization?

PR #405, opened 2023-07-10
Add Stable Diffusion V2 support to DirectML examples

PatriceVignola:


examples/directml/stable_diffusion/stable_diffusion.py
```diff
@@ -188,13 +193,30 @@ def optimize(
 
     model_info = dict()
 
-    for submodel_name in ("text_encoder", "vae_encoder", "vae_decoder", "safety_checker", "unet"):
+    models_without_safety_checker = [
+        "stabilityai/stable-diffusion-2",
```
jstoecker, line 197: You should be able to programmatically check if the model has a checker and avoid baking model names using something like this:

```python
pipeline.safety_checker is None
```

examples/directml/stable_diffusion/stable_diffusion.py
```diff
@@ -257,15 +285,24 @@ def optimize(
...
-    parser.add_argument("--model_id", default="runwayml/stable-diffusion-v1-5", type=str)
+    parser.add_argument(
+        "--model_id",
+        default="runwayml/stable-diffusion-v1-5",
+        choices=(
+            "CompVis/stable-diffusion-v1-4",
+            "runwayml/stable-diffusion-v1-5",
+            "sayakpaul/sd-model-finetuned-lora-t4",
+            "stabilityai/stable-diffusion-2",
+        ),
```
jstoecker, line 304: This will prevent optimizing arbitrary LoRA models. It's probably not a big deal, but perhaps instead just warn the user if their choice isn't in this list of tested models.

examples/directml/stable_diffusion/user_script.py
```diff
@@ -148,22 +160,30 @@ def unet_load(model_name):
     return model
 
 
-def unet_conversion_inputs(model):
-    return tuple(unet_inputs(1, torch.float32).values())
+def unet_conversion_inputs_512(model):
```
jstoecker, line 163: We can avoid multiple functions by storing the image size in a global variable, and `stable_diffusion.py` can change the default value of that to 512 or 768 depending on the model requirements.

jstoecker (changes requested):

jstoecker (approved):

PatriceVignola:
@trajepl Is it possible to re-run the CI? Thanks!

PR #406, opened 2023-07-10
🎞️ Evaluate LLaMA Torch Model and Onnx Model simultaneously

trajepl:
## Describe your changes
For LLaMa model optimization, the input data for torch model and onnx model is different.
- torch model's input: input_ids, attention_mask.
- onnx model's input: input_ids, attention_mask, past_key_values[1..26]
Also as user cannot specify different dataloader function to provide different input data based on model type in `user_script.py`.
Then it will fail if we evaluate torch model and onnx model at the same time.
In this PR, we filter the torch model input based on io_config to evaluate onnx and torch model at the same run for LLaMA.

### DIFF
--- a/examples/open_llama/user_script.py
+++ b/examples/open_llama/user_script.py
@@ -4,20 +4,22 @@
 # --------------------------------------------------------------------------
 import torch
 
+from olive.constants import Framework
 
-# Helper latency-only dataloader that creates random tensors with no label
+
 class RandomDataLoader:
-    def __init__(self, create_inputs_func, batch_size, torch_dtype):
+    def __init__(self, create_inputs_func, batch_size, torch_dtype, model_framework=Framework.PYTORCH):
         self.create_input_func = create_inputs_func
         self.batch_size = batch_size
         self.torch_dtype = torch_dtype
+        self.model_framework = model_framework
 
     def __getitem__(self, idx):
         label = None
-        return self.create_input_func(self.batch_size, self.torch_dtype), label
+        return self.create_input_func(self.batch_size, self.torch_dtype, self.model_framework), label
 
 
-def dummy_inputs(batch_size, torch_dtype):
+def dummy_inputs(batch_size, torch_dtype, model_framework=Framework.PYTORCH):
     past_sequence_length = 1
     attention_mask_sequence_length = 1
     sequence_length = 2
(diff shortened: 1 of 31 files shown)


olive/model/hf_utils.py
```diff
@@ -32,7 +32,7 @@ def task_or_model_class_required(cls, v, values):
         if values["model_name"]:
             if not v and not values.get("task", None):
                 raise ValueError("Either task or model_class must be specified")
-            return v
+        return v
```
trajepl, line 35: Fix: when model_name is none, model_calss or task can not be set.

Based on following logics, model_name can be None.
image

olive/evaluator/olive_evaluator.py
```diff
@@ -576,6 +576,12 @@ def _evaluate_latency(
         session = model.prepare_session(inference_settings=self.get_inference_settings(metric), device=device)
 
         input_data, _ = next(iter(dataloader))
+        # filter out the input data that is not used in the model
+        input_names = model.get_io_config().get("input_names", {})
```
guotuofeng, line 580: why the input data for latency is required but accuracy doesn't need?

olive/model/__init__.py
```diff
@@ -670,21 +670,14 @@ def to_json(self, check_object: bool = False):
...
-            model_path=model_path,
+            **(kwargs or {}),
         )
         self.model_components = model_components
 
-    def load_model(self, rank: int = None):
-        raise NotImplementedError()
-
-    def prepare_session(self, inference_settings: Dict[str, Any], device: Device, rank: int = None):
-        raise NotImplementedError()
```
guotuofeng, line 686: is this stuff related to the PR?

guotuofeng:
It seems the small PR includes a lot of unrelated changes, could we split them?

guotuofeng:
with https://github.com/microsoft/Olive/pull/407, I suppose only model and user script need changes

guotuofeng (approved):

examples/open_llama/user_script.py
```diff
@@ -4,20 +4,22 @@
 # --------------------------------------------------------------------------
 import torch
 
+from olive.constants import Framework
```
xiaoyu-work, line 7: I'm okay to keep this for this example for now. But we should be aware of using Olive internal code in user script.

devang-ml: IMO, We don't want to use this path.

devang-ml (commented):
- Is it possible for user to provide separate user config for PT model and ONNX model? They know which pass is operating on which type of model?
- Do we want to force using same "metrics" for PT model as well as ONNX model?

examples/open_llama/user_script.py
trajepl: image

For customized evaluation function, we face the same question about using Olive code in user script.

I may think it is fine(not a strong preference) but we should maintain a nice tutorial to tell user how to use.

From user's side, I may think that there is not a big gap better separate user config and some internal Olive flags. 
But from Olive's side, as model type(torch, onnx, snpe, openvino and etc.) increasing, the separate user config may contain:
1. [model_type]_dataloader_func
2. [model_type]_evaluation_func
3. [model_type]_model_loader
In this way, olive have to maintain more similar configs for different model types.

PR #409, opened 2023-07-11
Unify data dir mechanism used by all the places in Olive

guotuofeng:
## Describe your changes
Currently, the following passes and evaluators using the data_dir:
* all quantization pass
* perf-tuning
* QAT including train_data_dir and val_data_dir
* metric config

The absolute path processing make the integration to AzureML very painful. We introduce the uniform data_root, which requires all the data used by all techniques should be put under the same data root directory. And the data_root should be able to passed to Olive from command line.

### DIFF
--- a/docs/source/tutorials/how_to_add_optimization_pass.md
+++ b/docs/source/tutorials/how_to_add_optimization_pass.md
@@ -39,6 +39,7 @@ takes an `AcceleratorSpec` as input and returns `Dict[str, PassConfigParam]`.
 
 - `required` : whether the parameter is required
 - `category`: The param category. It could be the following values:
+
     * `object` : whether the parameter is an object/function. If so, this parameter accepts the object or a string with the
     name of the object in the user script. The type must include `str`.
     * `path` : whether the parameter is a path. If so, this file/folder will be uploaded to the host system.
--- a/docs/source/tutorials/how_to_add_optimization_pass.md
+++ b/docs/source/tutorials/how_to_add_optimization_pass.md
@@ -38,11 +38,11 @@ takes an `AcceleratorSpec` as input and returns `Dict[str, PassConfigParam]`.
 - `type_` : type of the parameter
 
 - `required` : whether the parameter is required
-
-- `is_object` : whether the parameter is an object/function. If so, this parameter accepts the object or a string with the
+- `category`: The param category. It could be the following values:
+    * `object` : whether the parameter is an object/function. If so, this parameter accepts the object or a string with the
     name of the object in the user script. The type must include `str`.
-
-- `is_path` : whether the parameter is a path. If so, this file/folder will be uploaded to the host system.
+    * `path` : whether the parameter is a path. If so, this file/folder will be uploaded to the host system.
+    * `data`: whether the parameter is a data path, which will be used to do data path normalization based on the data root.
 
 - `description` : description of the parameter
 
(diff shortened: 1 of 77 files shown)


docs/source/tutorials/how_to_add_optimization_pass.md
```diff
@@ -39,11 +39,11 @@ takes an `AcceleratorSpec` as input and returns `Dict[str, PassConfigParam]`.
...
 - `required` : whether the parameter is required
-
-- `is_object` : whether the parameter is an object/function. If so, this parameter accepts the object or a string with the
+- `category`: The param category. It could be the following values:
+    * `object` : whether the parameter is an object/function. If so, this parameter accepts the object or a string with the
     name of the object in the user script. The type must include `str`.
-
-- `is_path` : whether the parameter is a path. If so, this file/folder will be uploaded to the host system.
+    * `path` : whether the parameter is a path. If so, this file/folder will be uploaded to the host system.
+    * `data`: whether the parameter is a data path, which will be used to do data path normalization based on the data root.
```
trajepl, line 46: I saw it can be `none`, in which case we need that case?

docs/source/tutorials/how_to_add_optimization_pass.md
```diff
@@ -39,11 +39,11 @@ takes an `AcceleratorSpec` as input and returns `Dict[str, PassConfigParam]`.
 - `type_` : type of the parameter
 
 - `required` : whether the parameter is required
-
-- `is_object` : whether the parameter is an object/function. If so, this parameter accepts the object or a string with the
+- `category`: The param category. It could be the following values:
+    * `object` : whether the parameter is an object/function. If so, this parameter accepts the object or a string with the
```
trajepl, line 43: To add blank space after `category`L42 and `data`L46.

olive/engine/engine.py
```diff
@@ -254,6 +254,7 @@ def register_pass(
     def run(
         self,
         input_model: OliveModel,
+        data_root: str = None,
```
trajepl, line 257: Double check: data_root accepts only `str` and all the operation about resource type will be after `normalize_data_path`, right?

guotuofeng: yes

olive/workflows/run/config.py
```diff
@@ -55,6 +55,7 @@ class RunConfig(ConfigBase):
     azureml_client: AzureMLClientConfig = None
     input_model: ModelConfig
     systems: Dict[str, SystemConfig] = None
+    data_root: str = None
```
xiaoyu-work, line 58: Can you also update the documentation about data_root config and how it is related to data_dir for each pass and evaluator?

olive/cache.py
```diff
@@ -165,6 +166,44 @@ def get_local_path(resource_path: Optional[ResourcePath], cache_dir: Union[str,
...
+    if not data_dir_str:
+        data_full_path = data_root
+    elif Path(data_dir_str).is_absolute():
+        data_full_path = data_dir_str
+    else:
+        if data_root:
+            if isinstance(data_dir, ResourcePath) and data_dir.is_azureml_resource():
+                raise ValueError("could not append AzureML data to data_root")
+
+            # we cannot use Path to join the path. If the data_root is something like: azureml://, then Path will
```
xiaoyu-work, line 188: Does this case mean data_root an AML data, and data_dir is the relative path to AML data root path?

trajepl: Based on Mike's design doc.
The logic will be like:
1. if data_root is given, but data_dir not, just use data_root.
2. if data_root is not given, but data_dir is, use data_dir.
3. When both of them are set, concatenate as data_root / data_dir.

In this way, there won't be too many absolute data path.

xiaoyu-work: I mean this logic works for both local data and AML data?

Also, does this logic works if the user wants to use both AML data and local data? Seems data_root can only be one of them?

guotuofeng: If user want to use both AML data and local data at the same time, he need to specify the absolute local path.

docs/source/tutorials/how_to_add_optimization_pass.md
guotuofeng: it is the default value

olive/workflows/run/config.py
guotuofeng: sure, will do

olive/cache.py
guotuofeng: the documentation demonstrate how to use both of them

trajepl (approved):

PR #418, opened 2023-07-17
Add Aml sku support for AMLSystem

xiaoyu-work:
## Describe your changes

Add Aml sku support for AMLSystem


olive/systems/azureml/aml_system.py
```diff
@@ -92,6 +97,27 @@ def _create_environment(self, docker_config: AzureMLDockerConfig):
             return Environment(image=docker_config.base_image, conda_file=docker_config.conda_file_path)
         raise ValueError("Please specify DockerConfig.")
 
+    def _create_compute(self, aml_compute_config: AmlComputeConfig):
```
guotuofeng, line 100: what's the consideration that we need create aml compute? IMHO, the compute creation should be user's responsibility?

guotuofeng: @xiaoyu-work , should the aml compute creation happens when running python -m olive.workflows.run --setup stage?

olive/systems/system_config.py
```diff
@@ -15,6 +15,15 @@
 from olive.systems.common import AzureMLDockerConfig, LocalDockerConfig, SystemType
 
 
+class AmlComputeConfig(ConfigBase):
+    name: str = "olive-default"
+    size: str = "STANDARD_DS3_v2"
```
guotuofeng, line 20: user need explicitly provide name, size, location instead of using default ones

olive/systems/system_config.py
```diff
@@ -30,11 +39,18 @@ class DockerTargetUserConfig(TargetUserConfig):
...
-    aml_compute: str
+    aml_compute: str = None
+    aml_compute_config: AmlComputeConfig = None
     aml_docker_config: AzureMLDockerConfig
     instance_count: int = 1
     is_dev: bool = False
 
+    @validator("aml_compute_config", pre=True)
+    def _validate_aml_compute_config(cls, v, values):
+        if v is None and values.get("aml_compute") is None:
```
guotuofeng, line 50: should not create default AmlComputeConfig, users need provide the name, size, location explicitly

olive/systems/azureml/aml_system.py
xiaoyu-work: I would like we create the compute at runtime as this is a part of the process of olive workflow.

olive/systems/azureml/aml_system.py
```diff
@@ -92,6 +97,27 @@ def _create_environment(self, docker_config: AzureMLDockerConfig):
...
+            cluster_basic = AmlCompute(
+                name=aml_compute_config.name,
+                type="amlcompute",
+                size=aml_compute_config.size,
+                location=aml_compute_config.location,
+                min_instances=aml_compute_config.min_instances,
+                max_instances=aml_compute_config.max_instances,
+                idle_time_before_scale_down=aml_compute_config.idle_time_before_scale_down,
+            )
+            ml_client.begin_create_or_update(cluster_basic).result()
```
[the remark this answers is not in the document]
xiaoyu-work: If the compute with the same name exists, we will get an error here and throw OliveSystemException.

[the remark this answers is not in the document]
trajepl: I may think that: 
1. every computer created by Olive should be shutdown properly after the optimization done
2. If there is the same name, it means the aml compute instance is exactly same with the existed one. Why we throw the OliveSystemException?
3. How to reuse the exited aml compute:
     - if the existed compute is running with other optimization job (should wait or throw the error?
     - if the existed compute is not activated, we can directly leverage that?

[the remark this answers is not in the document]
xiaoyu-work: The original thought is, the functionalities of `aml_compute` and `aml_compute_config` are different. The `aml_compute` config means the user wants to use the existing one, and `aml_compute_config` means the user wants to create a new one with user-provided name, SKU, etc. If the compute name exists in `aml_compute_config` and reusing that compute with `aml_compute_config`, this will confuse the boundaries between these 2 configurations. (`aml_compute_config` this term may confuse the user unless we have a clear documentation about it).

But I agree we should consider shutting it down or deleting it after the workflow run.

azureml_scripts/manage_compute_instance.py
```diff
@@ -0,0 +1,142 @@
...
+except ImportError:
+    raise ImportError("azure-ai-ml is not installed. Please install azure-ai-ml packages to use this script.")
+
+try:
+    from azure.identity import AzureCliCredential, DefaultAzureCredential, InteractiveBrowserCredential
+except ImportError:
+    raise ImportError("azure-identity is not installed. Please install azure-identity packages to use this script.")
+
+
+def get_args():
```
guotuofeng, line 20: >get_args [](http://example.com/codeflow?start=4&length=8)

need we put this script into olive wheel?

xiaoyu-work: Good point. Added to Olive wheel.

guotuofeng (approved):

PR #419, opened 2023-07-18
☂️Opt in input model eval

trajepl:
## Describe your changes
Currently, if you want to get the evaluation result from input_model, they need to set:
1. `evaluation_only`: in this case, olive only evaluate input model without passes run.
2. or to set metric `goal` to tell the metric tolerance based on the baseline.

Several customers would love to compare the input model metrics with output model's metrics by default. So in this PR, we make input model is evaluated as long as there is the given metrics. 
If user do not need to evaluate input model, they can manually set `evalute_input_model` as `False`.

With the PR, the default behaviors. is like:
1. evaluate input model with given metrics.
2. if user gives the metric goal, olive read the result in step 1 from cache.
3. then olive will run search or no_search passes w.r.t given metrics.

   Replace  `evaluation_only` with `evaluate_input_model` and set False by default.

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -259,7 +259,7 @@ class Engine:
         packaging_config: Optional[PackagingConfig] = None,
         output_dir: str = None,
         output_name: str = None,
-        evaluation_only: bool = False,
+        evaluate_input_model: bool = True,
     ):
         """
         Run all the registered Olive passes on the input model and produce one or more candidate models.
@@ -270,7 +270,7 @@ class Engine:
             output_dir: output directory for the output model
             output_name: output name for the output model, if output_name is provided, the output
                 model will be saved to engine's output_dir with the prefix of output_name.
-            evaluation_only: if evaluation_only is True, run the evaluation on the input model and return the results.
+            evaluate_input_model: if evaluate_input_model is True, run the evaluation on the input model.
 
         Return:
             if search strategy is None, all passes are run in the order they were registered.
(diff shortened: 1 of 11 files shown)


olive/engine/engine.py
```diff
@@ -299,18 +299,24 @@ def run(
...
-                    result_name = f"{prefix_output_name}metrics"
+                    logger.info(f"Input model evaluation results: {results}")
+                    result_name = f"{prefix_output_name}_input_model_metrics"
                     results_path = output_dir / f"{result_name}.json"
                     with open(results_path, "w") as f:
                         json.dump(results.to_json(), f, indent=4)
+                    logger.info(f"Saved evaluation results of input model to {results_path}")
                     outputs[accelerator_spec] = results
-                elif self.no_search:
+                    if not self.passes:
```
xiaoyu-work, line 315: In this case, does the Olive only run for evaluating the input model?

xiaoyu-work:
In zipfile packaging mode, we have a json file to store metrics called `metrics.json`. Can you please add input model metric to this file if we both need input model evaluation and zipfile packaging? https://github.com/microsoft/Olive/blob/main/olive/engine/packaging/packaging_generator.py#L122-L125

olive/engine/engine.py
trajepl: Yes, as there is no any passes were given.

trajepl:
> In zipfile packaging mode, we have a json file to store metrics called `metrics.json`. Can you please add input model metric to this file if we both need input model evaluation and zipfile packaging? https://github.com/microsoft/Olive/blob/main/olive/engine/packaging/packaging_generator.py#L122-L125

image
Added

xiaoyu-work:
> > In zipfile packaging mode, we have a json file to store metrics called `metrics.json`. Can you please add input model metric to this file if we both need input model evaluation and zipfile packaging? https://github.com/microsoft/Olive/blob/main/olive/engine/packaging/packaging_generator.py#L122-L125
> 
> image Added

Thanks! Seems accuracy dropped xd

xiaoyu-work (approved):

trajepl:
> > > In zipfile packaging mode, we have a json file to store metrics called `metrics.json`. Can you please add input model metric to this file if we both need input model evaluation and zipfile packaging? https://github.com/microsoft/Olive/blob/main/olive/engine/packaging/packaging_generator.py#L122-L125
> > 
> > 
> > image Added
> 
> Thanks! Seems accuracy dropped xd

Ahahah, no worries. It is from unit test where we use dummy model with forked value.

PR #427, opened 2023-07-21
Whisper: use quick merge in prepost to support large models 

jambayk:
## Describe your changes
Based on https://github.com/microsoft/onnxruntime-extensions/pull/496, this PR uses `quick_merge` to merge the pre, core and post models instead of `onnx.compose.merge_models` which fails when the models being merged are large (>2GB). The failure is because `onnx.compose.merge_models` calls `onnx.checker.check_model` internally on a large modelproto. 

Also perfom model check on the saved file path instead of the loaded model to avoid the same issue in Olive. 

### DIFF
--- a/olive/passes/utils/whisper_prepost.py
+++ b/olive/passes/utils/whisper_prepost.py
@@ -248,11 +248,70 @@ def _postprocessing(name: str) -> onnx.ModelProto:
     return fn_decoder.onnx_model
 
 
+# this is copied over from https://github.com/microsoft/onnxruntime-extensions/pull/496 to avoid dependency on
+# unreleased version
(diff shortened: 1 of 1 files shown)


xiaoyu-work:
The original PR from onnxruntime-extension also updated the whisper example. Can you double check if we can simplify our whisper example also?

jambayk:
> The original PR from onnxruntime-extension also updated the whisper example. Can you double check if we can simplify our whisper example also?

The other changes requires the latest code from onnxruntime-extensions. 
i think we should make the other changes when onnxruntime-extensions has a new pypi release with this changes.

xiaoyu-work (approved):

PR #428, opened 2023-07-21
🛺 Auto infer accelerators with EPs

trajepl:
## Describe your changes

From many users' feedback, when they ensure cuda/trt ep is run on gpu certainly, they have to write the accelerators configs to tell current device is [`gpu`]

This PR is used to remedy this. The logic is like:
If Olive can find unique accelerator for each EP provided by users, then use it unless if user set it explicitly.
For examples:

Eps, inferred acceleartors:
(["CPUExecutionProvider"], None),
(["CUDAExecutionProvider"], ["gpu"]),
(["CUDAExecutionProvider", "CPUExecutionProvider"], None), // as we find multi-accelerators(cpu/gpu/npu) for CPU ep.
(["DmlExecutionProvider", "CUDAExecutionProvider"], ["gpu"]),
(["QNNExecutionProvider", "CUDAExecutionProvider"], ["npu", "gpu"]),

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -105,7 +105,8 @@ class Engine:
                 accelerators = ["CPU"]
             else:
                 logger.debug(
-                    f"Use inferred accelerators {inferred_accelerators} from given execution providers {self.execution_providers}."
+                    f"Use inferred accelerators {inferred_accelerators} "
+                    f"from given execution providers {self.execution_providers}."
                 )
                 accelerators = inferred_accelerators
 
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -94,14 +94,20 @@ class Engine:
                 # for docker system and python system, we default use CPUExecutionProvider
                 execution_providers = ["CPUExecutionProvider"]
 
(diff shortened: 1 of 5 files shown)


olive/engine/engine.py
```diff
@@ -120,7 +128,13 @@ def __init__(
                     self.accelerator_specs.append(AcceleratorSpec(device, ep))
                     processed_ep.add(ep)
 
-        assert self.accelerator_specs, "No valid accelerator specified for target system."
+        assert self.accelerator_specs, (
+            "No valid accelerator specified for target system. "
+            "Please specify the accelerators in the target system or provide valid execution providers. "
+            f"Given execution providers: {self.execution_providers}. "
+            f"Current accelerators: {accelerators}."
+            f"Supported execution providers: {AcceleratorLookup.EXECUTION_PROVIDERS}."
```
leqiao-1, line 136: should here only show the providers mapped by the accelerators ?

trajepl: There is a case the accelerators is inferred from EPs. I think all mapping is good here.

leqiao-1 (approved):

PR #429, opened 2023-07-22
Add dolly-v2-3b support to the DML examples

PatriceVignola:


examples/directml/dolly_v2/user_script.py
```diff
@@ -29,10 +30,10 @@ def dolly_v2_inputs(batch_size, torch_dtype):
 
     for layer_index in range(32):
```
trajepl, line 31: Quick question, will the number of past_key_values be changed if user switch to other size of model? Should we also make it adaptable with different model size?

Seems like the num_hidden_layers is 36 for dolly 12b.

PatriceVignola: I can change it when I onboard the 12b model. I like onboarding the models one by one and making the scripts progressively more generic as needed.

trajepl (approved):

PR #430, opened 2023-07-23
Disable input model evaluation for DirectML examples

PatriceVignola:
Input model evaluation doesn't currently work for DirectML because Olive equates `gpu` accelerators to `cuda` pytorch devices, which doesn't work when torch cuda isn't installed (and it won't usually be when people are running DML examples).

This PR is meant to unblock those examples, but in the future we should have a path to fallback to the CPU or use pytorch-directml when using `"evaluate_input_model": true`.


trajepl:
Thanks for this fix. I did not realized that we have pytorch-directml before, will try to complete the fallback logics in the future.

trajepl (approved):

PR #434, opened 2023-07-24
🔍 Make OrtTransformersOptimization searchable

trajepl:
## Describe your changes
1. Make OrtTransformersOptimization searchable: use_gpu, only_onnxruntime 
2. validate search point for OrtTransformersOptimization  based on accelerator_spec

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -892,7 +892,7 @@ class Engine:
         output_model_path = str(output_model_path)
 
         # prune if invalid search_point
-        if not p.validate_search_point(pass_search_point) and not self.no_search:
+        if not p.validate_search_point(pass_search_point, accelerator_spec) and not self.no_search:
             output_model = PRUNED_CONFIG
         else:
             # run pass
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -338,7 +338,7 @@ class Pass(ABC):
             config[key] = value
         return self._config_class(**config).dict()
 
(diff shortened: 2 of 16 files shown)


olive/passes/olive_pass.py
```diff
@@ -340,7 +340,9 @@ def config_at_search_point(self, point: Dict[str, Any]) -> Dict[str, Any]:
             config[key] = value
         return self._config_class(**config).dict()
 
-    def validate_search_point(self, search_point: Dict[str, Any]) -> bool:
+    def validate_search_point(
+        self, search_point: Dict[str, Any], accelerator_spec: AcceleratorSpec, with_fixed_value: bool = False
```
leqiao-1, line 344: `with_fixed_value` should the default value be True here ?

olive/engine/engine.py
```diff
@@ -919,7 +919,7 @@ def _run_pass(
         output_model_path = str(output_model_path)
 
         # prune if invalid search_point
-        if not p.validate_search_point(pass_search_point) and not self.no_search:
+        if not p.validate_search_point(pass_search_point, accelerator_spec, with_fixed_value=True):
```
leqiao-1, line 922: `with_fixed_value` should the default value be False here ?

trajepl: If `False`(legacy way actually), we cannot validate the search point with fixed value. Let us say:
fp16 is a fixed value and user set is as True. Then if `with_fixed_value` is True, we can use the value of fp16 to filter CPU ep out.

leqiao-1 (approved):

PR #435, opened 2023-07-24
Huggingface DataConfig: Add support for `data_files` and `max_samples`

jambayk:
## Describe your changes
This PR exposes the `data_files` argument for load_dataset. Sometimes we only want specific files from the dataset to be used. 

`max_samples` is used during pre-processing to limit the maximum number of samples. Some datasets have a large number of samples which might be too much for evaluation or calibration purposes. 
In the future, we could provide other arguments such as `max_calibration_samples` or `max_calibration_batches` if needed. 

### DIFF
--- a/olive/data/component/load_dataset.py
+++ b/olive/data/component/load_dataset.py
@@ -23,7 +23,7 @@ def simple_dataset(data_dir, input_data, label_cols=None, **kwargs):
 
 
 @Registry.register_dataset()
-def huggingface_dataset(data_dir, data_name=None, subset=None, split="validation", **kwargs):
+def huggingface_dataset(data_dir, data_name=None, subset=None, split="validation", data_files=None, **kwargs):
     """
     This function is used to create a dataset from huggingface datasets
     """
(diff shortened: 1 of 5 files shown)


olive/data/component/load_dataset.py
```diff
@@ -23,7 +23,7 @@ def simple_dataset(input_data, label_cols=None, **kwargs):
 
 
 @Registry.register_dataset()
-def huggingface_dataset(data_name=None, subset=None, split="validation", **kwargs):
+def huggingface_dataset(data_name=None, subset=None, split="validation", data_files=None, **kwargs):
```
trajepl, line 26: I was thinking that huggingface `load_dataset` has many arguments for specific case. Then I add kwargs, expose `data_files` is okay for me. 

But is it possible to get it from kwargs? We can update the documents to tell users that they can set any arguments from `load_dataset`

jambayk: Can you share an example for how to provide kwargs to load_dataset? I don't see it in the documents.

jambayk: Because of how `params_config` is used here https://github.com/microsoft/Olive/blob/main/olive/data/config.py#L98, `kwargs` is empty for all functions. I think it will only be non-empty if the `component.params` is provided specifically https://github.com/microsoft/Olive/blob/main/olive/data/config.py#L23

jambayk: As discussed offline, I have created a work item and will look at simplifying kwargs in a followup PR.

trajepl (approved):

PR #444, opened 2023-07-26
`text-generation` task support and `Perplexity` metric

jambayk:
## Describe your changes
This PR adds support for `text-generation` task in huggingface dataset. Added new pre-processing and post-processing component for the task. The pre-preprocessing component the tokenizes the input text and splits it into chunks using randomization, non-overlapping blocks or sliding window depending on the arguments provided. 

Also adds Perplexity metric under accuracy type. 


olive/data/component/pre_process_data.py
```diff
@@ -122,7 +122,119 @@ def _tokenizer_and_align_labels(examples):
...
+    tokenized_inputs = {
+        "input_ids": [],
+        "target_ids": [],
+    }
+
+    random_seed = kwargs.get("random_seed")
+    if random_seed is None:
+        # no randomization, just use contiguous blocks of tokens
+        max_text = max_samples * seqlen if max_samples is not None else num_text
+        max_text = min(max_text, num_text)
```
trajepl, line 185: Different models have different tokenizers(encoders), like BPE, wordPiece, Unigram and etc. which may generate various sequence lengths.

seq_len means the tokenized sequence length at the most time from I know.

It is better to compare the length at the same level. **Not the word piece vs token**

https://huggingface.co/docs/transformers/tokenizer_summary

olive/data/component/pre_process_data.py
```diff
@@ -122,7 +122,119 @@ def _tokenizer_and_align_labels(examples):
...
+
+        num_tokens = input_ids.shape[1]
+        step = stride or seqlen
+        # loop over the number of tokens
+        # all inputs must be seqlen long
+        for begin_loc in range(0, num_tokens - seqlen, step):
+            # end_loc is the beginning of the next sequence
+            end_loc = begin_loc + seqlen
+            # get the input sequence
+            input_ids = encodings.input_ids[0, begin_loc:end_loc].clone()
```
trajepl, line 197: Why we only get the first sentence here? `input_ids[0, a:b]` should be `input_ids[:, a:b]`?

olive/data/component/pre_process_data.py
```diff
@@ -122,7 +122,119 @@ def _tokenizer_and_align_labels(examples):
...
+        max_text = max_samples * seqlen if max_samples is not None else num_text
+        max_text = min(max_text, num_text)
+        encodings = tokenizer(" ".join(split_text[:max_text]), return_tensors="pt")
+        input_ids = encodings.input_ids
+
+        num_tokens = input_ids.shape[1]
+        step = stride or seqlen
+        # loop over the number of tokens
+        # all inputs must be seqlen long
+        for begin_loc in range(0, num_tokens - seqlen, step):
```
trajepl, line 193: How we ensure `(num_tokens-seqlen) % step == 0` if we do not drop the inputs' tail?

olive/data/component/pre_process_data.py
```diff
@@ -122,7 +122,119 @@ def _tokenizer_and_align_labels(examples):
...
+        "input_ids": [],
+        "target_ids": [],
+    }
+
+    random_seed = kwargs.get("random_seed")
+    if random_seed is None:
+        # no randomization, just use contiguous blocks of tokens
+        max_text = max_samples * seqlen if max_samples is not None else num_text
+        max_text = min(max_text, num_text)
+        encodings = tokenizer(" ".join(split_text[:max_text]), return_tensors="pt")
```
trajepl, line 186: It may hurt PPL value if we drops some inputs, especially for some popular model(gpt2), user may say why the PPL value is different with the public report.

Or maybe you consider this somewhere?

olive/data/component/pre_process_data.py
```diff
@@ -122,7 +122,119 @@ def _tokenizer_and_align_labels(examples):
...
+def text_generation_huggingface_pre_process(
+    _dataset, model_name, input_cols, seqlen, stride=None, max_samples=None, **kwargs
+):
+    """Pre-process data.
+
+    Args:
+        _dataset (object): Data to be pre-processed.
+        model_name (str): Name of the huggingface model.
+        input_cols (list): List of input columns.
+        seqlen (int): Length of the sequence.
```
trajepl, line 141: Better to make these several `numbers` clearer.

olive/data/component/pre_process_data.py
jambayk: We are using `seqlen` at the token level. 
This part is used as a heuristic to speed up tokenization. Most of the time (maybe everytime), the number of tokens is more than the number of words. So, we are taken a chunk of text with the `max_text` length to only tokenize as much as we need. Otherwise, it takes a long time to tokenize all of the text when we are not going to use them all. 

You can see that `seqlen` of the encodings is used to get the actual inputs and targets.

olive/data/component/pre_process_data.py
jambayk: `encodings.input_ids` is has the batch dimension but it is always `1` in this case since we only give one `text`. We don't use `dataset.map` here that takes care of the batch dimension so we are removing the batch dimension. it gets added back by the dataloader based on the batchsize later.

olive/data/component/pre_process_data.py
jambayk: Range automatically drops the leftover that is less than step.

olive/data/component/pre_process_data.py
jambayk: Do you mean the `max_samples` dropping inputs? If users want to use the whole dataset, they can set `max_samples` to `None`. And it will use the entire text.

olive/data/component/pre_process_data.py
jambayk: Do you mean better descriptions for the parameters? I can do that

olive/data/component/pre_process_data.py
trajepl: Also, from my understanding, 

Different sentence may not be relevant with each other. In this [example](https://huggingface.co/docs/transformers/perplexity), they did with the same way to concatenate.
But in some cases, it is more like to calculate PPL per sentence then to get the mean-PPL.
image

I do not know which should be the best way to reproduce some model's benchmark report. 
Could you help validation that with a real model?

olive/data/component/pre_process_data.py
trajepl: Yes, I know. I was considering the following case:
image

olive/data/component/pre_process_data.py
trajepl: Thanks, I was thinking that you handle the sentence one-by-one. But as you mentioned above, there is only one text input for tokenizer.

olive/data/component/pre_process_data.py
jambayk: I was thinking about this case too. I have seen some examples where each sentence is padded or truncated and made into a sample. 
I am thinking of maybe introducing different modes. Something like:
- joint-block (join all sentences. No sliding window)
- joint-sliding-window (join all sentences, sliding window + context)
- joint-random (join all sentences, random start point)
- individual-block (each sentence is a sample. truncated or padded to max length)
- individual-random (each sentence is a sample. randomly choose sentences. truncated or padded to max length) 

I think that would cover most of the ways of measuring perplexity. 

Can I do it in a follow up PR? I need this dataset and ppl support for sparsity work item. and currently don't have time to implement all modes and validate with models.

olive/data/component/pre_process_data.py
jambayk: our range is from 0 to num_tokens-seqlen, so we always guarantee that the final part of the range has enough length. each part is guaranteed to be seqlen so we will also have `stride` context. 
Also, just to avoid any potential confusion, their seq_len = our num_tokens and max_length = seq_len. 

In their implementation, the first `trg_len=`max_length` which doesn't make sense. This means the whole input_ids is context. Looks like a bug.

olive/data/component/pre_process_data.py
jambayk: Sure! I will update these. Thanks!

olive/data/component/pre_process_data.py
trajepl: For sure.

trajepl (approved):

PR #450, opened 2023-07-28
🧆 Set Accelerator agnostic as False for passes need accelerators

trajepl:
## Describe your changes
 Set Accelerator agnostic as False for passes need accelerators

### DIFF
--- a/olive/passes/onnx/optimum_merging.py
+++ b/olive/passes/onnx/optimum_merging.py
@@ -22,6 +22,13 @@ class OptimumMerging(Pass):
     _accepts_composite_model = True
 
     @staticmethod
+    def is_accelerator_agnostic(accelerator_spec: AcceleratorSpec) -> bool:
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/optimum_merging.py
```diff
@@ -21,6 +21,13 @@ class OptimumMerging(Pass):
 
     _accepts_composite_model = True
 
+    @staticmethod
+    def is_accelerator_agnostic(accelerator_spec: AcceleratorSpec) -> bool:
```
guotuofeng, line 25: is this pass accelerator agnostic? I only see the EP when used to generate the model?

trajepl: image
For this case, EP is used to optimize model graph. I suspect that there might be different optimizations wrt EPs.

guotuofeng (approved):

PR #451, opened 2023-07-31
Keep `preds` and `targets` as torch tensor during accuracy evaluation

jambayk:
## Describe your changes
During accuracy evaluation, we convert the output and target tensors to list before computing the metric value. However, this method is expensive and not efficient when the tensors are large since we end up converting them back to torch tensors anyways. 
This is felt significantly when evaluating perplexity for large language models where the `preds` tensor is a large logit tensor of size `batch_size, seqlen, vocab` (like `1 X 2048 X 32000`). This is very slow. 

In this PR, we just move the tensors to `cpu` if not already on cpu and concatenate them using `torch.cat`. This has the same effect as `list.extend`. 

### DIFF
--- a/olive/evaluator/accuracy.py
+++ b/olive/evaluator/accuracy.py
@@ -7,7 +7,6 @@ from abc import abstractmethod
 from inspect import isfunction, signature
 from typing import Any, Callable, Dict, Type, Union
 
-import numpy as np
 import torch
 import torchmetrics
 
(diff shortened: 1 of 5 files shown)


olive/evaluator/accuracy.py
```diff
@@ -62,6 +61,13 @@ def _metric_config_from_torch_metrics(cls):
     def _default_config(cls) -> Dict[str, ConfigParam]:
         return cls._metric_config_from_torch_metrics()
 
+    @staticmethod
+    def prepare_tensors(preds, target, dtypes=torch.int):
+        dtypes = dtypes if isinstance(dtypes, (list, tuple)) else [dtypes, dtypes]
```
trajepl, line 66: how about to add assertion:
assert len(dtypes) == 2

guotuofeng, line 66: could we just use the dtype instead of [dtypes, dtypes]?

jambayk: done!

jambayk: The idea was to support either a len2 tuple/list or a single type. So that the types for preds and tensor can be different if wanted. 
I am using the `[dtypes, dtypes]` part to make the logic simpler (no additional conditions) for the following lines that refer to `dtypes[0]` and `dtypes[1]`.

olive/evaluator/olive_evaluator.py
```diff
@@ -334,8 +334,11 @@ def _inference(
                 # convert to dict of torch tensor
                 result = {name: torch.Tensor(res[i]) for i, name in enumerate(output_names)}
             outputs = post_func(result) if post_func else result
-            preds.extend(outputs.tolist())
-            targets.extend(labels.data.tolist())
+            # keep as numpy or torch arrays
+            preds.append(outputs.cpu())
```
guotuofeng, line 338: the append will make the list as a single item. That's the why the extend is used originally.

jambayk: torch.cat at line 340 and 341 does the same thing as noted in the PR description

trajepl (approved):

PR #454, opened 2023-07-31
Add `component_kwargs` in `DataConfig.params_config`

jambayk:
## Describe your changes
`DataConfig` supports multiple ways to pass parameters to the components. They are:
- `DataConfig.params_config` - the parameters for different components are provided as members of this dictionaries. Only the named parameters of components are updated using this. There is no way to specify component specific kwargs. 
- `DataConfig.components["component_name"].params` - We can provide component specific kwargs using this but it is hard to access. Moreover, there is not way to provide this for hf datatemplate or `HFConfig.datasets` for input model. 

This PR introduces a reserved member in `DataConfig.params_config` called `component_kwargs` where we can provide component specific kwargs. An example usage is:
```
{
    "name": "c4_train",
    "params_config": {
        "model_name": "gpt2",
        "task": "text-generation",
        "data_name": "allenai/c4",
        "subset": "allenai--c4",
        "split": "train",
        "data_files": {"train": "en/c4-train.00000-of-01024.json.gz"},
        "input_cols": ["text"],
        "component_kwargs": {
            "pre_process_data": {
                "max_samples": 128,
                "random_seed": 42,
                "seqlen": 2048,
            }
        }
    }
  }
```
This has automatic support for all templates and `HFConfig.dataset`. 

### DIFF
--- a/test/unit_test/workflows/mock_data/text_generation_dataset_random.json
+++ b/test/unit_test/workflows/mock_data/text_generation_dataset_random.json
@@ -1,0 +1,62 @@
+{
+    "input_model":{
+        "type": "PyTorchModel",
+        "config": {
+            "hf_config": {
+                "model_name": "gpt2",
+                "task": "text-generation",
+                "dataset": {
+                    "data_name": "ptb_text_only",
+                    "subset": "penn_treebank",
+                    "split": "train",
+                    "input_cols": ["sentence"],
+                    "component_kwargs": {
+                        "pre_process_data": {
+                            "seqlen": 1024,
+                            "max_samples": 5,
+                            "random_seed": 42
+                        }
+                    }
+                }
+            }
+        }
+    },
+    "evaluators": {
+        "common_evaluator": {
+            "metrics":[
+                {
+                    "name": "perplexity",
+                    "type": "accuracy",
+                    "sub_types": [
+                        {"name": "perplexity"}
+                    ]
+                }
+            ]
+        }
+    },
+    "passes": {
+        "conversion": {
+            "type": "OnnxConversion"
+        },
+        "quantization": {
+            "type": "OnnxQuantization"
+        },
+        "perf_tuning": {
+            "type": "OrtPerfTuning"
+        }
+    },
+    "engine": {
+        "log_severity_level": 0,
+        "search_strategy": {
+            "execution_order": "joint",
+            "search_algorithm": "tpe",
+            "search_algorithm_config": {
+                "num_samples": 3,
+                "seed": 0
+            }
+        },
+        "evaluator": "common_evaluator",
+        "clean_cache": true,
+        "cache_dir": "cache"
+    }
+}
--- a/olive/data/component/pre_process_data.py
+++ b/olive/data/component/pre_process_data.py
@@ -157,6 +157,8 @@ def text_generation_huggingface_pre_process(
     from datasets import Dataset as HFDataset
     from transformers import AutoTokenizer
 
+    assert seqlen is not None, "Must specify seqlen"
+
     # get tokenizer
     tokenizer = AutoTokenizer.from_pretrained(model_name)
 
--- a/olive/data/config.py
+++ b/olive/data/config.py
@@ -98,23 +98,30 @@ class DataConfig(ConfigBase):
     def fill_in_params(self):
         """
         Fill in the default parameters for each component.
-        1. if prams_config is not None, use the params_config to fill in the params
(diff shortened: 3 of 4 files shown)


test/unit_test/workflows/mock_data/text_generation_dataset_random.json
```diff
@@ -0,0 +1,62 @@
...
+        "config": {
+            "hf_config": {
+                "model_name": "gpt2",
+                "task": "text-generation",
+                "dataset": {
+                    "data_name": "ptb_text_only",
+                    "subset": "penn_treebank",
+                    "split": "train",
+                    "input_cols": ["sentence"],
+                    "component_kwargs": {
```
guotuofeng, line 13: in this example, I don't see the params_config is demonstrated in the PR description. How does it look like if both params_config and component_kwargs exist at the same time?

jambayk: This is equivalent to
```
{
    "name": "ptb",
    "params_config": {
        "model_name": "gpt2",
        "task": "text-generation",
        "data_name": "ptb_text_only",
        "subset": "penn_treebank",
        "split": "train"
        "input_cols": ["sentence"],
        "component_kwargs": {
            "pre_process_data": {
                "seqlen": 1024,
                 "max_samples": 5,
                 random_seed": 42
            }
        }
    }
  }
```
Similar to the one shown in the PR description. olive.workflows.run automatically converts hf_config.dataset to this format.

test/unit_test/workflows/mock_data/text_generation_dataset_random.json
```diff
@@ -0,0 +1,62 @@
+{
+    "input_model":{
+        "type": "PyTorchModel",
+        "config": {
+            "hf_config": {
+                "model_name": "gpt2",
+                "task": "text-generation",
+                "dataset": {
+                    "data_name": "ptb_text_only",
```
trajepl, line 9: There is the disk limitation in our pipeline. As this case will download new data which may bring the risk to trigger `out of space` error.

How about use bert model + glue dataset for test?

jambayk: There is another workflow which uses the same dataset https://github.com/microsoft/Olive/blob/main/test/unit_test/workflows/mock_data/text_generation_dataset.json. This is the same dataset with just some new args. 

So I think this might be okay? Also, I think PTB dataset is not too big.

test/unit_test/workflows/mock_data/text_generation_dataset_random.json
jambayk: It would be something like 
```
"raw_data": {
            "name": "raw_data",
            "type": "RawDataContainer",
            "components": {
                "post_process_data": {
                    "type": "inception_post_process"
                }
            },
            "params_config": {
                "data_dir": "data",
                "component_kwargs: {
                     "load_data": {
                        "input_names": ["input"],
                         "input_shapes": [[1, 299, 299, 3]],
                         "input_order_file": "input_order.txt",
                         "annotations_file": "labels.npy",
                      }
                 },
               "batch_size": 7
        }
```
But this is not necessary since all of the `load_data` args are name arguments in https://github.com/microsoft/Olive/blob/main/olive/data/component/load_dataset.py#L46. So they can be added directly to `params_config`. 
`component_args` is most useful when we want to provide values to `kwargs` in a function like https://github.com/microsoft/Olive/blob/main/olive/data/component/load_dataset.py#L26C101-L26C109.

guotuofeng: If we'd like to pass parameters to components.post_process_data in above example, need we define two sections? one for components and another for component_kwargs? If yes, it would be better if we could put all the definitions including  post_process_data and the params in consolidated place instead of spreading them.

jambayk: We can pass the parameters directly to the components.post_process_data.params. It is equivalent to passing it to component_kwargs. So two sections are not needed for the same component `post_process`. 
`component_kwargs` is just an additional way of providing kwargs without creating a `components` dict which is not possible for pytorch_model.hf_config.

guotuofeng (approved):

test/unit_test/workflows/mock_data/text_generation_dataset_random.json
jambayk: Resolved offline.

trajepl (approved):

PR #456, opened 2023-08-01
🎏 Multi passes run in same olive workflow

trajepl:
## Describe your changes
Provide the interface to let use set the multi pass flows to run in save olive workflow. 

The search strategy will traverse multi-passes sequentially:
for example, for the flows of `[[pass_a, pass_b, pass_c], [pass_a, pass_c]]`
1. `joint`: 
    - `[pass_a, pass_b, pass_c]`
    - `[pass_a, pass_c]`
3. `pass-by-pass`: 
    - `[[pass_a], [pass_b], [pass_c]]`
        - `[pass_a]`
        - `[pass_b]`
        - `[pass_c]`
    -  `[[pass_a], [pass_c]]`
        - `[pass_a]`
        - `[pass_c]`

Provide the interface to let use set the multi pass flows to run in save olive workflow. 
```json
"passes": {
    "onnx_conversion": {
        "type": "OnnxConversion",
        "config": {
            "target_opset": 13
        }
    },
        "config": {
            "model_type": "bert",
            "num_heads": 12,
            "hidden_size": 768,
            "float16": true
        }
    },
    "onnx_quantization": {
        "type": "OnnxQuantization",
        "config": {
            "data_dir": "data",
            "dataloader_func": "resnet_calibration_reader",
            "weight_type": "QUInt8"
        }
    }
},
"pass_flows": [
    ["onnx_conversion", "transformers_optimization"],
    ["onnx_conversion", "transformers_optimization", "onnx_quantization"],
    ["onnx_conversion", "onnx_quantization"],
]
```

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -154,6 +154,9 @@ class Engine:
         # {"pass_name": {"pass": pass, "host": host, "evaluator": evaluator, "clean_run_cache": clean_run_cache}}
         self.passes = OrderedDict()
 
+        self.pass_flows = None
+        self.pass_flows_search_spaces = None
+
         self.footprints = defaultdict(Footprint)
 
         self.azureml_client_config = self._config.azureml_client_config
@@ -194,6 +197,7 @@ class Engine:
             if clean_run_cache:
                 cache_utils.clean_pass_run_cache(pass_config["type"].__name__, cache_dir)
 
+        self.set_pass_flows(self.pass_flows)
         self._initialized = True
 
     def register(
@@ -266,6 +270,18 @@ class Engine:
             "evaluator": evaluator_config,
             "output_name": output_name,
         }
+
+    def set_pass_flows(self, pass_flows: List[List[str]]):
+        """
+        Construct pass flows from a list of pass names.
+        Args:
+            pass_flows: a list of pass names, each pass name is a string.
+        """
+        if not pass_flows:
+            assert self.pass_config, "Pass flows can only be set after passes are registered"
+            self.pass_flows = [list(self.pass_config.keys())]
+        else:
+            self.pass_flows = pass_flows
 
     def run(
         self,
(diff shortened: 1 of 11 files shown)


olive/engine/engine.py
```diff
@@ -420,7 +440,7 @@ def run_no_search(
             )
 
         # initialize the search strategy
-        self.search_strategy.initialize(self.pass_search_spaces, input_model_id, objective_dict)
+        self.search_strategy.initialize(self.pass_flows_search_spaces, input_model_id, objective_dict)
```
jambayk, line 443: So, for the no search case, it only runs the first flow? We only call `next_step` once, so it would return the first flow. 

Actually, it looks like there might be a bug in the `run_no_search` mode from the beginning. `next_step` with `pass-by-pass` execution order would only return the first pass. It only behaved as expected when the execution order is `joint`.

olive/strategy/search_strategy.py
```diff
@@ -122,7 +135,24 @@ def _next_search_group(self, init_model_id: Optional[str] = None) -> Optional[Se
         """
         Get the next search space group and initialize the search algorithm.
         """
-        # TODO: organize the state better and make execution order more flexible using a graph
+        if not (self._spaces_groups or self._pass_by_pass_sg):
```
jambayk, line 138: Could you add some comments to explain the different if cases in the logic? 

Please correct me if i understood any of the cases wrong:
138 : joint - all flows are exhausted. pass-pass-pass - all flows exhausted, current flow exhausted 
141: No input_model_id is provided for the first call to `_next_search_group`.

151: first call, initialize by getting first flow and original input model 
156. Get best model from the previous group/pass 
...

olive/engine/engine.py
jambayk: The potential bug is not because of this PR so it doesn't block it. 

But I am interested in whether only one flow or all flows are run in no search mode.

trajepl: Good catch. It is expected to be all flows.
I will fix this part. Thanks.

trajepl: Good catch. It is expected to be all flows.
I will fix this part. Thanks.

trajepl: Updated.

olive/engine/engine.py
```diff
@@ -420,68 +446,86 @@ def run_no_search(
...
+                _,
+                signal,
+                model_ids,
+            ) = self._run_passes(next_step["passes"], model, model_id, data_root, accelerator_spec)
+            # names of the output models of the passes
+            pass_output_names = [self.passes[pass_name]["output_name"] for pass_name, _ in next_step["passes"]]
+            pass_output_names = [f"{name}_{accelerator_spec}" if name else None for name in pass_output_names]
+
+            pass_flow = self.pass_flows[iter_num - 1]
+            result_prefix = "-".join(pass_flow)
```
jambayk, line 486: maybe we can use the pass_flow index (something like `flow-0`) instead of joining the whole flow? The name might be too long. 

Also, the file name in the whisper example https://github.com/microsoft/Olive/blob/main/examples/whisper/test_transcription.py#L70

trajepl: It may be not intuitive for user to find the corresponding passes if we use index. I switch it to a directory like:
output_dir / **pass_flow_joint_str** / filename.

Is that okay for you?

jambayk: This looks good to me. Thanks!

jambayk (approved):

PR #457, opened 2023-08-02
🐞 Bugfix: Fail to packaging olive output when save mode with external…

trajepl:
## Describe your changes
When the olive output model is saved with external data, our packaging logic will try to read the file which already be moved to other directories. This led to the failure when packaging.

### DIFF
--- a/olive/engine/packaging/packaging_generator.py
+++ b/olive/engine/packaging/packaging_generator.py
@@ -89,9 +89,13 @@ def _package_candidate_models(
                     model_config = pf_footprint.get_model_config(model_id)
                     onnx_file_name = model_config.get("onnx_file_name")
                     onnx_model = ONNXModel(temp_resource_path, onnx_file_name)
+                    model_name = Path(onnx_model.model_path).name
                     for file in Path(temp_resource_path.get_path()).iterdir():
-                        if file.name == Path(onnx_model.model_path).name:
-                            file_name = "model.onnx"
+                        if file.name.startswith(model_name):
+                            if file.name.endswith(".onnx"):
+                                file_name = "model.onnx"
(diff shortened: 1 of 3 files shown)


olive/engine/packaging/packaging_generator.py
```diff
@@ -89,9 +89,13 @@ def _package_candidate_models(
...
                     onnx_file_name = model_config.get("onnx_file_name")
                     onnx_model = ONNXModel(temp_resource_path, onnx_file_name)
+                    model_name = Path(onnx_model.model_path).name
                     for file in Path(temp_resource_path.get_path()).iterdir():
-                        if file.name == Path(onnx_model.model_path).name:
-                            file_name = "model.onnx"
+                        if file.name.startswith(model_name):
+                            if file.name.endswith(".onnx"):
+                                file_name = "model.onnx"
+                            elif file.name.endswith(".onnx.data"):
```
jambayk, line 97: Do you have an example of the failure? 
I think the previous logic should have worked since we are only renaming the `.onnx` file to `model.onnx` but keeping the original name for the other files. 
onnx external data doesn't work if the file name is changed. This logic assumes that the external data is `model.onnx.data`. Which is true most of the time but is not always true if user provided their own name during conversion.

trajepl: image
I see, looks like *.onnx.data will be recorded in model's proto info.

The bug is caused by line 93 where we will rename(move) the original model. When it comes two 2rd iteration on `x.onnx.data`, the `onnx_model.model_path` will fail.
image
image

trajepl: I just updated with temp model path, please help take a look.

PR #460, opened 2023-08-03
🦞 Expose Logits from inference for AUROC(=AUC) measurement

trajepl:
## Describe your changes
Current AUC computing is wrong which need the probabilities or logits for different classes.

This PR is used to expose Logits from inference for AUROC(=AUC) measurement


olive/evaluator/accuracy.py
```diff
@@ -114,28 +114,28 @@ def measure(self, preds, target):
...
 
-class AUC(AccuracyBase):
-    name: str = "auc"
+class AUROC(AccuracyBase):
+    name: str = "auroc"
 
-    def measure(self, preds, target):
-        preds_tensor, target_tensor = self.prepare_tensors(preds, target)
-        preds_tensor = preds_tensor.flatten()
+    def measure(self, preds, target, logits=None):
```
guotuofeng, line 127: preds are useless?

olive/evaluator/olive_evaluator.py
```diff
@@ -337,9 +338,11 @@ def _inference(
             # keep as numpy or torch arrays
             preds.append(outputs.cpu())
             targets.append(labels.cpu())
+            logits.append(result.cpu())
         preds = torch.cat(preds, dim=0)
         targets = torch.cat(targets, dim=0)
-        return preds, targets
+        logits = torch.cat(logits, dim=0)
+        return preds, targets, logits
```
guotuofeng, line 345: do we need define a class to contains all of them? is it possible that we need add other values in future?

olive/evaluator/olive_evaluator.py
```diff
@@ -614,25 +620,30 @@ def _inference(
...
         targets = []
+        logits = []
         device = PyTorchEvaluator._device_string_to_torch_device(device)
         if device:
             session.to(device)
         for input_data, labels in dataloader:
             input_data = tensor_data_to_device(input_data, device)
             result = session(**input_data) if isinstance(input_data, dict) else session(input_data)
+            if not isinstance(result, torch.Tensor) and getattr(result, "logits", None) is not None:
+                result = result.logits
```
jambayk, line 631: a lot of our post funcs are already implement assuming the original state of `result`. I think it might be better to save the logit separately and not overwrite the original result.

olive/evaluator/olive_evaluator.py
```diff
@@ -337,9 +338,11 @@ def _inference(
             # keep as numpy or torch arrays
             preds.append(outputs.cpu())
             targets.append(labels.cpu())
+            logits.append(result.cpu())
```
jambayk, line 341: result is not a torch tensor if the `len(output_names) >1`. It is going to be a dictionary of tensors.

olive/evaluator/olive_evaluator.py
jambayk: This condition can be used at line 637 directly.

olive/evaluator/olive_evaluator.py
guotuofeng: seconds this comment

olive/evaluator/olive_evaluator.py
guotuofeng: something like https://huggingface.co/docs/transformers/main/main_classes/output#transformers.utils.ModelOutput

guotuofeng: is this line impacted? https://github.com/microsoft/Olive/pull/460/files#diff-816af95cf16e31610395bb20947f1d73cc1ca67355964e074a8f55375c8139a7R123

guotuofeng: ditto https://github.com/microsoft/Olive/blob/main/test/unit_test/assets/user_script.py#L8C5-L8C16

olive/evaluator/accuracy.py
trajepl: Updated with tuple(preds, logits)

olive/evaluator/olive_evaluator.py
trajepl: Updated with tuple(preds, logits)

olive/evaluator/olive_evaluator.py

olive/evaluator/accuracy.py
```diff
@@ -77,15 +77,15 @@ def prepare_tensors(preds, target, dtypes=torch.int):
...
         raise NotImplementedError()
 
 
 class AccuracyScore(AccuracyBase):
     name: str = "accuracy_score"
 
-    def measure(self, preds, target):
-        preds_tensor, target_tensor = self.prepare_tensors(preds, target)
+    def measure(self, model_output, target):
+        preds_tensor, target_tensor = self.prepare_tensors(model_output[0], target)
```
jambayk, line 88: I think it is clearer if we access it using `model_outputs.preds`

trajepl: The main concern is about compatibility on tuple and named tuple. Once user want use Olive's evaluator, there is no need to pack the (preds, logits) with pre-defined name.

trajepl: But it may only appear on the case user rewrite the inference function which is not supported in Olive now. 

Updated with named tuple.

guotuofeng (approved):

PR #461, opened 2023-08-03
Automatic onnx conversion for opt, llama, gptneox. Infer `feature` from `task`

jambayk:
## Describe your changes
Olive uses `transformer.onnx.FeaturesManager` to automatically get onnx conversion args. However, huggingface has started using `optimum` for this instead and haven't added support for newer models in `transformers.onnx`. 

In this PR, we add support for `opt`, `llama` and `gptneox` (model type for dolly) by extending `FeaturesManager`. All of these models are text decoder model so they are using the same onnx config. 

Also, infer `feature` from `task` if not provided. We previously used `default` but that doesn't work as expected if the `task` is different from the default. Such as `sequence-classification` for gpt2 whose default is `causal-lm`. 

### DIFF
--- a/docs/source/overview/options.md
+++ b/docs/source/overview/options.md
@@ -101,7 +101,8 @@ case insensitive.
         - `task: [str]`: This is the task type for the model such as `text-classification`. The complete list of supported task can be found
         at [huggingface-tasks](https://huggingface.co/docs/transformers/v4.28.1/en/main_classes/pipelines#transformers.pipeline.task).
 
-        - `feature: [str]`: The ONNX export features. This is only needed for HuggingFace hub model. Default to `default`. You can find more info at [Export to ONNX](https://huggingface.co/docs/transformers/serialization)
+        - `feature: [str]`: The ONNX export features. This is only needed for HuggingFace hub model. It is inferred from `task` if not provided. You must provide the feature if you need past. For instance,
+        `"causal-lm-with-past"`. You can find more info at [Export to ONNX](https://huggingface.co/docs/transformers/serialization)
 
         - `model_class: [str]`: Instead of the `task`, the class of the model can be provided as well. Such as `DistilBertForSequenceClassification`
 
--- a/docs/source/overview/options.md
+++ b/docs/source/overview/options.md
@@ -101,8 +101,8 @@ case insensitive.
         - `task: [str]`: This is the task type for the model such as `text-classification`. The complete list of supported task can be found
         at [huggingface-tasks](https://huggingface.co/docs/transformers/v4.28.1/en/main_classes/pipelines#transformers.pipeline.task).
 
-        - `feature: [str]`: The ONNX export features. This is only needed for HuggingFace hub model. It is inferred from `task` if not provided. You must provide the feature if you need past. For instance,
-        `"causal-lm-with-past"`. You can find more info at [Export to ONNX](https://huggingface.co/docs/transformers/serialization)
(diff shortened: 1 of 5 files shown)


olive/model/hf_onnx_config.py
```diff
@@ -0,0 +1,107 @@
+from typing import Any, Mapping, Optional, OrderedDict
+
+from transformers import PreTrainedTokenizer, TensorType
+from transformers.onnx import OnnxConfigWithPast
+
+ADDITIONAL_MODEL_TYPES = {
```
trajepl, line 6: Could  you help add a few comments to tell what the element is ?
Like "model“: feature list, onnx config class?

trajepl: Also quick question: the `feature` looks like the tasks which current pretrained model can support.

olive/model/hf_onnx_config.py
```diff
@@ -0,0 +1,107 @@
...
+                If `True`, store values on dynamic axis 1, else on axis 2.
+
+        """
+        if direction not in ["inputs", "outputs"]:
+            raise ValueError(f'direction must either be "inputs" or "outputs", but {direction} was given')
+
+        name = "past_key_values" if direction == "inputs" else "present"
+        sequence_length_name = "past_sequence" if direction == "inputs" else "past_sequence + sequence"
+        for i in range(self.num_layers):
+            inputs_or_outputs[f"{name}.{i}.key"] = {0: "batch", 2: sequence_length_name}
```
trajepl, line 66: IIRC, kv cache may has following shapes:
1. (batch size, seq_length, hidden_size)
2. or (batch size, num_attention_head ,seq_length, hidden_size/num_attention_head ) (used in dolly/llama

Quick question if shape 1 is the case of inverted_values_shape == True?

olive/model/hf_onnx_config.py
jambayk: Yes, I can add some comments. 
That's right, the features are the tasks supported by the model. But transformers.onnx calls them differently. optimum doesn't anymore. That's also why I created the map from tasks to feature so the user doesn't need to provide it themselves unless needed.

olive/model/hf_onnx_config.py
jambayk: Yes, I think `inverted_values_shape==True` is for case 1. However, the additional models being added are for case 2 so I haven't see case 1 so far. 
The parent class `OnnxConfigWithPast` works for case 2 
https://github.com/huggingface/transformers/blob/main/src/transformers/onnx/config.py#L530. In the future, we might have to create or inherit a new class for case 1 if needed.

docs/source/overview/options.md
```diff
@@ -101,7 +101,8 @@ case insensitive.
         - `task: [str]`: This is the task type for the model such as `text-classification`. The complete list of supported task can be found
         at [huggingface-tasks](https://huggingface.co/docs/transformers/v4.28.1/en/main_classes/pipelines#transformers.pipeline.task).
 
-        - `feature: [str]`: The ONNX export features. This is only needed for HuggingFace hub model. Default to `default`. You can find more info at [Export to ONNX](https://huggingface.co/docs/transformers/serialization)
+        - `feature: [str]`: The ONNX export features. This is only needed for HuggingFace hub model. It is inferred from `task` if not provided. You must provide the feature if you need past. For instance,
```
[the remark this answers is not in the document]
jambayk: done!

olive/model/hf_onnx_config.py
jambayk: done!

trajepl (approved):

PR #464, opened 2023-08-07
Add SDXL DirectML example

PatriceVignola:


examples/directml/stable_diffusion_xl/stable_diffusion_xl.py
```diff
@@ -0,0 +1,369 @@
...
+        if weights_src_path.is_file():
+            weights_dst_path = dst_path.parent / (dst_path.name + ".data")
+            shutil.copyfile(weights_src_path, weights_dst_path)
+
+    print(f"The optimized pipeline is located here: {optimized_model_dir}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--model_id", default="stabilityai/stable-diffusion-xl-base-1.0", type=str)
```
jstoecker, line 295: Seems like most of this script is identical to the SD 1.x/2.x example. I'm not as familiar with the XL version, but would it be easy to simply extend the other sample based on the model_id? It looks like you might need to change the static dim names and perhaps some of the JSON. I could have missed something. Do you anticipate it being easier to have these scripts separated?

PatriceVignola: I anticipate having them separated will make it easier. For once, they use different pipelines and the types that the pipelines take as arguments are different.

As I enable inference for the refiner model, this will make the logic even more complex by nature (refiner works by first running the base XL model, taking the result and then running refiner on top of it), as well as having user scripts that are going to diverge.

I think the refiner model should also eventually move to the XL example, which should simplify the logic of the vanilla SD example.

jstoecker (dismissed):

jstoecker (approved):

PR #466, opened 2023-08-08
Add PrePostProcessor pipeline support

guotuofeng:
## Describe your changes
Add pre/post pipeline support by leveraging the onnxruntime-extension steps.


docs/source/features/passes/onnx.md
```diff
@@ -79,6 +79,103 @@ Please refer to [OrtTransformersOptimization](ort_transformers_optimization) for
...
+You can refer to [here](https://github.com/microsoft/onnxruntime-extensions/blob/main/onnxruntime_extensions/tools/Example%20usage%20of%20the%20PrePostProcessor.md) to see how to leverage `PrePostProcessor` to customize pre and post processing ops.
+
+Here are some examples to describe the pre/post processing which is exactly same with [superresolution](https://github.com/microsoft/onnxruntime-extensions/blob/main/onnxruntime_extensions/tools/add_pre_post_processing_to_model.py#L89)
+
+```json
+{
+    "pre": [
+        {"ConvertImageToBGR": {}},
+        {
+            "Resize": {
```
trajepl, line 92: Is there a central doc/list where we can find the pre/post ops list from onnxruntime-extensions? The examples are helpful, but how we guide user to complete the pre/post configs?

docs/source/features/passes/onnx.md
```diff
@@ -79,6 +79,103 @@ Please refer to [OrtTransformersOptimization](ort_transformers_optimization) for
...
+            "name": "image",
+            "data_type": "uint8",
+            "shape": ["num_bytes"],
+        }
+    ],
+    "target_opset": 16,
+}
+```
+
+* Olive introduce two placeholders to represent the model input/ouput shape dimension value: `__model_input__` and `__model_output__`.
```
trajepl, line 163: `introduces`.

Also, could you help put this line before the example? I was confusing what is the meaning of `__model_input__` when I read the example before this line.

docs/source/features/passes/onnx.md
```diff
@@ -79,6 +79,103 @@ Please refer to [OrtTransformersOptimization](ort_transformers_optimization) for
...
+        },
+        "io_map": [
+            ["Y1_uint8", 0, 0],
+            ["Cb1_uint8", 0, 1],
+            ["Cr1_uint8", 0, 2],
+        ],
+    }
+```
+* The `tool_command_args` will be used to describe the input parameters to create the `PrePostProcessor` instance. It is list of `PrePostProcessorInput`.
+  The `name` is the tensor name. The `data_type` and `shape` will be used to create the tensor type. The `shape` can be a list of integers or a list of string.
```
trajepl, line 178: What is the difference for list[int] vs list[str] for `shape`?

olive/passes/onnx/pipeline/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    "complex64": onnx.TensorProto.COMPLEX64,
+    "complex128": onnx.TensorProto.COMPLEX128,
+    "bfloat16": onnx.TensorProto.BFLOAT16,
+}
+
+
+def resolve_placeholder(model: onnx.ModelProto, param_value: dict):
+    resolved_param = param_value
+    arg_type = param_value.get("type")
+    if arg_type == "__model_input__":
```
trajepl, line 30: The code looks a bit duplicated.

olive/passes/onnx/pipeline/step_utils.py
```diff
@@ -0,0 +1,231 @@
...
+                },
+                "io_map": [["ConvertImageToBGR",0,0]]
+            }
+        }
+        ```
+    ## Config schema:
+
+    1. simple form:
+         ```
+         { }"step_name": {"param_name": param_value} }
```
trajepl, line 64: { *}*"step_name": {"param_name": param_value} }
Remove "}"

olive/passes/onnx/pipeline/step_utils.py
```diff
@@ -0,0 +1,231 @@
...
+                "tokenizer_param": {
+                    "type": "TokenizerParam",
+                    "params": {
+                        "vocab_or_file": "test_data/sentencepiece/test_model.model"
+                    }
+                }
+            }
+        }
+        ```
+       In this example, the type will be looked in the imported modules. So, please remember to import it before
```
trajepl, line 95: As Olive will be run in CLI model. Where to import the corresponding modules by users? in user script?

olive/passes/onnx/pipeline/step_utils.py
```diff
@@ -0,0 +1,231 @@
...
+                "dim_index": -2
+            }
+            ```
+        When these config is parsed, they will be converted to using the ModelProto shape information.
+
+    5. For list value, they will be treated to tuple in the function. User could explicitly specify the type as list if
+       the default tuple behavior is not expected. Instead of using `"param_name": param_value`, user could use the
+       following form by using `type` and `value`:
+            ```
+            "explicit_list": {
```
trajepl, line 124: Where would the `explicit_list` be used? Ops' arguments?

docs/source/features/passes/onnx.md
```diff
@@ -79,6 +79,103 @@ Please refer to [OrtTransformersOptimization](ort_transformers_optimization) for
 }
 ```
 
+`AppendPrePostProcessingOps` also support customize pre and post processing ops by leveraging the [onnxruntime-extension steps](https://github.com/microsoft/onnxruntime-extensions/tree/main/onnxruntime_extensions/tools/pre_post_processing/steps) and `PrePostProcessor`.
```
trajepl, line 82: `supports`
`customized` or `to customize`

docs/source/features/passes/onnx.md
guotuofeng: I can only find the full list by onnxruntime extension repo. And  there is no any document for the full list.

docs/source/features/passes/onnx.md
guotuofeng: will do

olive/passes/onnx/pipeline/__init__.py
guotuofeng: the parameter name is different. one is input_index and another is output_index

olive/passes/onnx/pipeline/step_utils.py
guotuofeng: good catch

olive/passes/onnx/pipeline/step_utils.py
guotuofeng: From the time being, Olive doesn't support user customize steps in user_script.py. We can consider add that support in future in case of user has the requirement.

The steps are imported in step_utils.py

olive/passes/onnx/pipeline/step_utils.py
guotuofeng: It is just for demo purpose. If user really want list instead of tuple, he/she could write parameter like this. In the current implementation, the list will be converted to tuple by default.

docs/source/features/passes/onnx.md
guotuofeng: it is used by PrePostProcessor  when creating the inputs. Here is how the shape is used. https://github.com/onnx/onnx/blob/main/onnx/helper.py#L1022-L1025

jambayk:
PR fix for the hanging test has been merged. Please push an empty commit, rebase or merge main to run ci with the fix. Thanks!

fs-eire:
I have a suggestion: can we name those steps in a way for users being able to immediately figure out whether it requires extension or not?

for example, currently we have steps like:
`ConvertImageToBGR`, `PixelsToYCbCr`, `CenterCrop`.

and we know that `ConvertImageToBGR` requires extension and the other 2 do not. so maybe rename them to:
`ext:ConvertImageToBGR`, `PixelsToYCbCr`, `CenterCrop`.

fs-eire:
would be better to introduce a [JSON schema](https://json-schema.org/) definition for the config file. may take some effort, but can greatly help users to write correct config.

guotuofeng:
> I have a suggestion: can we name those steps in a way for users being able to immediately figure out whether it requires extension or not?
> 
> for example, currently we have steps like: `ConvertImageToBGR`, `PixelsToYCbCr`, `CenterCrop`.
> 
> and we know that `ConvertImageToBGR` requires extension and the other 2 do not. so maybe rename them to: `ext:ConvertImageToBGR`, `PixelsToYCbCr`, `CenterCrop`.

I'd like not let user know whether the OP is extension required or not. Otherwise, he/she needs to figure out the implementation details before giving a name like ext:ConvertImageToBGR

guotuofeng:
> would be better to introduce a [JSON schema](https://json-schema.org/) definition for the config file. may take some effort, but can greatly help users to write correct config.

The schema could be found at 
https://github.com/microsoft/Olive/pull/466/files#diff-a9b0377d357b274360479840b53ee7a3be1bbb393fe32957748467dcb405fce4
https://github.com/microsoft/Olive/pull/466/files#diff-d79b6949a41a986d68cbb4f82f8b28d0d54eab11c33bf88cb57db59831660093

guotuofeng:
> > > I have a suggestion: can we name those steps in a way for users being able to immediately figure out whether it requires extension or not?
> > > for example, currently we have steps like: `ConvertImageToBGR`, `PixelsToYCbCr`, `CenterCrop`.
> > > and we know that `ConvertImageToBGR` requires extension and the other 2 do not. so maybe rename them to: `ext:ConvertImageToBGR`, `PixelsToYCbCr`, `CenterCrop`.
> > 
> > 
> > I'd like not let user know whether the OP is extension required or not. Otherwise, he/she needs to figure out the implementation details before giving a name like ext:ConvertImageToBGR
> 
> I see. Let me change my question: is there a way that I can know whether my config can generate a model supported in a specific environment (for example, web)?
> 
> because web is built with flags to exclude some extension operator (for example, DecodeImage/EncodeImage), is there a way that users can write a config for web without knowing those implementation details? If a checker can get a list of required operators for the whole prepostprocess pipeline it would be great. we can offer a list of all supported ext operators on web, so the checker can lookup the list and let the user know if the config is compatible with web.
> 
> What may happen now: users write a config, generate a model using that config, and fail to run that model on web. they need to troubleshoot: what went wrong?
> 
> What I expect to happen: users write a config, generate a model using that config, and Olive warn users that:
> 
> > this model is not compatible with web. if you want to use it with web, you cannot use 'ConvertImageToBGR'.

From the time being, Olive doesn't have the knowledge to tell user whether the model generated by the config could be run at web(including the extension OP or not). We have a plan to let user replace all these extension OPs with the built-in OP if these OP could be replaced seamlessly(he feature is not implemented yet). Otherwise, user need load the dll/so to allow the output onnx model could be run successfully. From another point of view, user who write the config should have the knowledge to choose which ops in order to allow the output model could be run successfully.

trajepl:
As a workaround, if it is only the case on ort-web, could we mentioned that in our docs? Like:
for some ops like 'ConvertImageToBGR' which require other extensions may be incompatible with ort-web, user need to exclude this kind of ops to generate proper models.

Also we can update ops list in the docs incrementally.

guotuofeng:
> As a workaround, if it is only the case on ort-web, could we mentioned that in our docs? Like: for some ops like 'ConvertImageToBGR' which require other extensions may be incompatible with ort-web, user need to exclude this kind of ops to generate proper models.
> 
> Also we can update ops list in the docs incrementally in the future.

That's a good point. @fs-eire, do you have any comments that we could add in the doc?

olive/passes/onnx/append_pre_post_processing_ops.py
```diff
@@ -4,18 +4,42 @@
...
+from olive.passes.onnx.pipeline import TENSOR_TYPE_MAP
 from olive.passes.pass_config import PassConfigParam
 
 
+class PrePostProcessorInput(ConfigBase):
+    name: str = Field(..., description="Name of the input.")
+    data_type: str = Field(..., description="Data type of the input.")
+    shape: List[Union[str, int]] = Field(..., description="Shape of the input.")
+
+    @validator("data_type", pre=True)
```
trajepl, line 27: Nitpick: I saw you have a PR to upgrade pydantic to >2.0.0. Please remember to update this.

trajepl (dismissed):

guotuofeng:
> I have a question regarding the pre/post steps:
> 
> who, or which repo defines the expected behavior (ie. What operator(s) to be inserted into the model and how to insert and how to set the attributes, etc.), if the operator(s) does not depends on ort-extension? Is it Olive, or ort-extension?
> 
> I would say that the repo who defines ( or, owns ) the pre-post- processing steps should document the information.
> 
> EDIT:
> 
> I checked the Olive repo. so from what I see, it seems that currently onnxruntime-extension defines all those steps, and Olive just call ort-extension to use whatever ort-extension may do (correct me if I am wrong). If everyone on the same page to agree that Olive is to just use a general interface to pass-through the configs to ort-extension and let ort-extension to do the actual work, it is OK to keep interface and document for Olive a general way. the detailed document should be put in ort-extension together with the definitions and implementation of the actual steps.

Please see comments from @devang-ml

olive/passes/onnx/append_pre_post_processing_ops.py
guotuofeng: yes

trajepl (approved):

PR #469, opened 2023-08-08
Fix OnnxConversion failure in GPTJ example

yuwenzho:
## Describe your changes
**summary:** 
Skip removing past_key_values in dummy input in OnnxConversion, which is required for hf model with past.

**Details:**
GPTJ example failed in OnnxConversion with error: `RuntimeError: The size of tensor a (8) must match the size of tensor b (18) at non-singleton dimension 3`
To quickly reproduce the error:
```python
from olive.model import PyTorchModel
from olive.passes.onnx.conversion import OnnxConversion
from olive.systems.local import LocalSystem

local_system = LocalSystem()
hf_config={"model_name": "hf-internal-testing/tiny-random-gptj",
                   "task": "text-generation",
                   "feature": "causal-lm-with-past"}
pytorch_model = PyTorchModel(hf_config=hf_config)
onnx_conversion_config = {}

p = create_pass_from_dict(OnnxConversion, onnx_conversion_config, disable_search=True)
output_folder = str("onnx")

# execute
onnx_model = local_system.run_pass(p, pytorch_model, None, output_folder)

```

The error is caused in [conversion.py#L115-L124](https://github.com/microsoft/Olive/blob/9ed6edb1163cbdba88cf198b0cb4ea3f6427829a/olive/passes/onnx/conversion.py#L115-L124). 

```diff
# some dummy inputs might not be used in the model, so we need to remove them
# this can happen when we are using an hf dataset to generate dummy inputs
# only handle dict for now since we cannot get the name of the input from a list/tuple
if isinstance(dummy_inputs, dict):
    dummy_input_keys = set(dummy_inputs.keys())
    unused_keys = dummy_input_keys - set(input_names)
-    # print(dummy_input_keys): {'past_key_values', 'attention_mask', 'input_ids'}
-    # print(set(input_names)): {'past_key_values.3.value', 'past_key_values.2.key', 'attention_mask', 'past_key_values.0.value', 'past_key_values.4.key', 'past_key_values.3.key', 'past_key_values.1.value', 'input_ids', 'past_key_values.1.key', 'past_key_values.4.value', 'past_key_values.2.value', 'past_key_values.0.key'}
-    # print(unused_keys): {'past_key_values'}
-    # past key values in dummy input are removed here mistakenly, which will cause torch.onnx.export failure.
    if unused_keys:
        logger.debug(f"Removing unused dummy inputs: {unused_keys}")
    for key in unused_keys:
        del dummy_inputs[key]
```

### DIFF
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -117,7 +117,23 @@ class OnnxConversion(Pass):
             # only handle dict for now since we cannot get the name of the input from a list/tuple
             if isinstance(dummy_inputs, dict):
                 dummy_input_keys = set(dummy_inputs.keys())
+
+                # handle dummy inputs for hf model with past, which has past_key_values
+                # match input names in `past_key_values.(hidden_layer_num).(key|value)` pattern
+                from transformers.modeling_utils import PreTrainedModel
+
+                if issubclass(type(pytorch_model), PreTrainedModel):
+                    for name, input in dummy_inputs.items():
+                        if isinstance(input, list):
+                            key_value_names = set(
+                                [f"{name}.{idx}.key" for idx in range(len(input))]
+                                + [f"{name}.{idx}.value" for idx in range(len(input))]
+                            )
+                            if key_value_names.issubset(set(input_names)):
+                                dummy_input_keys.discard(name)
+
                 unused_keys = dummy_input_keys - set(input_names)
+
                 if unused_keys:
                     logger.debug(f"Removing unused dummy inputs: {unused_keys}")
                 for key in unused_keys:
--- a/test/unit_test/passes/onnx/test_conversion.py
+++ b/test/unit_test/passes/onnx/test_conversion.py
@@ -1,16 +1,18 @@
 import tempfile
 from pathlib import Path
-from test.unit_test.utils import get_pytorch_model
+from test.unit_test.utils import get_hf_model_with_past, get_pytorch_model
+
+import pytest
 
 from olive.passes.olive_pass import create_pass_from_dict
 from olive.passes.onnx.conversion import OnnxConversion
 from olive.systems.local import LocalSystem
 
 
-def test_onnx_conversion_pass():
+@pytest.mark.parametrize("input_model", [get_pytorch_model(), get_hf_model_with_past()])
+def test_onnx_conversion_pass(input_model):
     # setup
     local_system = LocalSystem()
-    input_model = get_pytorch_model()
 
     p = create_pass_from_dict(OnnxConversion, {}, disable_search=True)
     with tempfile.TemporaryDirectory() as tempdir:
--- a/test/unit_test/utils.py
+++ b/test/unit_test/utils.py
@@ -70,6 +70,16 @@ def get_pytorch_model():
     )
 
 
+def get_hf_model_with_past():
+    return PyTorchModel(
+        hf_config={
+            "model_name": "hf-internal-testing/tiny-random-gptj",
+            "task": "text-generation",
+            "feature": "causal-lm-with-past",
+        }
+    )
+
+
 def get_pytorch_model_dummy_input(model):
     return torch.randn(1, 1)
 


yuwenzho:
My modification may not be robust. If you have any suggestions, please feel free to raise them.

olive/passes/onnx/conversion.py
```diff
@@ -118,6 +118,11 @@ def _run_for_config(
             if isinstance(dummy_inputs, dict):
                 dummy_input_keys = set(dummy_inputs.keys())
                 unused_keys = dummy_input_keys - set(input_names)
+
```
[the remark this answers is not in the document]
yuwenzho: I updated the matching method. Please help review, thanks! 

BTW, whether a new UT case needs to be added to [test_conversion.py](https://github.com/microsoft/Olive/blob/main/test/unit_test/passes/onnx/test_conversion.py) to cover this error?

olive/passes/onnx/conversion.py
```diff
@@ -117,7 +117,22 @@ def _run_for_config(
             # only handle dict for now since we cannot get the name of the input from a list/tuple
             if isinstance(dummy_inputs, dict):
                 dummy_input_keys = set(dummy_inputs.keys())
+
+                # handle dummy inputs for hf model with past, which has past_key_values
+                # match input names in `past_key_values.(hidden_layer_num).(key|value)` pattern
+                from transformers.modeling_utils import PreTrainedModel
```
guotuofeng, line 123: add try/catch since user might not install transformers.

olive/passes/onnx/conversion.py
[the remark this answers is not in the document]
guotuofeng: Yes, UT seems needed.

olive/passes/onnx/conversion.py
jambayk: `transformers` is a requirement for olive so this won't happen. Plus the new change is throwing importerror which it would do regardless

olive/passes/onnx/conversion.py
```diff
@@ -117,7 +117,25 @@ def _run_for_config(
             # only handle dict for now since we cannot get the name of the input from a list/tuple
             if isinstance(dummy_inputs, dict):
                 dummy_input_keys = set(dummy_inputs.keys())
+
+                # handle dummy inputs for hf model with past, which has past_key_values
+                # match input names in `past_key_values.(hidden_layer_num).(key|value)` pattern
+                try:
+                    from transformers.modeling_utils import PreTrainedModel
+
+                    if issubclass(type(pytorch_model), PreTrainedModel):
```
jambayk, line 126: not really a problem, but is there a reason you didn't do something like `isinstance(pytorch_model, PreTrainedModel)`?

test/unit_test/passes/onnx/test_conversion.py
```diff
@@ -22,3 +23,23 @@ def test_onnx_conversion_pass():
...
+    # test conversion for hf model with past
+    input_model = PyTorchModel(
+        hf_config={
+            "model_name": "hf-internal-testing/tiny-random-gptj",
+            "task": "text-generation",
+            "feature": "causal-lm-with-past",
+        }
+    )
+
+    p = create_pass_from_dict(OnnxConversion, {}, disable_search=True)
```
jambayk, line 36: since this is reusing the same code with only the input model different, would you be able to parameterize the test using the input model?

olive/passes/onnx/conversion.py
yuwenzho: I believe that `isinstance(pytorch_model, PreTrainedModel)` also works. I wrote it this way with reference to [transformers/onnx/convert.py#L481](https://github.com/huggingface/transformers/blob/main/src/transformers/onnx/convert.py#L481)

olive/passes/onnx/conversion.py
```diff
@@ -117,7 +117,25 @@ def _run_for_config(
...
+                        for name, input in dummy_inputs.items():
+                            if isinstance(input, list):
+                                dummy_input_keys.update(
```
jambayk, line 129: I don't think this line really does anything. Since `unused_keys = dummy_input_keys - set(input_names)`, it doesn't matter whether the new keys are added to `dummy_input_keys`. What matter is the `name` is not present in `dummy_input_keys`. 

i think a better logic could be something like: 

```python
key_value_names = set(
    [f"{name}.{idx}.key" for idx in range(len(input))] + [f"{name}.{idx}.value" for idx in range(len(input))]
)
if key_value_names.issubset(set(input_names)):
    dummy_input_keys.discard(name)
```
This makes it certain that we are dealing with key value list.

olive/passes/onnx/conversion.py
jambayk: I see. Was just curious if there was a reason for doing this. Thanks!

test/unit_test/passes/onnx/test_conversion.py
yuwenzho: Updated. Thanks a lot!

olive/passes/onnx/conversion.py
yuwenzho: Yes, you are right. Sorry I didn't consider this situation. Thanks for your comment, I updated this part.

jambayk:
Thanks for the fix! The failing test is not because of this PR. We will look into solving it and then we can retrigger the ci after the fix.

PR #468, opened 2023-08-08
✂️ Separate docker test case

trajepl:
## Describe your changes
Separate docker test case to avoid hanging on CI pipeline.

## Checklist before requesting a review
- [ ] Add unit tests for this change.
- [ ] Make sure all tests can pass.
- [ ] Update documents if necessary.
- [ ] Format your code by running `pre-commit run --all-files`
- [ ] Is this a user-facing change? If yes, give a description of this change to be included in the release notes.

PR #470, opened 2023-08-08
Separate docker test case

jambayk:
## Describe your changes
Separate docker test case to avoid hanging on CI pipeline. 

Same as #468 + additional commit to keep aml_system in test. Creating new PR since Jiapeng is unavailable to update the original PR. 

### DIFF
--- a/examples/test/test_bert_ptq_cpu_docker.py
+++ b/examples/test/test_bert_ptq_cpu_docker.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 import os
+import platform
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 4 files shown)


examples/test/test_bert_ptq_cpu_docker.py
```diff
@@ -0,0 +1,36 @@
...
+    example_dir = cur_dir / "bert"
+    os.chdir(example_dir)
+    yield
+    os.chdir(cur_dir)
+
+
+@pytest.mark.parametrize("search_algorithm", ["tpe"])
+@pytest.mark.parametrize("execution_order", ["joint"])
+@pytest.mark.parametrize("system", ["docker_system"])
+@pytest.mark.parametrize("olive_json", ["bert_ptq_cpu.json"])
```
xiaoyu-work, line 26: If we only test one value for these args, we can remove the parametrized annotation. Same to other tests.

xiaoyu-work (approved):

examples/test/test_bert_ptq_cpu_docker.py
jambayk: Yes, we can definitely clean up the parameterization for sure.

trajepl:
> ## Describe your changes
> Separate docker test case to avoid hanging on CI pipeline.
> 
> Same as #468 + additional commit to keep aml_system in test. Creating new PR since Jiapeng is unavailable to update the original PR.

I removed `aml_system` before is because that:
I saw there is a duplicated bert_cpu_aml e2e test job already.

image

PR #471, opened 2023-08-09
disable the pip cache when building docker images

guotuofeng:


jambayk (approved):

jambayk:
BTW should we add these to the other dockerfile we have too?

guotuofeng:
> BTW should we add these to the other dockerfiles we have too?

IMO, we can add this flag for all these files that has the issue.

PR #472, opened 2023-08-09
Fix DirectML examples

PatriceVignola:
## Describe your changes

- The `disable_search` argument in the config files was under `config`, but it should be one level above
- The SDXL README was updated to remove the "stable release" option since it only works from `main` for now
- Added the non-pareto footprints back, which are use by many DirectML examples and allow the intermediate steps to be investigated

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -588,6 +588,8 @@ class Engine:
             time_diff = time.time() - start_time
             self.search_strategy.check_exit_criteria(iter_num, time_diff, signal)
 
+        self.footprints[accelerator_spec].to_file(output_dir / f"{prefix_output_name}footprints.json")
+
         return self.get_pareto_frontier_footprints(
             accelerator_spec, output_model_num, objective_dict, output_dir, prefix_output_name
         )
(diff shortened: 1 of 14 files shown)


olive/engine/engine.py
```diff
@@ -588,6 +588,8 @@ def run_search(
             time_diff = time.time() - start_time
             self.search_strategy.check_exit_criteria(iter_num, time_diff, signal)
 
+        self.footprints[accelerator_spec].to_file(output_dir / f"{prefix_output_name}footprints.json")
```
PatriceVignola, line 591: I noticed this was recently removed in favor of only returning the pareto footprints, but I'd like to keep them somehow. It's a useful blueprint of all the steps that were processed; not just the pareto frontier step. For example, it allows the caller to query the location of the unoptimized models to do something else with them.

jambayk: This sounds good to me. It's definitely useful to have the full footprint saved as well.

jambayk (approved):
Thanks for catching the mistake! This LGTM. 

The failing ci job seems to be something to do with our workspace quota. not related to this PR.

PR #478, opened 2023-08-10
OnnxConversion support for gpu export

shaahji:
OnnxConversion support for gpu export

## Describe your changes

Move the model and the data to the device based on accelerator spec instead of forcing it to cpu.
Moved a function from PyTorchEvaluator to common utils so it's in shared location.

### DIFF
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -11,7 +11,7 @@ import onnx
 import torch
 
 from olive.common.config_utils import validate_config
-from olive.common.utils import tensor_data_to_device
+from olive.common.utils import device_string_to_torch_device, tensor_data_to_device
 from olive.hardware.accelerator import AcceleratorSpec
 from olive.model import CompositeOnnxModel, ONNXModel, PyTorchModel
 from olive.model.hf_utils import get_hf_model_io_config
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/conversion.py
```diff
@@ -75,8 +75,10 @@ def _run_for_config(
 
         # TODO: add e2e test for model on cpu but data on gpu; model on gpu but data on cpu
         # put pytorch_model and dummy_inputs at the same device
-        pytorch_model.to("cpu")
-        dummy_inputs = tensor_data_to_device(dummy_inputs, "cpu")
+        # device = 'cpu' if self.accelerator_spec.accelerator_type == Device.CPU else 'cuda'
+        device = device_string_to_torch_device(self.accelerator_spec.accelerator_type)
+        pytorch_model.to(device)
```
jambayk, line 80: Does the final model differ whether we use the `cpu` or `gpu`? Or is this just to speed up the model forward pass during tracing?

shaahji: Support for GPU based export is required because certain operators are implemented only for GPU and the export fails if attempted on CPU.

I can't compare the model output because the export fails on CPU because of lack of supported operators. Export performance was not a criteria for this change.

xiaoyu-work:
Not related to this PR, but we should also add gpu export for optimum conversion.

xiaoyu-work (approved):

olive/passes/onnx/conversion.py
jambayk, line 80: Can you put the device back on `cpu` once the export is done? the `pytorch_model` object persists in memory so we don't want it to keep using the `gpu` resources after this.

jambayk (changes requested):
Please move the model back to `cpu`

olive/passes/onnx/conversion.py
jambayk: Similar to https://github.com/microsoft/Olive/blob/main/olive/evaluator/olive_evaluator.py#L631

shaahji: Done!

guotuofeng (dismissed):

guotuofeng:
offline sync with @shaahji, this PR will be closed.

shaahji:
Closing in favor of https://github.com/microsoft/Olive/pull/482

PR #479, opened 2023-08-10
Few fixes for Quantization logic

shaahji:
Few fixes for Quantization logic

## Describe your changes

* Cleanup temp directories
* Force external data for preprocessing i.e. set _save_as_external_data_ to True
* Argument passed to _quant_preprocess was a Path object and was being forwarded to onnxruntime.save_model which when combined with _save_as_external_data_ fails to resolve to the correct path. External data end up being written to the current folder instead of putting it right next to the model file itself.

### DIFF
--- a/olive/passes/onnx/quantization.py
+++ b/olive/passes/onnx/quantization.py
@@ -377,9 +377,8 @@ class OnnxQuantization(Pass):
         # reload the model and save to output_model_path using the external data config
         # TODO: don't default to use_external_data_format=True if the loading and saving model makes
         # the pass inefficient
-        tmp_dir = tempfile.TemporaryDirectory(prefix="olive_tmp")
-        tmp_dir_path = Path(tmp_dir.name)
-        tmp_model_path = str(tmp_dir_path / Path(output_model_path).name)
+        new_tmp_dir = tempfile.TemporaryDirectory(prefix="olive_tmp")
+        tmp_model_path = str(Path(new_tmp_dir.name) / Path(output_model_path).name)
 
         if is_static:
             # get the dataloader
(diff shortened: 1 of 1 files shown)


olive/passes/onnx/quantization.py
```diff
@@ -377,8 +377,9 @@ def _run_for_config(
         # reload the model and save to output_model_path using the external data config
         # TODO: don't default to use_external_data_format=True if the loading and saving model makes
         # the pass inefficient
-        tmp_dir = tempfile.TemporaryDirectory(prefix="olive_tmp")
```
jambayk, line 380: We are using `tmp_dir` and not `self.tmp_dir` since we want the contents in `self.tmp_dir` to persist. It is used to store the `preprocessed` model created at line 326. So that the same input model doesn't need to be preprocessed at every run. 

So this is intentional.

olive/passes/onnx/quantization.py
```diff
@@ -417,16 +418,21 @@ def _run_for_config(
...
+    def _quant_preprocess(self, model: ONNXModel, output_model_path: Union[str, Path]) -> ONNXModel:
         from onnxruntime.quantization.preprocess import quant_pre_process
 
         try:
-            quant_pre_process(input_model_path=model.model_path, output_model_path=output_model_path, auto_merge=True)
+            quant_pre_process(
+                input_model_path=model.model_path,
+                output_model_path=str(output_model_path),
+                auto_merge=True,
+                save_as_external_data=True,
```
jambayk, line 434: This one makes sense since it is better to `save_as_external_data` even if not needed than for it to fail for large models. 

IIRC, i had some issues setting this to true always because of issues when loading the model.

olive/passes/onnx/quantization.py
shaahji: Fixed.

olive/passes/onnx/quantization.py
shaahji: I had similar problems and as it turns out it was because of the missing `str` for otput_model_path. `onnxruntime.save_model` doesn't support `Path` object and so the argument was not being resolved correctly.

olive/passes/onnx/quantization.py
jambayk: I don't think the issue here is that there are two tmp_dirs. It's the clean_up of `self.tmp_dir1` since that deletes the original preprocessed model. So the preprocessing would run again in the next run with the same input model.

shaahji: No, it's doesn't anymore because the preprocessed model is saved correctly to the output folder (instead of the working directory as it was earlier). All these path related issues were because of the incompatible `Path` argument to the `quant_pre_process` function. You can try running it locally to verify.

jambayk: Sorry, I mean that line 419 `self.tmp_dir1.cleanup()` would delete the preprocessed model that was saved in the current run. 
So in the next run of this pass with the same input model (during search), line `332` returns `false` and it runs the preprocessing again. 
if we won't do the cleanup for tmp_dir1 every run, then we can reuse the preprocessed model in subsequent runs of the same pass.

jambayk: Personally, I think we can just keep the `tmp_dir` related stuff as before. The `self.tmp_dir` is used for storing persistent files for the object. but the `tmp_dir` inside is used for actually temporary files.

olive/passes/onnx/quantization.py
jambayk: Oh I see. Thanks!

olive/passes/onnx/quantization.py
shaahji: Hmmm, removed the delete and added a note for future.

shaahji: I ran into both the issues (disk space and files getting deleted by OS) and that's what landed me around this code. I am working with a very large model and the data file is around 6GB and multiple copies adds up really fast.

jambayk: I will create a work item to track this. The tmp_dirs behave as expected on Linux. On windows, I have seen files getting deleted even though the reference to the tmp_dir still exists (so it should not have been deleted). 
We need to do some investigation into temp directory behaviors on different OS and come up with best practices. Maybe using context managers instead of object references when possible.

jambayk (approved):

PR #481, opened 2023-08-11
Separate `failed` and `invalid` pass run configs

jambayk:
## Describe your changes
Separate `PRUNED_CONFIG` into `INVALID_CONFIG` and `FAILED_CONFIG`. 
`INVALID_CONFIG` is used when the run config is invalid for a given accelerator spec.
`FAILED_CONFIG` is used when the run config is valid, but the pass run fails for some reason. 

This helps us make the search more efficient. the transformer optimizer pass was made `not accelerator_agnostic` even though the pass run itself doesn't depend on the accelerator. All information is already in the `run_config`. 
It is okay for a pass to be `accelerator_agnostic` but the search space to have points invalid for specific `accelerator_specs`. 

What matters when determining if a pass is `accelerator_agnostic` is if the pass run uses the accelerator specs and the output depends on it. 
In the transformer optimizer case, the `use_gpu` information already contains the necessary accelerator information so the run config has all information needed. 

### DIFF
--- a/olive/engine/config.py
+++ b/olive/engine/config.py
@@ -12,8 +12,12 @@ from olive.evaluator.olive_evaluator import OliveEvaluatorConfig
 from olive.strategy.search_strategy import SearchStrategyConfig
 from olive.systems.system_config import SystemConfig
 
-# pass search-point/config was pruned due invalid config or failed run
-PRUNED_CONFIG = "pruned-config"
+# pass search-point was pruned due to failed run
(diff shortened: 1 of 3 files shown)


guotuofeng:
@trajepl please take a look

trajepl (approved):

trajepl:
Thanks. It is really a good point not to cache the invalid configs. We should cache the results from running no matter failed or  success.

PR #486, opened 2023-08-11
Make pass list optional in config

xiaoyu-work:
## Describe your changes

Make pass list optional in config. If no pass provided, Olive will only evaluate input model.

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -278,8 +278,7 @@ class Engine:
             pass_flows: a list of pass names, each pass name is a string.
         """
         if not pass_flows:
-            assert self.pass_config, "Pass flows can only be set after passes are registered"
-            self.pass_flows = [list(self.pass_config.keys())]
+            self.pass_flows = [list(self.pass_config.keys())] if self.pass_config else []
         else:
             self.pass_flows = pass_flows
 
--- a/olive/workflows/run/config.py
+++ b/olive/workflows/run/config.py
@@ -66,7 +66,7 @@ class RunConfig(ConfigBase):
     evaluators: Dict[str, OliveEvaluatorConfig] = None
     pass_flows: List[List[str]] = None
     engine: RunEngineConfig
-    passes: Dict[str, RunPassConfig]
+    passes: Dict[str, RunPassConfig] = None
 
     @validator("input_model", pre=True)
     def insert_aml_client(cls, v, values):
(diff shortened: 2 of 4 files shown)


olive/engine/engine.py
```diff
@@ -309,7 +307,6 @@ def run(
                 1. Final model -> {output_dir}/{output_name}_{AcceleratorSpec}_model.onnx
                 2. JSON file -> {output_dir}/{output_name}_{AcceleratorSpec}_model.json
                 3. Evaluation results of the final model -> {output_dir}/{output_name}_{AcceleratorSpec}_metrics.json
-
```
guotuofeng, line 312: I add the blank lines to fix the warning when building docs

guotuofeng:
Do we really need only evaluate the input models in Olive?

olive/engine/engine.py
xiaoyu-work: I didn't realize this change. I'll add it back.

xiaoyu-work:
> Do we really need only evaluate the input models in Olive?

I will add model perf compare pipeline which will only evaluates input model with no pass added. Actually we added input model only evaluation before but not complete this: https://github.com/microsoft/Olive/blob/main/olive/engine/engine.py#L350

olive/engine/engine.py
```diff
@@ -294,7 +293,6 @@ def run(
     ):
         """
         Run all the registered Olive passes on the input model and produce one or more candidate models.
-
```
guotuofeng, line 297: ditto

guotuofeng: maybe you can rebase from main?

xiaoyu-work: fixed

olive/workflows/run/config.py
```diff
@@ -66,7 +66,7 @@ class RunConfig(ConfigBase):
     evaluators: Dict[str, OliveEvaluatorConfig] = None
     pass_flows: List[List[str]] = None
     engine: RunEngineConfig
-    passes: Dict[str, RunPassConfig]
+    passes: Dict[str, RunPassConfig] = None
```
guotuofeng, line 69: just curious if the empty dict {} works for original code without your changes. 
This is not related to your PR

xiaoyu-work: New `set_pass_flows` still needs a pass to be registered. But empty dict works after the change for `set_pass_flows`

trajepl: Oh, xiaoyu is right. In my previous changes, I missed this part.
image

trajepl (approved):

PR #488, opened 2023-08-12
Fix inc quantization pass bug

xiaoyu-work:
## Describe your changes
Fix inc quantization pass bug

### DIFF
--- a/olive/passes/onnx/inc_quantization.py
+++ b/olive/passes/onnx/inc_quantization.py
@@ -271,7 +271,7 @@ class IncQuantization(Pass):
         config.update(get_external_data_config())
         return config
 
-    def _set_eval_func(self, accuracy_metric, sub_type):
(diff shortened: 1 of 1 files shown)


guotuofeng:
would you please help add more description to clarify what's the bug?

trajepl:
Seems like the bug is invalid arguments from previous changes on `data_root`. The fix is ok for me.
Furthermore we should cover it in our test.

xiaoyu-work:
> would you please help add more description to clarify what's the bug?

The recent change for data_root was added to `evaluator.evaluate()` but inc quantization pass wasn't updated with this change..

xiaoyu-work:
> Seems like the bug is invalid arguments from previous changes on `data_root`. The fix is ok for me. Furthermore we should cover it in our test.

Yeah I was thinking cover more examples in a different pipeline so we can make sure all examples work.

guotuofeng (approved):

PR #493, opened 2023-08-15
Add `TorchTRTConversion` pass

jambayk:
## Describe your changes
This PR adds `TorchTRTConversion` pass which converts `nn.Linear` modules in transformer layers to `TRTModule` from `torch-tensorrt`. `torch-tensorrt` is an extension to `torch` where TensorRT engines can be used like normal `nn.Module`. 

We require this conversion to take advantage of `2:4` sparsity. 

~~I have not added unit test for the pass since the pass requires GPU and we don't have GPU support for ci tests yet. I will add a unit test in a follow up PR that uses aml system to run the pass and evaluation on GPU.~~

Add `TorchTRTConversion` pass.

### DIFF
--- a/olive/passes/pytorch/torch_trt_conversion.py
+++ b/olive/passes/pytorch/torch_trt_conversion.py
@@ -9,7 +9,7 @@ from typing import Any, Dict, List, Union
 import torch
 
 from olive.common.utils import tensor_data_to_device
-from olive.hardware.accelerator import AcceleratorSpec
+from olive.hardware.accelerator import AcceleratorSpec, Device
 from olive.model import PyTorchModel
 from olive.passes import Pass
 from olive.passes.olive_pass import PassConfigParam
@@ -18,8 +18,11 @@ from olive.passes.pytorch.sparsegpt_utils import _get_attr, get_layer_submodules
 logger = logging.getLogger(__name__)
 
 
-class SparseTRTConversion(Pass):
-    """Convert a PyTorch model to a sparse fp16 Pytorch model with TRT Modules."""
+class TorchTRTConversion(Pass):
+    """
+    Convert the nn.Linear modules in the transformer layers of a Hugging Face PyTorch model to TensorRT modules with
+    fp16 precision and sparse weights, if present.
+    """
 
     _requires_data_config = True
 
@@ -44,6 +47,14 @@ class SparseTRTConversion(Pass):
             ),
         }
 
+    def validate_search_point(
+        self, search_point: Dict[str, Any], accelerator_spec: AcceleratorSpec, with_fixed_value: bool = False
+    ) -> bool:
+        if accelerator_spec.accelerator_type != Device.GPU:
+            logger.info("TorchTRTConversion only supports GPU.")
+            return False
+        return True
+
     @torch.no_grad()
     def _run_for_config(
         self, model: PyTorchModel, data_root: str, config: Dict[str, Any], output_model_path: str
@@ -56,11 +67,11 @@ class SparseTRTConversion(Pass):
             raise ValueError(f"Unsupported model type: {model_type}. Supported types: {seqlens.keys()}")
 
         if not torch.cuda.is_available():
-            raise ValueError("SparseTRTConversion requires a GPU to run.")
+            raise ValueError("TorchTRTConversion requires a GPU to run.")
         device = "cuda"
 
         # load_data
-        assert config["data_config"] is not None, "Data config is required for SparseTRTConversion."
+        assert config["data_config"] is not None, "Data config is required for TorchTRTConversion."
         first_batch = self._data_config.to_data_container().get_first_batch(data_root_path=data_root)[0]
         first_batch = tensor_data_to_device(first_batch, device=device)
         batch_size = first_batch["input_ids"].shape[0]
--- a/olive/passes/pytorch/torch_trt_conversion.py
+++ b/olive/passes/pytorch/torch_trt_conversion.py
@@ -22,6 +22,9 @@ class TorchTRTConversion(Pass):
     """
     Convert torch.nn.Linear modules in the transformer layers of a Hugging Face PyTorch model to TensorRT modules with
     fp16 precision and sparse weights, if applicable.
+
+    The entire model is saved using `torch.save` and can be loaded using `torch.load`. Loading the model requires
+    `torch-tensorrt` and Olive to be installed.
     """
 
     _requires_data_config = True
--- a/olive/passes/pytorch/torch_trt_conversion.py
+++ b/olive/passes/pytorch/torch_trt_conversion.py
@@ -13,7 +13,13 @@ from olive.hardware.accelerator import AcceleratorSpec, Device
 from olive.model import PyTorchModel
 from olive.passes import Pass
 from olive.passes.olive_pass import PassConfigParam
-from olive.passes.pytorch.sparsegpt_utils import _get_attr, get_layer_submodules, get_layers, seqlens
+from olive.passes.pytorch.sparsegpt_utils import (
+    _get_attr,
+    get_layer_submodules,
+    get_layers,
+    seqlens,
+    validate_min_max_layers,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -102,8 +108,7 @@ class TorchTRTConversion(Pass):
         layers = get_layers(pytorch_model, model_type)
 
         # get layer information
-        min_layer = config["min_layer"] or 0
-        max_layer = config["max_layer"] or len(layers)
+        min_layer, max_layer = validate_min_max_layers(config["min_layer"], config["max_layer"], len(layers))
         layer_name_filter = config["layer_name_filter"] or []
         if isinstance(layer_name_filter, str):
             layer_name_filter = [layer_name_filter]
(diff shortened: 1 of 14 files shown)


olive/passes/pytorch/trt_utils.py
```diff
@@ -0,0 +1,67 @@
...
+def compile_trt_model(torch_module: torch.nn.Module, hidden_states: torch.Tensor, batch_size: int, seqlen: int):
+    """Compile a torch module to a TensorRT module.
+
+    :param torch_module: The torch module to compile. Only torch.nn.Linear modules are supported currently.
+    :param hidden_states: The input tensor to the torch module.
+    :param batch_size: The batch size of the input tensor.
+    :param seqlen: The maximum sequence length of the input tensor. seqlen dimension is treated as dynamic.
+    """
+
+    logging.getLogger("torch_tensorrt").setLevel(logging.ERROR)
```
guotuofeng, line 35: what's the consideration that we need set the logger level of torch_tensor? would you add some comments here?

olive/passes/pytorch/torch_trt_conversion.py
```diff
@@ -0,0 +1,172 @@
...
+        # disable cache
+        use_cache = pytorch_model.config.use_cache
+        pytorch_model.config.use_cache = False
+
+        # get module list of layers
+        layers = get_layers(pytorch_model, model_type)
+
+        # get layer information
+        min_layer = config["min_layer"] or 0
+        max_layer = config["max_layer"] or len(layers)
```
guotuofeng, line 106: do we need verify whether min_layer and max_layer is valid or not? what if user specify max_layer greater than len(layers)?

olive/passes/pytorch/trt_utils.py
jambayk: Yes, I can add comments. 
the torch tensorflow logs are very verbose so the olive logs get over populated. 
It produces many lines per module and there are around or so module per model.

olive/passes/pytorch/torch_trt_conversion.py
jambayk: Oh yes, I was thinking of the same but didn’t get around to adding this check. Also needed for the sparsegpt pass 

I will add it. Thanks!

guotuofeng:
I think you might mean integration test to test TRTConversion. Is it possible that mock compile_trt_model to test the logic in the TorchTRTConversion in unit test?

jambayk:
> I think you might mean integration test to test TRTConversion. Is it possible that mock compile_trt_model to test the logic in the TorchTRTConversion in unit test?

I wasn’t sure if it would fall under integration test if we use aml system + pass.
Yes, I can mock the function. That sounds like a good idea since that’s the only one that requires gpu. I can probably mock the cuda related code too.

olive/passes/pytorch/trt_utils.py
jambayk: Done!

olive/passes/pytorch/torch_trt_conversion.py
jambayk: Done!

test/unit_test/passes/pytorch/test_torch_trt_conversion.py
jambayk, line 1: @guotuofeng I have added the ut. Can you please take a look at this? Thanks!

test/unit_test/passes/pytorch/test_torch_trt_conversion.py
```diff
@@ -0,0 +1,85 @@
...
+@patch("torch.zeros", side_effect=mocked_torch_zeros)
+def test_sparsegpt(mock_is_cuda_available, mock_tensor_data_to_device, mock_module_to, mock_torch_zeros):
+    # setup
+    # mock trt utils since we don't have tensorrt and torch-tensorrt installed
+    mock_trt_utils = MagicMock()
+    mock_compile_trt_model = MagicMock()
+    mock_compile_trt_model.return_value = MockTRTLinearLayer()
+    mock_trt_utils.compile_trt_model = mock_compile_trt_model
+
+    sys.modules["olive.passes.pytorch.trt_utils"] = mock_trt_utils
```
guotuofeng, line 47: >sys.modules["olive.passes.pytorch.trt_utils"] = mock_trt_utils [](http://example.com/codeflow?start=4&length=62)

I suppose this line is to avoid import TensorRT related stuff in UT. right?

jambayk: yes, tenssort requires gpu, etc

test/unit_test/passes/pytorch/test_torch_trt_conversion.py
```diff
@@ -0,0 +1,85 @@
...
+        data_config = huggingface_data_config_template(model_name=model_name, task=task, **dataset)
+        config = {
+            "data_config": data_config,
+        }
+
+        p = create_pass_from_dict(TorchTRTConversion, config, disable_search=True)
+        output_folder = str(Path(tempdir) / "sparse")
+
+        # execute
+        model = local_system.run_pass(p, input_model, None, output_folder)
```
guotuofeng, line 76: >local_system.run_pass [](http://example.com/codeflow?start=16&length=21)

could the pass.run be more clear?

jambayk: The other pass tests do the same as this so I copied them here

test/unit_test/passes/pytorch/test_torch_trt_conversion.py
```diff
@@ -0,0 +1,85 @@
...
+# all the following patches are needed to run the test on CPU
+# make torch.cuda.is_available return True
+@patch("torch.cuda.is_available", return_value=True)
+# make tensor_data_to_device return the input tensor, ignore device
+@patch("olive.common.utils.tensor_data_to_device", side_effect=lambda x, y: x)
+# make torch.nn.Module.to do nothing
+@patch("torch.nn.Module.to", side_effect=lambda device: None)
+# replace device in kwargs with "cpu"
+@patch("torch.zeros", side_effect=mocked_torch_zeros)
+def test_sparsegpt(mock_is_cuda_available, mock_tensor_data_to_device, mock_module_to, mock_torch_zeros):
```
guotuofeng, line 39: >test_sparsegpt [](http://example.com/codeflow?start=4&length=14)

test_sparsegpt_happy_path be more clear since we might add other cases in future to test the invalid scenarios like max_layer invalid?

jambayk: Yeah, I can do that

olive/passes/pytorch/torch_trt_conversion.py
```diff
@@ -0,0 +1,177 @@
...
+            del info["handles"]
+
+        # convert submodules to trt modules
+        logger.debug(f"Converting layers {min_layer} to {max_layer}...")
+        for layer_index, info in layer_info.items():
+            logger.debug(f"Converting layer {layer_index}...")
+            for name, shape in info["input_shapes"].items():
+                input = torch.zeros(shape, dtype=torch.float16, device=device)
+                # create trt module
+                trt_module = compile_trt_model(info["submodules"][name], input, batch_size, seqlen)
```
guotuofeng, line 153: >compile_trt_model [](http://example.com/codeflow?start=29&length=17)

should we check whether the input submodule is torch.nn.Linear before calling compile_trt_model? or the pass caller need make sure it?

jambayk: This is guaranteed by line 123. 
all of the sub modules we access here are found by that line where we specifically ask for NN.linear

guotuofeng: If user need to make sure the parent model, should we add document for what kinds of model types supports?

guotuofeng: or this pass only used for sparsegpt scenario?

jambayk: Yes, we are only using it for the sparsegpt scenario. I can add documentation in the pass docstring

olive/passes/pytorch/torch_trt_conversion.py
```diff
@@ -0,0 +1,177 @@
...
+    get_layer_submodules,
+    get_layers,
+    seqlens,
+    validate_min_max_layers,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class TorchTRTConversion(Pass):
```
guotuofeng, line 27: >TorchTRTConversion [](http://example.com/codeflow?start=6&length=18)

does this pass support any type PT model of just some models defined in the layer map? Do we need add document to demonstrate what type of models this pass support?

test/unit_test/passes/pytorch/test_torch_trt_conversion.py
jambayk: Resolving this conversation since this is the case for all pass tests. Maybe we can make the change to all pass tests in a different PR if nedded. 
Please unresolve the conversation if you think otherwise.

olive/passes/pytorch/torch_trt_conversion.py
jambayk: Done. Added to the docs and docstring. 
I have plans to support non hf_config pytorch models in the future. But for now, it only supports PyTorch model with hf_config and some model types.

guotuofeng (approved):

PR #502, opened 2023-08-22
Bugfix: Insert `input_model` data_config and don't overwrite custom hf data_configs

jambayk:
## Describe your changes
Previously, `RunConfigs.data_configs` had some default configs. We remove the default configs and only insert a data config for the input model if it has `hf_config.dataset`. 

There was also a bug where all HF container data configs were being overwritten by the input model's dataset. But we don't want to do this. This has been fixed by only updating the `model_name` and `task` of the user provided hf data_configs if not already present. 

### DIFF
--- a/olive/workflows/run/config.py
+++ b/olive/workflows/run/config.py
@@ -10,7 +10,6 @@ from pydantic import validator
 from olive.azureml.azureml_client import AzureMLClientConfig
 from olive.common.config_utils import ConfigBase, validate_config
 from olive.data.config import DataConfig
-from olive.data.constants import DEFAULT_HF_DATA_CONTAINER_NAME, DefaultDataContainer
 from olive.data.container.huggingface_container import HuggingfaceContainer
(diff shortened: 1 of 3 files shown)


olive/workflows/run/config.py
```diff
@@ -10,7 +10,6 @@
 from olive.azureml.azureml_client import AzureMLClientConfig
 from olive.common.config_utils import ConfigBase, validate_config
 from olive.data.config import DataConfig
-from olive.data.constants import DEFAULT_HF_DATA_CONTAINER_NAME, DefaultDataContainer
 from olive.data.container.huggingface_container import HuggingfaceContainer
```
trajepl, line 13: We can remove `DEFAULT_HF_DATA_CONTAINER_NAME` from olive.data.constants as there is no case to use that.

trajepl:
Thanks for this fixing which remove the default data configs. It looks clearer.

trajepl (approved):

PR #505, opened 2023-08-23
`Pass._requires_data_config` -> `Pass._data_configs()`

jambayk:
## Describe your changes
`Pass` has a class attribute `_requires_data_config` that we use to add a config parameter with type `DataConfig` and name `data_config`. We also use this in workflow run config validation to resolve string name references or auto insert input models. 

However, this functionality is limited. We can only add one `DataConfig` parameter with name `data_config` and single description. Currently, all of the passes are using the same parameter definition. 

In this PR, `_requires_data_config` is replaced with a static method `_data_configs` where we can define data config parameters with their own names and descriptions. We can also specify whether the parameters are required and if we should auto insert input model data configs. 
This will be especially useful in passes that do training (fine-tuning), etc where we need multiple data configs for train and validation. 

### DIFF
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -18,10 +18,11 @@ from olive.model import CompositeOnnxModel, DistributedOnnxModel, OliveModel
 from olive.passes.pass_config import (
     PassConfigBase,
     PassConfigParam,
+    PassDataConfigParam,
     PassParamDefault,
     create_config_class,
-    get_data_config,
     get_user_script_config,
+    pass_data_configs_to_params,
 )
 from olive.resource_path import ResourcePath
 from olive.strategy.search_parameter import (
@@ -46,8 +47,6 @@ class Pass(ABC):
     registry: Dict[str, Type["Pass"]] = {}
     # True if pass configuration requires user script for non-local host support
     _requires_user_script: bool = False
-    # True if pass configuration requires data configuration which will leverage data container for pass execution
-    _requires_data_config: bool = False
     # True if the pass processes a composite model at once. Otherwise, the components of the
     # composite model will be processed individually.
     _accepts_composite_model: bool = False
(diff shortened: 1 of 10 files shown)


guotuofeng:
If users need specify the data_config by overriding the _data_configs explicitly, the method would be needless since users could add the config in _default_config?

olive/passes/olive_pass.py
```diff
@@ -214,6 +211,25 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
         """
         raise NotImplementedError()
 
+    @classmethod
+    def data_configs(cls) -> Dict[str, PassDataConfigParam]:
```
guotuofeng, line 215: could we consider using _default_config if user would like explicitly specify the data_config?

guotuofeng: the _default_config seems a little bit duplicate with _data_config method?

guotuofeng (changes requested):
could we unify data_config() with default_config()?

olive/passes/olive_pass.py
jambayk: Adding to default_config is possible too but it makes it hard to get the names of data_config parameters (since we need accelerator config to use default_config) 
And it doesn’t have the option to whether to auto insert from input model or not.

olive/passes/onnx/inc_quantization.py
```diff
@@ -272,6 +271,17 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
         config.update(get_external_data_config())
         return config
 
+    @staticmethod
+    def _data_configs() -> Dict[str, PassDataConfigParam]:
+        return {
+            "data_config": PassDataConfigParam(
```
trajepl, line 277: better to "calibration_data_config"? It may be clearer for user.

olive/passes/onnx/perf_tuning.py
```diff
@@ -342,6 +341,14 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
             ),
         }
 
+    @staticmethod
+    def _data_configs() -> Dict[str, PassDataConfigParam]:
+        return {
+            "data_config": PassDataConfigParam(
+                description="Data config for calibration, required if 'dataloader_func' is not provided."
```
trajepl, line 348: This is not for `calibration`, right? It is for latency metrics.

olive/passes/onnx/inc_quantization.py
jambayk: Yes, I considered changing the name but didn’t do it in this PR since I will have to update the examples haha

olive/passes/onnx/perf_tuning.py
jambayk: Oh yeah. I thought I updated the description here but it didn’t 
Thanks!

PR #506, opened 2023-08-23
Remove `Pass._requires_data_config`, resolve using `endswith`, don't auto-insert input model dataset

jambayk:
## Describe your changes
This PR is an update to [#506](https://github.com/microsoft/Olive/pull/505) with changes discussed offline. 

`Pass` has a class attribute `_requires_data_config` that we use to add a config parameter with type `DataConfig` and name `data_config`. We also use this in workflow run config validation to resolve string name references or auto insert input models. 

However, this functionality is limited. We can only add one `DataConfig` parameter with name `data_config` and single description. Currently, all of the passes are using the same parameter definition. 

We remove `_requires_user_config` class attribute from passes. DataConfig parameters must be defined along with other parameters in `_default_config`. Parameter names ending in `default_config` are reserved for such parameters. We use this to find and resolve DataConfig parameters in run config validation. 

Auto-insertion of the input model (if it's a pytorch model with hf_config.dataset) data config in pass configs is removed. This behavior only applies to a subset of pytorch input models and might be unexpected to the user. If the user wants to use the input model's data config, they can do so by assigning the value for the parameter as `INPUT_MODEL_DATA_CONFIG`. After this, it behaves like any other string name resolution. 

### DIFF
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -6,7 +6,7 @@ import inspect
 import logging
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any, Callable, Dict, Optional, Tuple, Type, Union
+from typing import Any, Callable, Dict, Optional, Tuple, Type, Union, get_args
 
 from pydantic import validator
 
@@ -20,7 +20,6 @@ from olive.passes.pass_config import (
     PassConfigParam,
     PassParamDefault,
     create_config_class,
-    get_data_config,
     get_user_script_config,
 )
 from olive.resource_path import ResourcePath
@@ -46,8 +45,6 @@ class Pass(ABC):
     registry: Dict[str, Type["Pass"]] = {}
     # True if pass configuration requires user script for non-local host support
     _requires_user_script: bool = False
-    # True if pass configuration requires data configuration which will leverage data container for pass execution
-    _requires_data_config: bool = False
     # True if the pass processes a composite model at once. Otherwise, the components of the
     # composite model will be processed individually.
     _accepts_composite_model: bool = False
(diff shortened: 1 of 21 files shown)


olive/workflows/run/config.py
```diff
@@ -50,7 +50,7 @@ def create_engine(self):
         return Engine(config)
 
 
-INPUT_MODEL_DATA_CONFIG_NAME = "olive_input_model_data_config"
+INPUT_MODEL_DATA_CONFIG = "INPUT_MODEL_DATA_CONFIG"
```
guotuofeng, line 53: what about use something start with double underscore? for example "__input_model_data_config__?

jambayk: Do they also provide the underscore value? or the constant name?

jambayk: It is the value. But I wasn't sure if we want user to provide the underscore value or constant name if we change it to underscore.

guotuofeng: IMHO, The constant name is Olive internally and user need only knowthe reserved names itself.

jambayk: That sounds good to me. I will update this. Thanks!

olive/passes/olive_pass.py
```diff
@@ -80,9 +77,12 @@ def __init__(
...
             self._user_module_loader = UserModuleLoader(self._config["user_script"], self._config["script_dir"])
-        if self._requires_data_config:
-            data_config = self._config.get("data_config") or {}
-            self._data_config = DataConfig(**data_config)
+
+        self._data_configs = {}
+        for k, v in self._config.items():
+            if v is None or not k.endswith("data_config"):
+                continue
+            self._data_configs[k] = v if isinstance(v, DataConfig) else DataConfig(**v)
```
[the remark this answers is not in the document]
jambayk: Yes, I think we can do that. I will update this.

olive/workflows/run/config.py
jambayk: Done! 
There are no documentation at all about data_configs so I think we can add the documentation in a later PR. Looks like it was not added when the concept was first introduced.

olive/passes/olive_pass.py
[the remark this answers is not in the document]
jambayk: Done!

guotuofeng (dismissed):
Please help update document for the reserved data_config

guotuofeng (approved):

PR #509, opened 2023-08-24
`test-generation` dataset: Define strategy for `corpus` datasets, introduce `pair` dataset type

jambayk:
## Describe your changes
Follow up to [the discussion](https://github.com/microsoft/Olive/pull/444#discussion_r1275799367) in https://github.com/microsoft/Olive/pull/444. 

There are many ways text-generation datasets are tokenized. We currently support multiple methods where the text are joined together and tokenized with or without a sliding window. The different methods were selected based on the combination of arguments. 

As I suggested in the linked discussion, this PR formalizes the different methods of tokenization into `corpus_strategy`. The text is joined or processed individually based on the strategy. 

We also introduce a second type of dataset for `text-generation` tasks called `pair`. Here, the dataset is expected to consist of two text columns: input and output. Such as prompt and completion in chatbot type scenario. Note that this is not a seq2seq scenerio since the model is still generative. The input and output are concatenated into a single sequence. The input might be ignored during evaluation (loss, metrics, etc) based on the arguments provided. 

Added some config classes to maintain and validate the different kwargs accepted by the preprocessor.


olive/data/component/pre_process_data.py
```diff
@@ -128,121 +137,458 @@ def _tokenizer_and_align_labels(examples):
...
+    def _check_custom(cls, v, field, values):
+        if "pair_format" not in values:
+            raise ValueError("Invalid pair_format")
+        if values["pair_format"] == TextGenPairFormat.CUSTOM and v is None:
+            raise ValueError(f"{field.name} must be specified when pair_format is CUSTOM")
+        return v
+
+
 @Registry.register_pre_process()
 def text_generation_huggingface_pre_process(
```
trajepl, line 258: Could we move those Params class and help function to a separated module? Like componnet.generation or something else. 
It is better to keep pre_process_data.py clear then user/dev can know which kind of function is used for data_config.

jambayk: Yes, I can move them to a separate file. I was also thinking that this file is going to become unwieldy as we add more tasks. But wasn't sure if I should move the preprocessing func also. 

I will only move the helper classes and functions as you suggested.

jambayk: Done!

trajepl (approved):

PR #512, opened 2023-08-25
Move reusable code from `sparsegpt_utils` to common and hf utils

jambayk:
## Describe your changes
`sparsegpt_utils` has some mappings and helper functions that are useful in other parts of Olive too. This PR refactors it and puts the reusable code into `olive.common.utils` and `olive.model.hf_utils`. 

Also cleaned up the imports in `hf_utils` by moving `transformers` imports to top of file. `transformers` is an olive requirement so we don't need to lazy import it unless it is a version specific import. 

### DIFF
--- a/olive/common/utils.py
+++ b/olive/common/utils.py
@@ -168,9 +168,9 @@ def get_attr(module, attr, fail_on_not_found=False):
     """
     Get attribute from module.
 
-    :param module: module to get attribute from
-    :param attr: attribute name, can be a string with dot notation. If empty, return modulenot
-    :param fail_on_not_found: if True, raise AttributeError if attribute is not found
+    :param module: module to get attribute from.
+    :param attr: attribute name, can be a string with dot notation. If empty, return module.
+    :param fail_on_not_found: if True, raise AttributeError if attribute is not found.
     :return: attribute
     """
     if not attr:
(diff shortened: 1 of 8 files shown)


olive/common/utils.py
```diff
@@ -162,3 +162,28 @@ def tensor_data_to_device(data, device: str):
...
+        # return module if attr is empty
+        return module
+
+    attr = attr.split(".")
+    for a in attr:
+        try:
+            module = getattr(module, a)
+        except AttributeError:
+            if fail_on_not_found:
+                raise AttributeError(f"Attribute {attr} not found.")
```
guotuofeng, line 186: I suppose raise is enough. If we want to re-raise the exception, we could use raise from e

jambayk: The error message is different from the original one. The original message only mentions `a` but we want to raise with `attr` in the message.

guotuofeng: how about [https://stackoverflow.com/questions/24752395/python-raise-from-usage](https://stackoverflow.com/questions/24752395/python-raise-from-usage)?

guotuofeng: [https://github.com/microsoft/OLive/blob/9028fd5dd6def368ce93dda14140fe01fff8b3ed/olive/passes/onnx/quantization.py#L412](https://github.com/microsoft/OLive/blob/9028fd5dd6def368ce93dda14140fe01fff8b3ed/olive/passes/onnx/quantization.py#L412)

jambayk: Thanks! just updated it. 
The `raise from` method is much better since it returns a better trace.

guotuofeng (approved):

guotuofeng (approved):

PR #518, opened 2023-08-29
🏴 Fetch hf model config and spread to ONNX model

trajepl:
## Describe your changes

1. Fetch huggingface model config if the input model is load from `from_pretrained`.
2. Spread the config to other onnx model which can leverage to fill out some parameters automatically.


olive/passes/onnx/transformer_optimization.py
```diff
@@ -21,6 +21,21 @@ class OrtTransformersOptimization(Pass):
     """Optimize transformer based models in scenarios where ONNX Runtime does not apply the optimization at load time.
     It is based on onnxruntime.transformers.optimizer."""
 
+    # To extend following list/map from huggingface config
+    # there is the priority order: NUM_HEADS_NAMES[0] and HIDDEN_SIZE_NAMES[0] are the first choice
+    # which means user can override the value in config file or hf_config.config
+    NUM_HEADS_NAMES = ["num_heads", "num_attention_heads", "n_head", "encoder_attention_heads"]
+    HIDDEN_SIZE_NAMES = ["hidden_size", "d_model", "n_embd"]
+    MODEL_TYPE_MAPPING = {
```
guotuofeng, line 29: >MODEL_TYPE_MAPPING [](http://example.com/codeflow?start=4&length=18)

could we unify these definitions with the hf_mapping?

docs/source/overview/options.md
```diff
@@ -113,8 +113,11 @@ case insensitive.
                 - `component_func: [str]`: The component function name will be loaded from `model_script`.
                 - `dummy_inputs_func: [str]`: The dummy input function name will be loaded from `model_script`.
 
-        - `config: [str]`: The config of the model can be provided as well. Such as `WhisperConfig`. See
-        [huggingface configurations](https://huggingface.co/docs/transformers/main_classes/configuration)
+        - `config: [dict]`: The config of the model can be provided as well. Such as `WhisperConfig`. See
```
jambayk, line 116: I didn't realize that we had a `hf_config.config` field. If this is true, I think we might have to update our `load_huggingface_model_from_task` and `get_hf_model_config` to use `hf_config.config` while calling `from_pretrained`.

olive/passes/onnx/conversion.py
```diff
@@ -189,7 +190,14 @@ def _convert_model_on_device(
         if device != "cpu":
             pytorch_model.to("cpu")
         # save the model to the output path and return the model
-        return model_proto_to_olive_model(onnx_model, output_model_path, config)
+        olive_onnx_model = model_proto_to_olive_model(onnx_model, output_model_path, config)
+        # inherit the hf_config from the original model
+        _hf_config = deepcopy(model.hf_config) or HFConfig()
+        _hf_config.config = _hf_config.config or {}
+        if hasattr(pytorch_model, "config"):
```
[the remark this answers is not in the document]
jambayk: Oh, I got confused. This is from the loaded pytorch model and not the olive pytorch model.

olive/passes/onnx/transformer_optimization.py
```diff
@@ -140,6 +155,25 @@ def _run_for_config(
         for key in get_external_data_config():
             del run_config[key]
 
+        if model.hf_config and model.hf_config.config:
```
jambayk, line 158: I would like the relationship between `model_name`, `model_path`, `config` to be better defined first. 
If we want to combine the config from `model_path`/`model_name` and `hf_config`, I think we have to load the `PreTrainedConfig` here correctly instead of just looking at `hf_config.config`. Since the config can also be loaded directly from model name or path. 

What I am thinking is we use the `hf_config.config` as kwargs when calling `from_pretrained` for both model and config load. https://github.com/huggingface/transformers/blob/main/src/transformers/configuration_utils.py#L435 
This meets the description given in the docs. 
However, there might be the case that `model_name` is not present. In which case we can just do one of:
- PretrainedConfig.from_dict`: not reliable since it returns a `PretrainedConfig` which might not have some fields. 
- Get `_name_or_path` from `hf_config.dict` and use it as model_name: probably the best option

Also there is no need to add `hf_config.config` if we already have model_name and model_path is None, in OnnxConversion.

jambayk: This PR https://github.com/microsoft/Olive/pull/520 adds `load_model` and `load_config_model` where we could also handle the above.

trajepl: > Also there is no need to add hf_config.config if we already have model_name and model_path is None, in OnnxConversion.

if we already have model_name and model_path is **not** None?

trajepl: Updated with auto_tune_config as we discussed offline.

examples/bert/bert_cuda_gpu.json
```diff
@@ -50,9 +50,6 @@
         "transformers_optimization": {
             "type": "OrtTransformersOptimization",
             "config": {
-                "model_type": "bert",
-                "num_heads": 12,
-                "hidden_size": 768,
                 "float16": true
             }
```
devang-ml, line 54: Presumably these are coming from the auto_tune_config  but I don't see it in the .json ?

olive/model/hf_utils.py
```diff
@@ -215,6 +215,7 @@ class HFConfig(ConfigBase):
     # TODO: remove model_class and only use task
     model_class: str = None
     components: List[HFComponent] = None
+    # TODO: double check if we need this field
     config: Dict[str, Any] = None
```
guotuofeng, line 219: remove it directly?

docs/source/overview/options.md
```diff
@@ -113,8 +113,8 @@ case insensitive.
...
```
guotuofeng, line 116: >config [](http://example.com/codeflow?start=11&length=6)

double confirm with @xiaoyu-work. I suppose we can remove it

olive/model/__init__.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         self.local_model_path = None
         self.composite_parent = None
         self.io_config = None
+        self.auto_tune_config = None
```
guotuofeng, line 66: >auto_tune_config [](http://example.com/codeflow?start=13&length=16)

@jambayk , should be better if we keep it globally like engine config instead of in model itself? 

"auto_tune": {
    "num_heads": 128,
    ""hidden_size: 64
}

examples/bert/bert_cuda_gpu.json
trajepl: It will fetch the config from huggingface model config firstly, if it is not found, the use the settings from auto_tune_config.

The progress is be like:
1. fuse auto_tune_config with huggingface model config(if have)
2. output model will inherit the configs from first steps.
3. pass to parse the required configs.

olive/passes/olive_pass.py
```diff
@@ -389,11 +389,21 @@ def run(
...
+                components.append(self.inherit_hf_config_from_input_model(model, output_model_components))
                 component_names.append(model.get_model_component_name(cidx))
             return CompositeOnnxModel(components, component_names, hf_config=model.hf_config)
         else:
-            return self._run_for_config(model, data_root, config, output_model_path)
+            output_model = self._run_for_config(model, data_root, config, output_model_path)
+            return self.inherit_hf_config_from_input_model(model, output_model)
+
+    def inherit_hf_config_from_input_model(self, input_model: OliveModel, output_model: OliveModel) -> OliveModel:
+        if hasattr(output_model, "hf_config") and not isinstance(output_model, PyTorchModel):
```
guotuofeng, line 401: >not isinstance(output_model, PyTorchModel): [](http://example.com/codeflow?start=50&length=43)

what about ONNXModel has hf_config?

examples/whisper/prepare_whisper_configs.py
```diff
@@ -74,10 +71,6 @@ def main(raw_args=None):
         "whisper_audio_decoder_dataloader" if not args.no_audio_decoder else "whisper_no_audio_decoder_dataloader"
     )
 
-    # update model specific values for transformer optimization pass
-    template_json["passes"]["transformers_optimization"]["config"]["num_heads"] = whisper_config.encoder_attention_heads
-    template_json["passes"]["transformers_optimization"]["config"]["hidden_size"] = whisper_config.d_model
```
guotuofeng, line 79: do we need remove the settings for whisper?

olive/passes/olive_pass.py
trajepl: Added todo for next step to fix.

examples/whisper/prepare_whisper_configs.py
trajepl: Yes, currently, we can parse this values from model configs.

guotuofeng (dismissed):

olive/model/hf_utils.py
trajepl: Done.

olive/passes/onnx/transformer_optimization.py
trajepl: Thanks. I do not realize we have the hf_mappings. Updated now.

olive/passes/onnx/transformer_optimization.py
```diff
@@ -140,6 +141,26 @@ def _run_for_config(
         for key in get_external_data_config():
             del run_config[key]
 
+        if model.hf_config:
+            model_config = model.hf_config.load_model_config().to_dict()
+            input_model_type = model_config.get("model_type", "")
+            _model_type = MODEL_TYPE_MAPPING.get(input_model_type, input_model_type)
+            assert _model_type in transformers_optimizer.MODEL_TYPES, (
```
jambayk, line 148: This assert might have to be moved after line 152. Otherwise, this assert will fail even though the user provide the value through `run_config["model_type"]`

trajepl: Thx! Updated.

xiaoyu-work (approved):

PR #519, opened 2023-08-29
🏉 Summarize pass run history

trajepl:
## Describe your changes

Summarize pass run history to one table which will be print in regular output and stored in output_dir/run_shistory_{eps}.txt
image

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -994,10 +994,9 @@ class Engine:
         host = self.host_for_pass(pass_id)
         if host.system_type != SystemType.AzureML:
             input_model = self._prepare_non_local_model(input_model)
+        run_start_time = datetime.now().timestamp()
         try:
-            run_start_time = datetime.now().timestamp()
             output_model = host.run_pass(p, input_model, data_root, output_model_path, pass_search_point)
-            run_end_time = datetime.now().timestamp()
         except OlivePassException as e:
             logger.error(f"Pass run_pass failed: {e}", exc_info=True)
             output_model = FAILED_CONFIG
@@ -1013,6 +1012,7 @@ class Engine:
             if self.no_search:
                 raise  # rethrow the exception if no search is performed
 
+        run_end_time = datetime.now().timestamp()
         # cache model
         self._cache_model(output_model, output_model_id)
 
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -6,6 +6,7 @@ import json
 import logging
 import time
 from collections import OrderedDict, defaultdict
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
@@ -379,6 +380,10 @@ class Engine:
                 raise
             except Exception as e:
                 logger.warning(f"Failed to run Olive on {accelerator_spec}: {e}", exc_info=True)
+
+        for eps in self.footprints.keys():
+            logger.info(f"Run history for {eps}:")
+            self.footprints[eps].summarize_run_history(output_dir / f"run_history_{eps}.json")
 
         if packaging_config:
             logger.info(f"Package top ranked {sum([len(f.nodes) for f in pf_footprints.values()])} models as artifacts")
@@ -990,7 +995,9 @@ class Engine:
         if host.system_type != SystemType.AzureML:
             input_model = self._prepare_non_local_model(input_model)
         try:
+            run_start_time = datetime.now().timestamp()
             output_model = host.run_pass(p, input_model, data_root, output_model_path, pass_search_point)
+            run_end_time = datetime.now().timestamp()
         except OlivePassException as e:
             logger.error(f"Pass run_pass failed: {e}", exc_info=True)
             output_model = FAILED_CONFIG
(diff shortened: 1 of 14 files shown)


olive/engine/engine.py
```diff
@@ -1019,6 +1026,8 @@ def _run_pass(
             parent_model_id=input_model_id,
             from_pass=pass_name,
             pass_run_config=pass_config,
+            start_time=run_start_time,
```
jambayk, line 1029: Maybe we should save the start and end times in the cached run json too? 

Otherwise, for cached runs, we won't have times for them. Like the record at line 966

requirements.txt
```diff
@@ -5,6 +5,7 @@ pandas
 pydantic<2.0.0
 protobuf<4.0.0
 pyyaml
+tabulate
```
guotuofeng, line 8: is it a hard requirement?

olive/engine/footprint.py
```diff
@@ -284,6 +284,35 @@ def plot_pareto_frontier(self, ranks=None, save_path=None, is_show=True, save_fo
...
+                model_id=model_id,
+                parent_model_id=node.parent_model_id,
+                from_pass=node.from_pass,
+                duration_sec=duration,
+                metrics=str(node.metrics.value) if node.metrics else None,
+            )
+            rls.append(tuple(run_history))
+        tab_rls = tabulate(rls, headers=headers, tablefmt="grid")
+        logger.info(f"run history:\n{tab_rls}")
+        with open(output_path, "w") as f:
```
guotuofeng, line 313: could we return the result and let the caller decide whether to print or write to file?

guotuofeng (changes requested):
better keep the dependency optional?

olive/engine/engine.py
trajepl: Done.

requirements.txt
trajepl: Removed.

olive/engine/footprint.py
trajepl: Updated.

olive/engine/engine.py
```diff
@@ -621,6 +627,21 @@ def get_pareto_frontier_footprints(
 
         return pf_footprints
 
+    def dump_run_history(self, run_history, output_path: str = None):
+        if not run_history:
+            logger.info("No run history to dump!")
```
jambayk, line 632: return here? otherwise it will fail in the next steps

olive/engine/engine.py
```diff
@@ -865,23 +890,34 @@ def _cache_run(
...
             except Exception as e:
                 logger.error(f"Failed to load run: {e}", exc_info=True)
-                output_model_id = None
-            return output_model_id
-        else:
-            return None
+                run_json = None
+        return run_json
+
+    def _load_output_model_id_from_run(
```
jambayk, line 909: looks like this method is not being used anywhere?

jambayk: Maybe we can combine this with `load_run_json` to return both output_model_id and run_json. if we still want to keep this logic around for future use.

olive/engine/engine.py
trajepl: good caceh. fixed.

olive/engine/engine.py
trajepl: Removed directly. Better to keep _load_run clear in my opinion. I changed to make _load_run return empty dict if _load_run failed.

olive/engine/engine.py
```diff
@@ -621,6 +627,22 @@ def get_pareto_frontier_footprints(
...
+            logger.info("No run history to dump!")
+            return
+        headers = run_history[0]._fields
+        try:
+            from tabulate import tabulate
+
+            formatted_rls = tabulate([tuple(rh) for rh in run_history], headers=headers, tablefmt="grid")
+            logger.info(f"run history:\n{formatted_rls}")
+        except ImportError:
+            logger.error("Please install tabulate for better run history output")
```
guotuofeng, line 641: warning?

olive/engine/engine.py
```diff
@@ -380,6 +381,11 @@ def run(
             except Exception as e:
                 logger.warning(f"Failed to run Olive on {accelerator_spec}: {e}", exc_info=True)
 
+        for eps in self.footprints.keys():
```
guotuofeng, line 384: ep instead of eps? I suppose the key is a single EP

olive/engine/engine.py
```diff
@@ -621,6 +627,22 @@ def get_pareto_frontier_footprints(
...
+        try:
+            from tabulate import tabulate
+
+            formatted_rls = tabulate([tuple(rh) for rh in run_history], headers=headers, tablefmt="grid")
+            logger.info(f"run history:\n{formatted_rls}")
+        except ImportError:
+            logger.error("Please install tabulate for better run history output")
+            formatted_rls = run_history
+        with open(output_path, "w") as f:
+            f.write(f"{formatted_rls}")
```
guotuofeng, line 644: if no tabulate installed, will the write work for the runhistory list?

jambayk:
LGTM as far as the parts I am concerned about. Please feel free to merge once Mike's comments are resolved and he approves. Thanks!

olive/engine/engine.py
jambayk: Since we are calling ` dump_run_history` for all runs and tabulate is optional, I think `info` might be better? Otherwise, we are giving error/warning over an optional dependency.

guotuofeng (approved):

PR #520, opened 2023-08-29
Add `model_loading_args`, `load_model` and `load_model_config` to `HFConfig`

jambayk:
## Describe your changes
Add a new field called `model_loading_args` to `HFConfig`. `transformers.PreTrainedConfig.from_pretrained` method can take many other keyword arguments. One such argument is `device_map` which can be used to load a large model to multiple gpus. This can be useful when evaluating accuracy metrics such as perplexity for large LLMs. 

New passes like QLoRA needs to save a hf model with quantization_config. The model is quantized during loading. 

Also added `load_model` and `load_model_config` methods to `HFConfig`. This way, we can use the hf_config directly instead of keep loading logic outside of it. I think we can also handle the point I brought up in [this discussion](https://github.com/microsoft/Olive/pull/518#discussion_r1309159989) in https://github.com/microsoft/Olive/pull/518 by expanding on these methods.

### DIFF
--- a/olive/model/hf_utils.py
+++ b/olive/model/hf_utils.py
@@ -38,7 +38,7 @@ class HFModelLoadingArgs(ConfigBase):
     torch_dtype: str = None
     # A map that specifies where each submodule should go. Refer to link in class docstring
     # str suffices for torch.dtype, otherwise serializer won't recognize it
-    device_map: Union[int, str, Dict[str, Union[int, str]]] = None
+    device_map: Union[int, str, Dict] = None
     # A dictionary device identifier to maximum memory
     max_memory: Dict = None
     # quant method to use (transformers.QuantizationMethod), e.g. "bitsandbytes", "gptq"
@@ -184,9 +184,9 @@ class HFConfig(ConfigBase):
         model_name_or_path = model_path or self.model_name
         loading_args = self.model_loading_args.get_loading_args() if self.model_loading_args else {}
         if self.task:
-            model = load_huggingface_model_from_task(self.task, model_name_or_path, loading_args)
+            model = load_huggingface_model_from_task(self.task, model_name_or_path, **loading_args)
         elif self.model_class:
-            model = load_huggingface_model_from_model_class(self.model_class, model_name_or_path, loading_args)
+            model = load_huggingface_model_from_model_class(self.model_class, model_name_or_path, **loading_args)
         return model
 
     def load_model_config(self, model_path: str = None):
@@ -196,7 +196,7 @@ class HFConfig(ConfigBase):
         return get_hf_model_config(model_name_or_path)
 
 
-def load_huggingface_model_from_task(task: str, name: str, kwargs: Dict[str, Any] = None):
+def load_huggingface_model_from_task(task: str, name: str, **kwargs):
     """Load huggingface model from task and name"""
     from transformers.pipelines import check_task
 
@@ -214,7 +214,6 @@ def load_huggingface_model_from_task(task: str, name: str, kwargs: Dict[str, Any
     class_tuple = class_tuple + model_class.get("pt", (AutoModel,))
 
     model = None
-    kwargs = kwargs or {}
     for model_class in class_tuple:
         try:
             model = model_class.from_pretrained(name, **kwargs)
@@ -250,7 +249,7 @@ def get_hf_model_config(model_name: str):
     return AutoConfig.from_pretrained(model_name)
 
 
-def load_huggingface_model_from_model_class(model_class: str, name: str, kwargs: Dict[str, Any] = None):
+def load_huggingface_model_from_model_class(model_class: str, name: str, **kwargs):
     """
     Load huggingface model from model_loader and name
     """
--- a/olive/model/hf_utils.py
+++ b/olive/model/hf_utils.py
@@ -10,7 +10,7 @@ from typing import Any, Callable, Dict, List, Optional, Union
 
 import torch
 import transformers
-from pydantic import validator
+from pydantic import Field, validator
 from transformers import AutoConfig, AutoModel, AutoTokenizer
 
 from olive.common.config_utils import ConfigBase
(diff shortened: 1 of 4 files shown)


olive/model/hf_utils.py
```diff
@@ -25,6 +27,137 @@ class HFComponent(ConfigBase):
     dummy_inputs_func: Union[str, Callable]
 
 
+class HFModelLoadingArgs(ConfigBase):
```
guotuofeng, line 30: kwargs?

olive/model/hf_utils.py
```diff
@@ -25,6 +27,137 @@ class HFComponent(ConfigBase):
...
+            "gptq": "4.32.0",
+        }
+        if quantization_method not in method_to_class_name:
+            raise ValueError(
+                f"Unsupported quantization method {quantization_method}. Supported methods are"
+                f" {list(method_to_class_name.keys())}"
+            )
+
+        try:
+            config_cls = getattr(transformers, method_to_class_name[quantization_method])
```
guotuofeng, line 142: >getattr [](http://example.com/codeflow?start=25&length=7)

getattr will not raise exception. what's the except scenario here?

olive/model/hf_utils.py
```diff
@@ -44,8 +178,25 @@ def task_or_model_class_required(cls, v, values):
...
+        return model
+
+    def load_model_config(self, model_path: str = None):
+        """Load model config from model_path or model_name"""
+        # TODO: handle self.config, probably need to pass it as kwargs
+        model_name_or_path = model_path or self.model_name
+        return get_hf_model_config(model_name_or_path)
+
+
+def load_huggingface_model_from_task(task: str, name: str, kwargs: Dict[str, Any] = None):
```
guotuofeng, line 199: >kwargs [](http://example.com/codeflow?start=59&length=6)

add **?

olive/model/hf_utils.py
```diff
@@ -98,11 +250,12 @@ def get_hf_model_config(model_name: str):
     return AutoConfig.from_pretrained(model_name)
 
 
-def load_huggingface_model_from_model_class(model_class: str, name: str):
+def load_huggingface_model_from_model_class(model_class: str, name: str, kwargs: Dict[str, Any] = None):
```
guotuofeng, line 253: >kwargs [](http://example.com/codeflow?start=73&length=6)

ditto to add double *

olive/model/hf_utils.py
jambayk: `BitsAndBytesConfig` and `GPTQConfig` are not present in old versions. So I am using getattr and try/except to check for import error

olive/model/hf_utils.py
jambayk: Yes, I can do that. I wasn't sure if we should except anything as `**kwargs` or have a dict the user to needs to pass.

olive/model/hf_utils.py
jambayk: Do you mean something like `HFModelLoadingKwargs`?

olive/model/hf_utils.py
```diff
@@ -25,6 +27,137 @@ class HFComponent(ConfigBase):
...
+    device_map: Union[int, str, Dict[str, Union[int, str]]] = None
+    # A dictionary device identifier to maximum memory
+    max_memory: Dict = None
+    # quant method to use (transformers.QuantizationMethod), e.g. "bitsandbytes", "gptq"
+    # must be provided if quantization_config is provided
+    quantization_method: str = None
+    # A dictionary of configuration parameters for quantization
+    quantization_config: Dict = None
+    # other kwargs
+    extra_args: Dict = None
```
guotuofeng, line 50: >extra_args [](http://example.com/codeflow?start=4&length=10)

what kinds of parameters that users need to know to be in extra_args or in named args in this class?

jambayk: There are many args here https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2074

olive/model/hf_utils.py
jambayk: Updated to use **kwargs instead dict param.

olive/model/hf_utils.py
jambayk: Updated

olive/model/hf_utils.py
guotuofeng: yeah, anyway you can ignore the comment

olive/model/hf_utils.py
guotuofeng: I suppose the old version will return None instead of raising exception for this case?

olive/model/hf_utils.py
guotuofeng: **kwargs is a dict actually

olive/model/hf_utils.py
jambayk: image

`getattr` raises AttributeError if a default value is not given.

olive/model/hf_utils.py
guotuofeng: What I mean is: should we add document about what kinds of args users need set by extra_args or by others like quantization_method? Since they seems not be consistent. 

For all the args defined in HFModelLoadingArgs, user can specify them directly. But for all other args, user need to set by using extra_args

olive/model/hf_utils.py
jambayk: Okay, thanks! I will keep it as it is for now. Since kwargs might give the idea that we can pass anything here. But here we need to pass `extra_args` dict if they want to pass more kwargs.

olive/model/hf_utils.py
jambayk: Sorry, I meant instead of asking for a dict called kwargs args to be passed line `func(kwargs)`, they can pass it like regular kwargs `func(**kwargs)`.

olive/model/hf_utils.py
guotuofeng: so, it seems the except is useless?

olive/model/hf_utils.py
```diff
@@ -25,6 +27,137 @@ class HFComponent(ConfigBase):
...
+        # copy args that can be directly copied
+        direct_copy_args = ["device_map", "max_memory"]
+        for arg in direct_copy_args:
+            if getattr(self, arg):
+                loading_args[arg] = deepcopy(getattr(self, arg))
+        # convert torch dtype to torch.dtype or "auto"
+        if self.torch_dtype:
+            loading_args["torch_dtype"] = self.get_torch_dtype()
+        # convert quantization_config to the config class
+        quantization_config = self.get_quantization_config()
```
guotuofeng, line 97: >quantization_config [](http://example.com/codeflow?start=8&length=19)

what about self.quantization_config has value?

olive/model/hf_utils.py
jambayk: Updated to use pydantic.Field with description

olive/model/hf_utils.py
```diff
@@ -25,6 +27,137 @@ class HFComponent(ConfigBase):
...
+                return getattr(torch, v)
+            except AttributeError as e:
+                raise ValueError(f"Invalid torch dtype {v}") from e
+        return v
+
+    def get_quantization_config(self):
+        if not self.quantization_method or not self.quantization_config:
+            return None
+        return self.dict_to_quantization_config(self.quantization_method, self.quantization_config)
+
```
guotuofeng, line 168: could we directly using self.get_quantization_config since the validation will call dict_to_quantization_config?

guotuofeng: in this way, the method seems can be removed?

olive/model/hf_utils.py
```diff
@@ -25,6 +27,137 @@ class HFComponent(ConfigBase):
...
+        # convert quantization_config to the config class
+        quantization_config = self.get_quantization_config()
+        if quantization_config:
+            loading_args["quantization_config"] = quantization_config
+        # add extra args
+        if self.extra_args:
+            loading_args.update(deepcopy(self.extra_args))
+        return loading_args
+
+    def get_torch_dtype(self):
```
guotuofeng, line 153: >get_torch_dtype [](http://example.com/codeflow?start=8&length=15)

could we merge the logic here with the validator?

olive/model/hf_utils.py
jambayk: validation actually calls `dict_to_quantization_config` but converts it back to dict at line 113. Otherwise, serialization and type checking becomes complicated. 

So I added this method to get the config object during model load time. Since `from_pretrained` doesn't actually work with dict for `gptq` quantization method.

olive/model/hf_utils.py
jambayk: the validator uses this helper function twice so I abstracted it. But I can make it a function inside the validator if you think it's better to not expose this method.

olive/model/hf_utils.py
guotuofeng: Actually, they are same regardless which way we use.

olive/model/hf_utils.py
jambayk: This is a similar case to the `get_quantization_config` method. `torch.dtype` is not directly serializable but `from_pretrained` needs the value as torch.dtype.

olive/model/hf_utils.py
jambayk: Resolving after offline conversion. tldr: there is no easy way to validate and serialize dataclasses (pydantic does things on top of them) + we don't have access to the parent class here for typing.

olive/model/hf_utils.py
jambayk: Similar comment to above

trajepl (approved):

PR #521, opened 2023-08-29
Batched tokenization for line-by-line. Manual special tokens for line-by-line, line-by-line-random

jambayk:
## Describe your changes
Batch tokenization for `line_by_line` corpus strategy. This can be helpful if the dataset is large and the tokenizer has an optimized batched tokenization. 

BOS and EOS tokens are added manually since some tokenizers (like llama tokenizer) don't add eos tokens. 
Refer to https://github.com/huggingface/transformers/issues/22794 for some background

### DIFF
--- a/olive/data/component/text_generation.py
+++ b/olive/data/component/text_generation.py
@@ -65,7 +65,7 @@ class TextGenParams(ConfigBase):
     # if false, cannot gaurantee all sequences are same length. data loader will have to handle this during collation
     pad_to_max_len: bool = True  # pad sequences to max_len, ignored for JOIN corpus strategy
     drop_short_sequences: bool = False  # drop sequences shorter than max_len. Mutually exclusive with pad_to_max_len
-    add_special_tokens: bool = True  # add special tokens, ignored for JOIN corpus strategy
+    add_special_tokens: bool = True  # add bos and eos tokens
 
     @validator("drop_short_sequences", always=True)
     def _check_padding(cls, v, values):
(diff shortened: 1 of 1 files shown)


olive/data/component/text_generation.py
```diff
@@ -259,7 +272,7 @@ def text_gen_corpus_pre_process(_dataset, tokenizer, all_kwargs):
                             max_length=args.source_max_len,
                             truncation=True,
                             padding="max_length" if args.pad_to_max_len else False,
-                            add_special_tokens=args.add_special_tokens,
+                            add_special_tokens=False,
```
guotuofeng, line 275: >False [](http://example.com/codeflow?start=47&length=5)

the parent else clause seems aligns with args.corpus_strategy == TextGenCorpusStrategy.LINE_BY_LINE

jambayk: Line 225 applied to both LINE_BY_LINE and LINE_BY_LINE_RANDOM since it's before the if/else. So, the behavior doesn't change.

jambayk: Updated PR title to make this clear

olive/data/component/text_generation.py
```diff
@@ -65,7 +65,7 @@ class TextGenParams(ConfigBase):
     # if false, cannot gaurantee all sequences are same length. data loader will have to handle this during collation
     pad_to_max_len: bool = True  # pad sequences to max_len, ignored for JOIN corpus strategy
     drop_short_sequences: bool = False  # drop sequences shorter than max_len. Mutually exclusive with pad_to_max_len
-    add_special_tokens: bool = True  # add special tokens, ignored for JOIN corpus strategy
+    add_special_tokens: bool = True  # add bos and eos tokens
```
trajepl, line 68: Do we need to split it as:
add_bos_tokens
add_eos_tokens
?

As you said, for llama model, the eos_token_id is missed. But for bos? Should we handle the bos and eos separately?

jambayk: I wasn't sure about whether we should separate the two. since for text-generation, we normally add both tokens.

jambayk: we pass `add_special_tokens=False` when calling the tokenizer. so, there won't be any duplicates. 

I can separate them into bos and eos. Will give us better control over this in the future if needed.

trajepl (approved):

PR #523, opened 2023-08-30
Add perf check pipeline

xiaoyu-work:
## Describe your changes

Add perf check pipeline.

- 4 models supported:
   - Intel/bert-base-uncased-mrpc
   - microsoft/deberta-base-mnli
   - distilbert": "distilbert-base-uncased-finetuned-sst-2-english
   - roberta_large": "roberta-large-mnli
- The pipeline will be run at 6:00 am every Friday.
- Each metric will be run 10 times and print the average.
- You can manually run it by `python run_performance_check.py --model_name bert`
- Pipeline link: https://aiinfra.visualstudio.com/Model%20optimization%20Toolkit/_build?definitionId=1265&_a=summary


Makefile
```diff
@@ -5,20 +5,24 @@ EXAMPLE_FOLDER             ?=
 EXAMPLE_NAME               ?=
 INSTALL_EXTRAS             ?=
 VERSION                    ?=
+MODEL_NAME				   ?=
```
guotuofeng, line 8: >				    [](http://example.com/codeflow?start=10&length=7)

use whitespace instead of tabs?

.azure_pipelines/performance.yaml
```diff
@@ -0,0 +1,41 @@
+schedules:
+- cron: 0 6 * * 5
```
trajepl, line 2: Do you have a quick estimation about the time costs for this perf_check pipeline?  Will the 0 6 * * 5 make our CI job crowed? How about put it in weekend? Like 0 0 * * 0, or 0 0 * * 6

Makefile
xiaoyu-work: Good catch. Fixed.

.azure_pipelines/job_templates/olive-performance-template.yaml
```diff
@@ -0,0 +1,46 @@
+parameters:
+  model_name: ''
+  pool: ''
+  device: 'cpu'
+
+jobs:
+- job: ${{ parameters.device}}_Model_Performance
+  timeoutInMinutes: 300
+  pool:
+    name: ${{ parameters.pool}}
```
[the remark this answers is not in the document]
xiaoyu-work: Good catch. Fixed.

.azure_pipelines/performance.yaml
xiaoyu-work: Roberta large takes the longest time to run which is around 2 hours https://aiinfra.visualstudio.com/Model%20optimization%20Toolkit/_build/results?buildId=347589&view=results. I set this pipeline to run on Friday morning to let the team can check this metric on the last workday of a week.

.azure_pipelines/performance_check/configs/bert_gpu.json
```diff
@@ -0,0 +1,85 @@
+{
```
devang-ml, line 1: How is this different from examples/bert/bert_cuda_gpu.json ?

xiaoyu-work: This is same as bert_cuda_gpu in our example.

.azure_pipelines/performance_check/run_performance_check.py
```diff
@@ -0,0 +1,245 @@
...
+    "distilbert": {"model_name": "distilbert-base-uncased-finetuned-sst-2-english", "task": "text-classification"},
+    "roberta_large": {"model_name": "roberta-large-mnli", "task": "text-classification"},
+}
+
+ACC_METRIC = {
+    "name": "accuracy",
+    "type": "accuracy",
+    "backend": "huggingface_metrics",
+    "sub_types": [{"name": "accuracy", "priority": 1, "goal": {"type": "max-degradation", "value": 0.01}}],
+    "user_config": {
```
jambayk, line 44: For the metrics, it should be possible to add the data_configs directly to the metrics so that we can use the data container directly without user scripts? Similar to https://github.com/microsoft/Olive/blob/main/examples/open_llama/open_llama_sparsegpt_gpu.json#L59. 

That way, we don't need to maintain user scripts for each model. Just need the data configs which can be extracted from the input model in the olive workflow json.

jambayk: `huggingface_data_config_template(model_name, task, **kwargs).to_json()` will give us the data_config json to use in the template.

xiaoyu-work: Good call. I can add data config to model config itself.

olive/model/__init__.py
```diff
@@ -588,7 +588,7 @@ def prepare_session(
         execution_providers: Union[str, List[str]] = None,
         rank: Optional[int] = None,
     ):
-        return self.load_model().eval()
+        return self.load_model(device).eval()
```
jambayk, line 591: I think it might be better to do the `to("cuda")` directly here instead of passing it to `load_model`. In this code, the `device` is only used when using hf_config to load the model. 
Also think it's better to not add `device` argument to the hf model loader func.

.azure_pipelines/performance_check/run_performance_check.py
```diff
@@ -0,0 +1,282 @@
...
+                    olive_config["input_model"]["config"]["model_path"] = str(
+                        Path(model_root_path / optimized_model / "model_quantized.onnx")
+                    )
+                elif optimized_model in ["optimum_o1", "optimum_o2", "optimum_o3", "optimum_o4"]:
+                    olive_config["input_model"]["config"]["model_path"] = str(
+                        Path(model_root_path / optimized_model / "model_optimized.onnx")
+                    )
+                elif optimized_model == "hf_pytorch":
+                    olive_config["input_model"]["type"] = "PyTorchModel"
+                    hf_config = {"hf_config": MODEL_NAME_TO_CONFIG_MAP[model_name]}
```
jambayk, line 203: I am interested in how the code works as expected since the hf_config is only being added for this case. For the other cases, there is no hf_config but the metrics are still working fine (when they don't have data_config to begin with). 

Maybe I am missing something here

jambayk: oh, looks like it might be a lucky instance where the `ACC_METRIC` and `LAT_METRIC` objects are modified by the first iteration. 
since the first iteration is `hf_pytorch`, the data config gets auto inserted from the input model and the other iterations use this. 

To be safe and correct, I think it might be better to update the metrics beforehand with the data_configs with something like:
```
ACC_METRIC["data_config"] = huggingface_config_template(model_name, task, **dataset)
```
That way the metrics already have the data config and don't rely on it being inserted by the first iteration.

olive/model/__init__.py
xiaoyu-work: The reason I didn't add "to("cuda")" here is I'm not sure how other `load_model` works and if they support `to("cuda")`, especially loading from module_loader.

.azure_pipelines/performance_check/run_performance_check.py
xiaoyu-work: Good catch! I also noticed the metrics were different since this change, but I'm not sure if this is related how we feed data to metric (theoretically no but this is where the difference starts). I'll update this part logic later.

olive/model/__init__.py
jambayk: btw, why is this change need in this PR? We do the `to("cuda")` step for the evaluator here https://github.com/microsoft/Olive/blob/main/olive/evaluator/olive_evaluator.py#L687 

I think we can assume that the loaded data is a torch model so they should have `to("cuda")` support.

.azure_pipelines/performance_check/run_performance_check.py
jambayk: During the workflow run validation we auto insert the data config from the input model if the input model has hf_config.dataset and the metric's `data_config` is `None`. 

Because the metrics added by reference, they are getting updated globally during the first run.

xiaoyu-work: It may cause longer Olive workflow run time but this shouldn't affect latency metric. Not sure why latency changes.

jambayk: I don't think the runtime gets affected since both accuracy and latency require data configs if there is no dataloader func. Also, both metrics will have the data_config auto-inserted. 

latency metric can try to get io_config from the onnx model itself but since they have dynamic shape, this would not work.

olive/model/__init__.py
jambayk: Maybe, we can do a try/except with the `to("cuda")` in the prepare session to be safe.

xiaoyu-work: Yeah you are right I just double confirmed huggingface supports both https://github.com/huggingface/optimum/issues/336. I'll remove this device switch logic since we can do this in evaluator.

jambayk: Thanks for checking!

guotuofeng, line 591: I think the model.load_model signature doesn't need to be changed after the model is loaded just by changing as the following line.
```python
model = self.load_model()
model.to("cuda") if device == Device.GPU else "cpu"
return model.eval()

```

guotuofeng: with above change, the hf_utils.py doesn't need to be changed since currently it only handle hf models.

guotuofeng (changes requested):
the PyTorchModel.load_model need not the device parameters since it currently only handle hf model and all other models is ignoring this paramater.

olive/model/__init__.py
guotuofeng: Another consideration is to make the load_model signature be consistent with other classes.

xiaoyu-work: I and Jambay had a similar discussion above. We decide to revert this change since evaluator applies device and it works for both hf model and pytorch model.

.azure_pipelines/performance_check/run_performance_check.py
```diff
@@ -0,0 +1,294 @@
...
+
+                olive_config["systems"]["local_system"]["config"]["accelerators"] = (
+                    ["cpu"] if device == "cpu" else ["gpu"]
+                )
+                olive_config["engine"]["cache_dir"] = str(Path(model_root_path / optimized_model / "cache"))
+                olive_config["engine"]["output_dir"] = str(Path(model_root_path / optimized_model / "output"))
+                olive_config["engine"]["execution_providers"] = (
+                    ["CPUExecutionProvider"] if device == "cpu" else ["CUDAExecutionProvider"]
+                )
+                olive_config["evaluators"]["common_evaluator"]["metrics"].append(ACC_METRIC)
```
jambayk, line 220: Thanks for the changes!
This is not a big issue currently but maybe it is safer if we deepcopy `hf_config`, `ACC_METRIC` and `LATENCY_METRIC` so that we are using fresh values for each iteration? Since these values are passed by reference and updated by the workflow run. 
It might be hard to debug in the future if the workflow run/validation makes unexpected changes to these values.

.azure_pipelines/performance_check/requirements-cpu.txt
jambayk, line 1: This PR which checks and fixes requirements files just merged https://github.com/microsoft/Olive/pull/548 
So it's failing the formatting test.

xiaoyu-work: Fixed.

.azure_pipelines/performance_check/run_performance_check.py
xiaoyu-work: Sounds good to me. I'll copy metrics here.

jambayk (approved):
Thanks! LGTM. Please feel free to merge once Mike is also happy with the changes and approves.

guotuofeng (dismissed):

.azure_pipelines/performance.yaml
```diff
@@ -0,0 +1,43 @@
+trigger: none
```
jambayk, line 1: BTW, since this pipeline takes time to run and we only plan to run this once a week.
should we turn of the pr trigger too? Like here? https://github.com/microsoft/Olive/blob/main/.azure_pipelines/build-doc.yaml#L12

jambayk (changes requested):
Waiting to resolve the discussion about pr trigger.

.azure_pipelines/performance.yaml
jambayk: Or add some branch and path filters to only trigger this PR when the relevant files in the `.azure_pipelines` subdir are updated by a PR

jambayk (approved):

PR #524, opened 2023-08-30
Add `adapter_path` to `PyTorchModel`

jambayk:
## Describe your changes
Some pytorch models such as huggingface models might have adapters, such as LoRA weights (https://huggingface.co/docs/peft/conceptual_guides/lora). This PR adds `adapter_path` to `PyTorchModel` which will be used to add the adapters to the model during load. Currently, we only support hugging face `peft` adapters. 

`adapter_path` can be a resource path like `model_path`. 

Updated the cache, docker system and azureml systems to handle adapter_path correctly. Refactored some of the code for `model_path` into helper methods so that we can reuse them for handling different resource path attributes. For now, we only have `model_path` and `adaptor_path`.

### DIFF
--- a/olive/model/__init__.py
+++ b/olive/model/__init__.py
@@ -9,7 +9,7 @@ import tempfile
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from pathlib import Path
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 
 import onnx
 import torch
@@ -31,6 +31,7 @@ from olive.resource_path import (
     ResourcePathConfig,
     ResourceType,
     create_resource_path,
+    resolve_local_resource,
 )
 from olive.snpe import SNPEDevice, SNPEInferenceSession, SNPESessionOptions
 from olive.snpe.tools.dev import get_dlc_metrics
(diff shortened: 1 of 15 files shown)


olive/model/__init__.py
```diff
@@ -513,6 +588,12 @@ def load_model(self, rank: int = None) -> torch.nn.Module:
             else:
                 raise ValueError(f"Unsupported model file format: {self.model_file_format}")
 
+        # we only have peft adapters for now
+        if self.adapter_path:
+            from peft import PeftModel
+
+            model = PeftModel.from_pretrained(model, self.adapter_path)
```
guotuofeng, line 595: are there any common way to handle this? This seems a little bit hacky. 

For these model loading need special handling, could we leverage the customer model loader by preparing couple of built-in loader functions?

guotuofeng:
>         model_loader: Union[str, Callable] = None,

basically, the customized model loader is used for any special case when loading model

---
Refers to: olive/model/__init__.py:498 in 0f6e50b. [](commit_id = 0f6e50b285bfcb535bd01fbc7b289eafb3a3af90, deletion_comment = False)

guotuofeng (changes requested):
Basically, adding an adapter for pytoch model and changing so many files seems be counter intuitive.

olive/model/__init__.py
devang-ml: Adapters are gaining wider adaption for LLM use cases.

We need to be prepared to support following scenarios in Olive:
- [PT Model, Adapter Weights] -> [PT Pass] -> [Merge Adapter weights] -> [ONNX Conversion] -> [ONNX Model] -> ...
- [PT Model, Adapter Weights] -> [PT Pass] -> [ONNX Conversion] -> [ONNX Model, ONNX Adapter Weights] -> ...
- [PT Model] -> [PT Pass] -> [PT Model, Adapter Weights] -> ....
- [ONNX Model] -> [ONNX Pass] -> [ONNX Model, Adapter Weights] -> ...

olive/model/__init__.py
```diff
@@ -56,68 +57,61 @@ def __init__(
...
         self.composite_parent = None
         self.io_config = None
+        # store resource paths
+        self.resource_paths = {}
+        # this is for storing local instances of resource paths
+        self.local_resource_paths = {}
+        resources = resources or {}
+        resources["model_path"] = model_path
+        for resource_name, resource_path in resources.items():
+            self.resource_paths[resource_name] = create_resource_path(resource_path)
```
guotuofeng, line 74: >resource_path [](http://example.com/codeflow?start=70&length=13)

need we check whether the resource_path is None or not?

jambayk: since all resources are optional, we don't need to check for None. `resource_paths` keep tracks of all resources even though the values are None. 
create_resource_path can handle None also

olive/model/__init__.py
```diff
@@ -56,68 +57,61 @@ def __init__(
...
-            self.model_resource_path.save_to_dir(download_path, overwrite=overwrite)
-        )
-        return self.local_model_path.get_path()
+        assert resource_name in self.resource_paths, f"{resource_name} is not a valid resource name."
+        resource_path = self.resource_paths[resource_name]
+        local_resource_path = self.local_resource_paths[resource_name]
+        local_path = resolve_local_resource(resource_path, local_resource_path)
+        if not local_path and resource_path:
+            raise ValueError(
+                f"{resource_name} is not local and not downloaded yet. Please call download_model() or"
```
guotuofeng, line 96: >download_model [](http://example.com/codeflow?start=83&length=14)

download_model is removed?

jambayk: Yes, like we discussed offline, the `download_model` method was never used. We also want to handle the model upload,download outside this model, so I removed this method.

test/integ_test/aml_model_test/test_aml_model.py
```diff
@@ -54,7 +54,7 @@ def test_aml_model_download():
     tmp_dir = tempfile.TemporaryDirectory()
     tmp_dir_path = Path(tmp_dir.name).resolve()
 
-    download_path = pytorch_model.download_model(tmp_dir_path)
```
guotuofeng, line 57: the method is removed

jambayk: removed the test as discussed offline

olive/model/__init__.py
```diff
@@ -461,11 +460,12 @@ def __init__(
         model_path: OLIVE_RESOURCE_ANNOTATIONS = None,
         model_file_format: ModelFileFormat = ModelFileFormat.PYTORCH_ENTIRE_MODEL,
         model_loader: Union[str, Callable] = None,
-        model_script: Union[str, Path] = None,
-        script_dir: Union[str, Path] = None,
```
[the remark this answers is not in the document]
jambayk: changed the type hits back to str | Path and added some validation in init.

guotuofeng (approved):

PR #533, opened 2023-09-01
🎏 Flag to identify model cannot be load by hf even if hf_config is not None

trajepl:
## Describe your changes

For torch model, if hf_config is not None, the loader will fall into huggingface model loader even if the model is local pytorch model(e.g, the output of TensorRTConversion).

Then we add a flag to identify model cannot be load by hf even if hf_config is not None. This flag is set by pass specifically. 

### DIFF
--- a/olive/model/__init__.py
+++ b/olive/model/__init__.py
@@ -466,6 +466,7 @@ class PyTorchModel(OliveModel):
         dummy_inputs_func: Union[str, Callable] = None,
         hf_config: Union[Dict[str, Any], HFConfig] = None,
         adapter_path: OLIVE_RESOURCE_ANNOTATIONS = None,
+        is_complete_hf_folder: bool = True,
     ):
         if not (
             isinstance(model_loader, Callable)
@@ -500,6 +501,10 @@ class PyTorchModel(OliveModel):
 
         # huggingface config
         self.hf_config = validate_config(hf_config, HFConfig) if hf_config else None
+        # if model_path is huggingface model name or complete huggingface model structure which
+        # contains model, config, tokenizer, and optionally adapter. That can be loaded by
+        # transformers.XXXForXXX.from_pretrained(model_path, ...)
+        self.is_complete_hf_folder = is_complete_hf_folder
 
     @property
     def script_dir(self) -> str:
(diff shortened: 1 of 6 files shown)


olive/model/__init__.py
```diff
@@ -466,6 +466,7 @@ def __init__(
         dummy_inputs_func: Union[str, Callable] = None,
         hf_config: Union[Dict[str, Any], HFConfig] = None,
         adapter_path: OLIVE_RESOURCE_ANNOTATIONS = None,
+        is_complete_hf_folder: bool = True,
```
guotuofeng, line 469: >is_complete_hf_folder [](http://example.com/codeflow?start=8&length=21)

could we use hf_config for this case by making it be Union[bool, Dict, HFConfig]?

guotuofeng: @jambayk

guotuofeng (changes requested):
let's wait comments from @jambayk

devang-ml (commented):
If a model is no longer a complete HF model then what's the use of hf_config ? Should the TorchTRTConversion pass set the hf_config to None ?

guotuofeng:
> If a model is no longer a complete HF model then what's the use of hf_config ? Should the TorchTRTConversion pass set the hf_config to None ?

@devang-ml, the reason for assigning hf_config to not None value is because transformer optimization pass would like to use original model config like num_heads/hidden_size for the converted onnx model if it is coming from HF model.

@trajepl / @jambayk 
There are several approach for current scenario:
1. save the TorchTRTConversion output as PYTORCH_TORCH_SCRIPT and when loading model choose torch.jit.load if the format is torch script and ignore the hf_config. In this way, we still use hf_config to contains the model config, which seems still not perfect since the hf_config is used for two scenarios. 1), load the pytorch model, 2), provide the model config. It seems go against https://en.wikipedia.org/wiki/Single-responsibility_principle.
2. Another solution is add a dedicated model_config in OliveModel so all kinds of model could leverage it in future. The responsibility boundary is clear: it is only for providing the model configuration regardless whether it is for model loading or other purpose.

Please let me know your feedbacks

guotuofeng:
@jambayk / @xiaoyu-work , 

What's the reason that adding hf_config in ONNX model in PR https://github.com/microsoft/Olive/pull/258?

trajepl:
> 2\. dedicated

Prefer option #2. For separated model config, I would like to complete it automatically if the model is from huggingface. More like the hf_config.config is previous discussion. 
And the separated method can keep the logic clearer.

PR #534, opened 2023-09-01
Add `ConfigWithExtraArgs`

jambayk:
## Describe your changes
This PR adds a new config base class called `ConfigWithExtraArgs` where the config values not defined in the class fields are collected in a field called `extra_args`. This behaves similar to `**kwargs` in functionality. It is useful for config classes where we expose some parameters but still want to accept extra arguments. The child classes can add their own validators on `extra_args` if needed.

### DIFF
--- a/olive/model/hf_utils.py
+++ b/olive/model/hf_utils.py
@@ -13,7 +13,7 @@ import transformers
 from pydantic import Field, validator
 from transformers import AutoConfig, AutoModel, AutoTokenizer
 
-from olive.common.config_utils import ConfigBase
+from olive.common.config_utils import ConfigBase, ConfigWithExtraArgs
 from olive.model.hf_mappings import MODELS_TO_MAX_LENGTH_MAPPING, TASK_TO_FEATURE
 from olive.model.model_config import IOConfig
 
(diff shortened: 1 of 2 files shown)


olive/model/hf_utils.py
```diff
@@ -27,7 +27,7 @@ class HFComponent(ConfigBase):
     dummy_inputs_func: Union[str, Callable]
 
 
-class HFModelLoadingArgs(ConfigBase):
+class HFModelLoadingArgs(ConfigWithExtraArgs):
```
xiaoyu-work, line 30: Do we have any other use case than `HFModelLoadingArgs`?

olive/model/hf_utils.py
```diff
@@ -78,8 +78,8 @@ class HFModelLoadingArgs(ConfigBase):
     extra_args: Dict = Field(
```
xiaoyu-work, line 78: This can be removed.

olive/model/hf_utils.py
jambayk: Yes, my qlora pass uses this feature too https://github.com/microsoft/Olive/blob/jambayk/qlora-dev/olive/passes/pytorch/qlora.py#L35. 

I can envision this being useful in the future too. Since, it sometimes feels harder to pass kwargs as a dict. Since we need to know what are defined fields and what are extra args.

olive/model/hf_utils.py
jambayk: I kept this since I wanted to override the description of the field. The behavior remains the same.

xiaoyu-work (approved):

PR #542, opened 2023-09-06
Ⓜ️ Model Attributes

trajepl:
## Describe your changes
1. introduce a new field for Olive Model: `model_attributes`
    - derived `model_attributes` from`hf_config` if there is `hf_config`
    - olive passes will inherit `model_attributes` from input_model. (`model.set_model_attributes(input_model.model_attributes)`)
2. remove hf_config from onnx related model

### DIFF
--- a/olive/model/__init__.py
+++ b/olive/model/__init__.py
@@ -57,12 +57,14 @@ class OliveModel(ABC):
         framework: Framework,
         model_file_format: ModelFileFormat,
         model_path: OLIVE_RESOURCE_ANNOTATIONS = None,
-        # resources other than model_path
+        # resources other than model_path, e.g.: path to model adapter
         resources: Optional[Dict[str, OLIVE_RESOURCE_ANNOTATIONS]] = None,
+        model_attributes: Optional[Dict[str, Any]] = None,
     ):
         self.framework = framework
         self.model_file_format = model_file_format
         self.composite_parent = None
+        self.model_attributes = model_attributes or {}
         self.io_config = None
         # store resource paths
         self.resource_paths: Dict[str, ResourcePath] = {}
@@ -79,6 +81,15 @@ class OliveModel(ABC):
     def model_path(self) -> str:
         """Return local model path."""
         return self.get_local_resource("model_path")
+
+    def extend_model_attributes(self, model_attr) -> Dict[str, Any]:
+        # self.model_attributes > model_attr, should respect self.model_attributes when there is conflict
+        # this method is mainly called by resulted model to extend model attributes from input_model.
+        # also resulted model config > input model config
+        self.model_attributes = self.model_attributes or {}
+        model_attr = model_attr or {}
+        model_attr.update(self.model_attributes)
+        self.model_attributes = model_attr
 
     def get_local_resource(self, resource_name: str) -> str:
         """
@@ -153,6 +164,7 @@ class OliveModel(ABC):
                 for resource_name, resource_path in self.resource_paths.items()
             },
         }
+        config["config"].update({"model_attributes": self.model_attributes})
         return serialize_to_json(config, check_object)
 
 
@@ -180,8 +192,14 @@ class ONNXModelBase(OliveModel):
         model_path: OLIVE_RESOURCE_ANNOTATIONS = None,
         inference_settings: Optional[dict] = None,
         use_ort_extensions: bool = False,
+        model_attributes: Optional[Dict[str, Any]] = None,
     ):
-        super().__init__(framework=Framework.ONNX, model_file_format=ModelFileFormat.ONNX, model_path=model_path)
+        super().__init__(
+            framework=Framework.ONNX,
+            model_file_format=ModelFileFormat.ONNX,
+            model_path=model_path,
+            model_attributes=model_attributes,
+        )
         self.inference_settings = inference_settings
         self.use_ort_extensions = use_ort_extensions
 
@@ -222,19 +240,18 @@ class ONNXModel(ONNXModelBase):
         onnx_file_name: Optional[str] = None,
         inference_settings: Optional[dict] = None,
         use_ort_extensions: bool = False,
-        hf_config: Union[Dict[str, Any], HFConfig] = None,
+        model_attributes: Optional[Dict[str, Any]] = None,
     ):
         super().__init__(
             model_path=model_path,
             inference_settings=inference_settings,
             use_ort_extensions=use_ort_extensions,
+            model_attributes=model_attributes,
         )
         self.onnx_file_name = onnx_file_name
         self.io_config = None
         self.graph = None
         self.all_graphs: Optional[List[GraphProto]] = None
-        # huggingface config
-        self.hf_config = validate_config(hf_config, HFConfig) if hf_config else None
 
         # if model_path is local folder, check for onnx file name
         model_resource_path = self.resource_paths["model_path"]
@@ -389,7 +406,7 @@ class ONNXModel(ONNXModelBase):
                 "onnx_file_name": self.onnx_file_name,
                 "inference_settings": self.inference_settings,
                 "use_ort_extensions": self.use_ort_extensions,
-                "hf_config": self.hf_config,
+                "model_attributes": self.model_attributes,
             }
         )
         return serialize_to_json(config, check_object)
@@ -466,6 +483,7 @@ class PyTorchModel(OliveModel):
         dummy_inputs_func: Union[str, Callable] = None,
         hf_config: Union[Dict[str, Any], HFConfig] = None,
         adapter_path: OLIVE_RESOURCE_ANNOTATIONS = None,
+        model_attributes: Optional[Dict[str, Any]] = None,
     ):
         if not (
             isinstance(model_loader, Callable)
@@ -485,6 +503,15 @@ class PyTorchModel(OliveModel):
             model_path=model_path,
             resources={"adapter_path": adapter_path, "script_dir": script_dir, "model_script": model_script},
         )
+        self.hf_config = validate_config(hf_config, HFConfig) if hf_config else None
+        if self.hf_config:
+            # priority: model_attributes < hf_config
+            # need call get_hf_model_config() after self.model_path is set
+            hf_model_config = self.get_hf_model_config().to_dict()
+            hf_model_config.update(model_attributes or {})
+            model_attributes = hf_model_config
+        self.model_attributes = model_attributes
+
         # ensure that model_script and script_dirs are local
         for resource_name in ["script_dir", "model_script"]:
             if self.resource_paths[resource_name]:
@@ -497,9 +524,6 @@ class PyTorchModel(OliveModel):
         self.dummy_inputs_func = dummy_inputs_func
 
         self.dummy_inputs = None
-
-        # huggingface config
-        self.hf_config = validate_config(hf_config, HFConfig) if hf_config else None
 
     @property
     def script_dir(self) -> str:
@@ -640,7 +664,7 @@ class PyTorchModel(OliveModel):
 
         return dummy_inputs
 
-    def get_model_config(self):
+    def get_hf_model_config(self):
         if self.hf_config is None:
             raise ValueError("HF model_config is not available")
         return self.hf_config.load_model_config(self.model_path)
@@ -699,6 +723,7 @@ class PyTorchModel(OliveModel):
                 "io_config": self.io_config,
                 "dummy_inputs_func": self.dummy_inputs_func,
                 "hf_config": self.hf_config,
+                "model_attributes": self.model_attributes,
             }
         )
         # convert script_dir and model_script to string
@@ -731,8 +756,14 @@ class SNPEModel(OliveModel):
         output_names: List[str],
         output_shapes: List[List[int]],
         model_path: OLIVE_RESOURCE_ANNOTATIONS = None,
+        model_attributes: Optional[Dict[str, Any]] = None,
     ):
-        super().__init__(framework=Framework.SNPE, model_file_format=ModelFileFormat.SNPE_DLC, model_path=model_path)
+        super().__init__(
+            framework=Framework.SNPE,
+            model_file_format=ModelFileFormat.SNPE_DLC,
+            model_path=model_path,
+            model_attributes=model_attributes,
+        )
         self.io_config = {
             "input_names": input_names,
             "input_shapes": input_shapes,
@@ -771,8 +802,14 @@ class TensorFlowModel(OliveModel):
         self,
         model_path: OLIVE_RESOURCE_ANNOTATIONS = None,
         model_file_format: ModelFileFormat = ModelFileFormat.TENSORFLOW_SAVED_MODEL,
+        model_attributes: Optional[Dict[str, Any]] = None,
     ):
-        super().__init__(model_path=model_path, framework=Framework.TENSORFLOW, model_file_format=model_file_format)
+        super().__init__(
+            model_path=model_path,
+            framework=Framework.TENSORFLOW,
+            model_file_format=model_file_format,
+            model_attributes=model_attributes,
+        )
 
     def load_model(self, rank: int = None):
         raise NotImplementedError()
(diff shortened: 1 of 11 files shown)


olive/model/__init__.py
```diff
@@ -80,6 +82,10 @@ def model_path(self) -> str:
         """Return local model path."""
         return self.get_local_resource("model_path")
 
+    def set_model_attributes(self, model_attr) -> Dict[str, Any]:
+        self.model_attributes = self.model_attributes or {}
+        self.model_attributes.update(model_attr or {})
```
guotuofeng, line 87: it is needless from the time being?

olive/model/__init__.py
```diff
@@ -497,9 +511,7 @@ def __init__(
         self.dummy_inputs_func = dummy_inputs_func
 
         self.dummy_inputs = None
-
-        # huggingface config
-        self.hf_config = validate_config(hf_config, HFConfig) if hf_config else None
+        self.set_model_attributes(model_attributes)
```
guotuofeng, line 514: >self.set_model_attributes(model_attributes) [](http://example.com/codeflow?start=8&length=43)

passing by init parameters?

olive/systems/azureml/aml_pass_runner.py
```diff
@@ -102,6 +102,7 @@ def main(raw_args=None):
 
     # run pass
     output_model = p.run(input_model, None, output_model_path)
+    output_model.set_model_attributes(input_model.model_attributes)
```
guotuofeng, line 105: >output_model.set_model_attributes(input_model.model_attributes) [](http://example.com/codeflow?start=4&length=63)

needless? I suppose the p.run will automatically propagate the model config?

guotuofeng:
>         model_json["config"]["hf_config"] = input_model_config["config"]["hf_config"]

do we need this?

---
Refers to: olive/systems/azureml/aml_pass_runner.py:112 in 8378d8e. [](commit_id = 8378d8ecbdd3853ab7d19e76017f53e62a3c04b5, deletion_comment = False)

guotuofeng (changes requested):
rebase from main

olive/model/__init__.py
trajepl: Updated

olive/model/__init__.py
trajepl: Updated

olive/systems/azureml/aml_pass_runner.py
trajepl: Yes, removed.

trajepl:
> > ```
> >     model_json["config"]["hf_config"] = input_model_config["config"]["hf_config"]
> > ```
> 
> do we need this?
> 
> Refers to: olive/systems/azureml/aml_pass_runner.py:112 in [8378d8e](https://github.com/microsoft/Olive/commit/8378d8ecbdd3853ab7d19e76017f53e62a3c04b5). [](commit_id = [8378d8e](https://github.com/microsoft/Olive/commit/8378d8ecbdd3853ab7d19e76017f53e62a3c04b5), deletion_comment = False)

Removed.

olive/passes/olive_pass.py
```diff
@@ -382,27 +382,20 @@ def run(
...
-
-    def inherit_hf_config_from_input_model(self, input_model: OliveModel, output_model: OliveModel) -> OliveModel:
-        # TODO: handle the case with local model path and huggingface model config for torch model
-        if hasattr(output_model, "hf_config") and not isinstance(output_model, PyTorchModel):
-            # not all models have hf_config
-            # Do not inherit hf_config from input model if output_model is PyTorchModel for time being.
-            if not output_model.hf_config and getattr(input_model, "hf_config", None):
-                output_model.hf_config = input_model.hf_config
+        output_model.extend_model_attributes(model.model_attributes)
         return output_model
```
devang-ml, line 399: Here, I am not sure how a pass is able to update the model_attributes.
I think, this should be pushed inside __run_for_confg()

olive/model/__init__.py
```diff
@@ -80,6 +82,15 @@ def model_path(self) -> str:
         """Return local model path."""
         return self.get_local_resource("model_path")
 
+    def extend_model_attributes(self, model_attr) -> Dict[str, Any]:
+        # self.model_attributes > model_attr, should respect self.model_attributes when there is conflict
```
devang-ml, line 86: It may not be an extension all the time.

guotuofeng: update should be better?

olive/passes/onnx/optimum_conversion.py
```diff
@@ -36,6 +36,7 @@ def _run_for_config(
 
         from optimum.exporters.onnx import main_export as export_optimum_model
 
+        model.extend_model_attributes_from_hf_config()
```
guotuofeng, line 39: >extend_model_attributes_from_hf_config [](http://example.com/codeflow?start=14&length=38)

this call should be part of model.__init__?

olive/model/__init__.py
```diff
@@ -509,6 +527,16 @@ def script_dir(self) -> str:
...
 
+    def extend_model_attributes_from_hf_config(self) -> Dict[str, Any]:
+        hf_model_config = {}
+        if self.hf_config:
+            # priority: model_attributes > hf_config
+            # need call get_hf_model_config() after self.model_path is set
+            # but if the input model_path is not local, we can't call get_hf_model_config()
+            hf_model_config = self.get_hf_model_config().to_dict()
+        hf_model_config.update(self.model_attributes)
+        self.model_attributes = hf_model_config
```
guotuofeng, line 538: I suppose all the logic here could be safely moved into constructor?

olive/model/__init__.py
```diff
@@ -509,6 +527,16 @@ def script_dir(self) -> str:
     def model_script(self) -> str:
         return self.get_local_resource("model_script")
 
+    def extend_model_attributes_from_hf_config(self) -> Dict[str, Any]:
```
guotuofeng, line 530: >extend_model_attributes_from_hf_config [](http://example.com/codeflow?start=8&length=38)

could be in if clause?

olive/passes/onnx/conversion.py
```diff
@@ -65,6 +65,7 @@ def _convert_model_on_device(
         output_model_path: str,
         device: str,
     ):
+        model.extend_model_attributes_from_hf_config()
```
guotuofeng, line 68: >extend_model_attributes_from_hf_config [](http://example.com/codeflow?start=14&length=38)

could we call it in pass.run?

olive/model/__init__.py
trajepl: I do not quit understand. More explanations?

olive/passes/onnx/conversion.py
trajepl: image
Like this?

guotuofeng: yes

olive/model/__init__.py
guotuofeng: IMHO, if the priority of hf_config/model_attr is different, the pass itself should be responsible do what ever merge and finally set the model_attr. 

From this point of view, a directly assignment is enough.

guotuofeng: Basically, the priority logic should not be responsibility of model. Instead, it should be in another place.

olive/passes/olive_pass.py
trajepl: `extend_model_attributes` will try merge output model's attributes and input model's attributes. When there is conflict, output model's attribute is always the first choice as pass already modified it.

olive/model/__init__.py
```diff
@@ -509,6 +527,16 @@ def script_dir(self) -> str:
...
+        hf_model_config = {}
```
guotuofeng: ditto here. The priority logic should be put another place.

olive/model/__init__.py
trajepl: `extend_model_attributes` with the priority order: self.model_attributes > model_attr
for me, the `update_model_attributes` with the priority order: self.model_attributes > model_attr
maybe we can add method `update_model_attributes` in case any pass requires in the future.

olive/passes/onnx/optimum_conversion.py
trajepl: We cannot put it under model.init as the input model can be in remote path where need be downloaded firstly.

olive/model/__init__.py
trajepl: Ditto.

olive/model/__init__.py
trajepl: Updated.

olive/model/__init__.py
trajepl: Let us discuss offline.

olive/model/__init__.py
guotuofeng: personally, I don't like the model maintain the logic here.

olive/passes/olive_pass.py
```diff
@@ -382,27 +385,20 @@ def run(
...
-            return self.inherit_hf_config_from_input_model(model, output_model)
-
-    def inherit_hf_config_from_input_model(self, input_model: OliveModel, output_model: OliveModel) -> OliveModel:
-        # TODO: handle the case with local model path and huggingface model config for torch model
-        if hasattr(output_model, "hf_config") and not isinstance(output_model, PyTorchModel):
-            # not all models have hf_config
-            # Do not inherit hf_config from input model if output_model is PyTorchModel for time being.
-            if not output_model.hf_config and getattr(input_model, "hf_config", None):
-                output_model.hf_config = input_model.hf_config
+        output_model.extend_model_attributes(model.model_attributes)
```
guotuofeng, line 401: a more concise and consolidated way to handle would likes:
```python
if output_model.model_attributes is None:
    hf_config = model.get_hf_model_config().to_dict() if isinstance(model, PyTorchModel) else {}
    input_attributes = model.model_attributes or {}
    hf_config.update(input_attributes)
    output_model.model_attributes = hf_config
```

guotuofeng: then we can safely remove all extend_model_attributes* method

guotuofeng (changes requested):
the priority logic should be move out of model itself.

olive/model/__init__.py
```diff
@@ -57,12 +57,14 @@ def __init__(
...
         model_path: OLIVE_RESOURCE_ANNOTATIONS = None,
-        # resources other than model_path
+        # resources other than model_path, e.g.: path to model adapter
         resources: Optional[Dict[str, OLIVE_RESOURCE_ANNOTATIONS]] = None,
+        model_attributes: Optional[Dict[str, Any]] = None,
     ):
         self.framework = framework
         self.model_file_format = model_file_format
         self.composite_parent = None
+        self.model_attributes = model_attributes or {}
```
guotuofeng, line 67: >or {} [](http://example.com/codeflow?start=49&length=5)

no need the empty dict

trajepl: Updated.

olive/passes/olive_pass.py
trajepl: Removed.

olive/model/__init__.py
trajepl: Updated to just run assignment after _run_for_config. We assume that once the pass modified the attributes, that is the final attributes. We should not overwrite it anymore with input model's attr.

olive/passes/olive_pass.py
trajepl: Updated with following logics:
1. we assume that the pass should return complete config even they modified it. That says the update logic is pushed into _run_for_config.
2. if output model's attributes is empty, we use the one in input model.

olive/passes/olive_pass.py
```diff
@@ -367,6 +367,12 @@ def run(
         """
         Run the pass on the model at a specific point in the search space.
         """
+        if isinstance(model, PyTorchModel) and model.hf_config:
+            hf_config = model.get_hf_model_config().to_dict()
+            input_model_attr = model.model_attributes or {}
+            hf_config.update(input_model_attr)
+            model.model_attributes = hf_config
+
```
devang-ml, line 375: Is there a reason to no do this in PyTorchModel __init__ ?

olive/passes/olive_pass.py
```diff
@@ -382,27 +388,24 @@ def run(
...
-    def inherit_hf_config_from_input_model(self, input_model: OliveModel, output_model: OliveModel) -> OliveModel:
-        # TODO: handle the case with local model path and huggingface model config for torch model
-        if hasattr(output_model, "hf_config") and not isinstance(output_model, PyTorchModel):
-            # not all models have hf_config
-            # Do not inherit hf_config from input model if output_model is PyTorchModel for time being.
-            if not output_model.hf_config and getattr(input_model, "hf_config", None):
-                output_model.hf_config = input_model.hf_config
+        # assumption: the model attributes from passes, if any, are more important than
+        # the input model attributes, we should not update/extend anymore outside of the pass run
+        output_model.model_attributes = output_model.model_attributes or model.model_attributes
```
xiaoyu-work, line 408: Where does `output_model.model_attributes` come from? Do we setup this somewhere?

olive/model/__init__.py
```diff
@@ -153,6 +155,7 @@ def to_json(self, check_object: bool = False):
                 for resource_name, resource_path in self.resource_paths.items()
             },
         }
+        config["config"].update({"model_attributes": self.model_attributes})
```
jambayk, line 158: I think it's cleaner if we add it directly after line 155?

olive/model/__init__.py
```diff
@@ -389,7 +397,7 @@ def to_json(self, check_object: bool = False):
                 "onnx_file_name": self.onnx_file_name,
                 "inference_settings": self.inference_settings,
                 "use_ort_extensions": self.use_ort_extensions,
-                "hf_config": self.hf_config,
+                "model_attributes": self.model_attributes,
```
jambayk, line 400: nit: This should already be present from line 394

olive/model/__init__.py
```diff
@@ -699,6 +709,7 @@ def to_json(self, check_object: bool = False):
                 "io_config": self.io_config,
                 "dummy_inputs_func": self.dummy_inputs_func,
                 "hf_config": self.hf_config,
+                "model_attributes": self.model_attributes,
```
jambayk, line 712: same as aboe

olive/passes/olive_pass.py
jambayk: I agree, I think it might be better to do this in the pytorch model init. 
This way, the serialized input model from the beginning has these model attributes. instead of the `model` object being dynamically updated by the pass run.

olive/passes/olive_pass.py
```diff
@@ -382,27 +388,24 @@ def run(
...
                 output_rank_model = self._run_for_config(input_rank_model, data_root, config, rank_output_path)
                 output_filepaths.append(output_rank_model.model_path)
-            return DistributedOnnxModel(output_filepaths, inference_settings=model.inference_settings)
+            output_model = DistributedOnnxModel(output_filepaths, inference_settings=model.inference_settings)
         elif isinstance(model, CompositeOnnxModel) and not self._accepts_composite_model:
             components = []
             component_names = []
             for cidx, child in enumerate(model.get_model_components()):
                 component_output_path = Path(output_model_path).with_suffix("") / str(cidx)
                 output_model_components = self._run_for_config(child, data_root, config, str(component_output_path))
```
jambayk, line 397: not related to this PR but the name `output_model_component` is confusing. since it is actually only once component.

olive/passes/olive_pass.py
trajepl: Oh, we can not put it under pytorch model.

The reason is that: the input model can be hosted in cloud path. But the `get_hf_mode_config` need the model and config with local path which requires to download model to local.  But in AML case, we expect the model and olive_run as all in AML. `Local` just collect the run results.

olive/passes/olive_pass.py
trajepl: Oh, yes.  See this: https://github.com/microsoft/Olive/blob/jiapli/model_attributes/olive/model/__init__.py#L67

olive/model/__init__.py
trajepl: Nope? The above config is got from python list expression. We cannot append it into a list expression.

trajepl: Or, do you mean other ways?

jambayk: Oh you are right. Thanks!

olive/passes/olive_pass.py
jambayk: Thanks! that makes sense. Then, I think this way of okay then since all of all information needed is also in the original model. Just not expanded out.

olive/passes/olive_pass.py
trajepl: I think you are talking about `output_model_compont[s]`, I remove that now.

jambayk: Thanks! There’s also some in onnx_conversion pass if you don’t mind.

trajepl: Updated.

jambayk (approved):

PR #543, opened 2023-09-06
Add vc support

xiaoyu-work:
## Describe your changes

Add vc support.

### DIFF
--- a/olive/systems/azureml/aml_system.py
+++ b/olive/systems/azureml/aml_system.py
@@ -73,6 +73,7 @@ class AzureMLSystem(OliveSystem):
         azureml_client_config: AzureMLClientConfig,
         aml_compute: str,
         aml_docker_config: Union[Dict[str, Any], AzureMLDockerConfig] = None,
+        resources: Dict = None,
         instance_count: int = 1,
         is_dev: bool = False,
         accelerators: List[str] = None,
@@ -81,6 +82,7 @@ class AzureMLSystem(OliveSystem):
     ):
         super().__init__(accelerators, olive_managed_env=olive_managed_env)
         self.instance_count = instance_count
+        self.resources = resources
         self.is_dev = is_dev
         self.requirements_file = requirements_file
         self.compute = aml_compute
@@ -272,7 +274,17 @@ class AzureMLSystem(OliveSystem):
         return {"pass_config": Input(type=AssetTypes.URI_FILE, path=pass_config_path), **pass_args}
 
     def _create_step(
-        self, name, display_name, description, aml_environment, code, compute, instance_count, inputs, script_name
+        self,
+        name,
+        display_name,
+        description,
+        aml_environment,
+        code,
+        compute,
+        resources,
+        instance_count,
+        inputs,
+        script_name,
     ):
         parameters = []
         for param, input in inputs.items():
@@ -283,12 +295,12 @@ class AzureMLSystem(OliveSystem):
         parameters.append("--pipeline_output ${{outputs.pipeline_output}}")
 
         cmd_line = f"python {script_name} {' '.join(parameters)}"
-
         component = command(
             name=name,
             display_name=display_name,
             description=description,
             command=cmd_line,
+            resources=resources,
             environment=aml_environment,
             code=str(code),
             inputs=inputs,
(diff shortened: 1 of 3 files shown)


devang-ml (commented):
- Please update options.md

guotuofeng:
could the description be more details? I don't know the background so there is no way to review it.

olive/systems/system_config.py
```diff
@@ -15,6 +15,16 @@
 from olive.systems.common import AzureMLDockerConfig, LocalDockerConfig, SystemType
 
 
+class VCConfig(ConfigBase):
+    instance_type: str
+    interactive: bool = False
+    imageVersion: str = ""
+    slaTier: str = "Standard"
+    priority: str = "Medium"
+    tensorboardLogDirectory: str = "/scratch/tensorboard_logs"
```
guotuofeng, line 24: default to None or required instead of the hardcode value?

olive/systems/system_config.py
```diff
@@ -15,6 +15,16 @@
 from olive.systems.common import AzureMLDockerConfig, LocalDockerConfig, SystemType
 
 
+class VCConfig(ConfigBase):
```
guotuofeng, line 18: do we need add documents/comments about the meaning of these fields?

olive/systems/system_config.py
xiaoyu-work: I'm testing the None value for these attributes now. If it supports None and the API can map a default value, I will update this logic.

olive/systems/system_config.py
xiaoyu-work: I will send the doc link to you offline.

olive/systems/system_config.py
guotuofeng, line 24: make the naming consistent by all use snake case(variable_part) instead of came case?

xiaoyu-work: I updated the PR and removed all other configs.

olive/systems/azureml/aml_system.py
```diff
@@ -81,7 +83,10 @@ def __init__(
         self._assert_not_none(aml_docker_config)
         aml_docker_config = validate_config(aml_docker_config, AzureMLDockerConfig)
         azureml_client_config = validate_config(azureml_client_config, AzureMLClientConfig)
+        if vc_config:
+            vc_config = validate_config(vc_config, VCConfig)
         self.azureml_client_config = azureml_client_config
+        self.vc_config = vc_config
```
trajepl, line 89: simplify to?
`self.vc_config = validate_config(vc_config, VCConfig) if vc_config else None`

olive/systems/azureml/aml_system.py
```diff
@@ -278,12 +293,20 @@ def _create_step(
         parameters.append("--pipeline_output ${{outputs.pipeline_output}}")
 
         cmd_line = f"python {script_name} {' '.join(parameters)}"
+        resources = None
+        if vc_config:
+            resources = {
+                "instance_type": vc_config.instance_type,
+                "instance_count": instance_count,
+                "properties": {"AISuperComputer": {**vc_config.extra_args} if vc_config.extra_args else {}},
```
trajepl, line 301: Double check: is this `AISuperComputer` a fixed value always?

olive/systems/azureml/aml_system.py
xiaoyu-work: Done.

olive/systems/azureml/aml_system.py
xiaoyu-work: Yes I think so.

olive/systems/azureml/aml_system.py
devang-ml: There is no need to validate vc_config. Just pass it as it is resource. Let user provide proper dict.

xiaoyu-work: @devang-ml this validation only turns a dict to VCConfig object. We only validate `instance_type` here.

devang-ml:
> could the description be more details? I don't know the background so there is no way to review it.

+1

devang-ml:
> > could the description be more details? I don't know the background so there is no way to review it.
> 
> +1

Something along the line "enable virtual clusters" ?

xiaoyu-work:
> > > could the description be more details? I don't know the background so there is no way to review it.
> > 
> > 
> > +1
> 
> Something along the line "enable virtual clusters" ?

I synced with Mike offline about what this PR does. Also `virtual cluster` is not a public concept of Azure Machine Learning.

jambayk (approved):

PR #545, opened 2023-09-07
Add `QLoRA` pass 

jambayk:
## Describe your changes
Add QloRA pass which performs the fine-tuning approach proposed in https://arxiv.org/abs/2305.14314. Based on the original code at https://github.com/artidoro/qlora/blob/main/qlora.py. 

Added examples, test and docs.

Also ordered the test requirements in alphabetical order. 

## Checklist before requesting a review
- [x] Add unit tests for this change.
- [x] Make sure all tests can pass.
- [x] Update documents if necessary.
- [x] Format your code by running `pre-commit run --all-files`
- [x] Is this a user-facing change? If yes, give a description of this change to be included in the release notes.
Add `QLoRA` pass for fine-tuning language models using the method proposed in https://arxiv.org/abs/2305.14314.

PR #549, opened 2023-09-08
Use `pass.run` and automatic `tmpdir` in pass unit tests

jambayk:
## Describe your changes
This is a follow up to the discussion https://github.com/microsoft/Olive/pull/545#discussion_r1319314454 to simplify the pass unit tests. 

`LocalSystem` is not needed to run the pass since we can run the pass directly. 
Also use `tmpdir` that is automatically provided by pytest instead of creating our own temporary directory. 

### DIFF
--- a/test/unit_test/passes/inc/test_inc_quantization.py
+++ b/test/unit_test/passes/inc/test_inc_quantization.py
@@ -3,7 +3,6 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 import platform
-import tempfile
 from pathlib import Path
 
 import pytest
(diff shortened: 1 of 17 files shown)


guotuofeng (approved):

test/unit_test/passes/inc/test_inc_quantization.py
```diff
@@ -15,62 +14,57 @@
...
-        # assert
-        assert quantized_model.model_path.endswith(".onnx")
-        assert Path(quantized_model.model_path).exists()
-        assert Path(quantized_model.model_path).is_file()
-        assert "QLinearConv" in [node.op_type for node in quantized_model.load_model().graph.node]
-
-
-def get_onnx_model(tempdir):
-    local_system = LocalSystem()
+def test_inc_quantization(tmpdir):
```
trajepl, line 22: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures
Seems the `tmpdir` return legacy [py.path.local](https://py.readthedocs.io/en/latest/path.html) objects rather than standard [pathlib.Path](https://docs.python.org/3/library/pathlib.html#pathlib.Path) objects.

The documents suggest to use `tmp_path`

jambayk: Thanks! Follow up PR to use this https://github.com/microsoft/Olive/pull/550

PR #561, opened 2023-09-13
Move model preparation logic out of model itself

guotuofeng:
## Describe your changes
From the time being, the OliveModel is used in Olive to prepare the optimization. It should use model config instead of model itself. For example, if the model is hosted in AML, Engine need create the model instance for run. later, it will need manually set the download model path to model instance. This kind of behavior hinder the lifecycle of the model. In case of remote model path scenario, we cannot call any model path related method in constructor. Such as https://github.com/microsoft/Olive/pull/542#discussion_r1320161984.

In this PR, I split the responsibility of model itself and model preparation to different place. The assumption is whenever the model instance is created, the model path should already be downloaded and can directly access.

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -333,59 +333,61 @@ class Engine:
                 # generate search space and initialize the passes for each hardware accelerator
                 self.setup_passes(accelerator_spec)
 
-            # hash the input model
-            input_model_id = self._init_input_model(input_model_config)
-            self.footprints[accelerator_spec].record(model_id=input_model_id)
+                # hash the input model
+                input_model_id = self._init_input_model(input_model_config)
+                self.footprints[accelerator_spec].record(model_id=input_model_id)
 
-            try:
-                if evaluate_input_model:
-                    prefix_output_name = (
-                        f"{output_name}_{accelerator_spec}_" if output_name is not None else f"{accelerator_spec}"
-                    )
-                    assert self.evaluator_config is not None, "evaluate_input_model is True but no evaluator provided"
-                    results = self._evaluate_model(
-                        input_model_config, input_model_id, data_root, self.evaluator_config, accelerator_spec
-                    )
-                    logger.info(f"Input model evaluation results: {results}")
-                    result_name = f"{prefix_output_name}_input_model_metrics"
-                    results_path = output_dir / f"{result_name}.json"
-                    with open(results_path, "w") as f:
-                        json.dump(results.to_json(), f, indent=4)
-                    logger.info(f"Saved evaluation results of input model to {results_path}")
-                    outputs[accelerator_spec] = results
-                    if not self.passes:
-                        logger.debug("No passes registered, return input model evaluation results.")
-                        return outputs
-
-                if self.no_search:
-                    output, model_ids = self.run_no_search(
-                        input_model_config,
-                        input_model_id,
-                        data_root,
-                        accelerator_spec,
-                        output_dir,
-                        output_name,
-                    )
-                    if output:
-                        outputs[accelerator_spec] = output
-                        pf_footprints[accelerator_spec] = self.footprints[accelerator_spec].get_footprints_by_model_ids(
-                            model_ids
+                try:
+                    if evaluate_input_model:
+                        prefix_output_name = (
+                            f"{output_name}_{accelerator_spec}_" if output_name is not None else f"{accelerator_spec}"
                         )
-                else:
-                    footprint = self.run_search(
-                        input_model_config,
-                        input_model_id,
-                        data_root,
-                        accelerator_spec,
-                        output_dir,
-                        output_name,
-                    )
-                    outputs[accelerator_spec] = footprint
-                    pf_footprints[accelerator_spec] = footprint
-            except EXCEPTIONS_TO_RAISE:
-                raise
-            except Exception as e:
-                logger.warning(f"Failed to run Olive on {accelerator_spec}: {e}", exc_info=True)
+                        assert (
+                            self.evaluator_config is not None
+                        ), "evaluate_input_model is True but no evaluator provided"
+                        results = self._evaluate_model(
+                            input_model_config, input_model_id, data_root, self.evaluator_config, accelerator_spec
+                        )
+                        logger.info(f"Input model evaluation results: {results}")
+                        result_name = f"{prefix_output_name}_input_model_metrics"
+                        results_path = output_dir / f"{result_name}.json"
+                        with open(results_path, "w") as f:
+                            json.dump(results.to_json(), f, indent=4)
+                        logger.info(f"Saved evaluation results of input model to {results_path}")
+                        outputs[accelerator_spec] = results
+                        if not self.passes:
+                            logger.debug("No passes registered, return input model evaluation results.")
+                            return outputs
+
+                    if self.no_search:
+                        output, model_ids = self.run_no_search(
+                            input_model_config,
+                            input_model_id,
+                            data_root,
+                            accelerator_spec,
+                            output_dir,
+                            output_name,
+                        )
+                        if output:
+                            outputs[accelerator_spec] = output
+                            pf_footprints[accelerator_spec] = self.footprints[
+                                accelerator_spec
+                            ].get_footprints_by_model_ids(model_ids)
+                    else:
+                        footprint = self.run_search(
+                            input_model_config,
+                            input_model_id,
+                            data_root,
+                            accelerator_spec,
+                            output_dir,
+                            output_name,
+                        )
+                        outputs[accelerator_spec] = footprint
+                        pf_footprints[accelerator_spec] = footprint
+                except EXCEPTIONS_TO_RAISE:
+                    raise
+                except Exception as e:
+                    logger.warning(f"Failed to run Olive on {accelerator_spec}: {e}", exc_info=True)
 
         for accelerator_spec in self.footprints.keys():
             logger.info(f"Run history for {accelerator_spec}:")
@@ -1103,7 +1105,7 @@ class Engine:
 
     def _evaluate_model(
         self,
-        model: ModelConfig,
+        model_config: ModelConfig,
         model_id: str,
         data_root: str,
         evaluator_config: OliveEvaluatorConfig,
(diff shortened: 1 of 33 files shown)


trajepl:
Please rebase main branch again to sync the change from Lei's multi-ep commit.

olive/model/__init__.py
```diff
@@ -68,19 +67,25 @@ def __init__(
...
 
     @property
     def model_path(self) -> str:
         """Return local model path."""
-        return self.get_local_resource("model_path")
+        return self.get_resource("model_path")
+
+    def set_resources(self, resources: Dict[str, OLIVE_RESOURCE_ANNOTATIONS]):
+        for resource_name, resource_path in resources.items():
+            if resource_path is not None:
```
trajepl, line 81: The logic looks duplicate with `self.set_resource`. Can we simplify as:
```python
for resource_name, resource_path in resource.items():
    self.set_resource(resouce_name, resource_path)
```

guotuofeng: actually, they are not same, the set_resource has validation to make sure the resource_name exist before update.

olive/model/__init__.py
```diff
@@ -183,6 +172,14 @@ def validate_type(cls, v):
             raise ValueError(f"Unknown model type {v}")
         return v
 
+    def get_resource_keys(self):
+        cls = REGISTRY[self.type.lower()]
```
jambayk, line 176: Is this equivalent to just returning `self.resource_keys`? The class attribute should be accessible using `self.` also.

olive/model/__init__.py
jambayk: Maybe it can be renamed to `validate_resources` to avoid confusion?

jambayk: maybe we can use a different name like `add_resources` or something to avoid confusion? Since `set_resource` assumes the resource name is already there?

olive/model/__init__.py
```diff
@@ -831,8 +833,9 @@ def __init__(self, model_path: OLIVE_RESOURCE_ANNOTATIONS, model_attributes: Opt
             model_attributes=model_attributes,
         )
         # check if the model files (xml, bin) are in the same directory
-        if self.resource_paths["model_path"].is_local_resource():
-            _ = self.model_config
+        model_path = create_resource_path(self.resource_paths["model_path"])
```
jambayk, line 836: nit: `self.get_resource("model_path")` might be better?

olive/model/__init__.py
jambayk: Please ignore this. Did not see that this is a method of `ModelConfig` and not `OliveModel`.

olive/systems/docker/utils.py
```diff
@@ -88,36 +87,14 @@ def create_metric_volumes_list(
     return mount_list
 
 
-def create_resource_path_mount(
-    resource_path: ResourcePath, local_resource_path: ResourcePath, container_root_path: Path
-) -> (str, str):
-    # no need to mount if resource path is None or string name
-    if not resource_path or resource_path.is_string_name():
```
jambayk, line 95: This case is still needed. There is nothing to mount for None or string resources

olive/systems/local.py
```diff
@@ -22,28 +22,31 @@ def __init__(self, accelerators: List[str] = None):
...
 
     def evaluate_model(
-        self, model: OliveModel, data_root: str, metrics: List[Metric], accelerator: AcceleratorSpec
+        self, model_config: ModelConfig, data_root: str, metrics: List[Metric], accelerator: AcceleratorSpec
     ) -> MetricResult:
         """
         Evaluate the model
         """
-        if isinstance(model, CompositeOnnxModel):
+        if model_config.type == "CompositeOnnxModel":
```
jambayk, line 43: It might be safer to check with `.lower()` since the keys in the registry is in lower case and the `type` is supposed to be case insensitive. 
Maybe we can also add the `.lower()` step in the type validation to be consistent. 
image

olive/model/__init__.py
guotuofeng: will use self.model_path

olive/systems/docker/utils.py
guotuofeng: will add it.

olive/systems/local.py
guotuofeng: I prefer to keep the original type user specified in ModelConfig. Instead, we could add lower() whenever it need check user inputs. 

Will add lower() when we explicitly check the type.

jambayk:
Can you also update this part in `cache.py`? https://github.com/microsoft/Olive/blob/main/olive/cache.py#L238 
Now, there is no need to create model just to get the resource_keys. 

Would also be safer to do a `.get` at this line https://github.com/microsoft/Olive/blob/main/olive/cache.py#L241 since we cannot guarantee that the workflow input model's config has all keys. 

Thanks!

olive/engine/engine.py
```diff
@@ -814,31 +815,34 @@ def _load_model(self, model_id: str) -> Union[OliveModel, str]:
...
-    def _prepare_non_local_model(self, model: OliveModel) -> OliveModel:
+    def _prepare_non_local_model(self, model_config: ModelConfig) -> ModelConfig:
         """
         Prepare models with non-local model path for local run by downloading the model resources to cache
         """
-        for resource_name, resource_path in model.resource_paths.items():
+        resource_keys = model_config.get_resource_keys()
+        for resource_name in resource_keys:
+            resource_path = model_config.config.get(resource_name)
+            resource_path = create_resource_path(resource_path)
```
jambayk, line 828: This can be simplfied using `model_config.get_resource_paths` like in other places in this PR.

olive/engine/engine.py
```diff
@@ -1014,18 +1022,19 @@ def _run_pass(
         # run pass
         host = self.host_for_pass(pass_id)
         if host.system_type != SystemType.AzureML:
-            input_model = self._prepare_non_local_model(input_model)
+            input_model_config = self._prepare_non_local_model(input_model_config)
```
guotuofeng, line 1025: @jambayk , do we need move the model download logic to OliveSystem? if yes, how could we pass the cache directory?

guotuofeng: maybe we can add OliveSystem.prepare_model(cache_dir) as a signature?

jambayk: but I don't think that's high priority since we only have aml system that has a different preparation (does nothing).

olive/engine/engine.py
guotuofeng: will change it

guotuofeng:
> Can you also update this part in `cache.py`? https://github.com/microsoft/Olive/blob/main/olive/cache.py#L238 Now, there is no need to create model just to get the resource_keys.
> 
> Would also be safer to do a `.get` at this line https://github.com/microsoft/Olive/blob/main/olive/cache.py#L241 since we cannot guarantee that the workflow input model's config has all keys.
> 
> Thanks!

will directly use model_config.get_resource_paths(), In this way, we need not call dict.get anymore.

jambayk (dismissed):

olive/model/__init__.py
```diff
@@ -93,41 +98,28 @@ def set_resource(self, resource_name: str, resource_path: Union[Path, str, Resou
...
+            resource_path = resolved_resource_path.get_path()
+
+        self.resource_paths[resource_name] = resource_path
+
+    def get_resource(self, resource_name: str) -> str:
         """
         Get local path of a resource.
 
         :param resource_name: name of the resource.
         :return: local path.
```
trajepl, line 116: The `return` may be not be local path, right?

guotuofeng: the resource always be local path

olive/model/__init__.py
```diff
@@ -258,10 +258,8 @@ def __init__(
         self.graph = None
         self.all_graphs: Optional[List[GraphProto]] = None
 
-        # if model_path is local folder, check for onnx file name
-        model_resource_path = self.resource_paths["model_path"]
-        if model_resource_path and model_resource_path.type == ResourceType.LocalFolder:
-            self.get_onnx_file_path(model_resource_path.get_path(), self.onnx_file_name)
+        # check for onnx file name since it will do validation
+        _ = self.model_path
```
trajepl, line 262: Will the assignment check the path name aotumatically? A bit confused here.

trajepl: I see, it is a method with property decorator.

guotuofeng: yes, you can see get_onnx_file_path implementation. I don't change the implementation and just refactoring

jambayk: Yes, you can see the definition of the `model_path` property at line 302

guotuofeng: I suppose @jambayk add this method?

olive/model/__init__.py
trajepl: Ic, looking other parts now.

olive/model/__init__.py
```diff
@@ -301,7 +299,8 @@ def get_onnx_file_path(model_path: str, onnx_file_name: Optional[str] = None) ->
     @property
     def model_path(self) -> str:
         model_path = super().model_path
-        model_path = self.get_onnx_file_path(model_path, self.onnx_file_name) if model_path else None
+        if model_path and Path(model_path).is_dir():
```
jambayk, line 302: This part doesn't actually need to be changed. `self.get_onnx_file_path` already handles the `file` case and also ensures that `model_path` exists.

guotuofeng: just see https://github.com/microsoft/Olive/commit/1f8427642e06c2927ba9bd35561faf0cbf2f81f7#diff-2e0c7ca9e36053f5b553ea13c7c506d26abced76ff0efd0b9f8b85043972cce3R322 has this check

guotuofeng: https://github.com/microsoft/Olive/blob/main/olive/model/__init__.py#L263

guotuofeng: actually, I am a little bit confused about the condition is not same originally when call get_onnx_file_path

jambayk: Yeah, iirc, when I added it, I wanted to make sure the local folder had the .onnx file inside it. But looking at it now, there was no need to distinguish between local file and local folder since `self.get_onnx_file_path` handles both cases. 

Not really an issue. Just makes the logic cleaner.

olive/model/__init__.py
trajepl: Yes thanks, I get the checking callback stack now.

olive/model/__init__.py
trajepl, line 302: No need to check is dir or not? Line 274 in this PR will check that.

trajepl: Oh, seems you talked that in above comments.

guotuofeng: will remove it

jambayk (approved):

Issue #557, opened 2023-09-13
[Bug]: Olive installs the stable GPU version of ORT even though the nightly GPU version of ORT is already installed


jsoto-gladia:
### What happened?

related to the issue [here](https://github.com/microsoft/onnxruntime/issues/17393)

**description**
It appears Olive installs the stable GPU version of ORT even though the nightly GPU version of ORT is already installed

**steps to reproduce**
On a v100s machine, I prepare my pyenv with :
```bash
jsoto@v100s:/$ git clone https://github.com/microsoft/Olive.git
jsoto@v100s:/Olive$ cd Olive/
jsoto@v100s:/Olive$ pyenv virtualenv 3.9 aceituna
jsoto@v100s:/Olive$ pyenv shell aceituna
```

When in aceituna pyenv, I install all the necessary packages with:
```bash
(aceituna) jsoto@v100s:/Olive$ cd examples/whisper
(aceituna) jsoto@v100s:/Olive/examples/whisper$ python -m pip install -r requirements.txt
(aceituna) jsoto@v100s:/Olive/examples/whisper$ python -m pip uninstall -y onnxruntime ort-nightly onnxruntime-gpu ort-nightly-gpu
(aceituna) jsoto@v100s:/Olive/examples/whisper$ python -m pip install ort-nightly --index-url https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/
```

When running the configuration preparation, I do:
```bash
(aceituna) jsoto@v100s:/Olive/examples/whisper$ python prepare_whisper_configs.py --model_name openai/whisper-tiny --multilingual
(aceituna) jsoto@v100s:/Olive/examples/whisper$ python -m olive.workflows.run --config whisper_gpu_fp32.json 
```
I see that Olive does not use my ort-nightly-gpu and install and uses onnxruntime-gpu;
```bash
[...]

Issue #508, opened 2023-08-24
[Bug]: INVALID_GRAPH /  Error Node (BeamSearch_node) has input size 12 not in range [min=5, max=10] when trying to build a non-openai provided Whisper model.


vmitro:
### What happened?

When trying to build an ONNX model without multilingual support, using instructions on the [whisper example](https://github.com/microsoft/Olive/blob/main/examples/whisper/README.md) page and supplying a non-openai repo (e.g. aware-ai/whisper-tiny-german) the `test_transcription.py` fails. Tested using venv.

```
(olive-env) (base) vmitro@v3629:~/projects/olive_stuff/Olive/examples/whisper$ python prepare_whisper_configs.py --model_name aware-ai/whisper-tiny-german --no_audio_decoder
(olive-env) (base) vmitro@v3629:~/projects/olive_stuff/Olive/examples/whisper$ python -m olive.workflows.run --config whisper_cpu_int8.json --setup
[2023-08-23 23:43:57,572] [INFO] [run.py:112:dependency_setup] The following packages are required in the local environment: ['onnxruntime']
[2023-08-23 23:43:57,572] [INFO] [run.py:116:dependency_setup] onnxruntime is already installed.
(olive-env) (base) vmitro@v3629:~/projects/olive_stuff/Olive/examples/whisper$ python -m olive.workflows.run --config whisper_cpu_int8.json
[2023-08-23 23:44:07,042] [WARNING] [config_utils.py:270:validate_config] Keys {'disable_search'} are not part of OrtTransformersOptimizationConfig. Ignoring them.
[2023-08-23 23:44:07,077] [DEBUG] [engine.py:675:resolve_goals] Resolving goals: {'latency': {'avg': None}}
[2023-08-23 23:44:07,077] [DEBUG] [engine.py:694:resolve_goals] No baseline got as no goal is provided the the goal is threshold
[2023-08-23 23:44:07,078] [DEBUG] [engine.py:475:run_no_search] Step no search with search point {'OnnxConversion': {}, 'OrtTransformersOptimization': {}, 'OnnxDynamicQuantization': {}, 'InsertBeamSearch': {}, 'AppendPrePostProcessingOps': {}} ...
[2023-08-23 23:44:07,078] [INFO] [engine.py:939:_run_pass] Running pass OnnxConversion
[2023-08-23 23:44:07,078] [DEBUG] [engine.py:957:_run_pass] Loading model from cache ...
[...]

PR #563, opened 2023-09-13
Drop python 3.7, Fix `onnxruntime` dependency setup, Whisper: Add EP and FAQ

jambayk:
## Describe your changes
Officially drop python 3.7 since it has already reached end of life. Most olive dependencies have already dropped python 3.7 support so olive was probably incompatible with it. 

`onnxruntime` packages are distributed using different names but they are all installed under the same `onnxruntime` namespace.
So, we cannot check if the required package is installed using the `try: __import__` method. This PR uses `importlib.metadata` to check the local ort installations and log the required steps needed. We don't want to uninstall packages ourselves so we will just log the command for the user to run. 

Addresses the original issue in https://github.com/microsoft/Olive/issues/557 

Also update Whisper example to provide and use accelerator+ep correctly.
Added FAQ for common issues like https://github.com/microsoft/Olive/issues/508

Update: Also fixes https://github.com/microsoft/Olive/issues/571

### DIFF
--- a/examples/whisper/README.md
+++ b/examples/whisper/README.md
@@ -128,3 +128,16 @@ python test_transcription.py --config whisper_cpu_int8.json
 
 `--task` Optional. Whether to perform X->X speech recognition ('transcribe') or X->English translation ('translate'). Default is `transcribe`. Only used
 when `--multilingual` is provided to `prepare_whisper_configs.py`
+
+## FAQ
+The following are some common issues that may be encountered when running this example.
+1. `INVALID_GRAPH / Error Node (BeamSearch_node) has input size 12 not in range [min=5, max=10]` or similar error when running inference using the optimized model.
+This is likely due to:
+    - Mismatch between the versions of onnxruntime used to run the optimization workflow and the inference. To fix this, please use the same version of
+    onnxruntime for both.
+    - Mismatch between the versions of onnxruntime used in a previously cached run and the currently installed version. To fix this, please delete the `cache` folder
+    and run the workflow again.
+
+2. Whenever you install a new version of onnxruntime (such as ort-nightly), you may need to delete the `cache` folder and run the workflow again. This is because the cache doesn't
+distinguish between different versions of onnxruntime and will use the cached models from a previous run. There might be incompatibilities between the cached models and the new
+version of onnxruntime.
--- a/olive/workflows/run/run.py
+++ b/olive/workflows/run/run.py
@@ -7,7 +7,7 @@ import logging
 import os
 import subprocess
 from pathlib import Path
-from typing import Union
+from typing import Dict, Union
 
 import onnxruntime as ort
 
@@ -76,6 +76,12 @@ def dependency_setup(config):
             "TorchTRTConversion": EXTRAS.get("torch-tensorrt"),
         },
     }
+    ORT_PACKAGE_TO_EP_MAPPING = {
+        "onnxruntime": "CPUExecutionProvider",
+        "onnxruntime-gpu": "CUDAExecutionProvider",
+        "onnxruntime-directml": "DmlExecutionProvider",
+        "onnxruntime-openvino": "OpenVINOExecutionProvider",
+    }
 
     local_packages = []
     remote_packages = []
@@ -111,13 +117,16 @@ def dependency_setup(config):
     # install missing packages to local or tell user to install packages in their environment
     logger.info(f"The following packages are required in the local environment: {local_packages}")
     for package in set(local_packages):
-        try:
-            __import__(package)
-            logger.info(f"{package} is already installed.")
-        except ImportError:
-            logger.info(f"Installing {package}...")
-            subprocess.check_call(["python", "-m", "pip", "install", "{}".format(package)])
-            logger.info(f"Successfully installed {package}.")
+        if package in ORT_PACKAGE_TO_EP_MAPPING:
+            check_local_ort_installation(ORT_PACKAGE_TO_EP_MAPPING, package)
+        else:
+            try:
+                __import__(package)
+                logger.info(f"{package} is already installed.")
+            except ImportError:
+                logger.info(f"Installing {package}...")
+                subprocess.check_call(["python", "-m", "pip", "install", "{}".format(package)])
+                logger.info(f"Successfully installed {package}.")
     if remote_packages:
         logger.info(
             "Please make sure the following packages are installed in {} environment: {}".format(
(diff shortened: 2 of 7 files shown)


olive/workflows/run/run.py
```diff
@@ -76,6 +76,12 @@ def dependency_setup(config):
             "TorchTRTConversion": EXTRAS.get("torch-tensorrt"),
         },
     }
+    ORT_PACKAGE_TO_EP_MAPPING = {
```
guotuofeng, line 79: it seems this map is some kind of duplicated with https://github.com/microsoft/OLive/blob/8dc23995537b5db607a4e5b9a43859095134a885/olive/systems/utils.py#L66

jambayk: I saw that mapping too but wasn't sure if we can reuse this since the mapping is reverse. and the reverse is not 1 to 1.

examples/whisper/README.md
```diff
@@ -128,3 +128,16 @@ python test_transcription.py --config whisper_cpu_int8.json
...
+The following are some common issues that may be encountered when running this example.
+1. `INVALID_GRAPH / Error Node (BeamSearch_node) has input size 12 not in range [min=5, max=10]` or similar error when running inference using the optimized model.
+This is likely due to:
+    - Mismatch between the versions of onnxruntime used to run the optimization workflow and the inference. To fix this, please use the same version of
+    onnxruntime for both.
+    - Mismatch between the versions of onnxruntime used in a previously cached run and the currently installed version. To fix this, please delete the `cache` folder
+    and run the workflow again.
+
+2. Whenever you install a new version of onnxruntime (such as ort-nightly), you may need to delete the `cache` folder and run the workflow again. This is because the cache doesn't
+distinguish between different versions of onnxruntime and will use the cached models from a previous run. There might be incompatibilities between the cached models and the new
```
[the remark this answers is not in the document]
jambayk: The grammar here is correct since it's a noun. Otherwise, the sentence would need to change to `the cached models and the new version might be incompatible`.

examples/whisper/README.md
```diff
@@ -128,3 +128,16 @@ python test_transcription.py --config whisper_cpu_int8.json
...
 when `--multilingual` is provided to `prepare_whisper_configs.py`
+
+## FAQ
+The following are some common issues that may be encountered when running this example.
+1. `INVALID_GRAPH / Error Node (BeamSearch_node) has input size 12 not in range [min=5, max=10]` or similar error when running inference using the optimized model.
+This is likely due to:
+    - Mismatch between the versions of onnxruntime used to run the optimization workflow and the inference. To fix this, please use the same version of
+    onnxruntime for both.
+    - Mismatch between the versions of onnxruntime used in a previously cached run and the currently installed version. To fix this, please delete the `cache` folder
+    and run the workflow again.
```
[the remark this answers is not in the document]
jambayk: I did to add this part since we generated the config jsons using a template. and generally, want the cache to be used. Otherwise, every run after this change would clean the cache.

olive/workflows/run/run.py
```diff
@@ -76,6 +76,12 @@ def dependency_setup(config):
...
+        "onnxruntime": "CPUExecutionProvider",
+        "onnxruntime-gpu": "CUDAExecutionProvider",
```
xiaoyu-work, line 81: How to handle the other GPU EP cases in this map?

jambayk: For now, it's only cuda for gpu. We could add an `ep` command line argument in the future if needed.

jambayk: This is just a heuristic. Since `onnxruntime-gpu` package should always have cuda ep. 

There's already some TODOs for figuring out the correct ort package https://github.com/microsoft/Olive/blob/main/olive/workflows/run/run.py#L99 

But that is orthogonal to this PR's change. This one is after the required package has been inferred. This only checks if the inferred package is present.

olive/workflows/run/run.py
```diff
@@ -183,3 +192,32 @@ def run(config: Union[str, Path, dict], setup: bool = False, data_root: str = No
...
+    all_packages = list(ep_mapping.keys())
+    all_nightly_packages = []
+    for package in all_packages:
+        nightly_package_parts = ["ort-nightly"]
+        if "-" in package:
+            nightly_package_parts.append(package.split("-")[1])
+        all_nightly_packages.append("-".join(nightly_package_parts))
+    all_packages.extend(all_nightly_packages)
+    logger.info(f"Uninstalling all onnxruntime packages: {all_packages}")
+    subprocess.check_call(["python", "-m", "pip", "uninstall", "-y"] + all_packages)
```
xiaoyu-work, line 217: I'm okay we install new packages for user, but I'm not sure if we could help user decide to automatically uninstall existing packages for user. If the package is incompatible, I prefer raising an error and giving them a suggestion what to do here.

jambayk: I can comment out the uninstall part but it will probably lead to package clashes. Because of how ort packages all of their different flavors using different pip packages but they all install into the same `onnxruntime` namespace. 
So, we don't know what ort package is already installed.  

Maybe for the ort package we don't do any of the installation and uninstallations? And just print the commands they have to run? 
- uninstall all existing ort packages
- install required ort package

olive/workflows/run/run.py
xiaoyu-work: Ah if this is only used for validation, that is fine.

olive/workflows/run/run.py
xiaoyu-work: Yeah we can explain this in error message that different onnxruntime packages share the same namespace so they have to uninstall all onnxruntime package and only install the one they need.

jambayk: should we raise an error or print the recommanded commands?

xiaoyu-work: We can print the recommended commands for user, as we can get all onnxruntime packages installed in his environment: https://github.com/microsoft/Olive/blob/main/olive/engine/packaging/packaging_generator.py#L206-L207 so we know what they should uninstall.

jambayk: Updated to use `importlib_metadata.package_distributions()`. Didn't use `pkg_resources` since it's deprecated and `importlib` is recommended.

olive/workflows/run/run.py
jambayk: This mapping doesn't exist anymore. Found a better way to find the package.

olive/workflows/run/run.py
```diff
@@ -9,10 +9,11 @@
 from pathlib import Path
 from typing import Union
 
+import importlib_metadata
```
guotuofeng, line 12: it seems it is third party packages?

jambayk: oh yeah, I thought it was the same as `importlib.metadata` but it's actually a backport https://pypi.org/project/importlib-metadata/

`importlib.metadata` is only available in python 3.8 and doesn't have `packages_distributions`.

What do you think we should do? 
- add `importlib_metadata` as a dependency? 
- or use `importlib.metadata` and mimic the `packages_distributions` functionality

jambayk: python < 3.8 are already deprecated and most of our other dependencies need python >= 3.8. So I don't think we need to worry about versions < 3.8?

jambayk: Updated as per offline discussion

setup.py
```diff
@@ -47,7 +47,6 @@ def get_extra_deps(rel_path):
     "Topic :: Software Development :: Libraries :: Python Modules",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.7",
```
guotuofeng, line 50: do we need enforce python version? https://stackoverflow.com/questions/19534896/enforcing-python-version-in-setup-py

guotuofeng: https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

jambayk: thanks! added

olive/workflows/run/run.py
```diff
@@ -183,3 +192,56 @@ def run(config: Union[str, Path, dict], setup: bool = False, data_root: str = No
...
+        return
+
+    # uninstall all ort packages
+    logger.info("There are one or more onnxruntime packages installed in your environment.")
+    uninstall_command = ["python", "-m", "pip", "uninstall", "-y"] + local_ort_packages
+    uninstall_command = " ".join(uninstall_command)
+    logger.info(f"Please run '{uninstall_command}' to uninstall all existing onnxruntime packages.")
+
+    # install the package
+    logger.info(f"Then install {package_name} by running 'python -m pip install {package_name}'.")
```
guotuofeng, line 229: forget to install?

jambayk: We only log the command to the user. 
Since they have to first uninstall all other ort packages and then install the correct one

jambayk: I had a discussion with Xiaoyu above and we thought it's better for olive to not uninstall packages and just tell the user

jambayk: Updated after offline discussion

guotuofeng (approved):

PR #300, opened 2023-05-25
add packaging mlflow model option

leqiao-1:
## Describe your changes
user can export model in mlflow format if set extra config in packaging

## Checklist before requesting a review
- [ ] Add unit tests for this change.
- [ ] Make sure all tests can pass.
- [x] Update documents if necessary.
- [x] Format your code by running `pre-commit run --all-files`
- [x] Release notes: Engine - packaging part: add support to package output ONNX model in mlflow format.

PR #564, opened 2023-09-13
Add packaging mlflow model option.

xiaoyu-work:
## Describe your changes

Add packaging mlflow model option. This PR is a follow up RP for https://github.com/microsoft/Olive/pull/300 as there is no package conflict now. 

### DIFF
--- a/olive/engine/packaging/packaging_config.py
+++ b/olive/engine/packaging/packaging_config.py
@@ -22,3 +22,4 @@ class PackagingConfig(ConfigBase):
 
     type: PackagingType = PackagingType.Zipfile
     name: str = "OutputModels"
+    export_in_mlflow_format: bool = False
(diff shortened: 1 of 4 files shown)


olive/engine/packaging/packaging_config.py
```diff
@@ -22,3 +22,4 @@ class PackagingConfig(ConfigBase):
 
     type: PackagingType = PackagingType.Zipfile
     name: str = "OutputModels"
+    export_in_mlflow_format: bool = False
```
guotuofeng, line 25: is it possible we export other format? if yes, we might need change the name like export_format or something else?

xiaoyu-work: As for format, I don't think we have other cases for now. As for model type, even though Olive doesn't limit output model type, we only focus on onnx model now.

leqiao-1 (approved):

PR #566, opened 2023-09-14
fix model to_json bug since the resource is just string now

guotuofeng:
## Describe your changes
PyTorchModel.to_json  will call resource.get_path. This would not work since the resource is just string now. The PR just to fix the bug.


trajepl:
Besides the fixed line, should we mark `self.resource_paths` as `Dict[str, str]`?
image

guotuofeng:
> Besides the fixed line, should we mark `self.resource_paths` as `Dict[str, str]`? image

good catch

trajepl (dismissed):

trajepl (approved):

PR #572, opened 2023-09-15
refactor engine run for each accelerator

guotuofeng:
## Describe your changes
Move the engine run for each accelerator to a dedicated method since it introduce a lot of indent in Python, which is hard to maintain. 

BTW, during the refactoring process, I found : 
https://github.com/microsoft/Olive/pull/419 introduce a behavior which is not consistent with original one. 
If there is no pass only the first accelerator will be used for evaluation input model. Not sure it is a bug. 

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -329,64 +329,28 @@ class Engine:
 
         for accelerator_spec in self.accelerator_specs:
             with self.create_managed_environment(accelerator_spec):
-                # generate search space and initialize the passes for each hardware accelerator
-                self.setup_passes(accelerator_spec)
+                run_result = self.run_accelerator(
+                    input_model_config, data_root, output_dir, output_name, evaluate_input_model, accelerator_spec
+                )
 
-                # hash the input model
-                input_model_id = self._init_input_model(input_model_config)
-                self.footprints[accelerator_spec].record(model_id=input_model_id)
+                if run_result is None:
+                    continue
 
-                try:
-                    if evaluate_input_model:
-                        prefix_output_name = (
-                            f"{output_name}_{accelerator_spec}_" if output_name is not None else f"{accelerator_spec}"
+                if evaluate_input_model and not self.passes:
+                    # for evaluate input model only, return the evaluation results
+                    # TODO: need check whether the evaluation results are valid since it will only evaluate input model
+                    # once and use the same evaluation results for all accelerators
+                    return run_result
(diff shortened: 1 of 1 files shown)


olive/engine/engine.py
```diff
@@ -406,6 +374,65 @@ def run(
...
+                )
+                logger.info(f"Input model evaluation results: {results}")
+                result_name = f"{prefix_output_name}_input_model_metrics"
+                results_path = output_dir / f"{result_name}.json"
+                with open(results_path, "w") as f:
+                    json.dump(results.to_json(), f, indent=4)
+                logger.info(f"Saved evaluation results of input model to {results_path}")
+                if not self.passes:
+                    logger.debug("No passes registered, return input model evaluation results.")
+                    return results
```
trajepl, line 410: The results line L410, L413, L422 are different. Should we unify these output with same formatting?

olive/engine/engine.py
```diff
@@ -406,6 +374,65 @@ def run(
...
+                    input_model_config, input_model_id, data_root, self.evaluator_config, accelerator_spec
+                )
+                logger.info(f"Input model evaluation results: {results}")
+                result_name = f"{prefix_output_name}_input_model_metrics"
+                results_path = output_dir / f"{result_name}.json"
+                with open(results_path, "w") as f:
+                    json.dump(results.to_json(), f, indent=4)
+                logger.info(f"Saved evaluation results of input model to {results_path}")
+                if not self.passes:
+                    logger.debug("No passes registered, return input model evaluation results.")
```
trajepl, line 409: For the case you mentioned, when there is no passes, the input model only be evaluated once. It should be a bug.
How to extract the logics of input model evaluation function out of `run_accelerator?`

guotuofeng: Actually, I am looking forward to unifying them. However, from the current footprint implementation, it seems a little bit hard to unify them. 

How could we do it?

guotuofeng: Basically, the run_accelerator is invoked under multiple EP context manager. If the evaluation model doesn't support multiple EP, it is ok to move it out of run_accelerator. Otherwise, we need make all of these logic under the context manager.

trajepl: Agree. The evaluation is bound with accelerate spec. Making all these logics under the same context is OK for me.

For unify the results, will do it in following PR.

trajepl (approved):

PR #574, opened 2023-09-16
Text-generation: Support text templates, Special tokens in corpus join, Batched processing

jambayk:
## Describe your changes
Improve text-generation dataset preprocessing using the following updates:
- Allow user to provide f-string templates to combine multiple text columns. This enables templates such as `"### Question: {prompt} \n### Answer: {response}"` 
- Allow adding special tokens to each example before joining them in corpus join strategies. This enables data "packing" like huggingface sft trainer does (https://github.com/huggingface/trl/blob/main/trl/trainer/sft_trainer.py#L90, https://github.com/huggingface/trl/blob/main/trl/trainer/utils.py#L464) 
- Process the texts in batches instead of joining them and processing as a single string. Tokenizers might be optimized for batched tokenization so we should take advantage of it. 

### DIFF
--- a/olive/data/component/text_generation.py
+++ b/olive/data/component/text_generation.py
@@ -87,7 +87,12 @@ class TextGenCorpusParams(TextGenParams):
     text_template: str = None
     corpus_strategy: TextGenCorpusStrategy = TextGenCorpusStrategy.LINE_BY_LINE
     stride: int = None  # required when corpus_strategy is JOIN_SLIDING_WINDOW
-    joiner: str = " "  # delimiter to use when joining the rows of the input columns.
+    # text to join the rows of input columns when corpus_strategy is JOIN
+    # joined as "{text_col1} {joiner} {text_col2} ..."
+    # if None, joined with a space
+    # can also provide __eos_token__ to use eos_token as the joiner
+    joiner: str = None
+    processing_batch_size: int = 1024  # number of examples to process at a time
     random_seed: int = None  # random seed for LINE_BY_LINE_RANDOM and JOIN_RANDOM
     random_retries: int = (
         10  # number of resamples to try before giving up when a sample is too short for RANDOM strategies
@@ -154,8 +159,9 @@ def text_gen_corpus_pre_process(_dataset, tokenizer, all_kwargs):
     text_template = args.text_template
     if text_template is None:
         text_template = " ".join(["{" + col + "}" for col in args.text_cols])
-    _dataset = apply_template(_dataset, args.text_cols, "text", text_template)
+    _dataset = apply_template(_dataset, args.text_cols, "text", text_template, remove_cols=True)
     text_list = _dataset["text"]
+    total_examples = len(text_list)  # total number of examples
 
     tokenized_inputs = {
         "input_ids": [],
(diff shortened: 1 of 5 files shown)


olive/data/component/text_generation.py
```diff
@@ -463,3 +505,12 @@ def extract_alpaca_dataset(example):
     # remove unused columns, keep only input and output
     dataset = dataset.remove_columns([col for col in dataset.column_names if col not in ["input", "output"]])
     return dataset
+
+
+def apply_template(dataset, cols, new_col, template, remove_cols=False):
+    """Apply template to column in dataset."""
+    dataset = dataset.map(
+        lambda x: {new_col: template.format(**{col: x[col] for col in cols})},
```
trajepl, line 513: Could you help give some introductions, like: the placeholder col in template must be from dataset?

olive/data/component/text_generation.py
```diff
@@ -118,7 +128,9 @@ class TextGenPairParams(TextGenParams):
 
     pair_format: TextGenPairFormat = TextGenPairFormat.DEFAULT
     input_col: str = None  # required when pair_format is CUSTOM
+    input_template: str = None  # a python f-string template for the input, input is replaced with {input}
```
guotuofeng, line 131: >input_template [](http://example.com/codeflow?start=4&length=14)

from semantic point of view, the input_template and input_col seems represent the same thing?

olive/data/component/text_generation.py
```diff
@@ -79,10 +79,20 @@ def _check_padding(cls, v, values):
 class TextGenCorpusParams(TextGenParams):
     """Parameters for text generation task with 'corpus' dataset type."""
 
-    text_cols: list  # list of text columns
+    # list of text columns, for each row, columns are concatenated together using template
+    # for join corpus_strategy, all rows are then concatenated together using joiner
+    text_cols: list
+    # a python f-string template for the text, the text in each column is replaced with {column_name}
+    # is not provided, the text in each column is joined with a space
+    text_template: str = None
```
guotuofeng, line 87: >text_template [](http://example.com/codeflow?start=4&length=13)

a custom function might be better than f-string template?

olive/data/component/text_generation.py
guotuofeng: the user scenario for f-string template seems be limited.

olive/data/component/text_generation.py
jambayk: I plan to support custom formatting function in the future. But that needs more things like user script module and modifying the data container. Since all of these are absorbed in the kwargs, I think it will be okay to support multiple ways:
- column names
- fstring template
- formatting func

olive/data/component/text_generation.py
jambayk: Like I mention in https://github.com/microsoft/Olive/pull/574#discussion_r1329091884, I am planning to support multiple different ways to set the input text. 
The f-string template is limited but it covers a lot of potential use cases while being easy to provide. for custom formatting func, the user needs to write and provide a user script, script dir, etc.

olive/data/component/text_generation.py
```diff
@@ -118,7 +128,9 @@ class TextGenPairParams(TextGenParams):
 
     pair_format: TextGenPairFormat = TextGenPairFormat.DEFAULT
     input_col: str = None  # required when pair_format is CUSTOM
+    input_template: str = None  # a python f-string template for the input, input is replaced with {input}
     output_col: str = None  # required when pair_format is CUSTOM
+    output_template: str = None  # a python f-string template for the output, output is replaced with {output}
```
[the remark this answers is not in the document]
jambayk: The "pair" and "corpus" + template are similar. However, pair dataset knows that there are two columns in each example and is able to truncate each sentence separately. We can also choose to make the input sequence if needed which is not possible in a generic corpus strategy. 

The pair dataset works similar to this data collator https://github.com/artidoro/qlora/blob/main/qlora.py#L448

olive/data/component/text_generation.py
```diff
@@ -65,7 +65,7 @@ class TextGenParams(ConfigBase):
     # if false, cannot gaurantee all sequences are same length. data loader will have to handle this during collation
     pad_to_max_len: bool = True  # pad sequences to max_len, ignored for JOIN corpus strategy
     drop_short_sequences: bool = False  # drop sequences shorter than max_len. Mutually exclusive with pad_to_max_len
-    add_special_tokens: bool = True  # add bos and eos tokens
+    add_special_tokens: bool = True  # add bos and eos tokens to each sequence
```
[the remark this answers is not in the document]
jambayk: For the max sequence length, I already have a dev branch to extract it from the tokenizer if not provided by the user.

olive/data/component/text_generation.py
jambayk: Pushed a commit to make only one required. The priority is `template` > `column name`

olive/data/component/text_generation.py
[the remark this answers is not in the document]
jambayk: Hf also has a similar setup but they use data collator and no column specific max lengths. https://github.com/huggingface/trl/blob/main/trl/trainer/utils.py#L56 

I am thinking of combining pair and corpus types under the same text-gen preprocessing. and to provide optional params to enable the column specific masking and truncation. 
But I have to do some design brainstorming on this and will do a follow up PR for the changes.

olive/data/component/text_generation.py
guotuofeng: could we add TODO there?

jambayk: added!

guotuofeng: BTW, I don't see the example use the template. are there any code snippet to demonstrate how they are used?

jambayk: There is no example in the code yet but I have a local example 
```
data_config = {
    "name": "tiny-code-python",
    "params_config": {
        "model_name": model_name,
        "task": "text-generation",
        "data_name": "parquet",
        "data_files": "tiny-codes-python.parquet",
        "split": "train",
        "component_kwargs": {
            "pre_process_data": {
                "dataset_type": "corpus",
                "corpus_strategy": "join",
                "text_template": "### Question: {prompt} \n### Answer: {response}", 
                "max_samples": 1000,
                "source_max_len": 1024,
            }
        }
    }
}
```
It is not complete yet so I am waiting to add some other changes before adding this example.

guotuofeng: could we postpone the change for template into your examples so the reviewer or other guys would have the knowledge when looking at the commit history in future? Or we need add examples in this PR. 

In this way, we can guarantee the feature is completed when it is merged into main.

guotuofeng: the context bug could be a separated PR.

jambayk: I will add the example in this PR then. I need this PR to get merged so I can make follow up changes. Otherwise, the PR becomes too big.

guotuofeng: Basically, each PR should be as minimal as possible but keep it complete. At the same time, every PR should not include other unrelated changes. For example, the bug fix in this PR.

jambayk: Sometimes, it's hard to include a complete example since the full example depends on other features which in turn depends on the current PR. 
It's easier if we have unit tests for the components but currently don't have them for dataset components. 

For this PR I will add an example with a custom component that will be updated later to remove the custom component once that feature is added.
I will also create a separate PR for the bug-fix.

guotuofeng: as an opposite example, I have no idea about what the context bug it is since the file include a lot of other changes unless I carefully look at every line changes.

jambayk: Separate PR for context length https://github.com/microsoft/Olive/pull/581

guotuofeng: > Sometimes, it's hard to include a complete example since the full example depends on other features which in turn depends on the current PR. It's easier if we have unit tests for the components but currently don't have them for dataset components.
> 
> For this PR I will add an example with a custom component that will be updated later to remove the custom component once that feature is added. I will also create a separate PR for the bug-fix.

If there is no real dataset, an unit test should be enough by mocking up the dataset.

jambayk: Added example. Will rebase with main when the bugfix PR merges.

jambayk: rebased to main

trajepl (approved):

Issue #573, opened 2023-09-15
[Bug]: OnnxQuantization


akarym-sl:
### What happened?

When running first OnnxQuantization pass with default parameters and a pass then with QUInt weight and activation type, the model parameters are not quantized to QUInt. 
To clarify, running pass: 

- OnnxConversion->OnnxQuantization (QUInt8) 

yields different accuracy than running two passes:

- OnnxConversion->OnnxQuantization
- OnnxConversion->OnnxQuantization (QUInt8)

### Version?

0.3.1

guotuofeng:
@akarym-sl , what's the config json you used to run the optimization for quantization?

akarym-sl:
Here is the config. For the second case I prepend ["onnx_conv", "onnx_quant"] to the "pass_flows" list
```json
{
    "input_model":{
        "type":"PyTorchModel",
        "config":{
            "model_path":"model.pt",
            "model_loader":"load_state_dict",
            "model_script":"save.py",
            "dummy_inputs_func":"get_dummy_inputs",
            "io_config":{
                "input_names":[
                    "input"
                ],
                "output_names":[
                    "output"
                ],
                "dynamic_axes":{
                    "input":{
                        "0":"batch"
                    },
                    "output":{
                        "0":"batch"
                    }
                }
            }
        }
    },
    "systems":{
        "local_system":{
            "type":"LocalSystem",
            "config":{
                "accelerators":[
                    "cpu"
                ]
            }
        }
    },
    "evaluators":{
        "custom_evaluator":{
            "metrics":[
                {
                    "name":"custom",
                    "type":"custom",
                    "user_config":{
                        "user_script":"user_script.py",
                        "batch_size":1,
                        "dataloader_func":"create_dataloader",
                        "evaluate_func":"evaluate"
                    },
                    "sub_types":[
                        {
                            "name":"latency",
                            "priority":1,
                            "higher_is_better":false
                        },
                        {
                            "name":"accuracy",
                            "priority":2,
                            "higher_is_better":true
                        }
                    ]
                }
            ]
        }
    },
    "engine":{
        "clean_cache":true,
        "cache_dir":".cache",
        "output_dir":"optimization",
        "host":"local_system",
        "target":"local_system",
        "execution_providers":[
            "CPUExecutionProvider"
        ],
        "evaluator":"custom_evaluator",
        "evaluate_input_model":false
    },
    "passes":{
        "onnx_conv":{
            "type":"OnnxConversion",
            "config":{
[...]

PR #577, opened 2023-09-18
🐛 Distinguish pass instance with name not cls name

trajepl:
## Describe your changes

This PR is used fix following issue where the same pass with different config appears in one olive run config.
https://github.com/microsoft/Olive/issues/573

[Root Cause]: 
Olive uses the pass's class name as key to identify the pass instance. When there are passes with same pass_class, the first one defined in olive run config will be picked always.
https://github.com/microsoft/Olive/blob/main/olive/engine/engine.py#L436

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -436,13 +436,17 @@ class Engine:
     def setup_passes(self, accelerator_spec: AcceleratorSpec):
         # clean the passes
         self.passes.clear()
-        for config in self.pass_config.values():
+        for name, config in self.pass_config.items():
             pass_cls: Type[Pass] = config["type"]
             pass_cfg = config["config"]
             pass_cfg = pass_cls.generate_search_space(accelerator_spec, pass_cfg, config["disable_search"])
             p = pass_cls(accelerator_spec, pass_cfg, config["disable_search"])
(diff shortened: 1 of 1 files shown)


olive/engine/engine.py
```diff
@@ -436,13 +436,17 @@ def run_accelerator(
...
             pass_cfg = config["config"]
             pass_cfg = pass_cls.generate_search_space(accelerator_spec, pass_cfg, config["disable_search"])
             p = pass_cls(accelerator_spec, pass_cfg, config["disable_search"])
             self.register_pass(
-                p, host=config["host"], evaluator_config=config["evaluator"], output_name=config["output_name"]
+                p,
+                host=config["host"],
+                evaluator_config=config["evaluator"],
+                output_name=config["output_name"],
+                name=name,
```
guotuofeng, line 449: >name=name, [](http://example.com/codeflow?start=16&length=10)

the parameter name is after the instance. Move there to keep consistency

trajepl: Thanks, I just treat it as kw arguments then ignore the order.
Will wait the first around CI test results. Then to fix it.

guotuofeng (approved):

PR #580, opened 2023-09-18
🥋 Unify engine run returning

trajepl:
## Describe your changes

This PR is used to unify the returning of `run_no_search` and `run_search` with olive footprints.

More discussions here: https://github.com/microsoft/Olive/pull/572

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -7,6 +7,7 @@ import logging
 import time
 from collections import OrderedDict, defaultdict
 from contextlib import contextmanager
+from copy import deepcopy
 from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Type, Union
@@ -325,7 +326,6 @@ class Engine:
         output_dir.mkdir(parents=True, exist_ok=True)
 
         outputs = {}
-        pf_footprints = {}
 
         for accelerator_spec in self.accelerator_specs:
             with self.create_managed_environment(accelerator_spec):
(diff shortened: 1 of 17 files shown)


olive/engine/engine.py
```diff
@@ -558,23 +540,21 @@ def run_no_search(
...
-                # output_model_json is the last model only if the flow is not pruned
-                output["model"] = output_model_json
-                if signal is not None:
-                    output["metrics"] = signal
-            else:
-                output = None
-            flows_output[tuple(pass_flow)] = output
-        return flows_output, output_model_ids
+        output_model_ids = list(output_models.keys())
+        fp_outputs = deepcopy(self.footprints[accelerator_spec].create_footprints_by_model_ids(output_model_ids))
```
guotuofeng, line 552: >deepcopy [](http://example.com/codeflow?start=21&length=8)

why do we need deepcopy? I suppose the create method will return new object. right?

trajepl: Good catche. The create method is renamed from get_footprints_by_model_ids where the dict is used to assign to a new variable, but not the new object. I will try to move the deepcopy logics to creat methods.

olive/engine/footprint.py
```diff
@@ -58,8 +59,8 @@ def __init__(
         objective_dict: Dict = None,
         is_marked_pareto_frontier: bool = False,
     ):
-        self.nodes = nodes or OrderedDict()
-        self.objective_dict = objective_dict or {}
+        self.nodes = deepcopy(nodes) or OrderedDict()
+        self.objective_dict = deepcopy(objective_dict) or {}
```
guotuofeng, line 63: It would be better to move deepcopy into the create metthod? for other footprint creation, the deepcopy is needless if the node is temp created?

olive/engine/footprint.py
```diff
@@ -153,7 +154,7 @@ def mark_pareto_frontier(self):
             self.nodes[k].is_pareto_frontier = cmp_flag
         self.is_marked_pareto_frontier = True
 
-    def get_footprints_by_model_ids(self, model_ids):
+    def create_footprints_by_model_ids(self, model_ids):
```
guotuofeng, line 157: need we rename get_pareto_frontier?

trajepl: Double checked that, after the get_pareto_frontier, olive_engine will change the nodes in some cases. 
Then seems we should rename get_pareto_frontier too.

Updated.

guotuofeng (approved):

PR #494, opened 2023-08-16
add multiple ep support

leqiao-1:
## Describe your changes
add support to tune and evaluate models across different execution providers 

## Checklist before requesting a review
- [x] Add unit tests for this change.
- [ ] Make sure all tests can pass.
- [ ] Update documents if necessary.
- [ ] Format your code by running `pre-commit run --all-files`
- [x] Is this a user-facing change? If yes, give a description of this change to be included in the release notes.
User can let olive manage environment for multiple eps.

PR #582, opened 2023-09-19
Add openllama arc example

xiaoyu-work:
## Describe your changes

* Add openllama arc example


xiaoyu-work:
Will update documentation shortly

xiaoyu-work:
There should be a bug about multiple EP. Will take a look later.

olive/engine/engine.py
```diff
@@ -121,7 +121,7 @@ def __init__(
             skip_get_available_ep = False
             if (
                 self.target.system_type in (SystemType.AzureML, SystemType.Docker, SystemType.PythonEnvironment)
-                and self.target.olive_managed_env
+                or self.target.olive_managed_env
```
guotuofeng, line 124: the logic is wrong

guotuofeng (changes requested):
the engine change need be reverted. 

Please refer to https://github.com/microsoft/Olive/pull/494/commits/c8bf6902f5c8c496491d8d75ad6c118516fe96a1 for more information

olive/engine/engine.py
jambayk: `or` looks right to me. 
We don't want to use the local `ort.get_available_providers` if the system is not local or managed env. https://github.com/microsoft/Olive/blob/main/olive/hardware/accelerator.py#L71 

Otherwise for non-managed, non-local systems, we would be using the available eps from the local environment.

xiaoyu-work: Yeah the reason I made this change just like Jambay's comment. We shouldn't get available eps from local environment for non-managed and non-local system.

guotuofeng: If we'd like not get available EP for docker and azureml, we need explicitly logic to handle this. The reason Lei add this check is because the python environment will fail since the skip is true and the available EP for python env is not correct.

jambayk: Created this PR to get the correct available eps https://github.com/microsoft/Olive/pull/591 

Most tests pass. Just need to make sure the docker tests pass.

examples/open_llama/README.md
```diff
@@ -91,6 +91,11 @@ Note: You must be logged in to HuggingFace using `huggingface-cli login` to down
 
 Requirements file: [requirements-qlora.txt](requirements-qlora.txt)
 
+### Optimizing Open Llama Model with Azure Arc
+This workflow optimizes Open Llama model on Azure ML compute, and evaluate output models on your device. Please connect your device to Azure Arc by following instruction: [Self-hosted Kubernetes cluster](https://microsoft.github.io/Olive/tutorials/azure_arc.html)
+
+This example config file is [open_llama_arc.json](open_llama_arc.json).
+
```
jambayk, line 98: Can you add a requirements file? This example requires the azureml extras to run.

xiaoyu-work: Good call. Added.

jambayk (approved):

guotuofeng (approved):

PR #587, opened 2023-09-20
⏮️ Preprocess for quant as true for bert model

trajepl:
## Describe your changes

Enable quantization preprocess by default for bert model, as we know that would avoid some bad samples from olive optimization for bert cases. 

### DIFF
--- a/.azure_pipelines/performance_check/configs/bert.json
+++ b/.azure_pipelines/performance_check/configs/bert.json
@@ -58,6 +58,7 @@
         "quantization": {
             "type": "OnnxQuantization",
             "config": {
+                "quant_preprocess": true,
                 "data_config": "__input_model_data_config__"
             }
         },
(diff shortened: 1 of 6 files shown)


jambayk (approved):
Thanks! Can you update the PR description before merging with the reason for this so we have a record of it in the commit history?

trajepl:
> Thanks! Can you update the PR description before merging with the reason for this so we have a record of it in the commit history?

done! thanks.

.azure_pipelines/performance_check/configs/bert.json
```diff
@@ -58,6 +58,7 @@
         "quantization": {
             "type": "OnnxQuantization",
             "config": {
+                "quant_preprocess": true,
```
guotuofeng, line 61: could we add this knowledge when generating auto-tune config?

trajepl: Yes~ We should. It is a good experience to run pre process for quantization.

guotuofeng (approved):

PR #588, opened 2023-09-20
Update ort-nightly c package download logic

xiaoyu-work:
## Describe your changes

* There is no straight download link for ort-nightly c package. User should download it by themselves on Azure DevOps link. Updated logic here to skip downloading ort-nightly pacakges.
* Add other ep packages support


olive/engine/packaging/packaging_generator.py
```diff
@@ -262,24 +262,20 @@ def _package_onnxruntime_packages(tempdir, pf_footprint: Footprint):
...
-            _download_c_packages(False, is_nightly, ort_version, cs_download_path)
+        # Download CPP && CS onnxruntime package
+        lang_list = ["cpp", "cs"]
+        for language in lang_list:
+            ort_download_path = tempdir / "ONNXRuntimePackages" / language
+            ort_download_path.mkdir(parents=True, exist_ok=True)
+            if is_nightly:
+                _skip_download_c_package(ort_download_path)
+            else:
+                if should_package_ort_cpu:
```
guotuofeng, line 273: >should_package_ort_cpu [](http://example.com/codeflow?start=19&length=22)

should we download DML/OpenVino EP?

xiaoyu-work: Good call. I'll update this with dml and openvino.

olive/engine/packaging/packaging_generator.py
```diff
@@ -262,24 +262,20 @@ def _package_onnxruntime_packages(tempdir, pf_footprint: Footprint):
...
+                    download_path = str(ort_download_path / f"microsoft.ml.onnxruntime.{ort_version}.nupkg")
+                    _download_c_packages(True, ort_version, download_path)
+                if should_package_ort_gpu:
+                    download_path = str(ort_download_path / f"microsoft.ml.onnxruntime.gpu.{ort_version}.nupkg")
+                    _download_c_packages(False, ort_version, download_path)
```
trajepl, line 278: nitpick: maybe we can simplify as:
```python
if should_package_ort_gpu:
    download_path = str(ort_download_path / f"microsoft.ml.onnxruntime.gpu.{ort_version}.nupkg")
else:
    download_path = str(ort_download_path / f"microsoft.ml.onnxruntime.{ort_version}.nupkg")

_download_c_packages(should_package_ort_gpu, ort_version, download_path)
```

olive/engine/packaging/packaging_generator.py
```diff
@@ -315,24 +311,21 @@ def _download_ort_extensions_package(use_ort_extensions: bool, download_path: st
...
     else:
-        if is_nightly:
-            urllib.request.urlretrieve(NIGHTLY_C_GPU_LINK.substitute(ort_version=ort_version), download_path)
-        else:
-            urllib.request.urlretrieve(STABLE_C_GPU_LINK.substitute(ort_version=ort_version), download_path)
+        urllib.request.urlretrieve(STABLE_C_GPU_LINK.substitute(ort_version=ort_version), download_path)
+
+
+def _skip_download_c_package(package_path):
+    warning_msg = (
```
trajepl, line 324: From my understanding, even we download the ort package, user still need to install them manually, right? 
Then why not just provide the install command like: pip install ortxxxx==xxx?
I may forget the reason why we need to download the package.

Please correct me if not so.

olive/engine/packaging/packaging_generator.py
xiaoyu-work: I've refactored this.

olive/engine/packaging/packaging_generator.py
xiaoyu-work: This will save the effort to download packages by themselves. Also, this could be useful for those who want to deploy the output model, so we can provide them a downloaded onnxruntime package.

guotuofeng:
>     is_stable = True if onnxruntime_pkg else False

is_nightly == False should be same with is_stable==True. right?

---
Refers to: olive/engine/packaging/packaging_generator.py:193 in 104b52d. [](commit_id = 104b52dd691ea08317a04c718d17dc5871511fdb, deletion_comment = False)

olive/utils.py
```diff
@@ -0,0 +1,10 @@
+def get_package_name_from_ep(execution_provider):
+    PROVIDER_PACKAGE_MAPPING = {
+        "CPUExecutionProvider": ("onnxruntime", "ort-nightly"),
+        "CUDAExecutionProvider": ("onnxruntime-gpu", "ort-nightly-gpu"),
+        "TensorrtExecutionProvider": ("onnxruntime-gpu", "ort-nightly-gpu"),
+        "RocmExecutionProvider": ("onnxruntime-gpu", "ort-nightly-gpu"),
```
guotuofeng, line 6: >ort-nightly-gpu [](http://example.com/codeflow?start=54&length=15)

seems we can replace onnxruntime with "ort-nightly" exception openvino?

Could we mapping to stable version and do the string replace in package except openvino?

xiaoyu-work: Even that is the pattern, I still treat them as different packages, like openvino is not following this, and I believe rocm package (to be released) will not follow this pattern as well.

xiaoyu-work:
> > ```
> > is_stable = True if onnxruntime_pkg else False
> > ```
> 
> is_nightly == False should be same with is_stable==True. right?
> 
> Refers to: olive/engine/packaging/packaging_generator.py:193 in [104b52d](https://github.com/microsoft/Olive/commit/104b52dd691ea08317a04c718d17dc5871511fdb). [](commit_id = [104b52d](https://github.com/microsoft/Olive/commit/104b52dd691ea08317a04c718d17dc5871511fdb), deletion_comment = False)

This is used to handle the case when the user has both (maybe unintentionally) ort stable and ort nightly installed. So is_nightly == False not quite same as is_stable==True.

olive/utils.py
jambayk, line 1: can we move this to somewhere in `olive.common`? That's where we have been keeping all olive utils. 

Also needs the copyright header

xiaoyu-work: Ah I missed olive.common. Will move to this folder.

jambayk (approved):

PR #591, opened 2023-09-21
Implement `get_supported_execution_providers` for all systems

jambayk:
## Describe your changes
Implement `get_supported_execution_providers` method for all systems so that for non-managed systems. 
Olive engine doesn't use this method for remote systems but this method might be useful in the system or for debugging purposes where users want to check the available providers in their remote systems.

Added some debug logs during the resolution of accelerator specs.

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -122,6 +122,7 @@ class Engine:
             elif self.target.system_type == SystemType.Docker:
                 # for docker system we default use CPUExecutionProvider
                 execution_providers = ["CPUExecutionProvider"]
+        logger.debug(f"Initial execution providers: {execution_providers}")
 
         self.execution_providers = execution_providers
 
@@ -139,6 +140,7 @@ class Engine:
                     f"from given execution providers {self.execution_providers}."
                 )
                 accelerators = inferred_accelerators
+        logger.debug(f"Initial accelerators: {accelerators}")
 
         ep_to_process = set(self.execution_providers)
         # Flatten the accelerators to list of AcceleratorSpec
@@ -159,9 +161,12 @@ class Engine:
             supported_eps = AcceleratorLookup.get_execution_providers_for_device_by_available_providers(
                 device, available_eps
             )
+            logger.debug(f"Supported execution providers for device {device}: {supported_eps}")
             for ep in ep_to_process.copy():
                 if ep == "CPUExecutionProvider" and device != "cpu" and is_cpu_available:
-                    logger.info("ignore the CPUExecutionProvider for non-cpu device")
+                    logger.info(
+                        "Ignore the CPUExecutionProvider for non-cpu device since cpu accelerator is also present."
+                    )
                 elif ep in supported_eps:
                     self.accelerator_specs.append(AcceleratorSpec(device, ep))
                     ep_to_process.remove(ep)
@@ -172,6 +177,9 @@ class Engine:
             f"Given execution providers: {self.execution_providers}. "
             f"Current accelerators: {accelerators}."
             f"Supported execution providers: {AcceleratorLookup.EXECUTION_PROVIDERS}."
+        )
+        logger.info(
+            f"Running workflow on accelerator specs: {','.join([str(spec) for spec in self.accelerator_specs])}"
         )
         if ep_to_process:
             logger.warning(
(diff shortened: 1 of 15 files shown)


olive/engine/engine.py
```diff
@@ -85,16 +85,13 @@ def __init__(
         if execution_providers is None:
             execution_providers = self._config.execution_providers
 
-        # verify the AzureML system have specified the execution providers
-        # Please note we could not use isinstance(target, AzureMLSystem) since it would import AzureML packages.
-        if self.target.system_type == SystemType.AzureML and execution_providers is None:
-            raise ValueError("AzureMLSystem requires execution providers to be specified.")
+        # for managed environment, the user must specify the execution providers
+        if self.target.olive_managed_env and execution_providers is None:
+            raise ValueError("Managed environment requires execution providers to be specified.")
```
xiaoyu-work, line 90: Is another main idea of `olive_managed_env` that user doesn't need to worry about ep and we will try all possible eps for them? I think that was why we skip get_supported_execution_providers call for `olive_managed_env`.

olive/systems/azureml/aml_system.py
```diff
@@ -621,5 +644,71 @@ def _create_metric_component(
 
         return metric_component
 
+    def get_supported_execution_providers(self) -> List[str]:
```
xiaoyu-work, line 647: I'm not sure we should create a new job just for available ep fetching. I prefer 2 options here: 1. user still needs to provide desired ep for aml system. 2. somehow update our aml runner script logic to support testing all available eps together in one script.

olive/engine/engine.py
```diff
@@ -112,22 +109,29 @@ def __init__(
...
-            skip_get_available_ep = False
-            if (
-                self.target.system_type in (SystemType.AzureML, SystemType.Docker, SystemType.PythonEnvironment)
-                and self.target.olive_managed_env
-            ):
-                skip_get_available_ep = True
-            supported_eps = AcceleratorLookup.get_execution_providers_for_device(device, skip_get_available_ep)
+            if self.target.olive_managed_env:
+                # get all the execution providers supported by the device
+                supported_eps = AcceleratorLookup.get_execution_providers_for_device(device, skip_get_available_ep=True)
```
guotuofeng, line 121: >get_execution_providers_for_device [](http://example.com/codeflow?start=50&length=34)

better use a separated method to query EP for multiple EP scenario since get_execution_providers_for_device is called in perf-tuning and model preparation as well

olive/systems/azureml/aml_system.py
guotuofeng: Yeah, I agree with xiaoyu. The dedicated job seems a little bit heavy. for azureml/docker system, I still prefer let user specify the EP explicitly to simplify the get available EP logic.

guotuofeng: I create PR https://github.com/microsoft/Olive/pull/592 for same purpose.

olive/engine/engine.py
jambayk: In this PR I didn't want to change any of the original behavior. Since Mike already merged his PR to correct the accelerator spec resolution, this change doesn't exist in this PR anymore.

olive/engine/engine.py
jambayk: This changed doesn't exist anymore since I rebased to main after your PR.

olive/systems/azureml/aml_system.py
jambayk: Thanks this makes sense to me. I rebased to main with your merged PR. 
I think we can still merge this PR since I made some code improvements to the systems and the `get_supported_execution_providers` methods is still useful. In case the user wants to debug and check the available eps in a remote system.

guotuofeng: I am not still sure whether we need get available EPs for both AzureML and docker. 
better confirm with @devang-ml

jambayk: I think it's okay to have this method. Might be useful in the future or if a user wants to check what eps their remote system supports.

jambayk: It's already implemented and doesn't change any of the current engine or olive behavior.

jambayk:
Closing since the we don't want to support these methods for the remote systems.

PR #592, opened 2023-09-21
normalize get available EPs behavior

guotuofeng:
## Describe your changes
Only get available EP in local and python environment. For AzureML and docker we need let user specify them

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -82,6 +82,31 @@ class Engine:
         else:
             self.target = LocalSystem()
 
+        self.setup_accelerators(execution_providers)
+
+        # default evaluator
+        self.evaluator_config = None
+        if evaluator_config is not None:
+            self.evaluator_config = evaluator_config
+        elif self._config.evaluator is not None:
+            self.evaluator_config = self._config.evaluator
+
+        # dictionary of passes
+        self.pass_config = OrderedDict()
+
+        # {"pass_name": {"pass": pass, "host": host, "evaluator": evaluator, "clean_run_cache": clean_run_cache}}
+        self.passes = OrderedDict()
+
+        self.pass_flows = None
+        self.pass_flows_search_spaces = None
+
+        self.footprints = defaultdict(Footprint)
+
+        self.azureml_client_config = self._config.azureml_client_config
+
+        self._initialized = False
+
+    def setup_accelerators(self, execution_providers):
         if execution_providers is None:
             execution_providers = self._config.execution_providers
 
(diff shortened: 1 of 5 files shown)


olive/engine/engine.py
```diff
@@ -85,19 +85,21 @@ def __init__(
...
             execution_providers = self._config.execution_providers
 
-        # verify the AzureML system have specified the execution providers
-        # Please note we could not use isinstance(target, AzureMLSystem) since it would import AzureML packages.
-        if self.target.system_type == SystemType.AzureML and execution_providers is None:
-            raise ValueError("AzureMLSystem requires execution providers to be specified.")
-        elif execution_providers is None:
-            if self.target.system_type in (SystemType.Local, SystemType.PythonEnvironment):
+        if not execution_providers:
+            if self.target.olive_managed_env:
```
xiaoyu-work, line 114: I had a same question in Jambay's PR: is another main idea of `olive_managed_env` that user doesn't need to worry about ep and we will try all possible eps for them? I think that was why we skip `get_supported_execution_providers` call for `olive_managed_env`. Or this is a part of Olive auto optimization?

guotuofeng: yes, for `olive_managed_env` we will create intermediate system and install all the needed ORT.

xiaoyu-work: I mean in this case, should we try all possible eps by device? Does user need to provide eps he wants?

guotuofeng: user need specify what's the EP he/she want to run against multiple EP. and there is no need to try all possible EP

olive/engine/engine.py
```diff
@@ -82,22 +82,49 @@ def __init__(
...
+                # verify the AzureML system have specified the execution providers
+                # Please note we could not use isinstance(target, AzureMLSystem) since it would import AzureML packages.
+                raise ValueError("AzureMLSystem requires execution providers to be specified.")
+            elif self.target.system_type in (SystemType.Local, SystemType.PythonEnvironment):
                 execution_providers = self.target.get_supported_execution_providers()
-            else:
-                # for docker system and python system, we default use CPUExecutionProvider
+            elif self.target.system_type == SystemType.Docker:
+                # for docker system we default use CPUExecutionProvider
                 execution_providers = ["CPUExecutionProvider"]
```
trajepl, line 124: From my understanding, in this PR, you want to put off the logic to get EPs if the current system's ORT version unknow(like aml, docker).
Local: SystemType.Local, SystemType.PythonEnvironment
Remote: AML, Docker

Then if so, why we give the CPU eps for docker as default? Should we let user specify for docer system?

olive/engine/engine.py
```diff
@@ -114,17 +141,24 @@ def __init__(
...
-                and self.target.olive_managed_env
-            ):
-                skip_get_available_ep = True
-            supported_eps = AcceleratorLookup.get_execution_providers_for_device(device, skip_get_available_ep)
+            if self.target.olive_managed_env:
+                available_eps = AcceleratorLookup.get_managed_supported_execution_providers(device)
+            elif self.target.system_type in (SystemType.Local, SystemType.PythonEnvironment):
+                available_eps = self.target.get_supported_execution_providers()
+            elif self.target.system_type == SystemType.Docker:
+                # TODO: do we need allow docker system support other execution providers?
```
trajepl, line 154: Yes, the TODO item is what I am thinking. See the above comment.

trajepl (approved):

PR #594, opened 2023-09-21
Add formatting function support in text-gen processsing

jambayk:
Allow users to provide formatting function for text-generation. 


docs/source/tutorials/configure_data.rst
```diff
@@ -415,3 +415,19 @@ The above case shows to rewrite all the components in data config. But sometime,
...
             - json_config https://github.com/microsoft/Olive/blob/main/test/unit_test/test_data_root.py#L107
+
+.. note::
+    The components will be called with the following arguments along with any additional keyword arguments provided in the config:
+        - load_dataset: `data_dir` (keyword, optional)
+        - pre_process_data: `dataset` (positional)
+        - post_process_data: `output` (positional)
+        - dataloader: `dataset` (positional)
+
+    The components can get access to the `user_script` and `script_dir` from the data config by adding them as named arguments (positional or keyword) to the component function.
```
trajepl, line 426: Could we distinguish the argument of `user_script` and `script_dir`(component level) with the ones in data_config level.

trajepl: It is a bit confused when I first see this comment(if I do not see the following `text_generation_huggingface_pre_process`), why we expose user_script in component level.

jambayk: do you mean the component has it's own user script and script dir? 
For this, I think it will become problematic in the future if the system is not local. We won't know what all to mount to the system.
So, I thought it's easy to use the same user script for the config and the components. 

Even now, our data configs are not fully compatible with aml and docker system since the user script and script dirs are not mounted at all. when we fix this, it will be even more problematic if components have their own user script.

jambayk: maybe we can make it clear by using a special argument name like `__data_config_user_script__` and `__data_config_script_dir__`?

jambayk: Might have to do something similar to https://github.com/microsoft/Olive/pull/409 in the future for user script and script dir too.

jambayk: rolled back this change as discussed offline.

trajepl:
It is still in Draft status. Any updates?

jambayk:
> It is still in Draft status. Any updates?

I was waiting to make a PR with unit tests for hf dataset preprocessing but unfortunately haven't had the time to do it yet. Since there is no current use case for the features added in this PR. It completes a TODO from before

Will open this PR for review and merge it if possible.

olive/data/config.py
```diff
@@ -98,30 +98,35 @@ def _update_default_component(self):
...
+        Priority is: component_kwargs > params_config > component.params > default value
         """
         from inspect import signature
 
-        self.params_config = self.params_config or {}
+        self.params_config = deepcopy(self.params_config) or {}
         component_kwargs = self.params_config.pop("component_kwargs", {})
         for k, v in self.components.items():
             component = Registry.get_component(k, v.type)
             # 1. use the params_config["component_kwargs"] to update component.params
```
trajepl, line 117: Remove this line to align with the steps you re-write.

trajepl:
It looks good to me. Feel free to ping me for approving once the comment is fixed. Thanks!

olive/data/config.py
jambayk: I think this comment is correct. this line and line 121 together perform step 1 described in line 1. We have to do the update first (line 118) and then skip any potential updates after that line 122)

trajepl (approved):

PR #598, opened 2023-09-22
🍵 Unit tests for same pass with different config in one olive config

trajepl:
## Describe your changes
Unit tests for same pass with different config in one olive config. To cover this case  https://github.com/microsoft/Olive/issues/573


guotuofeng:
this unit test is cover what we talked today about the bug fix. right?

trajepl:
Yes, add two different configs for one pass, then give 2 pass flows. 
This test is used to assert the two kind of pass instance are all run successfully.

image

guotuofeng (approved):

PR #599, opened 2023-09-22
🌊 Only_onnxruntime to false when opt_level > 1

trajepl:
## Describe your changes
1. Only_onnxruntime to false when opt_level > 1
2. set quantization calibration_method to minmax as the bug from ort 1.16.0 https://github.com/microsoft/onnxruntime/issues/17619


olive/passes/onnx/transformer_optimization.py
```diff
@@ -59,8 +59,19 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
             "only_onnxruntime": PassConfigParam(
                 type_=bool,
                 default_value=False,
-                searchable_values=Boolean(),
-                description="Whether only use onnxruntime to optimize model, and no python fusion.",
+                searchable_values=Conditional(
+                    parents=("opt_level",),
+                    support={
+                        (2,): Boolean([False]),
+                        (0,): Boolean(),
```
jambayk, line 66: not an issue but the default value for other params can be set by using https://github.com/microsoft/Olive/blob/main/olive/strategy/search_parameter.py#L127.

trajepl: ahah, thanks! I even think it is a bug where we do not have the default searchable options.

olive/strategy/search_parameter.py
```diff
@@ -80,8 +80,8 @@ class Boolean(Categorical):
     >>> Boolean()
     """
 
-    def __init__(self):
-        super().__init__([True, False])
+    def __init__(self, support: List[bool] = [True, False]):
```
jambayk, line 83: this has the same issue as using mutable objects as default values. 

if we only want one option, we can just use `Categorical([False])` instead. `Boolean` was created as a shorthand for `Categorical([True, False])`

examples/bert/bert_ptq_cpu.json
```diff
@@ -59,6 +60,7 @@
                 "quant_preprocess": true,
                 "per_channel": false,
                 "reduce_range": false,
+                "calibrate_method": "MinMax",
```
guotuofeng, line 63: add TODO here

trajepl: hard to comment for Json, I add some comment here.
image

olive/strategy/search_parameter.py
trajepl: Fixed.

olive/strategy/search_parameter.py
```diff
@@ -80,7 +80,7 @@ class Boolean(Categorical):
     >>> Boolean()
     """
 
-    def __init__(self):
+    def __init__(self, support: List[bool] = None):
         super().__init__([True, False])
```
guotuofeng, line 84: not pass default arrays, Instead, better pass None and let the Categorical handle Non

trajepl: Emm, Boolean() is used to directly create Categorical([True, False]). If we pass None, we can directly use Categorical.

(If the arguments `support` confuses you, I removed in latest commit as it is the zombie code.

guotuofeng: I mean 
image

olive/strategy/search_parameter.py
```diff
@@ -313,9 +313,10 @@ def json_to_search_parameter(json: Dict[str, Any]) -> SearchParameter:
     if search_parameter_type == "Categorical":
         return Categorical(json["support"])
     if search_parameter_type == "Conditional" or search_parameter_type == "ConditionalDefault":
-        stop_condition = lambda x: (  # noqa: E731
-            isinstance(x, dict) and x.get("olive_parameter_type") == "SearchParameter"
-        )
+
+        def stop_condition(x):
+            return isinstance(x, dict) and x.get("olive_parameter_type") == "SearchParameter"  # noqa: E731
```
guotuofeng, line 318: > # noqa: E731 [](http://example.com/codeflow?start=94&length=13)

no need this one?

trajepl: Good catch. This line is auto-formatted from my local linter plugin. But somehow the `noqa` was not removed.

guotuofeng (approved):

PR #600, opened 2023-09-22
Remove duplicate values from pytorch model json

jambayk:
## Describe your changes
Pytorch model adds values from the hf model config into model attributes to make them easily accessible. However, this makes the model json messy/confusing with values in model_attributes that are already in `model_config`. This PR removes information from model_attributes that is already in the hf_config. 

### DIFF
--- a/olive/model/__init__.py
+++ b/olive/model/__init__.py
@@ -725,11 +725,16 @@ class PyTorchModel(OliveModel):
                 "hf_config": self.hf_config,
             }
         )
-        # convert script_dir and model_script to string
-        # the original config has them as serialized ResourcePath
-        for resource_name in ["script_dir", "model_script"]:
-            if self.resource_paths[resource_name]:
-                config["config"][resource_name] = self.get_resource(resource_name)
+        # clean up redundant information in model_attributes
+        config["config"].pop("model_attributes", None)
+        # using a copy of self.model_attributes since config["config"]["model_attributes"] is already
+        # serialized and might not match self.model_attributes
+        model_attributes = deepcopy(self.model_attributes)
+        if model_attributes and self.hf_config:
+            for key, value in self.get_hf_model_config().to_dict().items():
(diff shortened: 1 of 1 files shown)


olive/model/__init__.py
```diff
@@ -737,11 +737,16 @@ def to_json(self, check_object: bool = False):
...
+        # clean up redundant information in model_attributes
+        config["config"].pop("model_attributes", None)
+        # using a copy of self.model_attributes since config["config"]["model_attributes"] is already
+        # serialized and might not match self.model_attributes
+        model_attributes = deepcopy(self.model_attributes)
+        if model_attributes and self.hf_config:
+            for key, value in self.get_hf_model_config().to_dict().items():
+                if key in model_attributes and model_attributes[key] == value:
+                    del model_attributes[key]
+        config["config"]["model_attributes"] = model_attributes or None
```
trajepl, line 749: If the input model is PytorchModel, olive will dump model_config directly without model_attributes no matter hf_config is None or not.

Is uour change used to cover the case where:
1. output model is pytorch model
2. output model might change the model attributes
3. output model has hf_config
right?

As in olive, for input model, olive dumps config from ModelConfig.to_json() not Model.to_json()

Curious that which case will trigger the inconsistent model_attributes.

jambayk: This is for when the output model is pytorch model for passes like qlora. It does not change the model attributes and only adds quantization config and adapter path. 

Since the model already has `hf_config`, it becomes messy or duplicated attributes if we directly save the model_attributes which were updated during init by using hf_config. So I wanted to remove the dynamically added attributed before serialization.

jambayk: We don't have any passes that change the model attributes yet. but that is something we might have to tackle in the future. For example a pruning pass changes the model architecture and hf_config but the model attributes still has old values. 

But this PR is not for that. It's just to clean up the json.

trajepl (approved):

PR #606, opened 2023-09-25
Add AML curated model support

xiaoyu-work:
## Describe your changes

Add AML curated model support.

### DIFF
--- a/examples/deberta/azureml_registry_config.json
+++ b/examples/deberta/azureml_registry_config.json
@@ -1,0 +1,92 @@
+{
+    "input_model":{
+        "type": "PyTorchModel",
+        "config": {
+            "model_path": {
+                "type": "azureml_registry_model",
+                "config": {
+                    "registry_name": "azureml",
+                    "name": "microsoft-deberta-base-mnli",
+                    "version": "9"
+                }
+            },
+            "model_file_format": "PyTorch.MLflow",
+            "hf_config": {
+                "model_name": "microsoft/deberta-base-mnli",
+                "task": "text-classification",
+                "dataset": {
+                    "data_name":"glue",
+                    "subset": "mnli_matched",
+                    "split": "validation",
+                    "input_cols": ["premise", "hypothesis"],
+                    "label_cols": ["label"],
+                    "batch_size": 1,
+                    "max_samples": 100,
+                    "component_kwargs": {
+                        "pre_process_data": {
+                            "align_labels": true
+                        }
+                    }
+                }
+            }
+        }
+    },
+    "evaluators": {
+        "common_evaluator": {
+            "metrics":[
+                {
+                    "name": "latency",
+                    "type": "latency",
+                    "sub_types": [
+                        {"name": "avg", "priority": 1}
+                    ]
+                }
+            ]
+        }
+    },
+    "passes": {
+        "conversion": {
+            "type": "OnnxConversion",
+            "config": {
+                "target_opset": 13
+            }
+        },
+        "transformers_optimization": {
+            "type": "OrtTransformersOptimization",
+            "disable_search": true,
+            "config": {
+                "model_type": "bert",
+                "num_heads": 12,
+                "hidden_size": 768,
+                "float16": false
+            }
+        },
+        "quantization": {
+            "type": "OnnxQuantization",
+            "config": {
+                "data_config": "__input_model_data_config__"
+            }
+        },
+        "perf_tuning": {
+            "type": "OrtPerfTuning",
+            "config": {
+                "data_config": "__input_model_data_config__"
+            }
+        }
+    },
+    "engine": {
+        "search_strategy": {
+            "execution_order": "joint",
+            "search_algorithm": "tpe",
+            "search_algorithm_config": {
+                "num_samples": 3,
+                "seed": 0
+            }
+        },
+        "clean_cache": true,
+        "evaluator": "common_evaluator",
+        "execution_providers": ["CPUExecutionProvider"],
+        "cache_dir": "cache",
+        "output_dir" : "models/microsoft-deberta"
+    }
+}
--- a/olive/azureml/azureml_client.py
+++ b/olive/azureml/azureml_client.py
@@ -52,13 +52,6 @@ class AzureMLClientConfig(ConfigBase):
 
     @validator("aml_config_path", always=True)
     def validate_aml_config_path(cls, v, values):
-        if v is None:
-            if values.get("subscription_id") is None:
-                raise ValueError("subscription_id must be provided if aml_config_path is not provided")
-            if values.get("resource_group") is None:
-                raise ValueError("resource_group must be provided if aml_config_path is not provided")
-            if values.get("workspace_name") is None:
-                raise ValueError("workspace_name must be provided if aml_config_path is not provided")
         if v is not None:
             if not Path(v).exists():
                 raise ValueError(f"aml_config_path {v} does not exist")
@@ -89,6 +82,12 @@ class AzureMLClientConfig(ConfigBase):
         logging.getLogger("azure.identity").setLevel(logging.ERROR)
 
         if self.aml_config_path is None:
+            if self.subscription_id is None:
+                raise ValueError("subscription_id must be provided if aml_config_path is not provided")
+            if self.resource_group is None:
+                raise ValueError("resource_group must be provided if aml_config_path is not provided")
+            if self.workspace_name is None:
+                raise ValueError("workspace_name must be provided if aml_config_path is not provided")
             return MLClient(
                 credential=self._get_credentials(),
                 subscription_id=self.subscription_id,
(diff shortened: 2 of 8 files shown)


olive/resource_path.py
```diff
@@ -344,6 +350,66 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
         return str(new_path)
 
 
+class AzureMLRegistryModel(ResourcePath):
```
jambayk, line 353: since `save_to_dir` for this class is very similar to that of `AzureMLModel`, could you abstract the logic to avoid duplicate code? Perhaps using a common base classs or helper function for download?

olive/azureml/azureml_client.py
```diff
@@ -101,6 +101,22 @@ def create_client(self):
                 credential=self._get_credentials(), path=self.aml_config_path, read_timeout=self.read_timeout
             )
 
+    def create_registry_client(self, registry_name: str):
```
guotuofeng, line 104: could we merge it with current create_client? It seems most of the logic is same.

olive/resource_path.py
xiaoyu-work: Done.

olive/azureml/azureml_client.py
xiaoyu-work: I updated this logic. Now registry client doesn't need workspace info.

olive/azureml/azureml_client.py
```diff
@@ -41,7 +41,7 @@ class AzureMLClientConfig(ConfigBase):
     # but DefaultAzureCredential accept kwargs as parameters, it is hard to validate the config.
     # so we just provide a dict here and let the user to provide the correct config following the doc.
     default_auth_params: Optional[Dict[str, Any]] = Field(
-        None,
+        {},
```
jambayk, line 44: It is probably fine since pydantic doesn't appear to use the default value by reference. But it might be safer to use `Field(default_factory=dict, ...)` to be safe? In case there are unexpected issues with using a mutable object as default?

olive/resource_path.py
```diff
@@ -34,7 +35,11 @@ def __str__(self) -> str:
 
 
 LOCAL_RESOURCE_TYPES = [ResourceType.LocalFile, ResourceType.LocalFolder]
-AZUREML_RESOURCE_TYPES = [ResourceType.AzureMLModel, ResourceType.AzureMLDatastore, ResourceType.AzureMLJobOutput]
+AZUREML_RESOURCE_TYPES = [
+    ResourceType.AzureMLModel,
+    ResourceType.AzureMLDatastore,
+    ResourceType.AzureMLJobOutput,
```
jambayk, line 41: Is there a reason `AzureMLRegistryModel` is not added here?

olive/resource_path.py
```diff
@@ -329,21 +325,93 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
         logger.debug(f"Downloading model {self.config.name} version {self.config.version} to {new_path}.")
         from azure.core.exceptions import ServiceResponseError
 
+        max_tries = self.config.azureml_client.max_operation_retries if self.config.azureml_client else 3
```
jambayk, line 328: I don't think this line is needed since both children classes provide the value?

xiaoyu-work: Oh I forgot to remove this. Thanks!

olive/resource_path.py
```diff
@@ -329,21 +325,93 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
...
+        }
+
+    def get_path(self) -> str:
+        return (
+            f"azureml://registries/{self.config.registry_name}/models/{self.config.name}/versions/{self.config.version}"
+        )
+
+    def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: bool = False) -> str:
+        # azureml client
+        default_auth_params = {}
```
jambayk, line 401: nit: something like 
`azureml_client = self.config.azureml_client or AzureMLClientConfig()`
could allow us to not hardcode default values here. Can get it directly from `AzureMLClientConfig`

xiaoyu-work: This looks better for me. I doubted the defualt values for each fileds before but seems they are all good. Will update this.

olive/resource_path.py
xiaoyu-work: We are doing a couple of validations based on aml resource type like workspace validation, and if aml client exist validation for those aml user workspace assets. Those validations looks more like for user-aml-workspace. AML registry model basically doesn't need those validation. It is just a cloud storage place to download the model. So instead of updating existing validation logic, I removed it from aml resource type.

olive/azureml/azureml_client.py
xiaoyu-work: Should `default_factory` in pydantic mainly be used for a callable in order to generate a dynamic default value?

olive/resource_path.py
jambayk: Thanks! Makes sense to me. 
Can you update this mapping https://github.com/microsoft/Olive/blob/main/olive/systems/azureml/aml_system.py#L41 so that aml model can support aml registry model? 
the new type just needs to map to custom_model.

olive/azureml/azureml_client.py
jambayk: Looks like default_factory is not necessary. It's just that we have been avoiding using `{}` as default values (instead use config.x or {}) but we use other mutable objects as default values. 
So, it is probably fine to use `{}` as default.

olive/resource_path.py
xiaoyu-work: Will update this. Question: after preparing non-local model (download it), the resource type won't be updated to local resource right?

olive/azureml/azureml_client.py
```diff
@@ -89,38 +82,59 @@ def create_client(self):
...
-            credential.get_token("https://management.azure.com/.default")
-            logger.debug("Using DefaultAzureCredential")
-        except Exception:
-            # Fall back to InteractiveBrowserCredential in case DefaultAzureCredential not work
-            credential = InteractiveBrowserCredential()
-            logger.debug("Using InteractiveBrowserCredential")
-
-        return credential
+    @staticmethod
+    def create_registry_client(registry_name: str, **kwargs):
```
guotuofeng, line 106: >kwargs [](http://example.com/codeflow?start=53&length=6)

it seems the kwargs only use default_auth_params? if yes, we could make the parameter explicitly

olive/azureml/azureml_client.py
guotuofeng: I think None as default value should be better than a default empty dict. 

Instead, whenever using the default_auth_params we need do validation

olive/azureml/azureml_client.py
```diff
@@ -89,38 +82,59 @@ def create_client(self):
...
+        """Create an MLClient instance."""
+        from azure.ai.ml import MLClient
+
+        # set logger level to error to avoid too many logs from azure sdk
+        logging.getLogger("azure.ai.ml").setLevel(logging.ERROR)
+        logging.getLogger("azure.identity").setLevel(logging.ERROR)
+
+        default_auth_params = kwargs.get("default_auth_params", {})
+
```
guotuofeng, line 115: Move this kind of logic into _get_credentials

olive/resource_path.py
jambayk: The preparation only happens for non-aml system, so if both host and target are aml, it won't be downloaded.

olive/resource_path.py
```diff
@@ -335,15 +330,72 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
...
+    def get_path(self) -> str:
+        return (
+            f"azureml://registries/{self.config.registry_name}/models/{self.config.name}/versions/{self.config.version}"
+        )
+
+    def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: bool = False) -> str:
+        # azureml client
+        azureml_client_config = self.config.azureml_client or AzureMLClientConfig()
+
+        ml_client = AzureMLClientConfig.create_registry_client(
```
jambayk, line 393: this doesn't really matter, but if we use `aml_client_config.create_registry_client`, we won't need to expose the `default_auth_params` parameter. 
This also makes the `create_client` and `create_registry_client` signatures similar and cleaner.

olive/azureml/azureml_client.py
jambayk: I agree `None` as default looks better and follows what we do for other parameters in olive (where we use `None` instead and just do `param or {}` when needed).

olive/resource_path.py
xiaoyu-work: I'm not sure if removing `default_auth_params` for registry client will work for some scenarios that needs it. I personally sometimes needs this arg for aml client config locally.

jambayk: oh, I didn't mean to remove it entirely. 
What I meant is since we already have an object `azureml_client_config`, it seems cleaner to just do 
```
ml_client = azureml_client_config.create_registry_client(self.config.registry_name)
```

This way, `create_registry_client` doesn't need the `default_auth_params` since it can get it from it's own `self.default_auth_params` like `create_client` does.

examples/deberta/azureml_registry_config.json
```diff
@@ -0,0 +1,97 @@
+{
+    "azureml_client": {
```
jambayk, line 2: is this required here? since azureml_client is optional for registry model

olive/resource_path.py
xiaoyu-work: Yeah I missed that we have a aml client config instance now. I'll update this.

examples/deberta/azureml_registry_config.json
xiaoyu-work: Good catch! Removed.

jambayk (approved):
Thanks for making the changes!

jambayk (approved):

PR #610, opened 2023-09-26
🫒 Olive auto optimizer

trajepl:
## Describe your changes
 Olive auto optimizer poc

Add the infrastructure of olive auto optimizer to search passes configs/flows automatically. User can ignore passes configs now for bert models now.


olive/engine/engine.py
```diff
@@ -49,6 +52,7 @@ def __init__(
         target: Optional[OliveSystem] = None,
         evaluator_config: Optional[OliveEvaluatorConfig] = None,
         execution_providers: Optional[List[str]] = None,
+        auto_optimizer_config: Optional[AutoOptimizerConfig] = None,
```
guotuofeng, line 55: could we move the auto_optimizer_config out of the engine class? It seems Engine itself doesn't care about the auto_optimizer.

olive/engine/engine.py
```diff
@@ -344,8 +402,23 @@ def run(
 
         for accelerator_spec in self.accelerator_specs:
             with self.create_managed_environment(accelerator_spec):
+                if self.enable_auto_optimizer:
```
guotuofeng, line 405: engine itself should be transparent about whether the pass is from auto-optimizer or registered by user explicitly. 

From this point of view, the engine just do register and run. prepare_passes_from_auto_optimizer should be splited into pass generation and pass registration. For pass generation, it should be outside this class's responsibility.

trajepl: Will refactor it in following PR.

olive/engine/engine.py
trajepl: As we discussed offline, will refactor this part with accelerator_spec setting up together.

guotuofeng (changes requested):
Is the PR a draft or ready to review?

olive/auto_optimizer/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+            if not resource_path or resource_path.is_local_resource_or_string_name():
+                continue
+            downloaded_resource_path = cache_utils.download_resource(resource_path, self._config.cache_dir)
+            if downloaded_resource_path:
+                # set local resource path
+                model_config.config[resource_name] = downloaded_resource_path
+        model = model_config.create_model()
+        self.model_attr = model.model_attributes or {}
+        _model_type = self.model_attr.get("model_type", None)
+        self.model_type = MODEL_TYPE_MAPPING.get(_model_type, _model_type)
```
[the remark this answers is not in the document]
trajepl: It belongs to model attributes.
For huggingface model, the model_type can be extracted from model config automatically. User do not need to specify that. We use the model_type to adopt different optimization config template at this early stage.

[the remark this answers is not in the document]
devang-ml: What about non-huggingface models?

olive/auto_optimizer/config_template/bert.yaml
```diff
@@ -0,0 +1,16 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
devang-ml, line 2: We should not need config templates to determine a config parameters of a Pass.

olive/auto_optimizer/template_mapping.py
```diff
@@ -0,0 +1,36 @@
...
+import yaml
+
+logger = logging.getLogger(__name__)
+
+TEMPLATE_CONFIG_PATH = Path(__file__).resolve().parent / "config_template"
+
+model_template_mapping = {
+    "bert": [["OnnxConversion", "OrtTransformersOptimization", "OnnxQuantization", "OrtPerfTuning"]],
+    "open_llama": [["OptimumConversion", "OrtTransformersOptimization", "OptimumMerging"]],
+}
```
devang-ml, line 19: We should not need template mapping to determine if we need OnnxConversion or OptimumConversion.

olive/auto_optimizer/__init__.py
```diff
@@ -0,0 +1,232 @@
...
+            if "OrtPerfTuning" in pass_config:
+                # if trt ep, disable fp16 for transformers optimization but enable fp16 for perf_tuning
+                self.set_config_value_if_not_exist(pass_config["OrtPerfTuning"], "trt_fp16_enable", trt_fp16)
+                # enable_cuda_graph only for cuda ep
+                self.set_config_value_if_not_exist(pass_config["OrtPerfTuning"], "enable_cuda_graph", cuda_fp16)
+                self.set_config_value_if_not_exist(pass_config["OrtPerfTuning"], "io_bind", True)
+
+    def regulate_passes(self, pass_config, pass_flows):
+        # remove passes that are not supported by execution provider
+        to_del_passes = set(self.auto_optimizer_config.disabled_passes) or set()
```
devang-ml, line 193: This should be the responsibility of the Pass itself, not the auto_optimizer or engine.
Imagine a Pass contributed by external developer for external target.

devang-ml (commented):
Overall, Olive should not determine sequence of passes based on whether the model is 'bert' or 'llama'.  If the user is aware of the model architecture details then user knows what kind of quantization precision they'd like to try on so on. They are not the intended target of the auto-optimization feature.

The auto-optimization feature should select pass sequence based whether the model is PT model or HF model or ONNX model etc., and what is the target accelerator and what EP is being tried. So the templates used by auto-optimization should also be based on these constraints and not whether the model is llama or not.

olive/auto_optimizer/__init__.py
trajepl: This part is a bit different with the config pruning or filtering which are used to cut off invalid configs for a specific given pass.

The `regulate_passes` is about to globally regulate passes combinations, like the cases of :
- quantization and fp16 cannot appear simultaneously
- cuda fp16(transformer optimization) and trt fp16(perf_tuning) cannot appear at the same time.

It can improve the olive optimization/search efficiency with global regulating in advance.

trajepl:
> Overall, Olive should not determine sequence of passes based on whether the model is 'bert' or 'llama'. If the user is aware of the model architecture details then user knows what kind of quantization precision they'd like to try on and so on. They are not the intended target of the auto-optimization feature.
> 
> The auto-optimization feature should select pass sequence based whether the model is PT model or HF model or ONNX model etc., and what is the target accelerator and what EP is being tried. So the templates used by auto-optimization should also be based on these constraints and not whether the model is llama or not.

I totally agree that auto optimization feature should select passes based the constraints: model platform/target accelerator/Evaluation target etc. Based on this, the generic progress for auto optimization is like:
1. generate different pass sequences(flow) based on constraints + opt_level(or strategy)
2. regulate pass sequences and configs combination from#1 in advance
image

This PR is used to introduce this generic progress, at the same time I just given a very simple strategy to generate pass sequences **by model type(called opt_level == 0 in this PR) to look up template which we verified already is working well** 
image

Also even for opt_level==0, we can keep updating the templates for SD/whisper and other models we verified already. Not only `bert` or `llama`.

And the feature of auto optimization will be iterated actively with more generic strategies.

olive/auto_optimizer/config_template/bert.yaml
trajepl: These kind of model.yaml and tempalte_mapping.py are only for opt_level 0 where we collect the verified passes sequences(flows) and configs to speed up optimization.

olive/auto_optimizer/template_mapping.py
trajepl: Same with <model>.yaml

olive/auto_optimizer/config_template/bert.yaml
trajepl: And the feature of auto optimization will be iterated actively with more generic strategies.
image

devang-ml (commented):
In that case we want to skip the opt level 0 that looks for model_type. That is a distraction.

devang-ml (commented):
Regarding
>>> 1. generate different pass sequences(flow) based on constraints + opt_level(or strategy)


You don't need to "generate" different pass sequences. You can just "select" a template for given target/precision/ep. You would need fixed set of templates.

devang-ml (changes requested):
we may not need regulate feature.

trajepl:
Closed and collect the comments to refine. Will send it later.

PR #619, opened 2023-09-28
Few fixes to get MoE experts distribution to work after recent changes to Olive

shaahji:
Few fixes to get MoE experts distribution to work after recent changes to Olive

### DIFF
--- a/olive/passes/onnx/moe_experts_distributor.py
+++ b/olive/passes/onnx/moe_experts_distributor.py
@@ -43,6 +43,7 @@ _expert_pattern = [
 
 
 def _dump_graph(node: Message, filepath: str):
+    Path(filepath).parent.mkdir(parents=True, exist_ok=True)
     with open(filepath, "w") as strm:  # noqa: PTH123
         json.dump(MessageToDict(node), fp=strm, indent=2, sort_keys=True, separators=_json_separators)
         strm.flush()
--- a/examples/switch/config.json
+++ b/examples/switch/config.json
@@ -3,7 +3,7 @@
     "input_model":{
         "type": "OnnxModel",
         "config": {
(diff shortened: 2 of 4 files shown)


olive/passes/onnx/moe_experts_distributor.py
```diff
@@ -43,6 +43,7 @@
 
 
 def _dump_graph(node: Message, filepath: str):
+    Path(filepath).parent.mkdir(parents=True, exist_ok=True)
```
trajepl, line 46: Double check: do you expect to create folder in **parent folder** of **parent folder** of `filepath`? `cd ../, mkdir -p xxx`?

guotuofeng: it seems the parent folder is already created before calling the internal function?

trajepl: Yes, not sure it is by design or something else. Please help double confirm.

shaahji: _filepath_ is a path to a file not a directory. So _Path(filepath).parent_ would be a directory. _mkdir_ on that seems correct.

Create the directory in which to create the file. I don't see an issue here.

guotuofeng: The line seems needless by considering the parent folder is already created when the function is called. What's the scenario that need this line?

shaahji: The function is called from _run_ function and I don't see any calls to _mkdir_. `_dump_graph` is a debug utility function and I use it as such, for debugging. Having the call in one place to create the parent folders before writing out the file doesn't seem wrong.

trajepl (approved):

PR #607, opened 2023-09-26
Enable more linter in adapter config

guotuofeng:
## Describe your changes
Enable the following linters
* mypy
* noqa
* space
* tab
* toml-sort
## Checklist before requesting a review
- [ ] Add unit tests for this change.
- [ ] Make sure all tests can pass.
- [ ] Update documents if necessary.
- [ ] Format your code by running `pre-commit run --all-files`
- [ ] Is this a user-facing change? If yes, give a description of this change to be included in the release notes.

PR #634, opened 2023-10-07
Fix footprint metric goal bug

xiaoyu-work:
## Describe your changes

- Fix `if_goals_met` logic bug when `cmp_direction == -1`. 
- Fix grammar error.
- Add `lintrunner` and `lintrunner-adapters` to dev requirements.

### DIFF
--- a/examples/bert/bert_ptq_cpu.json
+++ b/examples/bert/bert_ptq_cpu.json
@@ -24,7 +24,7 @@
                     "type": "accuracy",
                     "backend": "huggingface_metrics",
                     "sub_types": [
-                        {"name": "accuracy", "priority": 1, "goal": {"type": "max-degradation", "value": 0.02}},
+                        {"name": "accuracy", "priority": 1, "goal": {"type": "max-degradation", "value": 0.01}},
                         {"name": "f1"}
                     ]
                 },
@@ -49,9 +49,12 @@
         },
         "transformers_optimization": {
             "type": "OrtTransformersOptimization",
+            "disable_search": true,
             "config": {
-                "float16": false,
-                "only_onnxruntime": false
+                "model_type": "bert",
+                "num_heads": 12,
+                "hidden_size": 768,
+                "float16": false
             }
         },
         "quantization": {
--- a/examples/bert/bert_ptq_cpu.json
+++ b/examples/bert/bert_ptq_cpu.json
@@ -24,7 +24,7 @@
                     "type": "accuracy",
                     "backend": "huggingface_metrics",
                     "sub_types": [
-                        {"name": "accuracy", "priority": 1, "goal": {"type": "max-degradation", "value": 0.01}},
+                        {"name": "accuracy", "priority": 1, "goal": {"type": "max-degradation", "value": 0.05}},
                         {"name": "f1"}
                     ]
                 },
@@ -32,7 +32,7 @@
                     "name": "latency",
                     "type": "latency",
                     "sub_types": [
-                        {"name": "avg", "priority": 2, "goal": {"type": "percent-min-improvement", "value": 10}},
+                        {"name": "avg", "priority": 2, "goal": {"type": "percent-min-improvement", "value": 5}},
                         {"name": "max"},
                         {"name": "min"}
                     ]
--- a/examples/bert/bert_ptq_cpu.json
+++ b/examples/bert/bert_ptq_cpu.json
@@ -24,7 +24,7 @@
                     "type": "accuracy",
                     "backend": "huggingface_metrics",
                     "sub_types": [
-                        {"name": "accuracy", "priority": 1, "goal": {"type": "max-degradation", "value": 0.01}},
+                        {"name": "accuracy", "priority": 1, "goal": {"type": "max-degradation", "value": 0.02}},
                         {"name": "f1"}
                     ]
                 },
@@ -32,7 +32,7 @@
                     "name": "latency",
                     "type": "latency",
                     "sub_types": [
-                        {"name": "avg", "priority": 2, "goal": {"type": "percent-min-improvement", "value": 20}},
+                        {"name": "avg", "priority": 2, "goal": {"type": "percent-min-improvement", "value": 10}},
                         {"name": "max"},
                         {"name": "min"}
                     ]
--- a/examples/bert/bert_ptq_cpu.json
+++ b/examples/bert/bert_ptq_cpu.json
@@ -49,13 +49,7 @@
         },
         "transformers_optimization": {
             "type": "OrtTransformersOptimization",
-            "disable_search": true,
-            "config": {
-                "model_type": "bert",
-                "num_heads": 12,
-                "hidden_size": 768,
-                "float16": false
-            }
+            "disable_search": true
         },
         "quantization": {
             "type": "OnnxQuantization",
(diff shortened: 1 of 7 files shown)


requirements-dev.txt
```diff
@@ -2,5 +2,7 @@
 addict
 black
 flake8
+lintrunner
```
guotuofeng, line 5: this is needless https://github.com/microsoft/Olive/blob/main/.github/workflows/lint.yml#L45

xiaoyu-work: I added this to dev file mainly for other contributors. Since we are using lint for PR validation, they should have this package installed from requirements-dev file. And know how to use it, which means we also need to update [Contributing](https://github.com/microsoft/Olive/blob/main/CONTRIBUTING.md#contributing)

guotuofeng: Then we'd better add another requirement file. the dev requirement is used by others like testing, which will result the two packages are installed for all the test code. Ideally, the two packages should only be installed when needed.

guotuofeng: On another hand, any PR would trigger the lint CI pipelines. I guess it is only used locally. right? If yes, we can add the two lines in readme.

guotuofeng (changes requested):
the linter packages should not be installed in dev requirements.

requirements-dev.txt
xiaoyu-work: We already have `requirements-test.txt` for test dependencies? This `requirements-dev.txt` is for dev purpose.

guotuofeng: Oh, sorry, it seems requirements-test.txt doesn't include requirements-dev.txt. My bad.

However, I am still prefer not put the two dependency into the requirements-dev.txt since besides installing them, the init is still needed. From this point of view, it would be better to add this stuff into readme.

guotuofeng: Another example: https://github.com/microsoft/onnxruntime/blob/main/requirements-dev.txt

xiaoyu-work: I think `requirements-dev.txt` in onnxruntime repo has different meaning. This file in onnxruntime repo means dev env for onnxruntime package in order to build onnxruntime from source, like [this](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/models/stable_diffusion/README.md?plain=1#L102). Since Olive doesn't need this feature, the packages now listed in our `requirements-dev.txt` are more like dev tools for contributers, like `black`, `flake8` or `pre-commit`. I agree we should update documentation about how to use lint, but I prefer those are documented in Contributing, like [this](https://github.com/microsoft/onnxruntime/blob/main/docs/Coding_Conventions_and_Standards.md?plain=1#L153).

Also I believe most of the packages in our `requirements-dev.txt` are not needed anymore. We can remove others but only keep lint packages.

guotuofeng: could the pre-commit https://github.com/microsoft/Olive/pull/607/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9 accomplish these things?

guotuofeng: Originally, the purpose of requirements-dev.txt is same with that one in ORT. The typical user scenario is:
* user git clone code
* pip install -r requirements-dev.txt.

requirements-dev.txt
```diff
@@ -2,5 +2,7 @@
 addict
```
guotuofeng, line 2: curious why do dev need the addict?

requirements-dev.txt
guotuofeng: If you add the linter runner here, please remember remove those in the workflow.yaml

xiaoyu-work: Do we need both pre-commit and lint? I thought we will use lint to replace pre-commit?

My expectations for each requirements files in Olive are:
- `requirements.txt`: Olive dependencies.
- `requirements-test.txt`: Olive tests dependencies to run tests.
- `requirements-dev.txt`: for contributors to help them to raise a PR. This should fully support necessary packages for each PR validation, which means this should include `requirements-test.txt`.

requirements-dev.txt
xiaoyu-work: Not sure why this was added. I can remove this.

requirements-dev.txt
xiaoyu-work: I can clean up the requirements files and remove unneeded packages.

xiaoyu-work: I will also update the contributing doc.

guotuofeng: @trajepl tried to remove the pre-commit last week. But failed. So from the time being, we need both. Jiapeng would have more context.

guotuofeng: Please help add one section like https://github.com/microsoft/Olive#formatting for enabling linter

guotuofeng: > Do we need both pre-commit and lint? I thought we will use lint to replace pre-commit?
> 
> My expectations for each requirements files in Olive are:
> 
> * `requirements.txt`: Olive dependencies.
> * `requirements-test.txt`: Olive tests dependencies to run tests.
> * `requirements-dev.txt`: for contributors to help them to raise a PR. This should fully support necessary packages for each PR validation, which means this should include `requirements-test.txt`.

Please make requirements-dev.txt doesn't include requirements-test.txt. Basically, requirements-dev.txt is mainly for develop purpose like run flake, black format, linter etc. The test is not mandatory since it is optional run

trajepl:
The CI failed for some e2e tests since you tight the constraints up then no expected model output. Please try to modify the metrics goal to ensure at least one model in olive outputs.

trajepl (dismissed):

examples/bert/bert_ptq_cpu.json
```diff
@@ -49,9 +49,12 @@
         },
         "transformers_optimization": {
             "type": "OrtTransformersOptimization",
+            "disable_search": true,
             "config": {
-                "float16": false,
```
guotuofeng, line 53: @trajepl , is this ok to remove only_onnxruntime?

trajepl: It is fine, the default value is False. Just ensure there is at least one model conducted from Olive run.

examples/bert/bert_ptq_cpu.json
```diff
@@ -49,9 +49,12 @@
...
-                "only_onnxruntime": false
+                "model_type": "bert",
```
trajepl, line 54: BTW, you can remove model_type/num_heads/hidden_size/float16 in this case. Olive will extract and fill in these values from model_attributes.

requirements-dev.txt
```diff
@@ -1,6 +1,6 @@
 -r requirements.txt
-addict
 black
 flake8
+lintrunner
+lintrunner-adapters
```
guotuofeng, line 5: do we need remove line https://github.com/microsoft/Olive/blob/main/.github/workflows/lint.yml#L46?

guotuofeng (approved):

examples/bert/bert_ptq_cpu.json
xiaoyu-work: Removed.

requirements-dev.txt
xiaoyu-work: Removed.

trajepl (approved):

PR #635, opened 2023-10-09
enable stable_diffusion_xl with fp16 optimization by skipping a set o…

zhangxiang1993:


examples/directml/stable_diffusion_xl/config_vae_decoder.json
```diff
@@ -75,7 +75,43 @@
                     "enable_packed_kv": true,
                     "enable_bias_add": false
                 },
-                "force_fp32_ops": ["RandomNormalLike"]
+                "force_fp32_ops": ["RandomNormalLike"],
+                "force_fp32_nodes": [
+                    "/decoder/up_blocks.2/upsamplers.0/conv/Conv",
```
zhangxiang1993, line 80: Starting from the Conv node, output is overflow by casting to float16. Wrapping this single node also causing overflow, so all the nodes after it need to be skipped.

examples/directml/stable_diffusion_xl/stable_diffusion_xl.py
```diff
@@ -267,14 +267,6 @@ def optimize(
         with open(script_dir / f"config_{submodel_name}.json", "r") as fin:
             olive_config = json.load(fin)
 
-        # TODO: Remove this once we figure out which nodes are causing the black screen
-        if is_refiner_model and submodel_name == "vae_encoder":
```
PatriceVignola, line 271: Your fix only targets vae_decoder right? So the issue in vae_encoder should still be there. We don't need to address vae_encoder in this PR, but we should keep that part disabled.

olive/passes/onnx/transformer_optimization.py
```diff
@@ -76,6 +76,9 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
             "force_fp32_ops": PassConfigParam(
```
guotuofeng, line 76: could we use op_block_list/node_block_list as the parameters so that the transformer optimization could be more generic?

PatriceVignola: This will break existing configs that use `force_fp32_ops`

examples/directml/stable_diffusion_xl/stable_diffusion_xl.py
zhangxiang1993: skipped the nodes in vae_decoder fixed the black screen issue, so vae_encoder should not be an issue anymore.

PatriceVignola: Did you test it with the refiner model, or only SDXL?

olive/passes/onnx/transformer_optimization.py
guotuofeng: we could add a TODO and change force_fp32_nodes to node_block_list

examples/directml/stable_diffusion_xl/config_vae_decoder.json
```diff
@@ -79,7 +79,43 @@
                         "GroupNorm": [0, 1, 2]
                     }
                 },
-                "force_fp32_ops": ["RandomNormalLike"]
+                "force_fp32_ops": ["RandomNormalLike"],
+                "force_fp32_nodes": [
+                    "/decoder/up_blocks.2/upsamplers.0/conv/Conv",
+                    "/decoder/up_blocks.3/resnets.0/conv_shortcut/Conv",
+                    "Transpose_NCHW_to_NHWC_22",
```
PatriceVignola, line 86: Hmm you're probably not using the most recent branch from main or ORT 1.16. These transposes should not be there anymore because `group_norm_channels_last` is set to false.

PatriceVignola (approved):

olive/passes/onnx/transformer_optimization.py
guotuofeng: @trajepl / @jambayk , it seems force_fp32_nodes is used as node block list. I prefer the name to be consistent with that of transformer optimization. what's you thought?

jambayk: I think in the context of the transformer optimization pass, `force_fp32_ops/nodes` make more sense. Otherwise, the names `op/node_block_list` don't give any information about their uses. 
This conflict is present because we are using this pass to do the actual "transformer optimization" and "float_to_float16" steps. `op/node_block_list` are inputs to the float_to_float16 tool so their names are not clear when used in this pass.

jambayk: Not related to this PR but we should also remove `force_fp16_inputs` from `optimization_options` (which is used for fusion options only) and move it to pass config level. 
It is not used by the `optimize_model` code. Only by the fp16 conversion tool https://github.com/search?q=repo%3Amicrosoft%2Fonnxruntime%20force_fp16_inputs&type=code.

examples/directml/phi_1/optimize
jambayk, line 1: Was this file accidentally added?

PR #638, opened 2023-10-12
Add LoRA pass

xiaoyu-work:
## Describe your changes

Add LoRA pass


test/unit_test/passes/pytorch/test_lora.py
```diff
@@ -8,16 +8,59 @@
...
 from olive.passes.olive_pass import create_pass_from_dict
-from olive.passes.pytorch import QLoRA
+from olive.passes.pytorch import LoRA, QLoRA
 
 
 def patched_find_all_linear_names(model):
     return ["k_proj", "v_proj", "out_proj", "q_proj", "fc1", "fc2"]
 
 
+def test_lora(tmp_path):
```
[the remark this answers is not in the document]
xiaoyu-work: Updated test.

olive/passes/pytorch/lora.py
```diff
@@ -223,6 +305,12 @@ def _run_for_config(
             )
             config.training_args.evaluation_strategy = "no"
 
+        # Grandient checkpointing seems to be enabled by specifying requires_grad_(True)
+        # for the first parameter of the model.
+        # However, LoRA training does not train that parameter
+        if config.training_args.gradient_checkpointing:
```
jambayk, line 311: is this part required for lora in general or only for qlora? For qlora, it is already handles by `prepare_model_for_kbits_training` https://github.com/huggingface/peft/blob/main/src/peft/utils/other.py#L66

olive/passes/pytorch/lora.py
```diff
@@ -114,43 +116,34 @@ def create_training_args(self) -> transformers.TrainingArguments:
...
+            "target_modules": PassConfigParam(type_=List[str], default_value=None, description="Target modules"),
+            "lora_alpha": PassConfigParam(
+                type_=float, default_value=16, description="The alpha parameter for Lora scaling."
             ),
-            "double_quant": PassConfigParam(
+            "lora_dropout": PassConfigParam(
+                type_=float, default_value=0.0, description="The dropout probability for Lora layers."
+            ),
+            "bias": PassConfigParam(type_=str, default_value="none", description="Bias type for Lora"),
+            "modules_to_save": PassConfigParam(
```
jambayk, line 134: should we expose `target_modules` and `modules_to_save` only in the actual `LoRA` pass? Since qlora ignores these parameters, it is confusing to expose them in the base config itself.

olive/passes/pytorch/lora.py
```diff
@@ -114,43 +116,34 @@ def create_training_args(self) -> transformers.TrainingArguments:
...
-    This pass only supports PyTorchModel with hf_config.
-    """
+class LoRABase(Pass):
+    """Base class for LoRA and QLoRA fine-tuning passes."""
 
     @staticmethod
     def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
         return {
-            # quantization parameters
-            "compute_dtype": PassConfigParam(
```
jambayk, line 129: I think this might still be useful for lora training. It can be used to set `torch_dtype` in the `model_loading_args` so that we can train the model in fp16 or bf16 if needed. 
Could rename the parameter to `torch_dtype` or something and update the description: data type for model weights and adapter weights. For 4bit quantized model, it is also the computation data type for the quantized modules.

olive/passes/pytorch/lora.py
```diff
@@ -187,26 +180,115 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
...
+            eval_dataset = split_data["test"]
+        else:
+            # eval data config has not been provided, and eval_dataset_size has not been provided
+            eval_dataset = None
+
+        return train_dataset, eval_dataset
+
+    def enable_lora(
+        self, model, tokenizer, task: str, config: ConfigBase, target_modules: Optional[List[str]]
     ) -> PyTorchModel:
```
jambayk, line 245: the output of this method is not `PytorchModel` but `PeftModel` or simply `torch.nn.Module`

jambayk: could you also add type hints for `model` and `tokenizer` so that we don't confuse the type for `model`? `run_for_config` has it as pytorchmodel but now it's a torch model or transformers model.

olive/passes/pytorch/lora.py
xiaoyu-work: Sounds good to me. Will update it.

olive/passes/pytorch/lora.py
xiaoyu-work: Yeah I just noticed that qlora doesn't need this. I'll update this.

olive/passes/pytorch/lora.py
xiaoyu-work: Will do

olive/passes/pytorch/lora.py
xiaoyu-work: Yes this is required for lora in general. Qlora handles this but general lora doesn't.

jambayk: I see. I was looking for normal lora fine-tuning examples but couldn't find one that just uses a transformers model.

xiaoyu-work: I guess this is specific for llama model maybe? I saw several issue in other github repos discussing this like https://github.com/lm-sys/FastChat/issues/581

jambayk:
A recent urgent update to qlora caused a merged conflict. I will takeover this PR and update it in the next couple of days.

devang-ml (approved):

olive/passes/pytorch/lora.py
```diff
@@ -191,26 +195,162 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
...
+        )
+
+        # cast lora modules to model's dtype, should be same as torch_dtype
+        for module in lora_model.modules():
+            if isinstance(module, LoraLayer):
+                module.to(lora_model.dtype)
+
+        return lora_model
+
+    def train_and_save_new_model(
```
guotuofeng, line 342: I don't see the device is used in training?

jambayk: Do you mean the gpu? It is enabled by `device_map="auto"` during model loading.

PR #641, opened 2023-10-12
🦙 llama2 optimization

trajepl:


examples/llama2/user_script.py
```diff
@@ -297,9 +297,9 @@ def collate_batch(self, batch):
 
             # Pad input data because all model inputs must have same shape
             pad_len = self.pad_max - input_ids.shape[0]
-            input_ids = F.pad(input_ids, (0, pad_len), value=1)
-            attention_mask = F.pad(attention_mask, (0, pad_len), value=0)
-            position_ids = F.pad(position_ids, (0, pad_len), value=0)
+            input_ids = F.pad(input_ids, (0, pad_len), value=1)  # pylint: disable=not-callable
```
guotuofeng, line 300: you can just put # pylint: disable=not-callable above the three lines. Then only one line comment is enough

olive/model/__init__.py
```diff
@@ -475,7 +475,7 @@ def __init__(
         model_loader: Union[str, Callable] = None,
         model_script: Union[str, Path] = None,
         script_dir: Union[str, Path] = None,
-        io_config: Union[Dict[str, Any], IOConfig] = None,
+        io_config: Union[Dict[str, Any], IOConfig, str] = None,
```
guotuofeng, line 478: do we need support Callable like line 479?

examples/llama2/user_script.py
```diff
@@ -0,0 +1,334 @@
...
+                }
+                if self.merged:
+                    inputs.pop("attention_mask", None)
+                label = labels.detach().cpu().numpy()
+
+                # Yield (inputs, label) tuple for Intel's Neural Compressor:
+                # https://github.com/intel/neural-compressor/blob/d4baed9ea11614e1f0dc8a1f4f55b73ed3ed585c/neural_compressor/quantization.py#L55-L62
+                yield (inputs, label)
+
+        except StopIteration:
```
guotuofeng, line 333: useless StopIteration since using for loop?

olive/model/__init__.py
trajepl: Sounds good.
Also for this: https://github.com/microsoft/Olive/blob/main/olive/model/hf_utils.py#L25

How about to implement it with separated PR?

examples/llama2/ort_converter_merged_llama2_cpu.json
```diff
@@ -0,0 +1,110 @@
...
+                        "dataloader_func": "dataloader_func_for_merged",
+                        "batch_size": 1
+                    }
+                }
+            ]
+        }
+    },
+    "passes": {
+        "conversion_merged": {
+            "type": "OnnxConversion",
```
guotuofeng, line 34: I don't see OptimumConversion?

olive/model/__init__.py
guotuofeng: Please add TODO

examples/llama2/ort_converter_merged_llama2_cpu.json
trajepl: Currently, OptimumConversion conduct two models which is not the focus now. We can turn our attention now to directly convert one merged model.

examples/llama2/user_script.py
justinchuby: FYI: Be sure to add `# pylint: enable=not-callable` after the lines. Otherwise the rule will be disabled for the rest of the file

guotuofeng: since the not-callable is false positive at most case, I am considering to disable it globally.

olive/passes/onnx/quantization.py
```diff
@@ -506,3 +506,51 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
...
+            raise OlivePassError("OnnxLlamaMatMulWeight4Quantizer is only supported in onnxruntime >= 1.17.0")
+
+        from onnxruntime.quantization.matmul_4bits_quantizer import MatMul4BitsQuantizer
+
+        quant = MatMul4BitsQuantizer(
+            model.load_model(), config["block_size"], config["is_symmetric"], config["nodes_to_exclude"]
+        )
+        quant.process()
+
+        new_tmp_dir = tempfile.TemporaryDirectory(prefix="olive_tmp")
```
jambayk, line 549: Is this tmp_dir necessary? 
Since the quantizer edits the model in place, I think there are two options:
```
onnx_model = model.load_model()
quant = ...
return model_proto_to_olive_model(onnx_model, ...)
```
or 
```
quant =  MatMul4BitsQuantizer(model.load_model(), ...)
...
return model_proto_to_olive_model(quant.model.model, ....)
```
Both would work but I think the first option makes it clearer without assuming anything about the structure of `quant`.

examples/llama2/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+Performs optimization pipeline:
+- CPU, FP32: *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model fp32*
+- CPU, INT8: *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model fp32 -> Onnx Dynamic Quantization*
+- CPU, INT4: *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model fp32 -> Onnx Block wise int4 Quantization*
+- GPU, FP32: *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model fp32*
+- GPU, FP16: *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model fp16 + Grouped Query Attention*
+- GPU, INT4: *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model fp16 + Grouped Query Attention -> Onnx Block wise int4 Quantization*
+
+**Note that**: Currently, grouped query attention is only supported on GPU with fp16 and it requires the cuda architecture >= 80. You can just set `use_gqa` to `false` in the config file to disable it.
```
devang-ml, line 12: Could Olive determine that based on the accelerator spec for the readymade systems?

examples/llama2/llama2_cpu.json
```diff
@@ -0,0 +1,110 @@
+{
+    "input_model":{
```
devang-ml, line 2: Instead of separate json files for cpu and gpu, could we just use one template json file and generate needed json file in llama2.py based on inputs ?

olive/passes/onnx/quantization.py
trajepl: image
The ONNXModel from ort quantization save the model with:
1. topological_srt
2. extra arguemnts like convert_attribute as True.
To align with the original implementation, I prefer to save with the `save_model_to_file` firstly, then save with Olive's method.

trajepl: Added TODO. We can refine this part later.
image

examples/llama2/README.md
trajepl: It is a good point. Currently cannot I think. But it should be easy to add some infos for cuda architecture.
image

examples/llama2/llama2_cpu.json
trajepl: Updated.

olive/passes/onnx/quantization.py
jambayk: Yes, I think we can expose the `convert_attribute` in our save helper function. 
In this case, `saving -> loading -> saving` undoes the original `convert_attribute` setting since it is only used during saving. After the model is loaded, the external data information is wiped out. So the original save to temp file is redundant. 

The only thing that a direct save doesn't do is the `topological sort`. This part, we can do ourselves like in https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/transformer_optimization.py#L207 

But this is not a blocker. Just thought its cleaner if we don't deal with temp files when not needed.

trajepl: Agree. Let us finish it in following PR.

examples/llama2/README.md
devang-ml: Let's do that in a separate PR

olive/passes/onnx/transformer_optimization.py
```diff
@@ -193,6 +205,18 @@ def _run_for_config(
             optimizer.convert_float_to_float16(
                 keep_io_types=config["keep_io_types"], op_block_list=op_block_list, force_fp16_inputs=force_fp16_inputs
             )
+            if config["use_gqa"]:
+                # Replace MultiHeadAttention with GroupQueryAttention and remove attention mask nodes
+                num_kv_heads = model.model_attributes.get("num_key_value_heads", None)
+                if num_kv_heads is None:
```
jambayk, line 211: Not a blocker: Should we make it possible to provide the value as a pass config parameter too?
I think there is an asymmetry that `num_head` and `hidden_size` can be pass config param or model attributes. 
Also, we don't usually provide model_attributes manually for non-automatic workflows.

olive/passes/onnx/transformer_optimization.py
```diff
@@ -202,3 +226,36 @@ def _run_for_config(
 
         # save the model to the output path and return the model
         return model_proto_to_olive_model(optimizer.model, output_model_path, config)
+
+    @staticmethod
+    def _replace_mha_with_gqa(model: "OnnxModel", past_seq_len: str = "past_sequence_length", kv_num_heads: int = 0):
```
jambayk, line 231: Is there a reason we cannot use `onnxruntime.transformers.convergeneration.replace_mha_with_gqa` directly? Do we want this option to still be usable for older versions of onnxruntime on the host?

jambayk (approved):

olive/passes/onnx/transformer_optimization.py
trajepl: Yes, we can. But the GQA currently only works for llama-70B as the num_heads!=num_kv_heads. 
We can complete the hf_mapping to introduce similar value for num_kv_heads.

olive/passes/onnx/transformer_optimization.py
trajepl: Yes, old version of ort do not contain the corresponding OPs, so that is not the problem. 

The reason I was thinking: the scripts under convert generation is like temp-solutions, also as this part of logics is quite simple, I directly copy it here.

olive/passes/onnx/quantization.py
```diff
@@ -506,3 +506,52 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
...
+        config.update(get_external_data_config())
+        return config
+
+    def _run_for_config(
+        self, model: ONNXModel, data_root: str, config: Dict[str, Any], output_model_path: str
+    ) -> ONNXModel:
+        from onnxruntime import __version__ as OrtVersion
+
+        if version.parse(OrtVersion) < version.parse("1.17.0"):
+            raise OlivePassError("OnnxLlamaMatMulWeight4Quantizer is only supported in onnxruntime >= 1.17.0")
```
guotuofeng, line 540: MatMul4BitsQuantizer?

olive/passes/onnx/quantization.py
```diff
@@ -506,3 +506,52 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
...
+
+        quant = MatMul4BitsQuantizer(
+            model.load_model(), config["block_size"], config["is_symmetric"], config["nodes_to_exclude"]
+        )
+        quant.process()
+
+        # TODO(trajep): add more options to save_model_to_file
+        new_tmp_dir = tempfile.TemporaryDirectory(prefix="olive_tmp")
+        tmp_model_path = str(Path(new_tmp_dir.name) / Path(output_model_path).name)
+        quant.model.save_model_to_file(tmp_model_path, config["save_as_external_data"])
```
guotuofeng, line 552: add the ".onnx" suffix in the file name

olive/passes/onnx/quantization.py
trajepl: https://github.com/microsoft/Olive/pull/669

PR #647, opened 2023-10-16
Copy DistributedOnnxModel generated models to output folder

shaahji:
## Describe your changes

Copy DistributedOnnxModel generated models to output folder.
Also, in the case of large models, copy data files to the output folder.

### DIFF
--- a/olive/model/__init__.py
+++ b/olive/model/__init__.py
@@ -747,10 +747,9 @@ class PyTorchModel(OliveModel):
 
 class OptimumModel(PyTorchModel):
     def __init__(self, model_components: List[str], **kwargs):
-        super().__init__(
-            model_file_format=ModelFileFormat.OPTIMUM,
-            **(kwargs or {}),
-        )
+        kwargs = kwargs or {}
+        kwargs["model_file_format"] = ModelFileFormat.OPTIMUM
+        super().__init__(**kwargs),
         self.model_components = model_components
 
     def to_json(self, check_object: bool = False):
@@ -896,8 +895,6 @@ class OpenVINOModel(OliveModel):
 
 
 class DistributedOnnxModel(ONNXModelBase):
-    resource_keys: ClassVar[list] = ["model_filepaths"]
-
     EXECUTION_PROVIDERS: ClassVar[dict] = {
         "cpu": ["CPUExecutionProvider"],
         "gpu": ["CUDAExecutionProvider", "CPUExecutionProvider"],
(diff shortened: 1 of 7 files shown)


olive/resource_path.py
```diff
@@ -219,6 +219,14 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
         # copy the resource to the new path
         if is_file:
             shutil.copy(self.config.path, new_path)
+
+            # Copy the data file, if exists
```
trajepl, line 223: For onnx model, if the external data's file name is changed after the `copy`, onnx model can not be loaded successfully.

Can we load model back and run onnx.save again?

shaahji: The file name is the same as it was saved. The logic here is building the data file name/path merely by appending ".data" to the corresponding onnx file path. This is still not a complete solution because the name of the data file is an input to the onnx.save method. None of the logic actually uses this but there is potential for this to fail.

This is a stop-gap solution to make the distributed onnx export work correctly. The correct solution need discussion as there are other scenarios that need to be covered as well (like weights for each operator is exported out as an independent file in which case the model is neither just one file (onnx), nor two files (onnx and onnx.data) but an entire folder with any number of files.

shaahji:
@jambayk Updated the commit using _model_path_ and _model_name_ format where _model_path_ is a directory and copying models from _cache_ to _models_ works as expected. This is a follow up from our email discussions.

olive/model/__init__.py
```diff
@@ -905,28 +904,37 @@ class DistributedOnnxModel(ONNXModelBase):
...
     ):
         super().__init__(
-            model_path=None,
+            model_path=model_path,
             inference_settings=inference_settings,
             use_ort_extensions=use_ort_extensions,
             model_attributes=model_attributes,
         )
-        self.model_filepaths = model_filepaths or []
+        self.add_resources({"model_name_pattern": model_name_pattern, "num_ranks": str(num_ranks)})
```
jambayk, line 920: `self.add_resources` is only used to add `file/folder` resources. This is so that we can keep track of these for uploading, etc. 
For `model_name_patttern` and `num_ranks`, they should be added directly as attribute during init: `self.model_name_pattern`, etc.

olive/model/__init__.py
```diff
@@ -959,15 +967,7 @@ def get_execution_providers(device: Device):
...
-        config = {
-            "type": self.__class__.__name__,
-            "config": {
-                "model_filepaths": self.model_filepaths,
-                "inference_settings": self.inference_settings,
-                "use_ort_extensions": self.use_ort_extensions,
-                "model_attributes": self.model_attributes,
-            },
-        }
+        config = super().to_json(check_object)
```
jambayk, line 970: after adding `model_name_pattern` and `num_ranks` as attributes, you can update `config["config"]` here with their names and values.

olive/passes/olive_pass.py
```diff
@@ -344,13 +344,17 @@ def run(
...
                 input_rank_model = model.load_model(rank)
-                rank_output_path = Path(output_model_path).with_suffix("") / str(rank)
-                output_rank_model = self._run_for_config(input_rank_model, data_root, config, rank_output_path)
-                output_filepaths.append(output_rank_model.model_path)
-            output_model = DistributedOnnxModel(output_filepaths, inference_settings=model.inference_settings)
+                rank_output_path = Path(output_model_path).with_suffix("") / model.ranked_model_name(rank)
+                self._run_for_config(input_rank_model, data_root, config, rank_output_path)
+
+            output_model = DistributedOnnxModel(
+                model_path=output_model_path,
```
jambayk, line 353: It might be safer to also use `Path(output_model_path).with_suffix("")` here? Just in case the `output_model_path` had a suffix?

shaahji: Done!

olive/model/__init__.py
shaahji: Done!

olive/model/__init__.py
shaahji: Done!

jambayk (approved):

jambayk:
The doc build failure looks spurious. We can manually rerun the failed job after the other jobs complete.

PR #649, opened 2023-10-16
Support for Llama2 model with sharding optimization

shaahji:
## Describe your changes

Support for Llama2 model with sharding optimization

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,22 +13,51 @@ repos:
     hooks:
       - id: black
         name: Format code
+        exclude: |
+          (?x)^(
+            examples/llama2/code/parallel_layers.py|
+            examples/llama2/code/patching_llama.py
+          )$
+
   - repo: https://github.com/PyCQA/flake8
     rev: 4.0.1
     hooks:
       - id: flake8
         name: Check PEP8
+        exclude: |
+          (?x)^(
+            examples/llama2/code/parallel_layers.py|
+            examples/llama2/code/patching_llama.py
+          )$
+
   - repo: https://github.com/pycqa/isort
     rev: 5.11.5
     hooks:
       - id: isort
         name: Format imports
+        exclude: |
+          (?x)^(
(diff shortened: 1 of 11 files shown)


.pre-commit-config.yaml
```diff
@@ -13,22 +13,51 @@ repos:
     hooks:
       - id: black
         name: Format code
+        exclude: |
+          (?x)^(
```
jambayk, line 17: instead of adding the paths here which is not maintainable, can you add something like `noqa: ` wherever necessary in the files? Those should ignore the warning and errors from pre-commit.

shaahji: The files for the implementation aren't owned by us. This was implemented by an external team, and I merely copied those files to introduce support for this model. I would prefer not to modify these files, so they are easily comparable/updated to the source. 

Ideally, I would like to just add something at the top of the file to say, _ignore me_. But I didn't find anything that does that. Do you know how to ignore entire files from linting? The files have way too many lint errors to ignore individually.

jambayk: for specific tools we can add `ruff: noqa` and/or `flake8: noqa` at the top of a file. I am not sure about all the tools in the pre-commit. 
If you have to update the pre-commit config yaml, can you use a global exclude instead of hook-wise exclude so that it is at least easier to maintain? 

@guotuofeng might be more familiar with better methods.

guotuofeng: @shaahji, you can add the exclude patten in https://github.com/microsoft/Olive/blob/main/pyproject.toml#L2 and https://github.com/microsoft/Olive/blob/main/.flake8 so that all these exclude patten could be used by both pre-commit and lintrunner. 

Current approach only apply to pre-commit

guotuofeng: This pre-commit-config.yaml is not used now

shaahji:
Abandoned in favor of #694 & #698

PR #653, opened 2023-10-17
🚧 Consistent dataloader for benchmark

trajepl:
## Describe your changes
Consistent dataloader for benchmark
Transforms.RandomCrop/Clip will generate different tensors every time call dataloader creation function. It bring the inconsistent model accuracy measurement.

This PR is a workaround used to ensure the input data is consist for torch model and onnx model.

### DIFF
--- a/examples/resnet/user_script.py
+++ b/examples/resnet/user_script.py
@@ -31,10 +31,19 @@ class CIFAR10DataSet:
 
     def setup(self, stage: str):
         transform = transforms.Compose(
-            [transforms.Pad(4), transforms.RandomHorizontalFlip(), transforms.RandomCrop(32), transforms.ToTensor()]
+            [
+                transforms.Pad(4),
+                # use CenterCrop to keep consistent dataset
+                transforms.CenterCrop(32),
+                # Ban RandomHorizontalFlip and RandomCrop to keep consistent dataset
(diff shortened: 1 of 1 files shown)


examples/resnet/user_script.py
jambayk, line 1: What do you think about setting random seeds in the example's test script? https://discuss.pytorch.org/t/random-transforms-not-consistent-with-the-same-seed/63816

trajepl: Thanks for offline syncing. Both of numpy/random/torch.random did not work on this case.

trajepl: Let us disable random transforms for this case.

jambayk (approved):

jambayk (approved):

examples/resnet/user_script.py
```diff
@@ -17,6 +20,14 @@
 # Common Dataset
 # -------------------------------------------------------------------------
 
+seed = 0
+
+np.random.seed(seed)
```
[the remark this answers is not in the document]
trajepl: https://pytorch.org/docs/stable/notes/randomness.html
Based on this, better to have pinned seed value for ramdom/np.random/torch random all.

PR #654, opened 2023-10-17
Text-gen: Optional attn mask. Loading args for hf model config. QLoRA: Update loading args 

jambayk:
## Describe your changes
Some enhancements to the QLoRA pass and related components. 

Example use case scenario is for fine-tuning [microsoft/phi-1_5](https://huggingface.co/microsoft/phi-1_5). In order to load the model and config for this model non-interactively, we have to pass `trust_remote_code=True` to `model_loading_args`. However, the QLoRA pass and `get_hf_model_config` both ignore it.
There are also cases where we have to provide `token` for models like [meta-llama/Llama-2-7b](https://huggingface.co/meta-llama/Llama-2-7b). 
Both components have been updated to use the model_loading_args correctly. 

`microsoft/phi-1_5` gives a warning about `attention_mask` during training. Text-gen preprocessing is updated to make `attention_mask` optional (opt out if needed). 

Bug Fix: Check `pad_token_id` for `None` since `not pad_token_id` is `True` for `pad_token_id == 0`. 

Renamed `dataset_name` to `data_name` in qlora example to use consistent naming. 

### DIFF
--- a/olive/passes/pytorch/qlora.py
+++ b/olive/passes/pytorch/qlora.py
@@ -269,7 +269,7 @@ class QLoRA(Pass):
         new_model.model = None
         # remove the device map since we don't want "auto" device map
         new_model.hf_config.model_loading_args.device_map = None
-        # remove model_overwrites from model_attributes
+        # remove model_overwrites from model_attributes since new config will be created
         if new_model.model_attributes:
             for k in QLoRA.model_overwrites:
                 if k in new_model.model_attributes:
--- a/olive/passes/pytorch/qlora.py
+++ b/olive/passes/pytorch/qlora.py
@@ -12,7 +12,7 @@ import tempfile
 from copy import deepcopy
 from functools import partial
 from pathlib import Path
-from typing import Any, Dict, List, Tuple, Union
+from typing import Any, ClassVar, Dict, List, Tuple, Union
 
 import torch
 import transformers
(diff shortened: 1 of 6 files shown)


olive/passes/pytorch/qlora.py
```diff
@@ -267,6 +269,12 @@ def _run_for_config(
         new_model.model = None
         # remove the device map since we don't want "auto" device map
         new_model.hf_config.model_loading_args.device_map = None
+        # remove model_overwrites from model_attributes since new config will be created
+        if new_model.model_attributes:
+            for k in QLoRA.model_overwrites:
+                if k in new_model.model_attributes:
+                    del new_model.model_attributes[k]
```
trajepl, line 276: Call .pop(k, None) would be more clear?

olive/passes/pytorch/qlora.py
```diff
@@ -122,6 +122,8 @@ class QLoRA(Pass):
     This pass only supports PyTorchModel with hf_config.
     """
 
+    model_overwrites: ClassVar[tuple] = ("torch_dtype", "device_map", "quantization_method", "quantization_config")
```
trajepl, line 125: It would be better to add a few comments to introduce this new field.

jambayk: Done!

olive/passes/pytorch/qlora.py
jambayk: Thanks! this looks cleaner too. Updated.

trajepl (dismissed):

trajepl (approved):

PR #658, opened 2023-10-18
Fix bugs in how free dimension session options are set:

gaugarg-nv:
## Describe your changes

- The “unet_text_embeds_size” free dimension should always be overridden with a value of 1280 irrespective of image resolution - this value is same as the text_encoder_2 hidden dimension for both SD-XL base and refiner model. Currently using a resolution other than 1024 results in error.
- With the optimum pipeline, the right keyword to pass session options is “session_options” and not “sess_options”. This change improves perf by 40% for 512x512 resolution and about 10% for 1024x1024 resolution.


gaugarg-nv:
@microsoft-github-policy-service agree company="NVIDIA"

PatriceVignola (approved):

guotuofeng (approved):

PR #662, opened 2023-10-20
🆓 Release GPU memory for torch model evaluation

trajepl:
## Describe your changes
In llamav2 tests, torch model occupied the gpu memory which leads to error when evaluate onnx model.

Release GPU memory for torch model evaluation
model.to("cpu") could not release gpu memory in time. Call torch.cuda.empty_cache() to cleanup.

### DIFF
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -698,9 +698,10 @@ class PyTorchEvaluator(OliveEvaluator, framework=Framework.PYTORCH):
         targets = torch.cat(targets, dim=0)
         logits = torch.cat(logits, dim=0)
         # move model to cpu
-        # don't want model to be kept on gpu since model persists and takes up gpu memory
         if device:
             session.to("cpu")
+        # only move to cpu cannot release gpu memory, call cuda.empty_cache() to release gpu memory
+        torch.cuda.is_available() and torch.cuda.empty_cache()
         return OliveModelOutput(preds=preds, logits=logits), targets
 
     def _evaluate_accuracy(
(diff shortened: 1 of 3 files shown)


olive/evaluator/olive_evaluator.py
```diff
@@ -698,9 +698,10 @@ def _inference(
         targets = torch.cat(targets, dim=0)
         logits = torch.cat(logits, dim=0)
         # move model to cpu
-        # don't want model to be kept on gpu since model persists and takes up gpu memory
         if device:
             session.to("cpu")
+        # only move to cpu cannot release gpu memory, call cuda.empty_cache() to release gpu memory
+        torch.cuda.is_available() and torch.cuda.empty_cache()
```
guotuofeng, line 704: better to use if clause to make the logic clear.

guotuofeng (dismissed):

olive/evaluator/olive_evaluator.py
jambayk, line 1: This is interesting. We didn't consider that torch and onnx don't share the same cache and torch might not release the cached memory. 

Can you update in these places too: 
https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/conversion.py#L193
https://github.com/microsoft/Olive/blob/main/olive/passes/pytorch/qlora.py#L271

jambayk: actually in the qlora pass, we might also have to delete `pytorch_model` too since there is still a reference to the `torch.nn.Module`.

trajepl: Thanks! Updated.

jambayk: Thanks!

jambayk (approved):

PR #664, opened 2023-10-25
raise the known failure exceptions when do perf-tuning

guotuofeng:
## Describe your changes
In perf-tuning, if the exception is known as ImportModuleError, we need not retry the next step. And we need throw the exception like in engine.

### DIFF
--- a/test/unit_test/passes/onnx/test_perf_tuning.py
+++ b/test/unit_test/passes/onnx/test_perf_tuning.py
@@ -40,4 +40,19 @@ def test_ort_perf_tuning_pass_with_dynamic_shapes(mock_get_io_config, tmp_path):
     with pytest.raises(TypeError) as e:
         # execute
         p.run(input_model, None, output_folder)
-        assert "ones() received an invalid combination of arguments" in str(e.value)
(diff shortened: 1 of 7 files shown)


trajepl (dismissed):

test/unit_test/passes/onnx/test_perf_tuning.py
```diff
@@ -40,4 +40,19 @@ def test_ort_perf_tuning_pass_with_dynamic_shapes(mock_get_io_config, tmp_path):
     with pytest.raises(TypeError) as e:
         # execute
         p.run(input_model, None, output_folder)
-        assert "ones() received an invalid combination of arguments" in str(e.value)
+    assert "ones() received an invalid combination of arguments" in str(e.value)
```
trajepl, line 43: Should the assert in pytest's context?

guotuofeng: no, the assert should be outside the context

guotuofeng: otherwise, the assert will be ignored.

trajepl (approved):

PR #666, opened 2023-10-26
raise ValueError if the model_type is None

guotuofeng:

### DIFF
--- a/olive/passes/onnx/transformer_optimization.py
+++ b/olive/passes/onnx/transformer_optimization.py
@@ -157,9 +157,12 @@ class OrtTransformersOptimization(Pass):
 
         if model.model_attributes:
             model_config = model.model_attributes
-            input_model_type = model_config.get("model_type", "")
-            _model_type = MODEL_TYPE_MAPPING.get(input_model_type, input_model_type)
-            run_config["model_type"] = run_config["model_type"] or _model_type
+            input_model_type = model_config.get("model_type")
+            if input_model_type:
+                model_type = MODEL_TYPE_MAPPING.get(input_model_type)
+            else:
+                model_type = None
+            run_config["model_type"] = run_config["model_type"] or model_type
             assert run_config["model_type"] in transformers_optimizer.MODEL_TYPES, (
                 f"Unsupported model type: {run_config['model_type']}, please select one from "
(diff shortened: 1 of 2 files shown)


trajepl (dismissed):

olive/passes/onnx/transformer_optimization.py
```diff
@@ -157,13 +157,12 @@ def _run_for_config(
...
         if model.model_attributes:
             model_config = model.model_attributes
-            input_model_type = model_config.get("model_type", "")
-            _model_type = MODEL_TYPE_MAPPING.get(input_model_type, input_model_type)
-            run_config["model_type"] = run_config["model_type"] or _model_type
-            assert run_config["model_type"] in transformers_optimizer.MODEL_TYPES, (
-                f"Unsupported model type: {run_config['model_type']}, please select one from "
-                "{transformers_optimizer.MODEL_TYPES} which need to be set under OrtTransformersOptimization.config"
-            )
+            input_model_type = model_config.get("model_type")
```
jambayk, line 160: not an issue but I think the original lines 160-162 do the same thing as the new `160-165`? `""` behaves the same as `None` in this use case. 

The only thing incorrect was that the final `model_type` check should have been outside this if condition

olive/passes/onnx/transformer_optimization.py
```diff
@@ -175,6 +174,13 @@ def _run_for_config(
                         run_config["hidden_size"] = model_config[hidden_size_name]
                         break
 
+        if run_config["model_type"] is None or run_config["model_type"] not in transformers_optimizer.MODEL_TYPES:
```
jambayk, line 177: Is the `None` check to make the logic clear? Since `None` would also fail the second condition.

olive/passes/onnx/transformer_optimization.py
guotuofeng: Yes, your are correct.

For the model_type variable, it is fine since it is only used when model.model_attributes have values.

olive/passes/onnx/transformer_optimization.py
guotuofeng: I explicitly check the None to make the logic clear.

jambayk (approved):

PR #679, opened 2023-11-01
Support custom root directory for temporary directories and files

jambayk:
## Describe your changes
Olive uses a lot of temporary directories which are created in the default tempory directory root of the system. However, we encounter issues with storage, permissions, files disappearing, etc.
This PR adds a new olive module called `olive.tmp_dir` which allows setting/user an environment variable `OLIVE_TMP_DIR_ROOT` that will be used as the root directory for all olive based temporary directories. 

Code that set their own `dir` are unchanged. Most of them use the destination dir as the `dir` since that allows for fast move through rename (shutil.move under the same file system is a rename). 

Allow user to set custom root directory for olive temporary directories and files using environment variable `OLIVE_TMP_DIR_ROOT` or workflow config option `tmp_dir_root`.


olive/workflows/run/run.py
```diff
@@ -144,6 +145,10 @@ def run(config: Union[str, Path, dict], setup: bool = False, data_root: str = No
     set_default_logger_severity(config.engine.log_severity_level)
     ort.set_default_logger_severity(config.engine.ort_log_severity_level)
 
+    # set tmp dir root
+    if config.engine.tmp_dir_root is not None:
+        set_tmp_dir_root(config.engine.tmp_dir_root)
```
guotuofeng, line 150: it seems we can only change the tempfile.tempdir like https://docs.python.org/3/library/tempfile.html#tempfile.tempdir? so there is no need tmp_dir.py?

guotuofeng (changes requested):
https://docs.python.org/3/library/tempfile.html#tempfile.tempdir should be able to meet our requirements.

olive/workflows/run/config.py
```diff
@@ -34,6 +34,7 @@ class RunEngineConfig(EngineConfig):
     packaging_config: PackagingConfig = None
     log_severity_level: int = 1
     ort_log_severity_level: int = 3
+    tmp_dir_root: str = None  # if None, will use the whatever is set in OLIVE_TMP_DIR_ROOT
```
guotuofeng, line 37: should this belong engine config or run config? BTW, need the parameter passed by command line like:
python -m olive.workflows.run --config <config_path> --temp_dir blalblabla?

jambayk:
closing in favor of https://github.com/microsoft/Olive/pull/683

PR #680, opened 2023-11-01
OnnxMatMul4Quantizer: Suppress quantizer logs, no tmp_dir

jambayk:
## Describe your changes
`MatMul4BitsQuantizer` uses logging.basicConfig to set level to INFO and prints very verbose logs at INFO level. Suppress these logs by manually setting the loggers level to `ERROR`. 
Update ort version requirement to `>=1.16.2` since the quantizer will be added to 1.16.2.

Remove the save to tmp_dir -> load -> save steps since it is not needed. We only need to sort the model topologically before saving the model to file. Refer to this discussion for more context https://github.com/microsoft/Olive/pull/641#discussion_r1373793462. 

`llama.py`: Correct description for `--only_config` option. Create a new user_script file for the workflow instead of updating the original one.

### DIFF
--- a/examples/llama2/llama2.py
+++ b/examples/llama2/llama2.py
@@ -92,7 +92,7 @@ def main(raw_args=None):
     config_name = f"llama2_{device}_{gqa}"
 
     # create user script
-    user_script_path = Path(__file__).parent / f"user_script.py"
+    user_script_path = Path(__file__).parent / "user_script.py"
     config_script_path = Path(__file__).parent / f"{config_name}_user_script.py"
     update_user_script(user_script_path, config_script_path, model_name)
 
@@ -102,7 +102,7 @@ def main(raw_args=None):
     template_json = json.loads(template_json)
 
     # dump config
-    with open(f"llama2_{device}_{gqa}.json", "w") as f:  # noqa: PTH123
+    with open(f"{config_name}.json", "w") as f:  # noqa: PTH123
         json.dump(template_json, f, indent=4)
 
     if not args.only_config:
--- a/examples/llama2/llama2.py
+++ b/examples/llama2/llama2.py
@@ -49,7 +49,7 @@ def get_args(raw_args):
         "--only_config",
         action="store_true",
         required=False,
-        help="Whether to use GQA(grouped query attention) instead of MHA(multi-head attention).",
+        help="Whether to only dump the config file without running the optimization.",
     )
 
     return parser.parse_args(raw_args)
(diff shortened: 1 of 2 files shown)


examples/llama2/llama2.py
```diff
@@ -87,12 +87,20 @@
...
-    # update user script
-    user_script_path = Path(__file__).parent / "user_script.py"
-    update_user_script(user_script_path, model_name)
-
     device = "gpu" if args.gpu else "cpu"
     gqa = "gqa" if args.use_gqa else "mha"
+    config_name = f"llama2_{device}_{gqa}"
+
+    # create user script
+    user_script_path = Path(__file__).parent / f"user_script.py"
```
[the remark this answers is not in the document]
jambayk: Resolved

olive/passes/onnx/quantization.py
```diff
@@ -536,24 +536,24 @@ def _run_for_config(
...
 
-        if version.parse(OrtVersion) < version.parse("1.17.0"):
-            raise OlivePassError("MatMul4BitsQuantizer is only supported in onnxruntime >= 1.17.0")
+        if version.parse(OrtVersion) < version.parse("1.16.2"):
+            raise OlivePassError("MatMul4BitsQuantizer is only supported in onnxruntime >= 1.16.2")
 
         from onnxruntime.quantization.matmul_4bits_quantizer import MatMul4BitsQuantizer
 
+        # MatMul4BitsQuantizer sets basicConfig to INFO and prints verbose logs as INFO, suppress them
+        # TODO(jambayk): Use the ort logging severity from the workflow if we expose it as environment variable
```
guotuofeng, line 545: Actually, the ort logging severity only impact the logging in c++ code. when I tried last week, I have to use getLogger().setLevel to change the root logger verbosity to make sure the messages could be logged out.

so, do we need change the root logger level so to output the ORT python messages?

jambayk: Yes, I looked at the ort logging severity only works for c++. 
I will thinking whether we would want to set the same logging level for ort python logging by setting the logger for the ort modules at the same level. 

We don't need to change the root logger. Just need to change the level for the logger for the given name space. 
I don't think it is recommended for packages to do anything to the root logger or use basic config. The ort package should not have done that too.

guotuofeng: In this week, I have to manually set the root logger level to troubleshoot why the quant_pre_process silently fail(the process exit silently). We need find way to config the ORT python logger level from troubleshooting point of view.

jambayk: This sounds good to me. If the logger is created using `logging.getLogger(__name__)`, it makes it easier for us since we could do `logging.getLogger("onnxruntime").setLevel(...)` to set it for ort. 
Or else, I don't know what else we could do without touching the root logger/

guotuofeng: Actually, I tried to logging.getLogger("onnxruntime").setLevel(...) but the log is not shown correctly. I don't spend time to investigate the root cause.

guotuofeng (approved):

PR #693, opened 2023-11-03
Add microsoft\phi_1 example

zhangxiang1993:


examples/directml/phi_1/README.md
jambayk, line 1: Please format the new code using lintrunner with the instructions at https://github.com/microsoft/Olive/blob/main/CONTRIBUTING.md#linting

The MYPY linter is slow so you could skip it using the following:
```
lintrunner --skip MYPY -a
```

examples/directml/phi_1/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+# Conversion to ONNX and Latency Optimization
+
+The easiest way to optimize the pipeline is with the `microsoft_phi.py` helper script:
+
+```
+python microsoft_phi.py --optimize
+```
+
+The language model phi-1 is large, and the optimization process is resource intensive. The optimization process can easily take more than 128GB of memory. You can still optimize the model on a machine with less memory, but you'd have to increase your paging file size accordingly and the conversion process will take significantly longer to complete (many hours).
```
jambayk, line 34: Is this comment still relevant since the model is much smaller than other examples like SD, Dolly, LlaMA, etc?

examples/directml/phi_1/microsoft_phi.py
```diff
@@ -0,0 +1,99 @@
+# from olive.workflows import run as olive_run
```
jambayk, line 1: Could you remove these commented code before the copyright header?

examples/directml/phi_1/microsoft_phi.py
```diff
@@ -0,0 +1,99 @@
...
+import os
+import shutil
+import sys
+from pathlib import Path
+
+import onnxruntime as ort
+from packaging import version
+from olive.workflows import run as olive_run
+
+# pylint: disable=redefined-outer-name
```
PatriceVignola, line 19: Why is this needed?

examples/directml/phi_1/phi_1_config.json
```diff
@@ -0,0 +1,78 @@
+{
```
PatriceVignola, line 1: The nomenclature in the directml examples folder is to prefix the `config` instead of having it as a suffix.

examples/directml/phi_1/microsoft_phi.py
zhangxiang1993: To disable the following warning:
Warning (PYLINT) W0621
    Redefining name 'script_dir' from outer scope (line 70) (redefined-outer-
    name)

examples/directml/phi/README.md
```diff
@@ -0,0 +1,74 @@
...
+# sample inputs
+inputs = tokenizer('''def print_prime(n):
+               """
+               Print all primes between 1 and n
+               """'''
+```
+
+```
+python .\optimize.py --inference
+def print_prime(n):
```
guotuofeng, line 48: the indentation should be same for two defs?

examples/directml/phi/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+Once the script successfully completes, the optimized ONNX pipeline will be stored under `models/optimized/microsoft/phi-1` or `models/optimized/microsoft/phi-1_5`.
+
+# Test Inference
+Re-running the script with `--inference`
+```
+# sample inputs
+inputs = tokenizer('''def print_prime(n):
+               """
```
guotuofeng, line 41: the left whitespace should be four?

examples/directml/phi_1/microsoft_phi.py
PatriceVignola: Sure, but the solution would be to rename the variable to not overwrite it then. I don't see where the redefining is happening either.

examples/directml/phi/microsoft_phi.py
```diff
@@ -0,0 +1,96 @@
...
+        assert conversion_footprint
+
+        model_info = {
+            "unoptimized": {
+                "path": Path(conversion_footprint["model_config"]["config"]["model_path"]),
+            },
+        }
+        print(f"Unoptimized Model : {model_info['unoptimized']['path']}")
+
+    # Create a copy of the unoptimized model directory, then overwrite with optimized models from the olive cache.
```
PatriceVignola, line 49: We don't have the optimized models yet, so we can get rid of the second part of the sentence.

PatriceVignola (approved):

PR #694, opened 2023-11-03
Implement DistributedPyTorchModel

shaahji:
## Description
Implement DistributedPyTorchModel

DistributedPyTorchModel is parallel to DistributedOnnxModel for PyTorch models.
Also, introducing an example, Llama2, as an use case.

### DIFF
--- a/olive/passes/pytorch/tensor_parallel.py
+++ b/olive/passes/pytorch/tensor_parallel.py
@@ -1,0 +1,101 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+# Automatically distribute PyTorch model using Tensor Parallelism
+# --------------------------------------------------------------------------
+
+import logging
+from abc import abstractmethod
+from pathlib import Path
+from typing import Any, Callable, ClassVar, Dict
+
+import torch
+from pydantic import validator
+
+from olive.hardware.accelerator import AcceleratorSpec
+from olive.model import DistributedPyTorchModel, PyTorchModel
+from olive.passes import Pass
+from olive.passes.olive_pass import PassConfigParam
+
+logger = logging.getLogger(__name__)
+
+
+class PyTorchTensorParallel(Pass):
+    RANKED_MODEL_NAME_FORMAT: ClassVar[str] = "model_{:02d}"
+
+    @staticmethod
+    def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        # Note : The default world_size should be the no of gpus (AceleratorSpec.Device == GPU)
+        # in the target OliveSystem
+        return {
+            "world_size": PassConfigParam(
+                type_=int,
+                default=2,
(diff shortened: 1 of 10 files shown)


olive/passes/pytorch/tensor_parallel.py
```diff
@@ -0,0 +1,101 @@
...
+        return {
+            "validate_distributor_config": validator("world_size", allow_reuse=True)(
+                PyTorchTensorParallel._validate_world_size
+            )
+        }
+
+    def _run_for_config(
+        self, model: PyTorchModel, data_root: str, config: Dict[str, Any], output_model_path: str
+    ) -> DistributedPyTorchModel:
+        self.world_size = int(config["world_size"])
```
jambayk, line 58: Since we want everything for a run to be configurable using the `config` dict and don't dynamically update attributes of a pass object if possible (want the pass to be independent of the model/run), would it be possible to pass `config["world_size"]` as arguments to whatever method that needs it? Like `split_weights` and `load_rank_weights`?

olive/passes/pytorch/tensor_parallel_llama2.py
```diff
@@ -0,0 +1,230 @@
...
+# --------------------------------------------------------------------------
+
+import logging
+import math
+from typing import Any, Dict, Optional, Tuple
+
+import torch
+import torch.nn.functional as F
+from transformers.activations import ACT2FN
+from transformers.models.llama.modeling_llama import LlamaAttention, LlamaMLP, repeat_kv, rotate_half
```
jambayk, line 15: I think importing some specific modules from a dependency like this at top level makes the whole package incompatible with older versions of `transformers`. For example, `transformers.models.llama` doesn't exist before version 4.28.0. 

What if we make this file as a helper util and move the actual pass to `tensor_parallel.py` and make it import the functions from here only where needed? Or whatever structure sounds reasonable to you.

jambayk: Like https://github.com/microsoft/Olive/blob/main/olive/passes/pytorch/qat_utils.py and https://github.com/microsoft/Olive/blob/main/olive/passes/pytorch/trt_utils.py

jambayk:
The lintrunner instructions are at https://github.com/microsoft/Olive/blob/main/CONTRIBUTING.md#linting

I find the MYPY linter to be slow, so I skip it using the following:
```
lintrunner --skip MYPY -a
```

olive/passes/pytorch/tensor_parallel_llama2.py
devang-ml: Valid concern. I'm ok to move it to a helper util file.
I expect in future we'll have PytorchTensorParallel pass variation for non-llama model also.

olive/passes/pytorch/tensor_parallel.py
shaahji: Done!

olive/passes/pytorch/tensor_parallel_llama2.py
shaahji: Done!

devang-ml (approved):

PR #698, opened 2023-11-06
Implement OnnxConversion for distributed python model

shaahji:
## Describe your changes

Implement OnnxConversion for distributed python model
Use llama2 as an example.


shaahji:
@devang-ml This is remaining change to convert llama2 model to ONNX. I have also added a script to compare the PyTorch output to distributed ONNX run.

examples/llama2/tensor_parallel_user_script.py
```diff
@@ -0,0 +1,134 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+from itertools import chain
+
+import torch
+import torch.nn.init
+from transformers import LlamaConfig
+
+
```
devang-ml, line 11: Do we need these separate user script file when we already got https://github.com/microsoft/Olive/blob/main/examples/llama2/user_script.py ?

devang-ml (commented):
Nit: We already have llama2 example that does ONNX conversion. Title of the PR should be to enable ONNX conversion for Distributed PYTorch model.

examples/llama2/tensor_parallel_user_script.py
shaahji: Removed the alternate implementation.

examples/llama2/tensor_parallel_config.json
jambayk, line 1: Can you add a section for this workflow in the README? Similar to the sections in this https://github.com/microsoft/Olive/blob/shaahji/distconv/examples/open_llama/README.md 

Also, does this workflow have external dependencies like mpi? Or is it already part of the olive dependencies?

olive/passes/onnx/conversion.py
```diff
@@ -24,6 +24,17 @@
 logger = logging.getLogger(__name__)
 
 
+# Note: This should move to OliveModel itself
+def get_io_config(model):
```
jambayk, line 28: Can you add a type hint for `model: PyTorchModel` and output?

olive/passes/onnx/conversion.py
```diff
@@ -195,22 +219,95 @@ def _convert_model_on_device(
...
+
+        torch.distributed.init_process_group(
+            "nccl", init_method="tcp://127.0.0.1:9876", world_size=world_size, rank=local_rank
+        )
+
+        output_filename = DistributedOnnxModel.DEFAULT_RANKED_MODEL_NAME_FORMAT.format(local_rank)
+        output_filepath = str(Path(output_dirpath) / output_filename)
+
+        input_model = DistributedPyTorchModel(**model_config)
+        pytorch_model = input_model.load_model(local_rank)
```
jambayk, line 263: maybe this can be called something like `olive_pytorch_model` to avoid confusion with `pytorch_model` which is of type `torch.nn.Module)

olive/passes/onnx/conversion.py
```diff
@@ -195,22 +219,95 @@ def _convert_model_on_device(
...
+
+        params = [
+            (
+                pass_config,
+                data_root,
+                model_config,
+                world_size,
+                device,
+                rank,
+                output_model_path,
```
jambayk, line 300: nit: the `output_model_path` could be resolved beforehand as 
`output_model_path = str(Path(output_model_path).with_suffix(""))` 

and use here and in the return part to be sure we are using the same value in both.

olive/passes/onnx/conversion.py
```diff
@@ -195,22 +219,95 @@ def _convert_model_on_device(
...
-    """Convert a PyTorch model to ONNX model using torch.onnx.export by using specific hardware device."""
+        return onnx_model
 
     @staticmethod
-    def is_accelerator_agnostic(accelerator_spec: AcceleratorSpec) -> bool:
-        return False
+    def _export_ranked_model(params):
+        pass_config, data_root, model_config, world_size, device, local_rank, output_dirpath = params
+
+        import os
```
justinchuby, line 232: Prefer importing on the top of the module in the global scope

olive/passes/onnx/conversion.py
```diff
@@ -55,26 +66,28 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
 
     def _run_for_config(
         self, model: PyTorchModel, data_root: str, config: Dict[str, Any], output_model_path: str
-    ) -> Union[ONNXModel, CompositeOnnxModel]:
-        return self._convert_model_on_device(model, data_root, config, output_model_path, "cpu")
```
guotuofeng, line 59: the original model conversion will always happen in CPU side. @trajepl should have some comments. I am not sure this behavior change has any impact.

guotuofeng: so, we don't need convert on GPU like https://github.com/microsoft/Olive/pull/478 anymore?

shaahji: We don't have an example where it's at the moment and as @devang-ml suggested it should be implemented differently. So, instead of fixing existing unused code, I chose to remove it.

examples/llama2/tensor_parallel_config.json
shaahji: We might want to organize the documentation for MPI examples together. Let me discuss this with @devang-ml and Emma and will introduce in a follow up PR.

olive/passes/onnx/conversion.py

olive/passes/onnx/conversion.py

olive/passes/onnx/conversion.py
shaahji: Done!

olive/passes/onnx/conversion.py

olive/passes/pytorch/tensor_parallel.py
```diff
@@ -70,7 +68,7 @@ def _run_for_config(
         # 4. Save the weights for each rank
         for rank in range(world_size):
             self.load_rank_weights(pytorch_model, rank, world_size)
-            output_model_name = PyTorchTensorParallel.RANKED_MODEL_NAME_FORMAT.format(rank)
+            output_model_name = DistributedPyTorchModel.DEFAULT_RANKED_MODEL_NAME_FORMAT.format(rank)
```
jambayk, line 71: It is not actually part of this PR but I think we should have done something like
```
try:
    line 62-74
finally:
    line 75-76
```
to ensure that the layers and initializers are reset even when there is an exception?

shaahji: Done!

jambayk (approved):

PR #699, opened 2023-11-06
🍨 Io binding with OrtValue & KV shared buffer

trajepl:
## Describe your changes
This pull request to the Olive project includes changes to improve the performance and efficiency of the codebase. The most important changes include adding a new method to efficiently bind inputs and outputs to the session, and adding new user configuration options to enable input/output binding to the session.

Main performance improvements:

* <a href="diffhunk://#diff-816af95cf16e31610395bb20947f1d73cc1ca67355964e074a8f55375c8139a7R315-R330">`olive/evaluator/olive_evaluator.py`</a>: Added a new method `prepare_io_bindings` to efficiently bind inputs and outputs to the session using either the CPU or GPU depending on the device. 

Configuration improvements:

* <a href="diffhunk://#diff-9518353a6765bba684599d6a43784130d1d59148103baff7550b6dbfdd387858L38-R39">`examples/bert/bert_cuda_gpu.json`</a>: Added new user configuration option `io_bind` to enable input/output binding to the session.
* <a href="diffhunk://#diff-26023f8bc1119b7b0914f257e10173362a37ab7b12e326893b2b575d400fff64L38-R39">`examples/bert/bert_trt_gpu.json`</a>: Added a new user configuration option `io_bind` to enable input/output binding to the session.

### DIFF
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -312,6 +312,22 @@ class OnnxEvaluator(OliveEvaluator, framework=Framework.ONNX):
             if k in input_names
         }
 
+    @staticmethod
+    def prepare_io_bindings(session, input_data, device):
+        """Convert input from numpy array to OrtValue."""
+        # TODO(trajep): add device_id to the OrtValue
+        from onnxruntime import OrtValue
+
+        io_bind_op = session.io_binding()
+        io_bind_device = "cuda" if device == "gpu" else "cpu"
+        for k, v in input_data.items():
+            new_v = OrtValue.ortvalue_from_numpy(v, io_bind_device)
+            io_bind_op.bind_ortvalue_input(k, new_v)
+        for item in session.get_outputs():
+            io_bind_op.bind_output(item.name, io_bind_device)
+
+        return io_bind_op
+
     def _inference(
         self,
         model: ONNXModel,
@@ -398,12 +414,7 @@ class OnnxEvaluator(OliveEvaluator, framework=Framework.ONNX):
         input_dict = OnnxEvaluator.format_input(input_data, io_config)
 
         if metric.user_config.io_bind:
-            io_bind_op = session.io_binding()
-            io_bind_device = "cuda" if device == "gpu" else "cpu"
-            for k, v in input_dict.items():
-                io_bind_op.bind_cpu_input(k, v)
-            for item in session.get_outputs():
-                io_bind_op.bind_output(item.name, io_bind_device)
+            io_bind_op = OnnxEvaluator.prepare_io_bindings(session, input_dict, device)
 
         for _ in range(warmup_num):
             if metric.user_config.io_bind:
@@ -545,12 +556,7 @@ class OnnxEvaluator(OliveEvaluator, framework=Framework.ONNX):
         input_feed = OnnxEvaluator.format_input(input_feed, io_config)
 
         if metric.user_config.io_bind:
-            io_bind_op = session.io_binding()
-            for k, v in input_feed.items():
-                io_bind_op.bind_cpu_input(k, v)
-            for item in session.get_outputs():
-                io_bind_op.bind_output(item.name, "cuda")
-
+            io_bind_op = OnnxEvaluator.prepare_io_bindings(session, input_feed, Device.GPU)
         latencies = []
         for i in range(warmup_num + repeat_test_num):
             MPI.COMM_WORLD.barrier()  # Synchronize before starting each run
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -291,8 +291,6 @@ class OliveEvaluator(ABC):
 
 
 class OnnxEvaluator(OliveEvaluator, framework=Framework.ONNX):
-    KV_CACHE_ORTVALUES = None
-
     def __init__(self):
         super().__init__()
 
@@ -315,7 +313,9 @@ class OnnxEvaluator(OliveEvaluator, framework=Framework.ONNX):
         }
 
     @staticmethod
-    def prepare_io_bindings(session, input_data, device, device_id: int = 0, shared_kv_buffer: bool = False):
+    def prepare_io_bindings(
+        session, input_data, device, device_id: int = 0, shared_kv_buffer: bool = False, kv_cache_ortvalues: dict = None
+    ):
         """Convert input from numpy array to OrtValue.
 
         session: ONNXRuntime session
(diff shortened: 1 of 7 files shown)


test/unit_test/passes/onnx/test_perf_tuning.py
```diff
@@ -18,6 +18,13 @@ def test_ort_perf_tuning_pass(config, tmp_path):
     p = create_pass_from_dict(OrtPerfTuning, config, disable_search=True)
     output_folder = str(tmp_path / "onnx")
 
+    # check config
+    # pylint: disable=protected-access
+    assert p._config["device"] is not None
```
guotuofeng, line 23: do we need add test cases that user explicitly specify provider_list?

trajepl: Updated. Add will check it in line 34.

test/unit_test/passes/onnx/test_perf_tuning.py
```diff
@@ -11,7 +12,14 @@
 from olive.passes.onnx import OrtPerfTuning
 
 
-@pytest.mark.parametrize("config", [{"input_names": ["input"], "input_shapes": [[1, 1]]}, {}])
+@pytest.mark.parametrize(
+    "config",
+    [
+        {"input_names": ["input"], "input_shapes": [[1, 1]]},
+        {"providers_list": ["CPUExecutionProvider", "CUDAExecutionProvider"], "device": "gpu"},
```
guotuofeng, line 19: need assert the EP used by perf tuning is the list besides checking none.

trajepl: Fixed in another PR.https://github.com/microsoft/Olive/pull/700

olive/evaluator/olive_evaluator.py
```diff
@@ -312,6 +314,49 @@ def format_input(input_data, io_config):
...
+
+        use_fp16 = any(v.dtype == np.float16 for v in input_data.values())
+        io_bind_op = session.io_binding()
+        io_bind_device = "cuda" if device == "gpu" else "cpu"
+
+        if shared_kv_buffer:
+            OnnxEvaluator.KV_CACHE_ORTVALUES = OnnxEvaluator.KV_CACHE_ORTVALUES or {}
+
+        for k, v in input_data.items():
+            if shared_kv_buffer and use_fp16 and ("cache" in k or "past_key_values" in k):
```
jambayk, line 339: Could you add a comment about `"cache"` being from the microsoft version model (llama) and `past_key_values` from the hf version of the model?

olive/evaluator/olive_evaluator.py
```diff
@@ -312,6 +314,49 @@ def format_input(input_data, io_config):
...
+                if k not in OnnxEvaluator.KV_CACHE_ORTVALUES:
```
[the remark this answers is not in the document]
guotuofeng: good catch for memory sake

olive/evaluator/olive_evaluator.py
```diff
@@ -312,6 +314,49 @@ def format_input(input_data, io_config):
...
+                    OnnxEvaluator.KV_CACHE_ORTVALUES[k] = OrtValue.ortvalue_from_numpy(v, io_bind_device, device_id)
```
guotuofeng, line 341: it seems we could move the cache logic into a separated class since they are not be evaluator responsibility. 
Instead, we only need call something like cache.update() here

olive/evaluator/olive_evaluator.py
```diff
@@ -312,6 +314,49 @@ def format_input(input_data, io_config):
...
+                else:
+                    OnnxEvaluator.KV_CACHE_ORTVALUES[k].update_inplace(v)
+                ort_v = OnnxEvaluator.KV_CACHE_ORTVALUES[k]
+            else:
+                ort_v = OrtValue.ortvalue_from_numpy(v, io_bind_device, device_id)
+            io_bind_op.bind_ortvalue_input(k, ort_v)
+
+        for item in session.get_outputs():
+            name = item.name
+            if shared_kv_buffer and use_fp16 and ("out" in name or "present" in name):
```
guotuofeng, line 351: move the model specific logic into a separated function/method: process_attention_outputs?

olive/evaluator/olive_evaluator.py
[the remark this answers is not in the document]
trajepl: Emm, I think the shared buffer is something like to pin a piece of GPU memory to accelerate generation(token) task.  That is why I added: `TODO(trajep): how shared_kv_buffer works with generation task in olive`

Note that, there is the operation to bind ortvalue output into shared buffer. Then actually I think in our latency measurement, the input should be chagned.
`io_bind_op.bind_ortvalue_output(name, OnnxEvaluator.KV_CACHE_ORTVALUES[input_name])`


We need a central attribute to ensure it is shared among multi-running. Maybe class attributes is too broader. Then how about instance attribute?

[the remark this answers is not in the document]
jambayk: For binding the output to the input, we don't need a central place to store the cache. The `prepare_io_binding` method can have a local `cache` dict for it. 
The cache only comes into play if we want to run the same model with different inputs. Even then, it depends on whether the gpu memory gets freed when the `io_bind_op` object goes out of scope. If it does, there is no need for a central cache since it gets refreshed at each iteration of the loop (over dataloader). 
The only advantage would could be speed up over the io_binding process itself since we won't need to allocate memory every time. If we do this, we could keep a cache dict inside `_evaluate_onnx_latency`.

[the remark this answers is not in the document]
trajepl: Let us not limit in Olive latency measurement firstly, the key advantage to enable shared buffer is to avoid memory allocation/release in every forward call. The kv cache should be updated_inplace since 2rd forward call. (actually shared buffer should be able cover all the transformers kv cache case not only GQA fp16.) 
For this goal, I think we are on the same line.

Actually from my side, I do not think shared buffer is required in our latency measurement.  It is more like the part of generation predication.

Then it comes to latency measurement, what if we enable shared buffer in **Olive**:
1. w/o shared buffer: the input are always be same including the kv cache.
2. with shared buffer: the input_ids, attention masks, position id are same, but kv cache got refreshed as it bind ortvalue output to shared buffer. 
Right?


But anyway in order to support the shared buffer in Olive(at least we have the capacity in Olive), we can leverage latency as a test. 
I would be agree the idea of `we could keep a cache dict inside _evaluate_onnx_latency.`
Like:

```python
def _evaluate_onnx_latency(...):
    ...
    kv_cache_buffer = {}
    prepare_io_bindings(..., kv_cache_buffer)
    ....
    del kv_cache
```

olive/evaluator/olive_evaluator.py
trajepl: Replace the central cache with local dict under `_evaluate_onnx_latency`.
And I used the feature of python dict reference to keep the kv_cache shared across the latency measurement.

Then it looks clear to update the python dict.

olive/evaluator/olive_evaluator.py
trajepl: See above comments.

jambayk (approved):

guotuofeng (approved):

PR #701, opened 2023-11-07
Add LLaMA 2 example for DirectML

PatriceVignola:
This adds the LLaMA 2 optimizations for DirectML with examples, and a sample ChatApp that was inspired from [this](https://github.com/microsoft/Llama-2-Onnx/tree/main/ChatApp) but stripped down to an MVP and customized for DirectML.

It also adds a "CompositePyTorchModel" which follows the same principle as the Composite Optimum model but for raw pytorch models instead.


olive/passes/onnx/conversion.py
```diff
@@ -137,19 +137,16 @@ def _convert_model_on_device(
             if isinstance(dummy_inputs, dict):
                 dummy_input_keys = set(dummy_inputs.keys())
 
-                # handle dummy inputs for hf model with past, which has past_key_values
+                # handle dummy inputs for model with past, which has past_key_values
                 # match input names in `past_key_values.(hidden_layer_num).(key|value)` pattern
-                from transformers.modeling_utils import PreTrainedModel
-
-                if issubclass(type(pytorch_model), PreTrainedModel):
```
PatriceVignola, line 144: We use the same cache format as huggingface here, so we can leverage this logic without restricting it only for `PreTrainedModel`

.pre-commit-config.yaml
```diff
@@ -27,6 +27,7 @@ repos:
     rev: v0.3.1
     hooks:
       - id: absolufy-imports
+        exclude: examples/directml/llama_v2/chat_app/
```
guotuofeng, line 30: need to add this line into pyproject.toml,

please refer to https://github.com/pylint-dev/pylint/blob/df0800ec22c108cb1db244b5c8c4b25db79c58ba/examples/pyproject.toml#L2 and
https://github.com/microsoft/Olive/blob/main/pyproject.toml#L180

olive/model/__init__.py
```diff
@@ -759,6 +759,43 @@ def to_json(self, check_object: bool = False):
         return serialize_to_json(config, check_object)
 
 
+class CompositePyTorchModel(PyTorchModel):
```
guotuofeng, line 762: I don't see where this class is used?

guotuofeng: what's the difference between this one and OptimumModel?

guotuofeng: It seems OptimumModel could be replaced with this one.

pyproject.toml
```diff
@@ -179,4 +179,3 @@ classmethod-decorators = ["classmethod", "pydantic.validator", "pydantic.root_va
 "examples/**" = ["INP001"]
 "olive/engine/packaging/sample_code/ONNXModel/python/**" = ["INP001"]
 "scripts/**" = ["INP001", "PTH123"]
-"examples/directml/llama_v2/chat_app/**"
```
guotuofeng: the item here is a key=["rules1 to be ignored", "rules2 to be ignored", ...]

PatriceVignola: But this is for ruff right? How can I stop `absolufy-import` from modifying files in that folder since it's a different hook?

guotuofeng: do you mean ignore for pylint? the ignore-paths is what you are looking for 
https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html
 https://github.com/pylint-dev/pylint/blob/3a4bb28304a276b47cf3b6f0ad9d3d5e39225813/examples/pyproject.toml#L45

olive/model/__init__.py
```diff
@@ -749,18 +749,52 @@ def to_json(self, check_object: bool = False):
...
+                del config_copy["name"]
+
+                self.model_component_names.append(model_name)
+                self.model_components[model_name] = validate_config(config_copy, ModelConfig).create_model()
+        else:
+            raise NotImplementedError
 
     def to_json(self, check_object: bool = False):
         config = super().to_json(check_object)
         config["config"].update({"model_components": self.model_components})
```
jambayk, line 783: I don't think the type of `self.model_components` is correct after the initialization? the original `model_components` was a list but `self.model_components` is a dict for `PYTORCH_ENTIRE_MODEL`.

pyproject.toml
guotuofeng: "examples/directml/llama_v2/chat_app/**" = ["TID252"]

https://docs.astral.sh/ruff/rules/#flake8-simplify-sim

trajepl:
For reference: 
in the coming release of optimum, the optimum exporter can export only the merged models.
https://github.com/huggingface/optimum/pull/1257/files

Also, with nightly ort/optimum, Olive optimize llama2 as following:
https://github.com/microsoft/Olive/tree/main/examples/llama2

PatriceVignola:
> what's the difference between this one and OptimumModel? could we leverage OptimumModel or unify them? It seems the two class is similiar.

I decoupled OptimumModel and CompositePyTorch model. They look similar at first glance, but they don't have much in common aside from the `model_components` part. OptimumModel is a very lightweight class that only needs to return strings of the actual model names which get handled by optimum in its own conversion pass, whereas CompositePyTorch model is a container of multiple other PyTorchModels that get handled recursively in the main OnnxConversion pass.

jambayk:
> > what's the difference between this one and OptimumModel? could we leverage OptimumModel or unify them? It seems the two class is similiar.
> 
> I decoupled OptimumModel and CompositePyTorch model. They look similar at first glance, but they don't have much in common aside from the `model_components` part. OptimumModel is a very lightweight class that only needs to return strings of the actual model names which get handled by optimum in its own conversion pass, whereas CompositePyTorch model is a container of multiple other PyTorchModels that get handled recursively in the main OnnxConversion pass.

That's what I thought too when looking at the code. 
On another note: I am thinking of removing the `OptimumModel` class entirely later. It is only used for `OptimumConversion` but I don't see a reason why we cannot just use `PyTorchModel` + hf_config and `OptimumConversion` + `model_components` as pass config. 
There are other olive passes that also only support pytorch models with hf_config.

olive/passes/onnx/optimum_conversion.py
```diff
@@ -33,6 +34,7 @@ def _run_for_config(
         self, model: OptimumModel, data_root: str, config: Dict[str, Any], output_model_path: str
     ) -> Union[ONNXModel, CompositeOnnxModel]:
         assert len(model.model_components) > 0
+        assert model.model_file_format == ModelFileFormat.OPTIMUM
```
guotuofeng, line 37: >ModelFileFormat.OPTIMUM [](http://example.com/codeflow?start=42&length=23)

I don't see the model_file_format is used for OPTIMUM. 

Do we really need it?

olive/model/__init__.py
```diff
@@ -762,6 +762,43 @@ def to_json(self, check_object: bool = False):
         return serialize_to_json(config, check_object)
 
 
+class CompositePyTorchModel(PyTorchModel):
+    def __init__(self, model_components: List[Dict[str, Any]], **kwargs):
+        kwargs = kwargs or {}
+        kwargs["model_file_format"] = ModelFileFormat.PYTORCH_ENTIRE_MODEL
```
guotuofeng, line 768: >ModelFileFormat.PYTORCH_ENTIRE_MODEL [](http://example.com/codeflow?start=38&length=36)

it will override the model_file_format if user specify it explicitly?

olive/passes/onnx/optimum_conversion.py
PatriceVignola: I didn't mean to touch this file actually, I forgot to revert t hose changes when I reverted the merge. I'll get rid of it.

olive/model/__init__.py
PatriceVignola: Can we use a `COMPOSITE_MODEL` for those then? Even if we don't explicity use it now, it can be useful to differentiate between models and model containers.

guotuofeng: I think it is ok to have a dedicated COMPOSITE_MODEL model file format for these composite models. 
How do you think? @jambayk

jambayk: Yes, that sounds good to me too.

guotuofeng (approved):

guotuofeng (approved):

PR #703, opened 2023-11-07
Accept device and dtype in `OnnxConversion`, Add `OnnxBnb4Quantization`, llama2 e2e qlora example

jambayk:
## Describe your changes
**OnnxConversion**
- The code has been refactored a bit so that the methods `_convert_model_on_device` and `_convert_distributed_model_on_device` have similar signature and behavior
- Doc strings for the methods
- The above also makes the logic for composite models simpler
- New config parameters added to make the pass more flexible:
  - `use_device`: users can specify what device to run conversion on. For example, there is enough gpu memory to speed up conversion or model is float16. 
  - `torch_dtype`: torch data type to cast the model to before conversion. For example, model is originally float16 but want to convert on cpu and run another pass later to convert to float16.  
- It is more aware about loading models using `hf_config`.
  - Check and update `torch_dtype`
  - Check if model is quantized using bitsandbytes. 
  
**OnnxBnb4Quantization**
- New quantization pass to quantize a model using nf4/fp4. Uses the `MatMulBnb4` quantizer and contrib op 
- It can handle both pass config and model attributes such as `quantized_modules` and `quantization_config` 
- Using this pass, we usually don't want to quantize all `MaMul` nodes, only the ones that were originally quantized in the source model to retain accuracy 

**llama2**
- Add E2E qlora + ort optimization workflow
- Update readme

qlora-e2e

`OnnxConversion` pass has new parameters `use_device` and `torch_dtype` for more flexible conversion.
New `OnnxBnb4Quantization` to quantize an ONNX model using `FP4/NF4` data types.


olive/common/utils.py
```diff
@@ -196,3 +214,21 @@ def get_attr(module, attr, fail_on_not_found=False):
...
+    :return: list of submodules names.
+    """
+    submodules = set()
+    for name, submodule in module.named_modules():
+        if isinstance(submodule, submodule_types):
+            if full_name:
+                submodules.add(name)
+            else:
+                submodules.add(name.split(".")[-1])
+    return list(submodules)
```
guotuofeng, line 234: yield is better than the list?

olive/model/__init__.py
```diff
@@ -577,6 +577,18 @@ def load_model(self, rank: int = None) -> torch.nn.Module:
 
         return model
 
+    def loads_model_from_hf_config(self) -> bool:
```
guotuofeng, line 580: the name seems like a load behavior but it is used as a flag. better to change the name

olive/common/utils.py
jambayk: This returns a list of unique values so I am not sure if `yield` works better. Also, it is currently only used as a list for the qlora pass and conversion pass. Both expect a list.

olive/passes/onnx/bnb_quantization.py
```diff
@@ -0,0 +1,124 @@
...
+            # extract quant_type from model_attributes if not specified in config
+            if not quant_type:
+                quant_config = model.model_attributes.get("quantization_config")
+
+                if quant_config.get("load_in_8bit"):
+                    raise ValueError("load_in_8bit is not supported. Only 4-bit quantization is supported.")
+
+                quant_type = quant_config.get("bnb_4bit_quant_type")
+                if not quant_type:
+                    raise ValueError("quant_type is required.")
```
guotuofeng, line 73: bnb_4bit_quant_type?

olive/model/__init__.py
jambayk: I used `loads` as an adjective and not verb. Maybe something like `model_loaded_from_hf_config` is clearer?

olive/passes/onnx/bnb_quantization.py
```diff
@@ -0,0 +1,124 @@
...
+            for attr in node.attribute:
+                if attr.type == onnx.AttributeProto.GRAPH:
+                    # recursive call to take care of sub-graph
+                    matmul_nodes += self._find_matmul_nodes(attr.g)
+                elif attr.type == onnx.AttributeProto.GRAPHS:
+                    for subgraph in attr.graphs:
+                        # recursive call to take care of sub-graph
+                        matmul_nodes += self._find_matmul_nodes(subgraph)
+            if node.op_type == "MatMul":
+                matmul_nodes.append(node.name)
```
guotuofeng, line 122: yield has better performance than loop.

olive/passes/onnx/bnb_quantization.py
jambayk: `qaunt_type` is the actual pass config parameter name. `bnb_4bit_quant_type` is only used to extract the value from quantization_config. 
If the pass cannot find it in the model attributes, then they should provide it as a pass config parameter.

examples/llama2/llama2_qlora.json
```diff
@@ -0,0 +1,135 @@
+{
+    "input_model":{
+        "type": "PyTorchModel",
+        "config": {
+            "model_script": "user_script.py",
+            "io_config": "get_merged_decoder_with_past_io_config",
```
guotuofeng, line 6: Nit: the io_config name is a little bit confused since it is a function. By contrast, the next line dummpy_inputs_func has a suffix func.

Are there any best practice to normalize these naming?

olive/model/__init__.py
guotuofeng: can_model_load_from_hf_config?

olive/common/utils.py
jambayk: Will address this in a follow up PR as discussed offline

olive/passes/onnx/bnb_quantization.py
jambayk: Will address this in a follow up PR as discussed offline

examples/llama2/llama2_qlora.json
jambayk: I think what happened is that some earlier PR made io_config: dict, str, func when it was originally only a dict. 
@trajepl has created a related TODO https://github.com/microsoft/Olive/blob/main/olive/model/hf_utils.py#L25

I can create a new work item to track this.

olive/passes/onnx/bnb_quantization.py
```diff
@@ -0,0 +1,124 @@
...
+from olive.hardware import AcceleratorSpec
+from olive.model import ONNXModel
+from olive.passes import Pass
+from olive.passes.onnx.common import get_external_data_config, model_proto_to_olive_model
+from olive.passes.pass_config import PassConfigParam
+
+logger = logging.getLogger(__name__)
+
+
+class OnnxBnb4Quantization(Pass):
```
guotuofeng, line 22: >OnnxBnb4Quantization [](http://example.com/codeflow?start=6&length=20)

do we need add unit test for the new pass? especially for config validation?

olive/model/__init__.py
jambayk: Done! Changed to `is_model_loaded_from_hf_config`

examples/llama2/llama2_qlora.json
```diff
@@ -0,0 +1,135 @@
...
+            "config": {
+                "save_as_external_data": true,
+                "all_tensors_to_one_file": true
+            }
+        }
+    },
+    "engine": {
+        "search_strategy": false,
+        "evaluator": "merged_evaluator",
+        "target": {
```
trajepl, line 125: You can skip to set system as GPU. It will be inferred automatically based on ep.

examples/llama2/llama2_qlora.json
```diff
@@ -0,0 +1,135 @@
...
+            "config": {
+                "save_as_external_data": true,
+                "all_tensors_to_one_file": true,
+                "model_type": "gpt2",
+                "opt_level": 0,
+                "only_onnxruntime": false,
+                "keep_io_types": false,
+                "float16": true,
+                "optimization_options": {
+                    "enable_rotary_embeddings": false
```
trajepl, line 110: Any reason why set rotary embeddings as False? Should we do the same thing in the template json?

olive/passes/onnx/bnb_quantization.py
```diff
@@ -0,0 +1,125 @@
...
+    """Quantize MatMul nodes in ONNX model using 4bit FP4/NF4 quantization."""
+
+    @staticmethod
+    def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        config = {
+            "quant_type": PassConfigParam(
+                type_=str,
```
trajepl, line 29: Nit: Better to set the default value? Somehow it can help on auto optimization.

olive/passes/onnx/quantization.py
```diff
@@ -530,7 +530,7 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
             ),
             "nodes_to_exclude": PassConfigParam(
                 type_=list,
-                default_value=[],
+                default_value=None,
```
trajepl, line 533: Double check: in this case `[]` should be fine?

examples/llama2/llama2_qlora.json
jambayk: The rotary embeddings do not work for this model. I think it is because the model graph has changed due to addition of the lora adapters and there are shape incompatibilities. (the error says it expect 3d input but got 4). 
But it works for the llama2.py example since it is the original model graph.

olive/passes/onnx/quantization.py
jambayk: `[]` also works but since we don't use `[]` as default value in other parts of the code base, I changed it to None. The quantizer can also handle None.

olive/passes/onnx/bnb_quantization.py
jambayk: This is similar to the `model_type`, `num_heads` etc in the transformers optimization pass. No default value is set so that we can check model_attributes if the pass config value is not given.

jambayk: Also, the bnb4 quantization is not really useful by itself as a model compression technique since accuracy is not good. It works best when used with fine-tuned lora adapters from qlora. So, we might not even try this pass in auto optimization.

examples/llama2/llama2_qlora.json
jambayk: Done!

guotuofeng (approved):

PR #704, opened 2023-11-08
Add Falcon example

xiaoyu-work:
## Describe your changes

Add Falcon example

### DIFF
--- a/examples/falcon/config.json
+++ b/examples/falcon/config.json
@@ -46,6 +46,7 @@
         },
         "transformers_optimization": {
             "type": "OrtTransformersOptimization",
+            "disable_search": true,
             "config": {
                 "float16": true
             }
--- a/examples/falcon/config.json
+++ b/examples/falcon/config.json
@@ -1,0 +1,70 @@
+{
+    "input_model":{
+        "type": "PyTorchModel",
+        "config": {
+            "hf_config": {
+                "model_name": "tiiuae/falcon-7b",
+                "task": "text-generation",
+                "dataset": {
+                    "data_name":"tiiuae/falcon-refinedweb",
+                    "split": "train",
+                    "input_cols": ["content"],
+                    "batch_size": 1
+                }
+            }
+        }
+    },
+    "systems": {
+        "local_system": {
+            "type": "LocalSystem",
+            "config": {
+                "accelerators": ["GPU"]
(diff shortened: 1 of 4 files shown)


examples/falcon/config.json
jambayk, line 1: Can you add a README and requirements file for this example?

xiaoyu-work: I don't think there is any additional packages required for this example. I will add a README here.

xiaoyu-work: Oh sorry maybe `onnxruntime` is the only one...

examples/gptj/requirements.txt
```diff
@@ -1,4 +1,3 @@
 datasets
 neural-compressor
 onnxruntime
-transformers
```
xiaoyu-work, line 4: `transformers` is a dependency of Olive

examples/falcon/config.json
jambayk: what about `datasets`?

examples/falcon/requirements.txt
```diff
@@ -0,0 +1 @@
+onnxruntime
```
jambayk, line 1: `onnxruntime-gpu`?

examples/falcon/config.json
xiaoyu-work: There is no separate script to use dataset here.

jambayk: Since this example is using hf datasets component, `datasets` becomes a requirement. https://github.com/microsoft/Olive/blob/main/olive/data/component/load_dataset.py#L26

xiaoyu-work: If this is a part of Olive logic, I prefer this package should be installed by `--setup` command. I also noticed that if we run `--setup` command, we don't need `onnxruntime-gpu` installed in `requirements.txt`. In general, seems `requirements.txt` is only needed for user_script package.

jambayk: It is true about the `onnxruntime-gpu` dependency. I updated it last time to check for eps and user environment.
But we currently don't have such check for `datasets`.

xiaoyu-work: Yeah, I updated requirements for now. I'll create a work item for this.

jambayk (approved):

PR #705, opened 2023-11-08
add trust_remote_code arg

xiaoyu-work:
## Describe your changes

Add trust_remote_code arg. 

Note: even with this change, I can still get a prompt asking about trusting the custom code or not. Need more investigation about this.

### DIFF
--- a/examples/falcon/config.json
+++ b/examples/falcon/config.json
@@ -53,14 +53,7 @@
         }
     },
     "engine": {
-        "search_strategy": {
-            "execution_order": "pass-by-pass",
-            "search_algorithm": "tpe",
-            "search_algorithm_config": {
-                "num_samples": 3,
-                "seed": 0
-            }
-        },
+        "search_strategy": false,
         "evaluator": "common_evaluator",
         "host": "local_system",
         "target": "local_system",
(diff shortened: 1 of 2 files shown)


jambayk (approved):
We already have an example that provides `trust_remote_code` as `True` by using the `extra_args` nature of this config. https://github.com/microsoft/Olive/blob/main/examples/phi/phi_qlora_tinycodes.json#L9

But it does not hurt to expose this arg as a parameter since it is important and common. Same for `token` which we could also expose.

examples/falcon/config.json
```diff
@@ -53,14 +53,7 @@
...
     "engine": {
-        "search_strategy": {
-            "execution_order": "pass-by-pass",
-            "search_algorithm": "tpe",
-            "search_algorithm_config": {
-                "num_samples": 3,
-                "seed": 0
-            }
-        },
+        "search_strategy": false,
```
jambayk, line 56: BTW, should the hf config of the input model be updated to use `trust_remote_code=True`?

xiaoyu-work: Falcon model doesn't need this but Yi model needs. I'm working on yi model.

PR #707, opened 2023-11-08
📶 Make engine optional to simplify the default user config

trajepl:
## Describe your changes
Make engine optional to simplify the default user config

### DIFF
--- a/olive/workflows/run/config.py
+++ b/olive/workflows/run/config.py
@@ -2,6 +2,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+import logging
 from pathlib import Path
 from typing import Dict, List, Union
 
@@ -18,6 +19,8 @@ from olive.model import ModelConfig
 from olive.passes import FullPassConfig, Pass
 from olive.resource_path import AZUREML_RESOURCE_TYPES
 from olive.systems.system_config import SystemConfig
+
+logger = logging.getLogger(__name__)
 
 
 class RunPassConfig(FullPassConfig):
@@ -62,8 +65,8 @@ class RunConfig(ConfigBase):
     data_root: str = None
     data_configs: Dict[str, DataConfig] = None
     evaluators: Dict[str, OliveEvaluatorConfig] = None
+    engine: RunEngineConfig = None
     pass_flows: List[List[str]] = None
-    engine: RunEngineConfig
     passes: Dict[str, RunPassConfig] = None
 
     @validator("input_model", pre=True)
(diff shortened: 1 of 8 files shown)


olive/workflows/run/config.py
```diff
@@ -80,6 +83,12 @@ def insert_aml_client(cls, v, values):
             v["config"]["model_path"]["config"]["azureml_client"] = values["azureml_client"]
         return v
 
+    @validator("engine", pre=True, always=True)
+    def default_engine_config(cls, v):
+        if not v:
+            v = RunEngineConfig()
+        return v
```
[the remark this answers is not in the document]
jambayk: If the `search_startegy` is not provided with `True` (for joint exhaustive search) or configuration dict, it is already no-search  by default 
https://github.com/microsoft/Olive/blob/main/olive/engine/config.py#L24
https://github.com/microsoft/Olive/blob/main/olive/workflows/run/config.py#L166
https://github.com/microsoft/Olive/blob/main/olive/engine/engine.py#L62

examples/test/test_bert_cuda_gpu.py
```diff
@@ -23,12 +23,20 @@ def setup():
 @pytest.mark.parametrize("execution_order", ["joint", "pass-by-pass"])
 @pytest.mark.parametrize("system", ["aml_system"])
 @pytest.mark.parametrize("olive_json", ["bert_cuda_gpu.json"])
+@pytest.mark.parametrize("no_engine_and_evaluator", [True, False])
```
jambayk, line 26: The test already takes a long time to run. I think it might be better to test different settings of the engine as unit tests? 
And keep the example tests as more like integration tests. Or move this `no_engine_and_evalutor` to a faster running test (without aml) like `bert_ptq_cpu`?

guotuofeng: yeah, unit test should be more appropriated than integration test.

olive/workflows/run/config.py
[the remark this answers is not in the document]
trajepl: Yes. There are the logics:
1. evaluate input model only evaluate_input_model and evaluator is not None
2. run search: no_search is False && search strategy is joint or exhaustive

When we give empty dict or default value to engine config, the search strategy is None, the evaluator is None.
In this way, following options are **disabled**:
1. evaluate input/output model
2. run passes with search
3. engine config is optional
image

olive/workflows/run/config.py
```diff
@@ -80,6 +83,12 @@ def insert_aml_client(cls, v, values):
             v["config"]["model_path"]["config"]["azureml_client"] = values["azureml_client"]
         return v
 
+    @validator("engine", pre=True, always=True)
+    def default_engine_config(cls, v):
+        if v is None:
+            v = {}
+        return v
```
guotuofeng, line 90: what's the default config? should we document the default engine config?

trajepl: https://microsoft.github.io/Olive/overview/options.html#engine-information 
See this, also updated with clear default value.

jambayk (approved):

PR #713, opened 2023-11-10
Add option to disable ORT postprocessing in OptimumMerging pass

PatriceVignola:
The OptimumMerging pass has a postprocessing process to run the model through ORT to do some additional optimizations, but doing this can hardcode some imports and opset versions into the ONNX model that are not even actually used, which makes the model incompatible with older ORT versions.

### DIFF
--- a/olive/passes/onnx/optimum_merging.py
+++ b/olive/passes/onnx/optimum_merging.py
@@ -28,7 +28,15 @@ class OptimumMerging(Pass):
 
     @staticmethod
     def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
-        return get_external_data_config()
+        config = {
+            "run_ort_postprocess": PassConfigParam(
+                type_=bool,
+                default_value=True,
+                description="Whether to run the model through ORT at the end to apply some optimizations.",
+            ),
+        }
+        config.update(get_external_data_config())
+        return config
 
     def _run_for_config(
         self, model: CompositeOnnxModel, data_root: str, config: Dict[str, Any], output_model_path: str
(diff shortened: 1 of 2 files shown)


olive/passes/onnx/optimum_merging.py
```diff
@@ -28,7 +28,15 @@ def is_accelerator_agnostic(accelerator_spec: AcceleratorSpec) -> bool:
 
     @staticmethod
     def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
-        return get_external_data_config()
+        config = {
+            "run_ort_postprocess": PassConfigParam(
```
guotuofeng, line 32: this seems be https://onnxruntime.ai/docs/performance/model-optimizations/graph-optimizations.html. 

Do we need other GraphOptimizationLevel values?

olive/passes/onnx/optimum_merging.py
```diff
@@ -61,12 +69,13 @@ def new_byte_size_func(_):
...
+        if config["run_ort_postprocess"]:
+            # Doing a dry run of ORT allows us to remove the initializers that were orphaned by the merging step
+            sess_options = onnxruntime.SessionOptions()
+            sess_options.graph_optimization_level = onnxruntime.GraphOptimizationLevel.ORT_ENABLE_BASIC
+            sess_options.optimized_model_filepath = output_model_path
 
-        execution_provider = self.accelerator_spec.execution_provider
-        onnxruntime.InferenceSession(output_model_path, sess_options, providers=[execution_provider])
+            execution_provider = self.accelerator_spec.execution_provider
+            onnxruntime.InferenceSession(output_model_path, sess_options, providers=[execution_provider])
```
guotuofeng, line 79: would the write file fail since the input model and optimized_model_filepath point to the same value?

guotuofeng: we might need add an unit test to cover this case. I am supposing it would fail

PatriceVignola: It doesn't fail, it's the default behavior since the beginning. It's able to write to the same location as the input (once the ONNX model is loaded in memory it doesn't keep a handle on it, so it's able to save to the same location).

olive/passes/onnx/optimum_merging.py
PatriceVignola: Optimizations should be done the merging step. The goal of this postprocessing ORT step was only to run the most basic optimizations for the purpose of removing unused initializers, and even for that purpose it doesn't seem to work 100% with some models.

olive/passes/onnx/optimum_merging.py
jambayk: I asked a similar question before but wanted to confirm with the external data. Do you know what happens to those? 
Does the graph optimization know what the original locations of the external data were? or they don't matter since it only cares about the .onnx file and the graph?

PatriceVignola: I don't think it cares about where the weights are located (since those shouldn't change), but the .onnx file has a reference to the location of external data anyway so it **should** be able to find it if needed.

guotuofeng: I am not sure whether the path is locked or not in Windows by considering the bug fix https://github.com/microsoft/onnxruntime/pull/17697. It would be better we can add unit test to cover it.

trajepl:
Gentle ping: do we still this PR? @PatriceVignola

jambayk:
Closing the PR for now since it's become stale.

PR #721, opened 2023-11-13
🧵 Throughput metric

trajepl:
## Describe your changes
Throughput metric


olive/evaluator/metric.py
```diff
@@ -114,7 +140,7 @@ def validate_sub_types(cls, v, values):
                 raise ValueError(f"higher_is_better must be specified for ranked custom metric: {v['name']}")
             return v
         # name
-        sub_type_enum = AccuracySubType if values["type"] == MetricType.ACCURACY else LatencySubType
+        sub_type_enum = MetricType.map_to_metric_type(values["type"])
```
guotuofeng, line 143: why do we need convert from values["type"] to sub_type_enum? could not we use something like:
```python
if values["type"] == MetricType.ACCURACY:
    v["name"] = AccuracySubType(v["name"])
    metric_config_cls = HuggingfaceMetrics.get_config_class()
```

olive/evaluator/metric.py
```diff
@@ -12,16 +12,28 @@
...
-from olive.evaluator.metric_config import LatencyMetricConfig, MetricGoal, get_user_config_class
+from olive.evaluator.metric_config import LatencyMetricConfig, MetricGoal, ThroughputMetricConfig, get_user_config_class
 
 logger = logging.getLogger(__name__)
 
 
 class MetricType(str, Enum):
     ACCURACY = "accuracy"
     LATENCY = "latency"
+    THROUGHPUT = "throughput"
```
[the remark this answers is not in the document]
trajepl: I was considering that before. But it will be hard to compare the metrics, as for latency, we expect lower is better, but for throughput, higher is better.
Also it will be more clear use separated way even we need to compute latency twice.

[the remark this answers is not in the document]
jambayk: Oh yeah, I forgot that the direction if opposite. Sounds good.

olive/evaluator/metric.py
trajepl: Updated.

docs/source/extending_olive/design.md
```diff
@@ -47,7 +47,7 @@ An Evaluator is used to evaluate a model on a specific **target** and return val
 
 To initialize an Evaluator, the user must provide a list of **Metrics** and the **target** system.
 
-Olive provides two in-build **Metrics**, accuracy and latency, along with the options of the user to provide their own
+Olive provides three in-build **Metrics**, accuracy, latency(ms) and throughput(tps, token per second), along with the options of the user to provide their own
```
guotuofeng, line 50: is the for generative model? how could we calculate the token number?

trajepl: Nope, not for generative models specially.  It is an estimation to see how many forward call can happen in one second.

olive/evaluator/olive_evaluator.py
```diff
@@ -313,6 +353,22 @@ def compute_latency(metric: Metric, latencies: Any) -> MetricResult:
             )
         return MetricResult.parse_obj(metric_res)
 
+    @staticmethod
+    def compute_throughput(metric: Metric, latencies: Any) -> MetricResult:
+        """Compute latency metrics."""
```
xiaoyu-work, line 358: Typo: latency -> throughput

olive/evaluator/olive_evaluator.py
```diff
@@ -313,6 +353,22 @@ def compute_latency(metric: Metric, latencies: Any) -> MetricResult:
...
+        latency_metrics = OliveEvaluator.latency_helper(latencies)
+        metric_res = {}
+        batch_size = metric.user_config.batch_size
+        for sub_type in metric.sub_types:
+            latency_sub_type_name = LatencySubType.MAX if sub_type.name == ThroughputSubType.MIN else LatencySubType.MIN
```
xiaoyu-work, line 363: Do we only have 2 throughput metrics? What about other subtypes for throughput?

trajepl: Good catch. Throughput has similar sub types with latency, but with different `higher_is_better`. For example:
1. for max throughput, we should use batch_size / latency_min.
2. for min throughput, we should use batch_size / latency_max.
3. for percentile@90 of thoughput(higher -> lower), we should use batch_size / latency_percentile@90(lower->higher)

olive/evaluator/olive_evaluator.py
trajepl: Fixed.

olive/evaluator/olive_evaluator.py
xiaoyu-work: Can you explain more about how to define "throughput" here? In which case for throughput "lower is better"?

trajepl: Throughput is always `higher is better`, that says `higher_is_better == True`. But for latency `higher_is_better == False`

> hroughput has similar sub types with latency, but with different higher_is_better

xiaoyu-work: If so, except `avg`, for other `Pn` throughput, I don't think it matches latency `Pn` metric. Should it be something like Latency `Pn` -> Throughput `P(100-n)`?

trajepl: Good question, but I think it is fine. 
latency P(75) means, 75% cases, the latencies number <= the value P(75). 
When it switches to throughput, it means 75% of cases, the throughput number >= the value P(75). 
The P(n) should always be the most bad case of n%. (but the `bad` for latency is higher, for throughput is lower)

xiaoyu-work: This makes sense to me. Thanks!

xiaoyu-work (approved):

olive/evaluator/olive_evaluator.py
trajepl: Thanks!

xiaoyu-work (approved):

nicolas-mng:
Nice! When will this be released?
Thanks

nicolas-mng:
I work behind some proxies which makes it difficult to `pip install` from a git repository directly but it can wait a bit :). Thank you for being reactive!

PR #727, opened 2023-11-15
Add dataconfig support for aml system

xiaoyu-work:
## Describe your changes

1. Add data config support for aml system. Some passes like LoRA or QLoRA need datasets. The data config should also be uploaded to aml job.
2. Add huggingface token support for hf model and hf data. This is needed for those jobs running on AML.
3. Fix minor bugs.


olive/data/component/load_dataset.py
```diff
@@ -27,6 +27,11 @@ def huggingface_dataset(data_dir, data_name=None, subset=None, split="validation
     """Create a dataset from huggingface datasets."""
     from datasets.utils.logging import disable_progress_bar, set_verbosity_error
 
+    if kwargs.get("huggingface_token") is not None:
+        from huggingface_hub import login
+
+        login(token=kwargs["huggingface_token"])
```
guotuofeng, line 33: we will pass the token as text in aml job? are there any security concerns?

olive/model/hf_utils.py
```diff
@@ -207,6 +207,7 @@ class HFConfig(ConfigBase):
     # feature is optional if task is specified and don't need past
     # else, provide feature such as "causal-lm-with-past"
     feature: str = None
+    token: str = None
```
guotuofeng, line 210: I am not sure it is right approach. But my gut feeling to keep a text token is not a good practice.

olive/data/component/load_dataset.py
trajepl: For the choice: save the token under `~/.cache/huggingface/token`, somehow it can be done by user out of Olive.

After set it in cache file, `get_token_permission` can return the right results even without token passed.
image

xiaoyu-work: @trajepl thanks for the heads up. Instead, I prefer directly getting token from local huggingface cache. I can make this change.

olive/data/container/data_container.py
```diff
@@ -38,6 +39,13 @@ def load_dataset(self, data_root_path: Optional[str] = None):
         data_dir = get_local_path_from_root(data_root_path, data_dir)
         if data_dir:
             params_config["data_dir"] = data_dir
+
+        if params_config.get("token", False):
+            from huggingface_hub import login
+
+            token = get_huggingface_token()
```
jambayk, line 46: is this manual login needed? I think just passing the value of `token` (either a boolean or the token string) to the `load_dataset` should work? 
https://huggingface.co/docs/datasets/v2.14.7/en/package_reference/loading_methods#datasets.load_dataset.token

https://github.com/microsoft/Olive/blob/main/olive/data/component/load_dataset.py#L35
https://github.com/microsoft/Olive/blob/main/examples/open_llama/lora_user_script.py#L14

There is no need for us to do the login.

jambayk: I think where you are getting a login requirement might be from the preprocessing component when it loads the tokenizer. 
In that case, we either need to expose `token` argument to the component or do a more centralized login. not here in the load dataset part.

jambayk: I personally prefer the "expose token" solution so that we don't need to login manually. The manual login assumes things about the location of the token but it might be stored in other places based on the transformers related environment variables or if they used git credentials.
https://huggingface.co/docs/huggingface_hub/package_reference/environment_variables#hfhome

jambayk: I was also thinking about these parameters for a while but didn't get the time to work on it. 

Just opened this PR https://github.com/microsoft/Olive/pull/728
Please let me know if it fulfills your requirements for login. For AzureML and Docker systems, we will have to think about it a bit more.

xiaoyu-work: I'm okay with your change for local system, but we still need str token for aml system. I'll double check with security team about this.

olive/systems/azureml/aml_system.py
```diff
@@ -373,6 +385,38 @@ def pass_runner_pipeline():
 
         return pass_runner_pipeline()
 
+    def _get_data_configs(self, pass_config: dict):
```
jambayk, line 388: Instead of hard coding `train_data_config` and `eval_data_config`, we should generalize the support for data configs like we do for path and data params. There are other passes that also have data configs. 

We could do something similar to `path_params` https://github.com/microsoft/Olive/blob/main/olive/passes/olive_pass.py#L92

xiaoyu-work: Besides train data config and eval data config, which kind of data config may be used in the passes?

jambayk: https://github.com/search?q=repo%3Amicrosoft%2FOlive+path%3A%2F%5Eolive%5C%2Fpasses%5C%2F%2F+DataConfig&type=code

xiaoyu-work: Will update this.

xiaoyu-work:
Will raise new PRs. Closed.

PR #728, opened 2023-11-16
`trust_remote_code` for hf data components, Default text-gen corpus dataset_type

jambayk:
## Describe your changes
This PR 
- exposes `trust_remote_code` arguments to huggingface data components that require user permission to run remote code. For input models with hf config model_loading_args, the value for `trust_remote_code` is auto inserted into the data configs without these specified. 
- makes the default value for `dataset_type` in `test_generation_pre_process` component to `corpus`. Previously, the data config set it to `None` if not given and this results in using `pair` strategy. 
- Casts attention mask to input_id's data type to ensure it has the expected data type. 
- Removes `"dataset_type":"corpus"` from the examples since it is the default and the most common type.
- Corrects the dataset config in falcon example. Dataset is also changed to `timdettmers/openassistant-guanaco` since the previous dataset is massive (>1TB) and we only need it for latency measurement. 
 
### DIFF
--- a/olive/model/hf_utils.py
+++ b/olive/model/hf_utils.py
@@ -75,7 +75,14 @@ class HFModelLoadingArgs(ConfigWithExtraArgs):
             " details for the supported parameters."
         ),
     )
-    # TODO(xiaoyuzhang): seems not working. need to check.
+    # for remote files that require authentication
+    token: Union[bool, str] = Field(
+        None,
+        description=(
+            "The token to use as HTTP bearer authorization for remote files. If `True`, will use the token generated "
+            " when running `huggingface-cli login`."
+        ),
+    )
     # whether to trust remote code
     trust_remote_code: bool = Field(
         None,
@@ -132,7 +139,7 @@ class HFModelLoadingArgs(ConfigWithExtraArgs):
     def get_loading_args(self):
         loading_args = {}
         # copy args that can be directly copied
-        direct_copy_args = ["device_map", "max_memory"]
+        direct_copy_args = ["device_map", "max_memory", "token", "trust_remote_code"]
         for arg in direct_copy_args:
             if getattr(self, arg):
                 loading_args[arg] = deepcopy(getattr(self, arg))
@@ -143,8 +150,6 @@ class HFModelLoadingArgs(ConfigWithExtraArgs):
         quantization_config = self.get_quantization_config()
         if quantization_config:
             loading_args["quantization_config"] = quantization_config
-        if self.trust_remote_code:
-            loading_args["trust_remote_code"] = self.trust_remote_code
         # add extra args
         if self.extra_args:
             loading_args.update(deepcopy(self.extra_args))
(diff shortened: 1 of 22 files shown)


olive/model/hf_utils.py
```diff
@@ -75,7 +75,14 @@ class HFModelLoadingArgs(ConfigWithExtraArgs):
             " details for the supported parameters."
         ),
     )
-    # TODO(xiaoyuzhang): seems not working. need to check.
+    # for remote files that require authentication
+    token: Union[bool, str] = Field(
```
guotuofeng, line 79: are there any security concerns if we keep text token?

jambayk: The user can always provide `True` to use the saved token. I provided `str` as an option since this parameter also accepts string. Our examples won't use string but the user can decide to use it if they want.

jambayk: Should I say `True` is recommended in order to not expose the token string in the config?

jambayk: Updated description to recommend `True`.

guotuofeng: I am ok with True. just have some concerns for text token

jambayk: Yes, I added more information in the description. We are not forcing the user to use string value, but they can if they want. 
This is the same as any code that uses hugging face loader.

guotuofeng: for local run, expose the text token is fine. if it is passed into azureml jobs, I am not sure whether it has any security issues

guotuofeng: are there any existing requirements for exposing the text token? if not, I suggest to disable it from the time being, we can consider add it when we really have specific requirements in future.

jambayk: Done! removed the str typing for now.

olive/model/hf_utils.py
```diff
@@ -75,7 +75,15 @@ class HFModelLoadingArgs(ConfigWithExtraArgs):
             " details for the supported parameters."
         ),
     )
-    # TODO(xiaoyuzhang): seems not working. need to check.
+    # for remote files that require authentication
+    # expose `str` type if needed in the future
+    token: bool = Field(
```
trajepl, line 80: double check:
1. token is None: huggingface will not contain any token in the http header for model requests.
2. token is True: huggingface will leverage the token user used in `huggingface-cli login`, if not, raise the error.

One observation: 
for llama2, it is a gated model which need user have the right permission.
1. I cannot load model by `from_pretrained` when I did not login, `token=None` now.
2. I run `huggingface-cli login`, then I can load model by `from_pretrained`, `token=None` now.

Just curious when I login hf, even the token is not True, I still can access.

trajepl: Not block this PR, just provide this information. :)

jambayk:
Converting to draft since `token` arg is not required as long as user has logged in to hf-hub.

olive/model/hf_utils.py
jambayk: As discussed offline, removed `token` from this PR since it's not needed for local systems. For remote systems, we can handle it later.

jambayk:
removed token from this PR since it's not needed for local systems. For remote systems, we can handle it later.

trajepl (approved):

PR #729, opened 2023-11-16
☸️ Whisper fp16 in ort 1.16.2

trajepl:
## Describe your changes
1. Changes follow ort 1.16.2 for whisper fp 16 cases
2. Topologically sort model before chaining and saving.

Olive optimization:
image

Right transcription: 
image

### DIFF
--- a/olive/passes/onnx/insert_beam_search.py
+++ b/olive/passes/onnx/insert_beam_search.py
@@ -11,7 +11,7 @@ from onnxruntime.transformers import onnx_model as ort_onnx_model
 from onnxruntime.transformers.convert_generation import get_shared_initializers
 from packaging import version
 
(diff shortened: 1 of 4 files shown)


olive/passes/onnx/insert_beam_search.py
```diff
@@ -123,6 +153,16 @@ def chain_model(
         )
         graph_outputs = [sequences]
 
+        if options["fp16"] and version_1_16:
```
jambayk, line 156: should this part be optional? Or is it always recommended?

trajepl: Good catch. It should be used always since ort 1.16.0

jambayk (approved):

PR #736, opened 2023-11-20
Add dataconfig support for aml system

xiaoyu-work:
## Describe your changes

Add data config support for aml system. Some passes like LoRA or QLoRA need datasets. The data config should also be uploaded to aml job.

### DIFF
--- a/olive/systems/azureml/aml_system.py
+++ b/olive/systems/azureml/aml_system.py
@@ -119,12 +119,13 @@ class AzureMLSystem(OliveSystem):
         ml_client = self.azureml_client_config.create_client()
         point = point or {}
         config = the_pass.config_at_search_point(point)
+        data_params = self._create_data_script_inputs_and_args(the_pass)
         pass_config = the_pass.to_json(check_object=True)
         pass_config["config"].update(the_pass.serialize_config(config, check_object=True))
 
         with tempfile.TemporaryDirectory() as tempdir:
             pipeline_job = self._create_pipeline_for_pass(
-                data_root, tempdir, model_config, pass_config, the_pass.path_params
+                data_root, tempdir, model_config, pass_config, the_pass.path_params, data_params
             )
 
             # submit job
@@ -304,6 +305,7 @@ class AzureMLSystem(OliveSystem):
         model_config: ModelConfig,
         pass_config: dict,
         pass_path_params: List[Tuple[str, bool, ParamCategory]],
+        data_params: Tuple[Dict, Dict],
     ):
         tmp_dir = Path(tmp_dir)
 
@@ -331,6 +333,7 @@ class AzureMLSystem(OliveSystem):
         inputs = {
             **self._create_model_inputs(model_resource_paths),
             **self._create_pass_inputs(pass_path_params),
+            **data_params[0],
             **accelerator_info,
         }
         # prepare outputs
@@ -361,6 +364,7 @@ class AzureMLSystem(OliveSystem):
         args = {
             **self._create_model_args(model_json, model_resource_paths, tmp_dir),
             **self._create_pass_args(pass_config, pass_path_params, data_root, tmp_dir),
+            **data_params[1],
             **accelerator_info,
         }
 
@@ -372,6 +376,38 @@ class AzureMLSystem(OliveSystem):
             return outputs
 
         return pass_runner_pipeline()
+
+    def _create_data_script_inputs_and_args(self, the_pass: Pass):
+        data_inputs = {}
+        data_args = {}
+        data_name_set = set()
+        for param, param_config in the_pass._config.items():
+            if param.endswith("data_config") and param_config.name not in data_name_set:
+                data_name_set.add(param_config.name)
+                if param_config.user_script:
+                    data_inputs.update(
+                        {f"{param_config.name}_user_script": Input(type=AssetTypes.URI_FILE, optional=True)}
+                    )
+                    data_args.update(
+                        {
+                            f"{param_config.name}_user_script": Input(
+                                type=AssetTypes.URI_FILE, path=param_config.user_script
+                            )
+                        }
+                    )
+                if param_config.script_dir:
+                    data_inputs.update(
+                        {f"{param_config.name}_script_dir": Input(type=AssetTypes.URI_FOLDER, optional=True)}
+                    )
+                    data_args.update(
+                        {
+                            f"{param_config.name}_script_dir": Input(
+                                type=AssetTypes.URI_FOLDER, path=param_config.script_dir
+                            )
+                        }
+                    )
+        logger.debug(f"Data inputs for pass: {data_inputs}, data args for pass: {data_args}")
+        return (data_inputs, data_args)
 
     def _run_job(
         self,
--- a/olive/systems/azureml/aml_system.py
+++ b/olive/systems/azureml/aml_system.py
@@ -8,7 +8,7 @@ import shutil
 import tempfile
 from copy import deepcopy
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, NamedTuple, Optional, Tuple, Union
 
 from azure.ai.ml import Input, Output, command
 from azure.ai.ml.constants import AssetTypes
@@ -48,6 +48,11 @@ RESOURCE_TYPE_TO_ASSET_TYPE = {
     ResourceType.AzureMLDatastore: None,
     ResourceType.AzureMLJobOutput: AssetTypes.CUSTOM_MODEL,
 }
+
+
+class DataParams(NamedTuple):
+    data_inputs: dict
+    data_args: dict
 
 
 def get_asset_type_from_resource_path(resource_path: ResourcePath):
@@ -306,7 +311,7 @@ class AzureMLSystem(OliveSystem):
         model_config: ModelConfig,
         pass_config: dict,
         pass_path_params: List[Tuple[str, bool, ParamCategory]],
-        data_params: Tuple[Dict, Dict],
+        data_params: DataParams,
     ):
         tmp_dir = Path(tmp_dir)
 
@@ -334,7 +339,7 @@ class AzureMLSystem(OliveSystem):
         inputs = {
             **self._create_model_inputs(model_resource_paths),
             **self._create_pass_inputs(pass_path_params),
-            **data_params[0],
+            **data_params.data_inputs,
             **accelerator_info,
         }
         # prepare outputs
@@ -365,7 +370,7 @@ class AzureMLSystem(OliveSystem):
         args = {
             **self._create_model_args(model_json, model_resource_paths, tmp_dir),
             **self._create_pass_args(pass_config, pass_path_params, data_root, tmp_dir),
-            **data_params[1],
+            **data_params.data_args,
             **accelerator_info,
         }
 
@@ -397,7 +402,7 @@ class AzureMLSystem(OliveSystem):
                     if data_config.script_dir:
                         update_dicts(data_config.name, "script_dir", data_config, AssetTypes.URI_FOLDER)
         logger.debug(f"Data inputs for pass: {data_inputs}, data args for pass: {data_args}")
-        return (data_inputs, data_args)
+        return DataParams(data_inputs, data_args)
 
     def _run_job(
         self,
--- a/olive/systems/azureml/aml_system.py
+++ b/olive/systems/azureml/aml_system.py
@@ -20,6 +20,7 @@ from olive.azureml.azureml_client import AzureMLClientConfig
 from olive.cache import normalize_data_path
 from olive.common.config_utils import ParamCategory, validate_config
 from olive.common.utils import retry_func
+from olive.data.config import DataConfig
 from olive.evaluator.metric import Metric, MetricResult
 from olive.hardware.accelerator import AcceleratorSpec
 from olive.model import ModelConfig
@@ -381,31 +382,20 @@ class AzureMLSystem(OliveSystem):
         data_inputs = {}
         data_args = {}
         data_name_set = set()
+
+        def update_dicts(name, key, script_attr, input_type):
+            data_inputs.update({f"{name}_{key}": Input(type=input_type, optional=True)})
+            data_args.update({f"{name}_{key}": Input(type=input_type, path=getattr(script_attr, key))})
+
         for param, param_config in the_pass._config.items():
-            if param.endswith("data_config") and param_config.name not in data_name_set:
-                data_name_set.add(param_config.name)
-                if param_config.user_script:
-                    data_inputs.update(
-                        {f"{param_config.name}_user_script": Input(type=AssetTypes.URI_FILE, optional=True)}
-                    )
-                    data_args.update(
-                        {
-                            f"{param_config.name}_user_script": Input(
-                                type=AssetTypes.URI_FILE, path=param_config.user_script
-                            )
-                        }
-                    )
-                if param_config.script_dir:
-                    data_inputs.update(
-                        {f"{param_config.name}_script_dir": Input(type=AssetTypes.URI_FOLDER, optional=True)}
-                    )
-                    data_args.update(
-                        {
-                            f"{param_config.name}_script_dir": Input(
-                                type=AssetTypes.URI_FOLDER, path=param_config.script_dir
-                            )
-                        }
-                    )
+            if param.endswith("data_config") and param_config is not None:
+                param_config = validate_config(param_config, DataConfig)
+                if param_config.name not in data_name_set:
+                    data_name_set.add(param_config.name)
+                    if param_config.user_script:
+                        update_dicts(param_config.name, "user_script", param_config, AssetTypes.URI_FILE)
+                    if param_config.script_dir:
+                        update_dicts(param_config.name, "script_dir", param_config, AssetTypes.URI_FOLDER)
         logger.debug(f"Data inputs for pass: {data_inputs}, data args for pass: {data_args}")
         return (data_inputs, data_args)
 
--- a/olive/systems/azureml/aml_system.py
+++ b/olive/systems/azureml/aml_system.py
@@ -389,13 +389,13 @@ class AzureMLSystem(OliveSystem):
 
         for param, param_config in the_pass._config.items():
             if param.endswith("data_config") and param_config is not None:
-                param_config = validate_config(param_config, DataConfig)
-                if param_config.name not in data_name_set:
-                    data_name_set.add(param_config.name)
-                    if param_config.user_script:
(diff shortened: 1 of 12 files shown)


olive/systems/azureml/aml_system.py
```diff
@@ -373,6 +378,36 @@ def pass_runner_pipeline():
 
         return pass_runner_pipeline()
 
+    def _create_data_script_inputs_and_args(self, data_configs: Optional[Dict[str, Any]] = None):
```
guotuofeng, line 381: just enable this feature so to change so huge files seems be a little bit hacky.

Are there any examples show the usage? Is there any better way to achieve similar functionality?

trajepl (commented):
Quick question: I assume that the data configs can be maintained by pass/evaluator themself.  Then the data configs should be visible to the system in pass/evaluator configs. Could you help introduce more about the use case of this PR?

olive/systems/azureml/aml_system.py
xiaoyu-work: This could be a quick design doc but I can summarize here:

Use case: some data configs may have `user_script` and `script_dir`, and some passes will use those data configs, like LoRA pass and some quantization passes. When running these passes on AML, only the pass config was submitted to the AML compute, with local `user_script` and `script_dir`, we need to upload all data configs user scripts and script dirs to AML, and in the meantime update the `user_script` and `script_dir` with AML file path in pass config.

So the challenges here are:
1. The data configs are various with different names and different user scripts and script dirs. Each data config's user script and script dir need to be uploaded separately to AML. So we will need to parse each user script and each script dir separately.
2. The data config was nested in the pass config, and only the pass config was submitted to the AML. The `run_pass` method for system has no knowledge about data configs. When creating AML job inputs and args for data config, we can easily get all non-None `user_script` and `script_dir` in data configs from olive workflow and add them to AML job, while when parsing back in aml pass runner, since data configs are nested in the pass config, we can hardly know which parser arguments are needed for this pass.
3. Then, we need to update data config in pass config with new AML file path.

To tackle challenge 1:
- We need data configs section now at system level (unless we collect all nested data configs from pass config, which is not preferrable).
- Since each `user_script` and each `script_dir` have to be uploaded to AML separately, we need a naming convention for each input, which can easily differentiate user script and script dir for each data config, and can be easily parsed back in aml system runner. So I decide to use `data_name_user_script` and `data_name_script_dir` for the input names.

To tackle challenge 2, there are a couple of options:
1. When submitting the job to AML, upload data configs as a file to AML job also. Go through each data config and figure out which data config has user script and script dir, then we can know the parser argument for all user scripts and script dirs.
Pros: Easy check on AML pass runner.
Cons: 1. A new file will be uploaded to aml. 2. DataConfig class now is not serializable. Need to support serialization for DataConfig class, which means we need to support serialization for all classes related to DataConfig class.

2. All checks are done in aml pass runner. Add a new  `ParamCategory.DATA_CONFIG` to each data config pass (do not reply on the param name like ending in `data_config`). Like how we do for [pass path now](https://github.com/microsoft/Olive/blob/main/olive/systems/azureml/aml_pass_runner.py#L43C23-L43C23), check if the pass has this `DATA_CONFIG` param and if this param config has `user_script` or `script_dir`.
Pros: No need change for DataConfig class. 
Cons: Tedious parsing logic in aml pass runner.

Above, I prefer the second approach.

To tackle challenge 3:
With the option 1 for challenge 2, even we can get all data configs param arguments for the pass, we still need to go over each pass parameter to find the correct data config parameter. But if we go with option 2 for challenge 2, we can update the `user_script` and `script_dir` aml path directly when we found it. So I still go for option 2.

xiaoyu-work:
> Quick question: I assume that the data configs can be maintained by pass/evaluator themself. Then the data configs should be visible to the system in pass/evaluator configs. Could you help introduce more about the use case of this PR?

I replied to Mike's comment above. Can you check this?

olive/systems/azureml/aml_system.py
jambayk: I agree with the solutions for challenge 2/3 but not for challenge 1. 

I don't think it is necessary to keep a system or engine level information for `data_configs`. There are multiple issues with it:
1. We cannot guarantee that the workflow config's `data_configs` member has all possible data config objects inside it. A pass config can also directly specify the data_config with referencing to the workflows `data_configs`. 
2. We are unnecessarily uploading all data configs for all passes even though they might not use them at all. 
3. data name is not a good key to use. We cannot guarantee that the user will not provide an incompatible name. This PR shows that we had to update the example to make it work.

I think it is better to do the same thing we do for `path_params`:
1. We already know what the data config params are.
2. There is not much nesting. It is just one level below the pass config so is the same as iterating over the pass config. This is already done in the pass runner and when you iterate over the `data_configs` in this PR. 
3. instead of using data_name as key, we can just use the pass param name. It is guaranteed to be unique for a pass since we would only use the data configs involved with the pass.

guotuofeng: for challenge 1, is it possible to have one consolidated user_script and script_dir for all the data?

guotuofeng:
> > Quick question: I assume that the data configs can be maintained by pass/evaluator themself. Then the data configs should be visible to the system in pass/evaluator configs. Could you help introduce more about the use case of this PR?
> 
> I replied to Mike's comment above. Can you check this?

I think a real user example would be much straightforward. maybe we need update examples?

olive/systems/azureml/aml_system.py
xiaoyu-work: @jambayk I thought about the cases you mentioned, but I decide to make our life easier, like maybe we should only allow the data config in data configs section, and have a naming rule for data configs. The reason I'm using data name as key rather param name is, I'm not sure if there will be any case that different data config params will use the same data config. If so, duplicated files and folders will be uploaded to AML. So regarding to uploading unnecessary data configs, I was also thinking doing a cross validation between data configs and the data configs in pass configs. But I'm open to this if we don't have any use case that same data config is used for different data config param in the pass.

xiaoyu-work: @guotuofeng I don't think that would work. Our current logic treat user_script and script_dir differently. We still need to differentiate them in AML pass runner if so.

olive/passes/olive_pass.py
```diff
@@ -139,14 +139,15 @@ def default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConf
...
+        # are of type DataConfig, Union[DataConfig, dict], ...
         # this requirement is on the pass developer but we can only check it here
         for param, param_config in config.items():
-            if param.endswith("data_config"):
+            if param_config.category == ParamCategory.DATA_CONFIG:
                 param_type = param_config.type_
                 assert param_type == DataConfig or DataConfig in get_args(
                     param_type
-                ), f"{param} ending with data_config must be of type DataConfig."
+                ), f"{param} must be of type DataConfig."
```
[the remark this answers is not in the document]
xiaoyu-work: If we have data configs at system level, we can resolve data config when before the pass run. Do we have any other use case for the nested data config in pass config?

xiaoyu-work:
> > > Quick question: I assume that the data configs can be maintained by pass/evaluator themself. Then the data configs should be visible to the system in pass/evaluator configs. Could you help introduce more about the use case of this PR?
> > 
> > 
> > I replied to Mike's comment above. Can you check this?
> 
> I think a real user example would be much straightforward. maybe we need update examples?

This could an example: for [llama2 qlora example](https://github.com/microsoft/Olive/blob/main/examples/llama2/llama2_qlora.json), instead of using local system, consider using aml system for this example.

guotuofeng:
> > > > Quick question: I assume that the data configs can be maintained by pass/evaluator themself. Then the data configs should be visible to the system in pass/evaluator configs. Could you help introduce more about the use case of this PR?
> > > 
> > > 
> > > I replied to Mike's comment above. Can you check this?
> > 
> > 
> > I think a real user example would be much straightforward. maybe we need update examples?
> 
> This could an example: for [llama2 qlora example](https://github.com/microsoft/Olive/blob/main/examples/llama2/llama2_qlora.json), instead of using local system, consider using aml system for this example.

it seems a single user_script.py is enough?

olive/systems/azureml/aml_system.py
guotuofeng: we could refactor our code to make it support this behavior. I suppose

xiaoyu-work: Can you explain more about "consolidated user_script and script_dir for all the data"?

xiaoyu-work:
> > > > > Quick question: I assume that the data configs can be maintained by pass/evaluator themself. Then the data configs should be visible to the system in pass/evaluator configs. Could you help introduce more about the use case of this PR?
> > > > 
> > > > 
> > > > I replied to Mike's comment above. Can you check this?
> > > 
> > > 
> > > I think a real user example would be much straightforward. maybe we need update examples?
> > 
> > 
> > This could an example: for [llama2 qlora example](https://github.com/microsoft/Olive/blob/main/examples/llama2/llama2_qlora.json), instead of using local system, consider using aml system for this example.
> 
> it seems a single user_script.py is enough?

Other cases may have more than one data configs referred in a same pass, with both user_script and script_dir.

guotuofeng:
> > > > > > Quick question: I assume that the data configs can be maintained by pass/evaluator themself. Then the data configs should be visible to the system in pass/evaluator configs. Could you help introduce more about the use case of this PR?
> > > > > 
> > > > > 
> > > > > I replied to Mike's comment above. Can you check this?
> > > > 
> > > > 
> > > > I think a real user example would be much straightforward. maybe we need update examples?
> > > 
> > > 
> > > This could an example: for [llama2 qlora example](https://github.com/microsoft/Olive/blob/main/examples/llama2/llama2_qlora.json), instead of using local system, consider using aml system for this example.
> > 
> > 
> > it seems a single user_script.py is enough?
> 
> Other cases may have more than one data configs referred in a same pass, with both user_script and script_dir.

it could, but we can enforce that all the script_dir and user_script.py should be put into a single file. This way, the Olive logic would be much simpler.

Current DataConfig has associated user_script and script_dir. we can move it out of DataConfig to upper layer. In this way, the aml job will only need upload the user_script once.

xiaoyu-work:
> > > > > > > Quick question: I assume that the data configs can be maintained by pass/evaluator themself. Then the data configs should be visible to the system in pass/evaluator configs. Could you help introduce more about the use case of this PR?
> > > > > > 
> > > > > > 
> > > > > > I replied to Mike's comment above. Can you check this?
> > > > > 
> > > > > 
> > > > > I think a real user example would be much straightforward. maybe we need update examples?
> > > > 
> > > > 
> > > > This could an example: for [llama2 qlora example](https://github.com/microsoft/Olive/blob/main/examples/llama2/llama2_qlora.json), instead of using local system, consider using aml system for this example.
> > > 
> > > 
> > > it seems a single user_script.py is enough?
> > 
> > 
> > Other cases may have more than one data configs referred in a same pass, with both user_script and script_dir.
> 
> it could, but we can enforce that all the script_dir and user_script.py should be put into a single file. This way, the Olive logic would be much simpler.
> 
> Current DataConfig has associated user_script and script_dir. we can move it out of DataConfig to upper layer. In this way, the aml job will only need upload the user_script once.
> 
> something like
> 
> image

My concern for this is user needs to have one user script for all modules codes such as models, evaluators, data config, etc. This will be hard for user to maintain. One possible solution is suggesting user to have one entrypoint script for all modules func but keep detailed code in script_dir. We can discuss more about which approach is better.

guotuofeng:
> > > > > > > > Quick question: I assume that the data configs can be maintained by pass/evaluator themself. Then the data configs should be visible to the system in pass/evaluator configs. Could you help introduce more about the use case of this PR?
> > > > > > > 
> > > > > > > 
> > > > > > > I replied to Mike's comment above. Can you check this?
> > > > > > 
> > > > > > 
> > > > > > I think a real user example would be much straightforward. maybe we need update examples?
> > > > > 
> > > > > 
> > > > > This could an example: for [llama2 qlora example](https://github.com/microsoft/Olive/blob/main/examples/llama2/llama2_qlora.json), instead of using local system, consider using aml system for this example.
> > > > 
> > > > 
> > > > it seems a single user_script.py is enough?
> > > 
> > > 
> > > Other cases may have more than one data configs referred in a same pass, with both user_script and script_dir.
> > 
> > 
> > it could, but we can enforce that all the script_dir and user_script.py should be put into a single file. This way, the Olive logic would be much simpler.
> > Current DataConfig has associated user_script and script_dir. we can move it out of DataConfig to upper layer. In this way, the aml job will only need upload the user_script once.
> > something like
> > image
> 
> My concern for this is user needs to have one user script for all modules codes such as models, evaluators, data config, etc. This will be hard for user to maintain. One possible solution is suggesting user to have one entrypoint script for all modules func but keep detailed code in script_dir. We can discuss more about which approach is better.

From the time being, all of the examples in Olive use a single user_script.py. That's the reason I propose the unify of user script. Yes, we can discuss and find out the ideal solution for this scenario.

olive/systems/azureml/aml_system.py
xiaoyu-work: @jambayk another option could be, iterating over the pass config to find the data configs that are used for this pass, but still keep data config name as param name to avoid duplicating. With this, we do not need data config arg for system. This would be better.

devang-ml: "We cannot guarantee that the workflow config's data_configs member has all possible data config objects inside it. "
It'd be fine to put this restriction and have pass specific data_config only refer the data config from the data_configs.

guotuofeng: > Can you explain more about "consolidated user_script and script_dir for all the data"?

we can add an assumption that all user written code need to be put into a single user_script for Olive. If there are multiple user_scripts file, the user need merge them into a single one to use Olive. 

This assumption should not a big deal for Olive users.

jambayk: @xiaoyu-work I am fine with your initial suggestion and devang's about data configs only being defined in the `data_configs` section. 

But that also seems like an extra restriction since we would need to iterate through the pass config anyway as you mentioned in 
> another option could be, iterating over the pass config to find the data configs that are used for this pass, but still keep data config name as param name to avoid duplicating. With this, we do not need data config arg for system. This would be better.
I agree this sounds like a better compromise since your original concern was about duplicated upload. We will however need to add some validations for the names of the data config so that can be used as command line arguments and also to make sure there are no duplicate names.

jambayk: @guotuofeng I think the work to have a single user script for a single workflow has greater scope than this PR and might be better dealt separately, preferably before this PR? This also came up in a previous discussion with @trajepl 

We have user scripts in many different parts of olive: models, metrics, pass configs, data configs, etc. We would need to come up with a way to provide the user script or module to the different components if they don't keep track of their own user scripts. We would also need to do this while also keeping the modular nature of the components. What do you think about this @devang-ml?

olive/passes/olive_pass.py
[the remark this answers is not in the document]
jambayk: The data configs are pass config parameters. So, we need to know what parameters data configs are so that we can resolve string names in the run config validation as well as perform some of the model name and task auto insertion.

olive/systems/azureml/aml_system.py
guotuofeng: Yes, you are correct. the single user script is another work item unrelated to the PR. I am ok with current approach to upload multiple user script way. 

However, we need not introduce the change at system level.

olive/extra_dependencies.json
```diff
@@ -33,13 +33,10 @@
...
     ],
     "lora": [
-        "accelerate",
-        "peft"
-    ],
-    "qlora": [
         "accelerate",
         "bitsandbytes",
-        "peft"
+        "peft",
```
jambayk, line 38: I have some upcoming changes to the lora passes which will make these requirements cleaner. If you are okay with it, I can do the change here and in `run.py` in my PR so that this PR only deals with the data configs part 
https://github.com/microsoft/Olive/compare/main...jambayk/ort-bf16

guotuofeng: yes, better to split the unrelated changes to another PR.

xiaoyu-work: Sure, please go ahead for your changes and PR.

olive/passes/olive_pass.py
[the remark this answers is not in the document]
xiaoyu-work: I added a new `ParamCategory.DATA_CONFIG` so we don't need to identify data config param by its name. If it is a dict, this step might be a little hacky.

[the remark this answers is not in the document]
jambayk: the paramcategory is only helpful inside the aml or docker systems since the pass is already instantiated with `accerator_spec`. For the workflow config json validation, we don't have access to the instantiated pass config so we don't know which params are data configs without some way. That's why after discussion with @guotuofeng, we decided to reserve `data_config` suffix for such params. https://github.com/microsoft/Olive/blob/main/olive/workflows/run/config.py#L204

Dicts are fine since the dict gets converted into a DataConfig object after validation by pydantic anyways. We can also remove `dict` from the `type_`. It was added to make it clear we can provide a dict as value instead but internally, since DataConfig is first in the Union, all dicts converted to DataConfig.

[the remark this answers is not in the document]
xiaoyu-work: I see. I will revert this change for now, and will investigate more on this.

olive/systems/azureml/aml_system.py
```diff
@@ -373,6 +378,27 @@ def pass_runner_pipeline():
...
+            if param.endswith("data_config") and param_config is not None:
+                data_config = validate_config(param_config, DataConfig)
+                if data_config.name not in data_name_set:
+                    data_name_set.add(data_config.name)
+                    if data_config.user_script:
+                        update_dicts(data_config.name, "user_script", data_config, AssetTypes.URI_FILE)
+                    if data_config.script_dir:
+                        update_dicts(data_config.name, "script_dir", data_config, AssetTypes.URI_FOLDER)
+        logger.debug(f"Data inputs for pass: {data_inputs}, data args for pass: {data_args}")
+        return (data_inputs, data_args)
```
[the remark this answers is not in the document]
xiaoyu-work: `_create_pipeline_for_pass` has all other args passed in one argument. So I unified data args to one tuple here.

olive/systems/azureml/aml_system.py
```diff
@@ -304,6 +306,7 @@ def _create_pipeline_for_pass(
         model_config: ModelConfig,
         pass_config: dict,
         pass_path_params: List[Tuple[str, bool, ParamCategory]],
+        data_params: Tuple[Dict, Dict],
```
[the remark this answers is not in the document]
xiaoyu-work: Sounds good to me. Will update this.

trajepl (approved):

PR #739, opened 2023-11-21
🌵 Complete optimum conversion/merge configs

trajepl:
## Describe your changes

Complete optimum conversion/merge configs.
Take conversion as example, the `fp16=True` means to load the torch model with float16 from huggingface model hub. Then the fp16 torch model is converted to onnx model. 
That is different the fp16 in transformer optimization.

### DIFF
--- a/olive/passes/onnx/optimum_conversion.py
+++ b/olive/passes/onnx/optimum_conversion.py
@@ -56,8 +56,8 @@ class OptimumConversion(Pass):
         if with_fixed_value:
             search_point = self.config_at_search_point(search_point or {})
 
-        if search_point.get("device") == "cuda" and not search_point.get("fp16"):
-            logger.info("OptimumConversion: fp16 is not set to True, but device is set to cuda.")
+        if search_point.get("fp16") and search_point.get("device") != "cuda":
+            logger.info("OptimumConversion: fp16 is set to True, but device is not set to cuda.")
             return False
 
         return True
--- a/olive/passes/onnx/optimum_conversion.py
+++ b/olive/passes/onnx/optimum_conversion.py
@@ -21,10 +21,29 @@ class OptimumConversion(Pass):
 
     @staticmethod
     def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        device = "cuda" if accelerator_spec.accelerator_type == "gpu" else "cpu"
         config = {
             "target_opset": PassConfigParam(
                 type_=int, default_value=14, description="The version of the default (ai.onnx) opset to target."
-            )
+            ),
+            "fp16": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                description="Whether to use fp16 precision to load torch model and then convert it to onnx.",
+            ),
+            "device": PassConfigParam(
+                type_=str, default_value=device, description="The device to use to do the export. Defaults to 'cpu'."
+            ),
+            "no_post_process": PassConfigParam(
+                type_=bool,
+                default_value=True,
+                description="Whether to skip post-processing the exported model.",
+            ),
+            "extra_args": PassConfigParam(
+                type_=dict,
+                default_value=None,
+                description="Extra arguments to pass to the `optimum.exporters.onnx.main_export` function.",
+            ),
         }
         config.update(get_external_data_config())
         return config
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/optimum_conversion.py
```diff
@@ -21,10 +21,29 @@ class OptimumConversion(Pass):
...
             "target_opset": PassConfigParam(
                 type_=int, default_value=14, description="The version of the default (ai.onnx) opset to target."
-            )
+            ),
+            "fp16": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                description="Whether to use fp16 precision to load torch model and then convert it to onnx.",
+            ),
+            "device": PassConfigParam(
```
jambayk, line 34: Should we set the default device to be `None` (or cpu) like with the conversion pass and make the user specify their own device if needed? Just like in conversion, we might want have gpu target but do conversion on CPU due to memory limits, etc.

jambayk: Conversion on cpu still seems to be recommended. GPU is only required for fp16

olive/passes/onnx/optimum_conversion.py
```diff
@@ -21,10 +21,29 @@ class OptimumConversion(Pass):
 
     @staticmethod
     def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        device = "cuda" if accelerator_spec.accelerator_type == "gpu" else "cpu"
         config = {
             "target_opset": PassConfigParam(
                 type_=int, default_value=14, description="The version of the default (ai.onnx) opset to target."
-            )
+            ),
+            "fp16": PassConfigParam(
```
jambayk, line 29: We could add a search point validation so that `fp16` can only be true if `"cuda" in device`.

olive/passes/onnx/optimum_conversion.py
```diff
@@ -41,22 +60,25 @@ def _run_for_config(
...
-            export_optimum_model(
-                model.model_path or hf_config.model_name,
-                output_model_path,
-                opset=config["target_opset"],
-                legacy=True,
-                no_post_process=True,
-            )
+        if version.parse(optimum_version.__version__) >= version.parse("1.14.0"):
+            # Optimum 1.14.0 needs to be run in legacy mode to support older versions of transformers
+            config["extra_args"]["legacy"] = True
```
jambayk, line 75: do you have plans to deprecate legacy support or make it default to `False`. Since optimum is planning to deprecate legacy mode, I think we should have legacy as opt-in instead of the default.

olive/passes/onnx/optimum_conversion.py
trajepl: Done.

olive/passes/onnx/optimum_conversion.py
trajepl: Yes, we should deprecate the legacy way by default and provide the options for users to enable for some corner cases in the future. 
I remembered that when I test the way with legacy=False for optimum 1.14.0, the open llama case failed for some reasons. 

Added the TODO item, will try to investigate the failure and remove legacy once the validations pass.

olive/passes/onnx/optimum_conversion.py
```diff
@@ -24,11 +27,41 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
...
         config.update(get_external_data_config())
         return config
 
+    def validate_search_point(
+        self, search_point: Dict[str, Any], accelerator_spec: AcceleratorSpec, with_fixed_value: bool = False
+    ) -> bool:
+        if with_fixed_value:
+            search_point = self.config_at_search_point(search_point or {})
+
+        if search_point.get("device") == "cuda" and not search_point.get("fp16"):
```
jambayk, line 59: I was thinking of the other direction, `if search_point.get("fp16") and "cuda" not in search_point.get("device")`. (or device=="cpu")
Similar to https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/conversion.py#L92.

Cuda is required for fp16 but not the other way around. We can still do fp32 on cuda.

olive/passes/onnx/optimum_conversion.py
jambayk: Thanks! this sounds good to me.

olive/passes/onnx/optimum_conversion.py
trajepl: Oh, yes. Updated now! Thanks!

jambayk: `search_point.get("device") == "cpu" and search_point.get("fp16")` is cleaner. optimum does the same https://github.com/huggingface/optimum/blob/main/optimum/exporters/onnx/__main__.py#L290

jambayk: Oh just saw your update. I think your check is fine too. Unless we want to support something like `"cuda:0"`

jambayk (approved):

olive/passes/onnx/optimum_conversion.py
trajepl: oh, cool. BTW, I previously try to add device id into Olive, but close that PR. There is a simple way to achieve that:
`CUDA_VISIBLE_DEVICES=1,2 python -m olive.workflows.run --config xxx`

test/unit_test/passes/onnx/test_optimum_conversion.py
```diff
@@ -12,14 +12,45 @@
...
+)
+def test_optimum_configs(config, is_valid, tmp_path):
+    input_model = get_optimum_model_by_hf_config()
+    # setup
+    p = create_pass_from_dict(OptimumConversion, config, disable_search=True)
+    output_folder = tmp_path
+
+    if not is_valid:
+        assert p.validate_search_point(config, None) is False
+        with pytest.raises(
```
jambayk, line 48: nit: is this part needed? 

I think matching a specific error message makes the test breakable if optimum changes the error message. Or is this the purpose? So that we can catch any incompatibility with new optimum changes?

trajepl: I also try to remove the match part in my first version of implementation. But as you said, once they change the code and have different supports on fp16, then we can catch that.

jambayk (approved):

PR #742, opened 2023-11-22
make Olive setup could be run without ORT installation

guotuofeng:
## Describe your changes
* refactor existing setup dependency logic to make sure the dml support be more easy to understand
* decouple olive setup with ORT.

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -9,7 +9,7 @@ from collections import OrderedDict, defaultdict
 from contextlib import contextmanager
 from datetime import datetime
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple, Type, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, Union
 
 import olive.cache as cache_utils
 from olive.common.config_utils import ConfigBase, validate_config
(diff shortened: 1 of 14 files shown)


olive/workflows/run/config.py
```diff
@@ -16,7 +16,7 @@
 from olive.engine.packaging.packaging_config import PackagingConfig
 from olive.evaluator.olive_evaluator import OliveEvaluatorConfig
 from olive.model import ModelConfig
-from olive.passes import FullPassConfig, Pass
+from olive.passes import FullPassConfig
```
jambayk, line 19: I don't think there is a need to make the import of `Pass` lazy? You already made `Pass` importable by making ort imports lazy. 

Even if `Pass` is not imported at top level, the validator `validate_pass_search` gets called anyways even in setup mode in `run.py:lines 210-213`

jambayk: Also any imports of/from `olive.passes` will automatically trigger the `__init__` imports in all sub modules anyways

guotuofeng: yes, you are right. will revert the change

olive/engine/engine.py
```diff
@@ -23,13 +23,15 @@
...
 from olive.hardware import AcceleratorLookup, AcceleratorSpec, Device
 from olive.model import ModelConfig
-from olive.passes.olive_pass import Pass
 from olive.strategy.search_strategy import SearchStrategy
 from olive.systems.common import SystemType
 from olive.systems.local import LocalSystem
 from olive.systems.olive_system import OliveSystem
 from olive.systems.utils import create_new_system_with_cache
 
+if TYPE_CHECKING:
```
jambayk, line 32: nit: are these changes also needed?

jambayk: Or this method is recommended since we are only using `Pass` for type hints?

guotuofeng: I add the TYPE_CHECKING here to make sure the explicit Pass import is optional for engine running.

jambayk: Sounds good to me. Thanks!

guotuofeng: as for best practice, any hint import should be put into TYPE_CHECKING.

jambayk (approved):

jambayk:
Merging this since I have some changes that have merge conflicts with this.

PR #743, opened 2023-11-22
🧇 Convert onnx model to target opset version

trajepl:
## Describe your changes

Convert onnx model to target opset version

### DIFF
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -476,6 +476,7 @@ class OnnxOpVersionConversion(Pass):
             # we have no good solution to skip the one pass once it is executed, just save the model again to
             # target output path
             logger.info(f"Model is already in target opset version {config['target_opset']}.")
+            return model
         else:
             model_proto = onnx.version_converter.convert_version(model_proto, config["target_opset"])
         return model_proto_to_olive_model(model_proto, output_model_path, config)
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -17,7 +17,14 @@ from packaging import version
 from olive.common.config_utils import validate_config
 from olive.common.utils import find_submodules, resolve_torch_dtype, tensor_data_to_device
 from olive.hardware import AcceleratorSpec
-from olive.model import CompositeOnnxModel, DistributedOnnxModel, DistributedPyTorchModel, ONNXModel, PyTorchModel
+from olive.model import (
+    CompositeOnnxModel,
+    DistributedOnnxModel,
+    DistributedPyTorchModel,
+    ONNXModel,
+    ONNXModelBase,
+    PyTorchModel,
+)
 from olive.model.hf_utils import HFModelLoadingArgs, get_hf_model_io_config
 from olive.model.model_config import IOConfig
 from olive.passes import Pass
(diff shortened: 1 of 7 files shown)


olive/passes/onnx/conversion.py
```diff
@@ -449,3 +456,34 @@ def _convert_distributed_model_on_device(
...
+                type_=int,
+                default_value=latest_opset_version,
+                description="The version of the default (ai.onnx) opset to target. Default: latest opset version.",
+            ),
+        }
+        config.update(get_external_data_config())
+        return config
+
+    def _run_for_config(
+        self, model: ONNXModelBase, data_root: str, config: Dict[str, Any], output_model_path: str
```
jambayk, line 477: Looks like the both the input and output model types are just `ONNXModel` like most of the other onnx passes? 
The composite and distributed cases are handled by https://github.com/microsoft/Olive/blob/main/olive/passes/olive_pass.py#L346

This pass only works with `ONNXModel` type.

test/unit_test/passes/onnx/test_conversion.py
```diff
@@ -24,3 +24,16 @@ def test_onnx_conversion_pass(input_model, tmp_path):
...
+def test_onnx_op_version_conversion_pass(target_opset, tmp_path):
+    input_model = get_onnx_model()
+    # setup
+    p = create_pass_from_dict(OnnxOpVersionConversion, {"target_opset": target_opset}, disable_search=True)
+    output_folder = str(tmp_path / "onnx")
+
+    onnx_model = p.run(input_model, None, output_folder)
+
+    # assert
+    assert Path(onnx_model.model_path).exists()
```
jambayk, line 39: could also load the model and verify the opset version?

trajepl: It should be. Updated.

olive/passes/onnx/conversion.py
trajepl: Removed the check.

jambayk (approved):

olive/passes/onnx/conversion.py
```diff
@@ -449,3 +449,33 @@ def _convert_distributed_model_on_device(
...
+        return config
+
+    def _run_for_config(
+        self, model: ONNXModel, data_root: str, config: Dict[str, Any], output_model_path: str
+    ) -> ONNXModel:
+        # get current models's opset version
+        model_proto = model.load_model()
+        model_opset_version = model_proto.opset_import[0].version
+        if model_opset_version == config["target_opset"]:
+            # we have no good solution to skip the one pass once it is executed, just save the model again to
```
jambayk, line 476: nit: for this part, we could also just return the input model directly if we don't want to save the model again. The perf tuning pass also does something similar by returning the same .onnx file.

trajepl: Oh, it is great. I forget the case of perf-tuning.  Updated now.

olive/passes/onnx/conversion.py
```diff
@@ -449,3 +449,34 @@ def _convert_distributed_model_on_device(
...
+            # target output path
+            logger.info(f"Model is already in target opset version {config['target_opset']}.")
+            return model
+        else:
```
jambayk, line 480: nit: more of a coding style/preference so no action needed. The code is clearer if 1) line 482 is under the else clause or 2) there is no else clause at all since the if clause returns

I don't think we have a specific preference for 1 or 2 in our code base. There is a mix of both. Personally, I use 2 for cases where the if clause returns.

jambayk (approved):

olive/passes/onnx/conversion.py
trajepl: Updated! Thanks!

jambayk (approved):

devang-ml (approved):

PR #747, opened 2023-11-22
Fix pytorch tensor parallel implementation

shaahji:
## Describe your changes

Fix pytorch tensor parallel implementation

* Previous verification was a false positive. The process was loading the entire model during the OnnxConversion pass resulting in exporting the same model i.e. the entire model for reach rank. The issue was found when attempting to export the Llama 70b model which kept running out of memory and the number of parameters reflected the error.

* Swap out of use of MPI in favor of multiprocessing since the latter is much easier to work with and debug. Important finding is that TensorParallel & OnnxConversion pass doesn't require MPI to succeed.

* Refactored the TensorParallel pass to work with multiprocessing jobs to speed up the process.

* Added OrtTransformersOptimization to Llama2 tensor parallel config
* Fixed Llama2 user script to account for world size to work with tensor parallel pass

### DIFF
--- a/examples/llama2/llama2_tensor_parallel.json
+++ b/examples/llama2/llama2_tensor_parallel.json
@@ -22,9 +22,11 @@
     },
     "passes": {
         "tensor_parallel": {
-            "type": "LlamaPyTorchTensorParallel",
+            "type": "PyTorchTensorParallel",
             "config": {
-                "world_size": 2
+                "user_script": "llama2_tensor_parallel.py",
+                "class_name": "LlamaPyTorchTensorParallel",
+                "world_size": 4
             }
         },
         "conversion": {
@@ -33,6 +35,19 @@
                 "target_opset": 17,
                 "save_as_external_data": true,
                 "all_tensors_to_one_file": true
+            }
+        },
+        "transformers_optimization_fp16": {
+            "type": "OrtTransformersOptimization",
+            "config": {
+                "save_as_external_data": true,
+                "all_tensors_to_one_file": true,
+                "model_type": "gpt2",
+                "opt_level": 0,
+                "only_onnxruntime": false,
+                "keep_io_types": false,
+                "float16": true,
+                "use_gqa": true
             }
         }
     },
--- a/examples/llama2/user_script.py
+++ b/examples/llama2/user_script.py
@@ -56,15 +56,12 @@ def get_merged_sample_with_past_kv_inputs(
     # By using a single function with no default values, we can avoid confusion and are deliberate about the sizes
     # can instead write dummy input functions like 'get_merged_decoder_with_past_dummy_inputs' if needed
 
-    if model_id:
-        config = LlamaConfig.from_pretrained(model_id)
-    else:
-        config = LlamaConfig.from_pretrained(model.hf_config.model_name)
-
+    config = model.get_hf_model_config() if model else LlamaConfig.from_pretrained(model_id)
+    world_size = config.world_size if hasattr(config, "world_size") else 1
     input_ids = torch.randint(low=0, high=config.vocab_size, size=(batch_size, seq_len), dtype=torch.int64)
     attention_mask = torch.ones(batch_size, past_seq_len + seq_len, dtype=torch.int64)
     position_ids = get_position_ids(attention_mask, past_seq_len=past_seq_len)
-    past_kv = get_past_kv_inputs(config, batch_size, past_seq_len, use_fp16=use_fp16)
+    past_kv = get_past_kv_inputs(config, batch_size, past_seq_len, use_fp16=use_fp16, world_size=world_size)
 
     return (input_ids, attention_mask, position_ids, past_kv)
 
@@ -80,12 +77,13 @@ def get_position_ids(attention_mask: torch.Tensor, past_seq_len: int):
     return position_ids[:, past_seq_len:]
 
 
-def get_past_kv_inputs(config: LlamaConfig, batch_size: int, past_seq_len: int, use_fp16: bool):
+def get_past_kv_inputs(config: LlamaConfig, batch_size: int, past_seq_len: int, use_fp16: bool, world_size: int = 1):
     """Get past_key_values for all layers.
 
     Shape of past_key_values is (batch_size, num_heads, past_seq_len, head_size).
     """
-    num_heads, head_size = config.num_key_value_heads, config.hidden_size // config.num_key_value_heads
+    num_heads = config.num_key_value_heads // world_size
+    head_size = config.hidden_size // config.num_attention_heads
     torch_dtype = torch.float16 if use_fp16 else torch.float32
     return [
         (
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -4,7 +4,7 @@
 # --------------------------------------------------------------------------
 import importlib
 import logging
-import os
+import multiprocessing
 import tempfile
 from copy import deepcopy
 from pathlib import Path
@@ -66,6 +66,12 @@ class OnnxConversion(Pass):
                     " will use the model as is."
                 ),
             ),
+            "parallel_jobs": PassConfigParam(
+                type_=int,
+                default=multiprocessing.cpu_count(),
+                required=False,
+                description="Number of parallel jobs. Defaulted to number of CPUs. Set it to 0 to disable.",
+            ),
         }
         config.update(get_external_data_config())
         return config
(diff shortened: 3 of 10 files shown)


examples/llama2/user_script.py
```diff
@@ -56,15 +56,18 @@ def get_merged_sample_with_past_kv_inputs(
     # By using a single function with no default values, we can avoid confusion and are deliberate about the sizes
     # can instead write dummy input functions like 'get_merged_decoder_with_past_dummy_inputs' if needed
 
-    if model_id:
+    if model:
+        config = model.load_model().config
+    elif model_id:
         config = LlamaConfig.from_pretrained(model_id)
     else:
         config = LlamaConfig.from_pretrained(model.hf_config.model_name)
```
jambayk, line 64: I think we can cover the required changes using `config = model.get_hf_model_config()` at line 64. This uses the `model_path` to load the llamaconfig without having to load the model entirely. 

Otherwise, the else condition never gets used since it's supposed to be the same condition as the new if.

shaahji: Done!

examples/llama2/llama2_tensor_parallel.json
```diff
@@ -34,6 +36,20 @@
                 "save_as_external_data": true,
                 "all_tensors_to_one_file": true
             }
+        },
+        "transformers_optimization_fp16": {
+            "type": "OrtTransformersOptimization",
+            "disable_search": true,
```
jambayk, line 42: nit: since `"search_strategy": false`, `disable_search` is redundant.

olive/passes/pytorch/tensor_parallel_llama2.py
```diff
@@ -201,69 +220,79 @@ def tp_llama_attention_forward(
...
+def restore_llama2_tensor_parallel_layers(originals: Dict[str, Any]):
+    from transformers.models import llama
+
+    llama.modeling_llama.LlamaMLP.__init__ = originals["mlp_init"]
+    llama.modeling_llama.LlamaAttention.__init__ = originals["attention_init"]
+    llama.modeling_llama.LlamaAttention.forward = originals["attention_forward"]
+    llama.modeling_llama.LlamaAttention.parallel_split = originals["parallel_split"]
+    llama.modeling_llama.apply_rotary_pos_emb = originals["apply_rotary_pos_emb"]
+
+    torch.nn.init.uniform_ = originals["uniform_"]
```
devang-ml, line 264: ?

olive/passes/onnx/conversion.py
```diff
@@ -420,34 +421,41 @@ def _convert_distributed_model_on_device(
         torch_dtype: Optional[torch.dtype] = None,
     ) -> DistributedOnnxModel:
         """Convert a DistributedPyTorchModel to a DistributedOnnxModel."""
-        from mpi4py.futures import MPIPoolExecutor
-
         pass_config = config
         model_config = model.to_json()["config"]
         world_size = model.num_ranks
         output_model_path = str(Path(output_model_path).with_suffix(""))
+        use_gpu = self.accelerator_spec.accelerator_type == Device.GPU
```
jambayk, line 428: should use_gpu just check `device`? We don't want the conversion pass to depend on the target accelerator type but pass `device` if needed.
device is cpu by default for regular conversion and cuda for distributed.

olive/passes/onnx/conversion.py
```diff
@@ -420,34 +421,41 @@ def _convert_distributed_model_on_device(
...
-            executor.map(OnnxConversion._export_ranked_model, params)
-            executor.shutdown()
+        max_parallel_jobs = min(world_size, config["parallel_jobs"] or multiprocessing.cpu_count())
+        if max_parallel_jobs <= 1:
+            results = [OnnxConversion._export_ranked_model(_) for _ in params]
+        else:
+            context = multiprocessing.get_context("spawn")
+            with context.Pool(processes=max_parallel_jobs) as pool:
+                results = pool.map(OnnxConversion._export_ranked_model, params)
+
```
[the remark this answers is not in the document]
shaahji: In case of multiprocessing jobs, the GPU is not dedicated to any specific process. The spawned processes could in theory be on separate CPU.

In the worst-case scenario, parallel jobs could be set to zero to run synchronously.

olive/passes/pytorch/tensor_parallel_llama2.py
shaahji: Did you mean to provide a comment?

olive/passes/onnx/conversion.py

olive/passes/onnx/conversion.py
```diff
@@ -357,58 +368,48 @@ def _export_ranked_model(params):
...
+            restore_llama2_tensor_parallel_layers,
         )
 
         output_filename = DistributedOnnxModel.DEFAULT_RANKED_MODEL_NAME_FORMAT.format(local_rank)
-        output_filepath = str(Path(output_dirpath) / output_filename)
+        output_filepath = ONNXModel.resolve_path(output_dirpath, output_filename)
 
-        input_model = DistributedPyTorchModel(**model_config)
-        olive_pytorch_model = input_model.load_model(local_rank)
+        restore_args = replace_llama2_tensor_parallel_layers()
```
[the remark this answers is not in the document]

olive/passes/onnx/conversion.py
```diff
@@ -137,7 +144,11 @@ def _export_pytorch_model(
         :param config: the config for the pass
         :param device: the device to use for conversion
         :param torch_dtype: the dtype to cast the model to before conversion
+        :param tempdir: directory to use for temporary files
         """
+        if torch.cuda.is_available():
+            torch.cuda.empty_cache()
```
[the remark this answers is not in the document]
shaahji: Done!

examples/llama2/llama2_tensor_parallel.json
shaahji: Done!

olive/passes/onnx/conversion.py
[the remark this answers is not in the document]
jambayk: just to be clear, is the multiprocessing option available:
-  only to ensure we don't overwhelm the cpu by running all jobs in parallel on a slower machine?
- Also to allow sharding a large model using fewer GPUs available than needed to do them in parallel?;
  - If this is the case, should the device id of the sharded model be something like process_idx % max_parallel_jobs? 
  
From the code, looks like it is only for the first scenario. If so, should we check beforehand that `torch.cuda.device_count() <= world_size`?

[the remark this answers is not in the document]
jambayk: FYI for the second scenario https://stackoverflow.com/questions/10190981/get-a-unique-id-for-worker-in-python-multiprocessing-pool

olive/passes/pytorch/tensor_parallel_layers.py
```diff
@@ -35,15 +35,15 @@ def __init__(
...
         factory_kwargs = {"device": device, "dtype": dtype}
         super().__init__()
         self.use_bias = bias
+        self.world_size = world_size
         self.in_features = in_features
-        self.out_features = out_features
-        self.rank_weights = []
-        self.rank_biases = []
+        self.out_features = out_features // self.world_size
         # We change from traditional `nn.Linear` and remove unecessary `torch.Tensor.transpose` operation
```
guotuofeng, line 47: typo

olive/passes/pytorch/tensor_parallel_layers.py
```diff
@@ -99,14 +89,15 @@ def __init__(
...
     ):
         factory_kwargs = {"device": device, "dtype": dtype}
         super().__init__()
         self.use_bias = bias
-        self.in_features = in_features
+        self.world_size = world_size
+        self.in_features = in_features // self.world_size
         self.out_features = out_features
-        self.rank_weights = []
         # We change from traditional `nn.Linear` and remove unecessary `torch.Tensor.transpose` operation
```
guotuofeng, line 101: typo

olive/passes/pytorch/tensor_parallel.py
```diff
@@ -50,51 +102,90 @@ def _validators() -> Dict[str, Callable]:
...
+
+            # 5. Save it out for each rank
+            pytorch_model.config.world_size = world_size
+            pytorch_model.save_pretrained(output_filepath)
+        finally:
+            # 6. Restore layers that were replaced
+            impl.restore_layers()
+
+        logger.debug(f"Successfully exported tensor parallel model for rank: {rank}, {output_filepath}")
+        return 1
```
guotuofeng, line 144: what's the 1 meaning?

olive/passes/onnx/conversion.py
```diff
@@ -357,58 +370,48 @@ def _export_ranked_model(params):
...
+                    device,
+                    torch_dtype,
+                    tempdir,
+                )
+        finally:
+            restore_llama2_tensor_parallel_layers(restore_args)
 
         # save the model to the output path and return the model
         model_proto_to_olive_model(ranked_onnx_model, output_filepath, pass_config)
+        return 1
```
guotuofeng, line 414: do we need the return value?

olive/passes/pytorch/tensor_parallel.py
guotuofeng: so, it is used as return code or boolean values? do we need add some document or comments for the return value?

shaahji: Added a comment.

PR #750, opened 2023-11-23
Complete `io_config`, Pass model to model funcs

jambayk:
## Describe your changes
`io_config` attribute can take both io_config dict and function names. These cases are not cleanly handled yet.
There is a special case where it is a string and is called with `hf_config.model_name`. But this is arbitrary and was initially implemented to make the whisper example work. 

In this PR:
- Handle all possible values for `io_config` attribute. 
- `PyTorchModel.get_io_config` method is added to get the io config from `io_config` and `hf_config` in decreasing order of priority. 
- `io_config`, `component_func` and `dummy_input_funcs` are called with `self` by the caller model. This makes them all standardized and not be called with some arbitrary arguments. 
- `HFConfig` updated to handle io_config and dummy_input creation from onnx config itself.

`component_func` and `io_config` functions in `PyTorchModel` are now called with the calling `OliveModel` itself.

### DIFF
--- a/olive/model/__init__.py
+++ b/olive/model/__init__.py
@@ -478,8 +478,7 @@ class PyTorchModel(OliveModel):
         model_loader: Union[str, Callable] = None,
         model_script: Union[str, Path] = None,
         script_dir: Union[str, Path] = None,
-        io_config: Union[Dict[str, Any], IOConfig] = None,
-        io_config_func: Union[str, Callable] = None,
+        io_config: Union[Dict[str, Any], IOConfig, str, Callable] = None,
         dummy_inputs_func: Union[str, Callable] = None,
         hf_config: Union[Dict[str, Any], HFConfig] = None,
         adapter_path: OLIVE_RESOURCE_ANNOTATIONS = None,
@@ -528,10 +527,9 @@ class PyTorchModel(OliveModel):
             ), "model_script must be a local file or a string name."
 
         # io config for conversion to onnx
-        self.io_config = validate_config(io_config, IOConfig).dict() if io_config else None
-        self.io_config_func = io_config_func
-        if self.io_config and self.io_config_func:
-            logger.warning("io_config_func is ignored since io_config is provided.")
+        self.io_config = (
+            validate_config(io_config, IOConfig).dict() if isinstance(io_config, (IOConfig, dict)) else io_config
+        )
 
         self.dummy_inputs_func = dummy_inputs_func
 
@@ -641,17 +639,17 @@ class PyTorchModel(OliveModel):
     def get_io_config(self) -> Dict[str, Any]:
         """Return io config of the model.
 
-        Priority: io_config > io_config_func > hf_config (using onnx_config)
+        Priority: io_config > hf_config (using onnx_config)
         """
-        if self.io_config:
+        if isinstance(self.io_config, dict):
             # io_config is provided
             return self.io_config
 
-        if self.io_config_func:
-            # io_config_func is provided
-            logger.debug("Using io_config_func to get io_config")
+        if isinstance(self.io_config, (str, Callable)):
+            # io_config is a string name or a callable
+            logger.debug(f"Calling {self.io_config} to get io_config")
             user_module_loader = UserModuleLoader(self.model_script, self.script_dir)
-            io_config = user_module_loader.call_object(self.io_config_func, self)
+            io_config = user_module_loader.call_object(self.io_config, self)
             return validate_config(io_config, IOConfig).dict()
 
         if self.hf_config and self.hf_config.task and not self.hf_config.components:
@@ -750,7 +748,6 @@ class PyTorchModel(OliveModel):
         return PyTorchModel(
             model_loader=model_loader,
             io_config=hf_component.io_config,
-            io_config_func=hf_component.io_config_func,
             dummy_inputs_func=hf_component.dummy_inputs_func,
             model_script=self.model_script,
             script_dir=self.script_dir,
@@ -765,7 +762,6 @@ class PyTorchModel(OliveModel):
                 "model_file_format": self.model_file_format,
                 "model_loader": self.model_loader,
                 "io_config": self.io_config,
-                "io_config_func": self.io_config_func,
                 "dummy_inputs_func": self.dummy_inputs_func,
                 "hf_config": self.hf_config,
             }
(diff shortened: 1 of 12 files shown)


olive/model/__init__.py
```diff
@@ -475,7 +475,8 @@ def __init__(
         model_loader: Union[str, Callable] = None,
         model_script: Union[str, Path] = None,
         script_dir: Union[str, Path] = None,
-        io_config: Union[Dict[str, Any], IOConfig, str] = None,
+        io_config: Union[Dict[str, Any], IOConfig] = None,
+        io_config_func: Union[str, Callable] = None,
```
guotuofeng, line 479: could we unify them into io_config?

jambayk: Yes, we can do it. That is simpler too.

I remembered in a previous discussion, you mentioned that `io_config` name is confusing when it is a func https://github.com/microsoft/Olive/pull/703#discussion_r1385845515, so I made them separate. 

But I think unifying them and making the type `Union[Dict, IOConfig, str, Callable]` makes the logic simpler. Even though io_config gets overloaded with two different types of values: the config itself or a function to load the config.

shaahji: I recommend unifying it to use just `io_config`.

jambayk: Yes, already pushed a commit to unify them. Fixing some bugs/tests now.

shaahji:
* Update documentation in docs/source/features/huggingface_model_optimization.md
* `encoder_decoder_init_dummy_inputs` and `decoder_dummy_inputs` in examples/whisper/code/user_script.py doesn't need the model. Only config is used.

olive/model/__init__.py
```diff
@@ -637,24 +636,49 @@ def prepare_session(
...
 
         # Priority: dummy_inputs_func > io_config.input_shapes > hf_config.dataset > onnx_config
         dummy_inputs = None
+        io_config = self.get_io_config()
         if self.dummy_inputs_func is not None:
             logger.debug("Using dummy_inputs_func to get dummy inputs")
             user_module_loader = UserModuleLoader(self.model_script, self.script_dir)
             dummy_inputs = user_module_loader.call_object(self.dummy_inputs_func, self)
-        elif self.io_config and self.io_config["input_shapes"]:
+        elif self.io_config and io_config["input_shapes"]:
```
guotuofeng, line 674: do we need this check?

jambayk: Yes, there is a comment below about checking `self.io_config`. We don't want to use io_config that is extracted from hf_config. 

Although, this can also be replaced by `elif io_config.get["input_shapes"]` since the io_config from hf_config never has `input_shapes`. https://github.com/microsoft/Olive/blob/main/olive/model/hf_utils.py#L362 

Depends on whether we want the logic to be clear or simpler.

guotuofeng: the check seems a little bit hacky since it is not used in the subsequent code.

The following line seems deadcode?

        if self.hf_config and self.hf_config.task and not self.hf_config.components:
            # hf_config is provided
            logger.debug("Using hf onnx_config to get io_config")
            return self.hf_config.get_io_config(self.model_path)

guotuofeng: It seems we need split the the get_io_config into two different methods. Once is only get from self.io_config, another is only get from hf_config. 

From the time being, the method has two responsibility, which violates the [https://en.wikipedia.org/wiki/Single-responsibility_principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)

jambayk: It is not dead code. This part is used if there is no `io_config` but we have `hf_config`. Just added a test case for this. 

This is also the reason we check for `self.io_config` since `hf_config` is used to generate io_config in a later `elif` clause.

jambayk: I don't think it is doing two responsibilities. It is doing the same thing as `get_dummy_inputs` where the `dummy_inputs` or `io_config` for a model can originate from different sources. The `hf_config` cases are special cases.

guotuofeng: I mean deadcode in get_dummy_inputs. This kind of behavior seems weird. We'd better refactor code to make sure the logic clear.

jambayk: it's not deadcode because we check io_config from `hf_config` doesn't have shapes.

guotuofeng: If it has only one responsibility, then the check for self.io_config will be needless. the check seems redundant if we don't look into how get_io_config is implemented.

jambayk: I don't see where anything is weird or unclear. The comments clearly say what the priority of io_config and dummy_inputs are. 
We already handle hf_config as special cases in model loading, etc. It is part of the pytorch model's config can be used for model loading, io_config, dummy_inputs etc so that the user doesn't need to provide them. 

We are just abstracting the different ways of getting the model, io_config, etc under the `load_model`, `get_dummy_inputs`, etc methods.
These logics need to be handled somewhere anyways. Better to do it in the model object itself.

jambayk: Updated as per offline discussion.

guotuofeng (approved):

olive/model/__init__.py
```diff
@@ -637,29 +636,71 @@ def prepare_session(
...
+
+        Priority: io_config > hf_config (using onnx_config)
+        """
+        io_config = None
+        if self.io_config:
+            # io_config is provided
+            io_config = self._resolve_io_config(self.io_config)
+        elif self.hf_config and self.hf_config.task and not self.hf_config.components:
+            # hf_config is provided
+            logger.debug("Using hf onnx_config to get io_config")
```
trajepl, line 672: Using hf *onnx*_config? It is pytorch model here.

olive/model/hf_utils.py
```diff
@@ -370,14 +382,9 @@ def get_hf_model_io_config(model_name: str, task: str, feature: Optional[str] =
...
-def get_hf_model_dummy_input(
-    model_name: str,
-    task: str,
-    feature: Optional[str] = None,
-    trust_remote_code: Optional[bool] = None,
-):
-    model_config = get_onnx_config(model_name, task, feature)
-    tokenizer = AutoTokenizer.from_pretrained(model_name, trust_remote_code=trust_remote_code)
+def get_hf_model_dummy_input(model_name: str, task: str, feature: Optional[str] = None, **kwargs):
+    model_config = get_onnx_config(model_name, task, feature, **kwargs)
```
trajepl, line 386: It might not be related with this PR. It looks weird to call `get_onnx_config` in `get_hf_model_dummy_input`

olive/model/__init__.py
trajepl: I found it in hf_utils. But somehow we may need to refine the name or logics here.
It looks weird to get `onnx config` under `get_io_config`.

jambayk: Like your comment below says, it's not related to this PR. It was already in use before this PR. The `onnx_config` comes from `transformers.onnx` https://github.com/huggingface/transformers/blob/main/src/transformers/onnx/config.py#L68. 

I have some plans to migrate from `transformers.onnx` to optimum since huggingface is not maintaining this module anymore. Maybe we can do the renaming then.

olive/model/hf_utils.py
jambayk: Same comment as above.

trajepl (approved):

PR #751, opened 2023-11-23
🧇 Unify the cache model name

trajepl:
## Describe your changes

For example in the case of onnx model components optimization workflow: conversion -> optimization -> mixed_precision -> etc.
Some passes like mixed_precision will forget the original model name(decoder.onnx, encoder.onnx etc) and use the default `main.onnx`.

This PR is used to unify the output cache model name.

Previous:
image

Now:

image

### DIFF
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -359,7 +359,7 @@ class Pass(ABC):
             components = []
             component_names = []
             for cidx, child in enumerate(model.get_model_components()):
-                component_output_path = Path(output_model_path).with_suffix("") / Path(child.model_path).name
+                component_output_path = Path(output_model_path).with_suffix("") / model.get_model_component_name(cidx)
                 output_model_component = self._run_for_config(child, data_root, config, str(component_output_path))
                 output_model_component.model_attributes = (
                     output_model_component.model_attributes or child.model_attributes
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -359,7 +359,7 @@ class Pass(ABC):
             components = []
             component_names = []
             for cidx, child in enumerate(model.get_model_components()):
-                component_output_path = Path(output_model_path).with_suffix("") / str(cidx)
+                component_output_path = Path(output_model_path).with_suffix("") / Path(child.model_path).name
                 output_model_component = self._run_for_config(child, data_root, config, str(component_output_path))
                 output_model_component.model_attributes = (
                     output_model_component.model_attributes or child.model_attributes
--- a/olive/passes/onnx/append_pre_post_processing_ops.py
+++ b/olive/passes/onnx/append_pre_post_processing_ops.py
@@ -84,7 +84,7 @@ class AppendPrePostProcessingOps(Pass):
     ) -> ONNXModel:
         from onnxruntime import __version__ as OrtVersion
 
-        output_model_path = ONNXModel.resolve_path(output_model_path)
+        output_model_path = ONNXModel.resolve_path(output_model_path, Path(model.model_path).name)
 
         tool_command = config.get("tool_command")
         if tool_command:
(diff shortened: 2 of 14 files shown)


olive/passes/onnx/conversion.py
```diff
@@ -345,7 +345,7 @@ def _convert_model_on_device(
         )
 
         # save the model to the output path and return the model
-        output_model_path = ONNXModel.resolve_path(output_model_path)
+        output_model_path = ONNXModel.resolve_path(output_model_path, Path(model.model_path).name)
```
jambayk, line 348: this idea of using the input model's file name only works for onnx -> onnx passes. Here, the change is not needed

trajepl: Yes, fix is coming.

olive/passes/onnx/append_pre_post_processing_ops.py
```diff
@@ -84,7 +84,7 @@ def _run_for_config(
     ) -> ONNXModel:
         from onnxruntime import __version__ as OrtVersion
 
-        output_model_path = ONNXModel.resolve_path(output_model_path)
+        output_model_path = ONNXModel.resolve_path(output_model_path, Path(model.model_path).name)
```
jambayk, line 87: Can you update `ONNXModel.resolve_path` to make sure it always has `.onnx` suffix in the `model_filename` part? 

All of these calls assume `Path(model.model_path).name` has `.onnx` suffix. It is true for all cases where olive creates the initial onnx model but it doesn't hurt to be safe.

olive/passes/onnx/conversion.py
```diff
@@ -106,7 +106,8 @@ def _run_for_config(
             for component_name in model.components:
                 # convert each component
                 component_model = model.get_component(component_name)
-                component_output_path = str(Path(output_model_path).with_suffix("") / component_name)
+                # component_name is set with Path(component_model.onnx).stem
+                component_output_path = str(Path(output_model_path).with_suffix("") / f"{component_name}.onnx")
```
jambayk, line 110: I think there it might be better to do `{component_name}/model.onnx` so that the external data files between the components don't clash. 
The external data files won't clash if `all_tensors_to_one_file` is `True` but for `False`, both components might have the same names for their tensor data files.

jambayk: So, the change here is not necessary imo

olive/passes/olive_pass.py
```diff
@@ -359,7 +359,7 @@ def run(
             components = []
             component_names = []
             for cidx, child in enumerate(model.get_model_components()):
-                component_output_path = Path(output_model_path).with_suffix("") / str(cidx)
+                component_output_path = Path(output_model_path).with_suffix("") / Path(child.model_path).name
```
jambayk, line 362: same as the other comment, i think it is better to keep the contents of the components in separate sub directories.

jambayk: We could enumerate over the component names to get a more descriptive sub dir name instead of `str(cidx)`

olive/passes/onnx/conversion.py
trajepl: Thanks. I did not realized that there will be clash for tensor data. Updated.

olive/passes/olive_pass.py
trajepl: Updated.

olive/passes/onnx/append_pre_post_processing_ops.py
trajepl: done.

jambayk (dismissed):

olive/passes/onnx/optimum_merging.py
```diff
@@ -72,8 +70,7 @@ def new_byte_size_func(_):
             ModelProto.ByteSize = prev_byte_size_func
 
         # onnx.save will fail if the directory doesn't already exist
-        Path(output_model_path).mkdir(parents=True, exist_ok=True)
-        output_model_path = os.path.join(output_model_path, "decoder_model_merged.onnx")
+        ONNXModel.resolve_path(output_model_path, "decoder_model_merged.onnx")
```
jambayk, line 73: this is missing `output_model_path = `

trajepl: fixed.

jambayk (approved):

PR #761, opened 2023-11-29
🧇 Warn the S8S8 QOperator quantization but not pruning

trajepl:
## Describe your changes
For some uses, they still want to try S8S8 with QOperator quantization config. But Olive directly prune it.
This PR is used to warn the S8S8 QOperator quantization but not pruning.

### DIFF
--- a/olive/passes/onnx/quantization.py
+++ b/olive/passes/onnx/quantization.py
@@ -304,7 +304,6 @@ class OnnxQuantization(Pass):
                 logger.warning(
                     "S8S8 with QOperator will be slow on x86-64 CPUs and should be avoided in general, try QDQ instead."
                 )
-                return True
             if config["EnableSubgraph"] is True:
                 logger.info("EnableSubgraph is not supported for static quantization.")
                 return False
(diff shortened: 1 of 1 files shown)


olive/passes/onnx/quantization.py
```diff
@@ -298,8 +298,13 @@ def validate_search_point(
...
             ):
-                logger.info("S8S8 with QOperator will be slow on x86-64 CPUs and should be avoided in general")
-                return False
+                # S8S8 with QOperator will be slow on x86-64 CPUs and should be avoided in general.
+                # https://onnxruntime.ai/docs/performance/model-optimizations/quantization.html#data-type-selection
+                # But we still allow it for users to try at their own risk. Olive just warns this to users.
+                logger.warning(
+                    "S8S8 with QOperator will be slow on x86-64 CPUs and should be avoided in general, try QDQ instead."
+                )
+                return True
```
jambayk, line 307: Should not return anything here since the next condition might be true.
Can just log the warning and let the next or final return happen.

trajepl: Updated.

jambayk (approved):

PR #771, opened 2023-12-02
Update PythonEnvironmentSystem

jambayk:
## Describe your changes
WIP. Creating a draft PR for CI checks.


olive/systems/python_environment/python_environment_system.py
```diff
@@ -81,8 +70,39 @@ def __init__(
...
+            # add extra args
+            for key, value in kwargs.items():
+                if value is None:
+                    continue
+                command += f" --{key} {value}"
+            # output path
+            output_path = tmp_dir_path / "output.json"
+            command += f" --output_path {output_path}"
+
+            run_subprocess(command, env=self.environ, check=True)
```
guotuofeng, line 100: >run_subprocess [](http://example.com/codeflow?start=12&length=14)

nit: do we need print the stdour/stderr into the logging file?

guotuofeng, line 100: >command [](http://example.com/codeflow?start=27&length=7)

I suppose the command will be splitted into array of string. right?

jambayk: yes the command split is handled by `run_subprocess` https://github.com/microsoft/Olive/blob/main/olive/common/utils.py#L24. It uses `shlex.split` which handles windows or posix so we can just provide the command string.

jambayk: yes, I can add this as debug log

jambayk:
Converting to draft as we want to keep the remote onnxruntime inference functionality in a new system and validate the mobilenet example again.

jambayk:
Closing for now. Not high priority so will resume later when the changes are needed.

Issue #777, opened 2023-12-04
[Bug]: local_resource_path.save_to_dir(save_dir, save_name, overwrite) failed


ningx-ms:
### What happened?

WSL2 + Ubuntu 20.04 as docker container. 

The adapter files are copied successfully and able to be loaded for inferencing. 
Maybe some file attributes failed to copy?

Note that in my setup, `workspace` and `project` is a folder on Windows file system, then mapped to a WSL Dev container.

Traceback (most recent call last):
  File "/opt/conda/lib/python3.9/site-packages/olive/engine/engine.py", line 412, in run_accelerator
    return self.run_no_search(
  File "/opt/conda/lib/python3.9/site-packages/olive/engine/engine.py", line 513, in run_no_search
    output_model_json = cache_utils.save_model(
  File "/opt/conda/lib/python3.9/site-packages/olive/cache.py", line 243, in save_model
    model_json["config"][resource_name] = local_resource_path.save_to_dir(save_dir, save_name, overwrite)
  File "/opt/conda/lib/python3.9/site-packages/olive/resource_path.py", line 223, in save_to_dir
    shutil.copytree(self.config.path, new_path)
  File "/opt/conda/lib/python3.9/shutil.py", line 569, in copytree
    return _copytree(entries=entries, src=src, dst=dst, symlinks=symlinks,
  File "/opt/conda/lib/python3.9/shutil.py", line 523, in _copytree
    raise Error(errors)
[...]

PR #779, opened 2023-12-05
Handle `shutil.Error` from `shutil.copytree`

jambayk:
## Describe your changes
As described in https://github.com/microsoft/Olive/issues/777, `shutil.copytree` fails with `shutil.Error` even though all files are copied. 
https://bugs.python.org/issue36823

This PR adds a new utils function `copy_dir` that wraps `shutil.copytree` to handle the above case. As a heuristic, we check that all required file names exist in the destination directory in the event of an error. 
 
### DIFF
--- a/olive/common/utils.py
+++ b/olive/common/utils.py
@@ -13,6 +13,7 @@ import shlex
 import shutil
 import subprocess
 import time
+from pathlib import Path
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


olive/common/utils.py
```diff
@@ -232,3 +233,52 @@ def find_submodules(module, submodule_types, full_name=False):
...
+        # only check files names so it is not foolproof
+        not_copied = [
+            member.relative_to(src_dir)
+            for member in all_files(src_dir, ignore)
+            if not (dst_dir / member.relative_to(src_dir)).exists()
+        ]
+        if not_copied:
+            raise RuntimeError(f"Failed to copy {not_copied}") from e
+        else:
+            logger.debug(
```
guotuofeng, line 281: warning or error?

jambayk: changed to warning since that makes more sense. That way, the user knows about the error even though we ignore it.

guotuofeng (approved):

PR #786, opened 2023-12-06
refactoring OliveModel and change its name to OliveModelHandler.

guotuofeng:
## Describe your changes
* Rename all OliveModel by adding suffix "Handler". For example PyTorchModel -> PyTorchModelHandler.
* Split the logic doesn't belong to model handler into separated mixins. HfConfigMixin, IoConfigMixin, DummyInputsMixin, etc.
* Merge the CompositeOnnxModel and CompositePyTorchModel into one CompositeModelHandler.
* Keep using the old "OliveModel" types

Basically, all the logic is keeping same except some logic simplify changes:
* Merge the Composite ONNX and PyTorch model. 
* Simplify the composite model enumerate logic by using a single method. 
* Same for original PyTorch model to get the hugging face model components. In the new implementation, only one get_hf_components is used to enumerate both name and model.
* As a side effect the merging composite onnx model and pytorch model, the original CompositePyTorch model conversion logic is in conversion.py, but with the new changes, the child model get and run for each child model will be done by the run of olive_pass.py 
* As a side effect of providing only single entry point to enumerate the model components, the insert_beam_search pass is updated by using the new get model component logic, so does the olive_pass.run. 

From the time being, the following logic is not tested
* Distributed models
* Composite PyTorch model


xiaoyu-work:
I have some general questions:

- I don't see that HF model logic is decoupled from PyTorch model? Should we have a separate HFModelHandler?
- For some mixin classes, such as `DummyInputsMixin`, `HfConfigMixin`, if they are only be used by PyTorchModelHandler, any reason it should be a minxin rather a regular attribute or method in the class? I'm okay to have them as mixins if some of them are reused in other model handlers, but now only pytorch model need them. A class attribute or method would be more clear.
- For some minxi classes, such as `JsonMixin` or `ResourceMixin` that are in `OliveModelHandler`. Any pros as a mixin comparing to a general method in `OliveModelHandler`? In general both cases look like an overkill here.

guotuofeng:
> I have some general questions:
> 
> * I don't see that HF model logic is decoupled from PyTorch model? Should we have a separate HFModelHandler?
> * For some mixin classes, such as `DummyInputsMixin`, `HfConfigMixin`, if they are only be used by PyTorchModelHandler, any reason it should be a minxin rather a regular attribute or method in the class? I'm okay to have them as mixins if some of them are reused in other model handlers, but now only pytorch model need them. A class attribute or method would be more clear.
> * For some minxi classes, such as `JsonMixin` or `ResourceMixin` that are in `OliveModelHandler`. Any pros as a mixin comparing to a general method in `OliveModelHandler`?
> 
> In general both 2 and 3 cases look like an overkill here.

these mixins is just make sure the related responsibilities be in their own class instead of mix them all together. Please refer to https://en.wikipedia.org/wiki/Single_responsibility_principle for more design principles.

xiaoyu-work:
> > I have some general questions:
> > 
> > * I don't see that HF model logic is decoupled from PyTorch model? Should we have a separate HFModelHandler?
> > * For some mixin classes, such as `DummyInputsMixin`, `HfConfigMixin`, if they are only be used by PyTorchModelHandler, any reason it should be a minxin rather a regular attribute or method in the class? I'm okay to have them as mixins if some of them are reused in other model handlers, but now only pytorch model need them. A class attribute or method would be more clear.
> > * For some minxi classes, such as `JsonMixin` or `ResourceMixin` that are in `OliveModelHandler`. Any pros as a mixin comparing to a general method in `OliveModelHandler`?
> > 
> > In general both 2 and 3 cases look like an overkill here.
> 
> these mixins is just make sure the related responsibilities be in their own class instead of mix them all together. Please refer to https://en.wikipedia.org/wiki/Single_responsibility_principle for more design principles.

I don't think my concerns go against single responsibility principle. A mixin is designed to add reusable functionality to multiple objects or classes. If a method is only used by one specific class, then creating a mixin might not be necessary. Instead, keeping it within the class where it belongs would likely be clearer.

olive/model/mixin/composite.py
```diff
@@ -0,0 +1,10 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+class CompositeMixin:
+    def set_composite_parent(self, cp):
+        self.composite_parent = cp
+
+    def get_composite_parent(self):
```
trajepl, line 9: Should we maintain the constructor function? Take CompositeMixin example, it use `self.composite_parent`. When it comes to subclass which inherts it like:
```python
Class A(CompositeMixin):
    ...
```

Then following code would fail:
```python
a = A()
a.get_composite_parent()
```
See this.
https://stackoverflow.com/questions/39469841/multiple-mixins-and-properties

guotuofeng: I prefer not have any `__init__` in mixin since the call of multiple __init__ will be much painful. https://www.reddit.com/r/learnpython/comments/ymet1a/writing_mixin_classes_whats_the_best_practice/

olive/model/handler/base.py
```diff
@@ -0,0 +1,66 @@
...
+    resource_keys: ClassVar[list] = ["model_path"]
+
+    def __init__(
+        self,
+        framework: Framework,
+        model_file_format: ModelFileFormat,
+        model_path: OLIVE_RESOURCE_ANNOTATIONS = None,
+        model_attributes: Optional[Dict[str, Any]] = None,
+    ):
+        self.framework = framework
```
trajepl, line 29: Could we add the comment to tell: 
if you add new Mixin into model handler, please make sure the member variable is initialized properly in model handler..

olive/model/config/hf_config.py
```diff
@@ -0,0 +1,215 @@
...
+from olive.common.config_utils import ConfigBase, ConfigWithExtraArgs
+from olive.common.pydantic_v1 import Field, validator
+from olive.common.utils import resolve_torch_dtype
+from olive.model.config.io_config import IoConfig
+
+logger = logging.getLogger(__name__)
+
+
+class HfComponent(ConfigBase):
+    name: str
```
devang-ml, line 21: Please add comment before each new class covering description, responsibilities & functionalities and config.json snippet example when relevant.

devang-ml (commented):
At high level, the proposed change looks good to me. Go ahead if details are OKed by other reviewers.

olive/model/config/hf_config.py
guotuofeng: sure, will do

olive/model/utils/onnx_utils.py
```diff
@@ -0,0 +1,69 @@
...
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+import logging
+from pathlib import Path
+from typing import Optional
+
+logger = logging.getLogger(__name__)
+
+
+def resolve_path(file_or_dir_path: str, model_filename: str = "model.onnx") -> str:
```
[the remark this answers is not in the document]
guotuofeng: make sense, will do

olive/model/config/registry.py
```diff
@@ -0,0 +1,36 @@
...
+
+        REGISTRY[model_type.lower()] = cls
+        cls.model_type = model_type
+        return cls
+
+    return decorator_model_class
+
+
+def get_model_handler(model_type):
+    if not is_valid_model_type(model_type):
```
trajepl, line 30: Maybe simplify to:
```python
REGISTRY.get(model_type.lower()) or raise ValueError(f"Unknown model type {model_type}")
```

olive/model/config/registry.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Returns:
+        cls: The class of register.
+    """
+
+    def decorator_model_class(cls):
+        if model_type in REGISTRY:
+            raise ValueError("Cannot have two model handlers with the same name")
+
+        REGISTRY[model_type.lower()] = cls
```
trajepl, line 22: Is it possible to get model_type automatically?
`model_type = model_type or model_type[:-len("handler")]`

I am not sure if it is totally 1:1 mapping between model_type and model_type's handler, like: f"{model_type}Handler"

guotuofeng: The model_type name will not be same with the class by removing suffix. I don't like make the tight coupled. By using the name explicitly, we could make the class name changed at anytime without breaking the back compatibility.

olive/model/handler/mixin/resource.py
```diff
@@ -0,0 +1,71 @@
...
+
+from olive.resource_path import OLIVE_RESOURCE_ANNOTATIONS, ResourcePath, ResourcePathConfig, create_resource_path
+
+logger = logging.getLogger(__name__)
+
+
+class ResourceMixin:
+    """Provide the resource functionalities for the model handler."""
+
+    resource_keys: Tuple[str] = None
```
jambayk, line 17: should the default value be `()` since `None` is not iterable?

Issue #709, opened 2023-11-09
[FR]: Hope Olive supports Pydantic version 2.0 or higher


tangcan1600:
### Proposal Summary

gradio 4.1.2 requires pydantic>=2.0, but you have pydantic 2.0a1 which is incompatible.
olive-ai 0.3.3 requires pydantic<2.0.0, but you have pydantic 2.0a1 which is incompatible.

gradio4.1.2 is not compatible with Pydantic 2.0 or earlier versions，hope olive can support Pydantic version 2.0 or higher，thanks

### What component(s) does this request affect?

- [ ] OliveModels
- [ ] OliveSystems
- [ ] OliveEvaluator
- [ ] Metrics
- [ ] Engine
- [ ] Passes
- [X] Other

guotuofeng:
@tangcan1600, could you clarify the scenario that need gradio 4.1.2? Is it possible that downgrade it to 3.x to unblock you?

tangcan1600:
> @tangcan1600, could you clarify the scenario that need gradio 4.1.2? Is it possible that downgrade it to 3.x to unblock you?

 I want to use the new  features in gradio 4.0, such as ChatInterface

tangcan1600:
I want to use the new features in gradio 4.0, such as ChatInterface

tangcan1600:
I want to use the new features in gradio 4.0, such as ChatInterface

tangcan1600:
> @tangcan1600, could you clarify the scenario that need gradio 4.1.2? Is it possible that downgrade it to 3.x to unblock you?
I want to use the new features in gradio 4.0, such as ChatInterface

guotuofeng:
there is not plan to upgrade to Pydantic 2 in the short term.

JKinx:
@tangcan1600 Olive is now compatible with pydantic v2. 

However, the version `2.0a1` you have in your environment is a pre-release and is missing the module we require to support v2. If you would like to use olive with pydantic v2, please update pydantic to a newer version and install olive from source:
```bash
pip install ---upgrade pydantic
pip install git+https://github.com/microsoft/Olive
```

jambayk:
@tangcan1600 Olive is now compatible with pydantic v2. 

However, the version `2.0a1` you have in your environment is a pre-release and is missing the module we require to support v2. If you would like to use olive with pydantic v2, please update pydantic to a newer version and install olive from source:
```bash
pip install --upgrade pydantic
pip install git+https://github.com/microsoft/Olive
```

PR #788, opened 2023-12-06
Add `pydantic_v1` module to make Olive `pydantic` version agnostic

jambayk:
## Describe your changes
Pydantic v2 had significant breaking changes which blocks migration to it. So, we pinned the version to v1. 

Pydantic v2 has a `v1` namespace inside it. 

This PR introduces `olive.common.pydantic_v1` module that handles the import of `v1` api so that Olive can be version agnostic wrt pydantic. This way, we won't have version conflict with client applications such as that in https://github.com/microsoft/Olive/issues/709. 

Note:
The doc builder still needs pydantic v1 and autodoc_pydantic v1 since we are using the v1 api.

### DIFF
--- a/olive/common/pydantic_v1.py
+++ b/olive/common/pydantic_v1.py
@@ -1,0 +1,19 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+"""Pydantic v1 compatibility module.
+
+Pydantic v2 has breaking changes that are not compatible with the current version of Olive.
+Migration Guide: https://docs.pydantic.dev/latest/migration/.
+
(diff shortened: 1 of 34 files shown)


olive/common/pydantic_v1.py
```diff
@@ -0,0 +1,19 @@
...
+
+Pydantic v2 has breaking changes that are not compatible with the current version of Olive.
+Migration Guide: https://docs.pydantic.dev/latest/migration/.
+
+In order to support both versions of Pydantic, we use this module to access pydantic's v1 API.
+"""
+
+try:
+    # pydantic v2
+    from pydantic.v1 import *  # noqa: F403
```
guotuofeng, line 16: would the namespace in v2 is full compatible with exist v1 stuff? did you try it? 

I guess there might be some features are not supported in v2 back-compatibility?

guotuofeng (approved):

PR #793, opened 2023-12-07
🫒 Auto optimization

trajepl:
## Describe your changes
Reopen the auto optimization PR with pass flows templates. 
image

Previous PR: https://github.com/microsoft/Olive/pull/610


trajepl:
@devang-ml For your comments:
> we may not need regulate feature.

With the pass flows templates, we define the possible pass combinations for given accelerators/EP/precisions. It is OK to not regulate the pass_flows. 
But the pass config is empty. For example for gpu_cuda_fp32 and gpu_cuda_fp16, there are both `OrtTransformersOptimization`, but we need assign different config to enable fp16 or not.

olive/auto_optimizer/config_template/available_pass_flows.yaml
```diff
@@ -0,0 +1,79 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+_variables:
+  - &fp32 ["OnnxConversion", "OrtTransformersOptimization", "OrtPerfTuning"]
+  - &fp16 ["OnnxConversion", "OrtTransformersOptimization", "OrtPerfTuning"]
```
devang-ml, line 8: Why not invert this mapping? List precision supported by each Pass.

devang-ml (commented):
We should not require user to provide precision for auto optimizer and create pass flows for all the precisions supported by the target accelerator. 

Let's aim to make auto optimizer just a Boolean flag.

olive/auto_optimizer/config_template/available_pass_flows.yaml
trajepl: Oh, the `&fp32`, `&inc_int8` and etc are just the reference to the pass_flows template which also are mapped with accelerate/ep(recall the pass flows table we talked before.)

Then this yaml file is used to describe the table I shared. It is the template which olive will retrieve according to accelerator/ep/precisions.
Revertting this mapping will flatten the pass_flows into pass level. Somehow we have to composite the pass into pass flows if revertting happens.

pyproject.toml
```diff
@@ -132,6 +132,7 @@ ignore = [
   "D406", # Ignore new line after section name
   "D407", # Ignore dashed-underline-after-section
   "N803", # Argument casing
+  "N805", # Checks for instance methods that use a name other than self for their first argument.
```
guotuofeng, line 135: this is for cls method?

test/unit_test/auto_optimizer/test_auto_optimizer.py
```diff
@@ -0,0 +1,111 @@
...
+        perf_opt_name = "OrtPerfTuning_trt_fp16" if expected_trt_fp16 else "OrtPerfTuning"
+        assert pass_config[trans_opt_name]["config"]["float16"] == expected_cuda_fp16
+        assert pass_config[perf_opt_name]["config"]["enable_cuda_graph"] == expected_cuda_fp16
+        assert pass_config[perf_opt_name]["config"]["trt_fp16_enable"] == expected_trt_fp16
+
+    @pytest.mark.parametrize(
+        "metrics, accelerator_spec, auto_optimizer_config, expected_pass_flows",
+        [
+            (
+                [get_accuracy_metric(AccuracySubType.ACCURACY_SCORE, goal_type="max-degradation")],
```
guotuofeng, line 71: could we postpone the get metrics in the test method? here is seems like a variable belongs to the module level instead of inside the test function.

olive/engine/engine.py
```diff
@@ -187,6 +197,44 @@ def setup_accelerators(self, execution_providers):
                 "Please consider installing an onnxruntime build that contains the relevant execution providers. "
             )
 
+    def prepare_passes_from_auto_optimizer(
```
guotuofeng, line 200: what's the output of the method?

olive/engine/config.py
```diff
@@ -32,3 +33,4 @@ class EngineConfig(ConfigBase):
     clean_cache: bool = False
     clean_evaluation_cache: bool = False
     plot_pareto_frontier: bool = False
+    auto_optimizer_config: Optional[AutoOptimizerConfig] = None
```
guotuofeng, line 36: should we move the config out of engine? I suppose it should be above the engine level?

olive/engine/engine.py
```diff
@@ -187,6 +197,44 @@ def setup_accelerators(self, execution_providers):
...
+            data_configs=data_configs,
+        )
+        pass_config, pass_flows = auto_optimizer.suggest()
+
+        if pass_config:
+            for pass_name, config in pass_config.items():
+                p = Pass.registry[config["type"].lower()]
+                self.register(p, config["config"], name=pass_name)
+            self.set_pass_flows(pass_flows)
+            # save suggested pass config/pass flows to output_dir/auto_optimizer/{accelerator_spec}.json
```
guotuofeng, line 228: when these files are used? should we use return values so that caller will use them?

olive/auto_optimizer/config_template/available_pass_flows.yaml
trajepl: Updated.

pyproject.toml
trajepl: In my local tests, it always show false alert on some instance method. Run linter runner init can fix.

olive/engine/engine.py
trajepl: Updated.

olive/engine/config.py
trajepl: Updated.

olive/engine/engine.py
trajepl: Removed in the new commit.

olive/workflows/run/run.py
```diff
@@ -182,16 +152,27 @@ def run_engine(config: RunConfig, data_root: str = None):
...
+        pass_flows = config.pass_flows or []
+        if not config.passes and not config.auto_optimizer_config.disable_auto_optimizer:
+            auto_optimizer = AutoOptimizer(
+                input_model,
+                engine.evaluator_config,
+                accelerator_spec,
+                config.auto_optimizer_config,
+                config.data_configs,
+            )
+            _passes, pass_flows = auto_optimizer.suggest()
```
guotuofeng, line 171: how about the following logic:
```python
pass_map = {}
if if not config.passes and not config.auto_optimizer_config.disable_auto_optimizer:
    for accelerator_spec in accelerator_specs:
        pass_map[[accelerato_spec]] = auto_optimizer.suggest()
else:
    pass_map[accelerator_specs] = (passes, pass_flows)

for accelerator_specs, (passes, pass_flows) in pass_map.items():
    engine.run(xxx)
```

guotuofeng: In this way, we could decouple the pass generation and engine run. And it is easy to test both logic by using UT easily.

trajepl: Updated.

olive/engine/engine.py
```diff
@@ -464,12 +389,18 @@ def setup_passes(self, accelerator_spec: AcceleratorSpec):
                 pass_search_spaces.append((pass_name, p.search_space()))
             self.pass_flows_search_spaces.append(pass_search_spaces)
 
+    def cleanup_passes(self):
```
guotuofeng, line 392: reset_passes?

guotuofeng:
> 

do we need update MANIFEST.in to include these files?

---
Refers to: olive/auto_optimizer/config_template/opt_level_passes.yaml:15 in 3430b4e. [](commit_id = 3430b4e9dbb0d025eb2fb1e95f3ce232ed993418, deletion_comment = False)

olive/auto_optimizer/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+    ):
+        # TODO(trajep): how to get model_attributes if the model is in AML workspace? Maybe we can ignore this for now?
+        self.input_model_config = input_model_config
+        self.evaluator_config = evaluator_config
+        self.accelerator_spec = accelerator_spec
+        self.auto_optimizer_config = auto_optimizer_config or AutoOptimizerConfig()
+        self.data_configs = data_configs or {}
+        self.initialize()
+
+    def initialize(self):
```
guotuofeng, line 74: private method?

olive/auto_optimizer/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+        model_config = deepcopy(self.input_model_config)
+        resource_paths = model_config.get_resource_paths()
+        for resource_name, resource_path in resource_paths.items():
+            if not resource_path or resource_path.is_local_resource_or_string_name():
+                continue
+            downloaded_resource_path = cache_utils.download_resource(resource_path, self._config.cache_dir)
+            if downloaded_resource_path:
+                # set local resource path
+                model_config.config[resource_name] = downloaded_resource_path
+        model = model_config.create_model()
```
guotuofeng, line 86: >model [](http://example.com/codeflow?start=8&length=5)

where is it used?

olive/auto_optimizer/regulate_mixins.py
```diff
@@ -0,0 +1,124 @@
...
+                                new_pf[j]: {
+                                    "type": trans_opt,
+                                    "config": {
+                                        "float16": cuda_fp16,
+                                        "use_gpu": True,
+                                    },
+                                }
+                            }
+                        )
+                    if perf_tuning == p:
```
guotuofeng, line 76: >if [](http://example.com/codeflow?start=20&length=2)

elif?

olive/auto_optimizer/__init__.py
guotuofeng, line 86: I suppose engine will cache the model?

olive/auto_optimizer/__init__.py
```diff
@@ -0,0 +1,138 @@
...
+        # if user can tolerate accuracy drop, we can enable more optimization
+        default_precisions = [Precision.FP32]
+        if self.is_accuracy_drop_tolerance:
+            # ignore int4 for now as it is not supported very well in onnxruntime
+            # enable it only when user explicitly set it
+            # default_precisions = [Precision.FP32, Precision.FP16, Precision.INT8, Precision.INT4]
+            default_precisions = [Precision.FP32, Precision.FP16, Precision.INT8]
+        self.auto_optimizer_config.precisions = self.auto_optimizer_config.precisions or default_precisions
+
+    def suggest(self):
```
guotuofeng, line 100: >suggest [](http://example.com/codeflow?start=8&length=7)

I suppose only this method need to be public and all other methods are private?

olive/auto_optimizer/regulate_mixins.py
```diff
@@ -0,0 +1,124 @@
...
+                if tuple(pf) not in unique_pass_flows:
+                    pass_flows.append(pf)
+                unique_pass_flows.add(tuple(pf))
+                for p in pf:
+                    if p not in pass_config:
+                        pass_config.update({p: {"type": p, "config": {}}})
+
+        return pass_config, pass_flows
+
+    def regulate_fp16(self, pass_config, pass_flows):
```
guotuofeng, line 35: >regulate_fp16 [](http://example.com/codeflow?start=8&length=13)

Basically, it a method is only invoked in the class itself, we can safely make it private. ditto for allow_precision

pyproject.toml
guotuofeng: Just curious which method except classmethod.

olive/auto_optimizer/__init__.py
trajepl: Removed and only keep input_model_config to regulate some passes roughly. Like for onnx model, we can skip conversion pass.

olive/auto_optimizer/regulate_mixins.py
trajepl: Set as private in latest commit.

olive/auto_optimizer/__init__.py
trajepl: Removed.

docs/source/tutorials/configure_auto_optimizer.rst
```diff
@@ -0,0 +1,263 @@
+.. _how_to_configure_auto_optimizer:
+
+[BETA] How To Configure Auto Optimizer
+=======================================
+
+Auto Optimizer is a tool that can be used to automatically search Olive passes combination based on:
+
+1. input model
+2. target device
+3. target precision: FP32, FP16, INT8, INT4 and etc.
```
guotuofeng, line 10: >FP32 [](http://example.com/codeflow?start=21&length=4)

I see the line 39 is lower case. should we make it lowe case here?

trajepl: Updated.

olive/auto_optimizer/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+    FP16 = "fp16"
+    INT8 = "int8"
+    INT4 = "int4"
+
+
+class AutoOptimizerConfig(ConfigBase):
+    # opt_level
+    # 0: look up template mapping
+    # TODO(anyone): support more opt_level
+    opt_level: int = 0
```
xiaoyu-work, line 33: For other optimization tools, `opt_level` usually means different precision. Do we want to follow the same concept here? Or will define a new `opt_level` for Olive.

olive/auto_optimizer/__init__.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # used to disable auto optimizer when user only want to evaluate input model
```
xiaoyu-work, line 35: If user only wants to evaluate input model, will they add `auto_optimizer_config` in configuration?

docs/source/tutorials/configure_auto_optimizer.rst
```diff
@@ -0,0 +1,262 @@
+.. _how_to_configure_auto_optimizer:
```
xiaoyu-work, line 1: Seems this auto optimizer only supports onnxruntime optimization with onnx model. Can we add this note to the documentation? For example, this auto optimizer doesn't support openvino model and snpe model optimization.

olive/auto_optimizer/regulate_mixins.py
```diff
@@ -0,0 +1,131 @@
...
+                unique_pass_flows.add(tuple(pf))
+                for p in pf:
+                    if p not in pass_config:
+                        pass_config.update({p: {"type": p, "config": {}}})
+
+        return pass_config, pass_flows
+
+    def _regulate_fp16(self, pass_config, pass_flows):
+        pass_config = pass_config or {}
+        is_gpu = self.accelerator_spec.accelerator_type == Device.GPU and self.accelerator_spec.execution_provider in [
```
xiaoyu-work, line 44: How are we going to support other gpu ep https://github.com/microsoft/Olive/blob/main/olive/hardware/accelerator.py#L56?

olive/auto_optimizer/__init__.py
trajepl: Yes, they need to disable auto optimizer with `disable_auto_optimizer` as True.

olive/auto_optimizer/regulate_mixins.py
trajepl: 1. For openvino and cpu ep in gpu device, we will ignore to regulate fp16.
2. For rocm, I am not sure how it works for fp16 which is converted from transformer optimization passes. If needed, we can just add it into the check lists.

olive/auto_optimizer/__init__.py
xiaoyu-work: I expect if they do not add `auto_optimizer_config` to config file, auto optimizer will be automatically disabled?

olive/auto_optimizer/regulate_mixins.py
xiaoyu-work: We add rocm use case when needed.

olive/auto_optimizer/__init__.py
devang-ml: We don't need to imply that opt_level impacts precision.

olive/auto_optimizer/config_template/precision_aware_pass_flows.yaml
```diff
@@ -0,0 +1,16 @@
...
+# --------------------------------------------------------------------------
+
+
+# Passes which can conduct precision aware optimization
+# e.g. OrtTransformersOptimization can be applied to fp32, fp16
+# but OnnxQuantization only works for int8
+OrtTransformersOptimization:
+  - fp32
+  - fp16
+
```
devang-ml, line 13: Why do we need a separate .yaml here? Just like OnnxMatMul4Quantizer lists 2 EPs in pass_capability.yaml the passes can list multiple precisions.

olive/auto_optimizer/template_mapping.py
```diff
@@ -0,0 +1,78 @@
...
+    # 2. OnnxQuantization and OrtTransformersOptimization_fp16 can not appear in
+    # the same pass flow as we should avoid to quantize a fp16 model
+    # etc.
+
+    pass_flows = []
+    for p in passes_candidates:
+        pass_flows.append([*LEADING_PASSES, p, *TRAILING_PASSES])
+    if not pass_flows or precision == "fp16":
+        pass_flows.append([*LEADING_PASSES, *TRAILING_PASSES])
+    return pass_flows
```
devang-ml, line 78: Instead of keeping LEADING_PASSES and TRAILING_PASSES list, could we just use the sequence from opt_level_passes directly? We need to avoid multiple places listing a pass to aid creation of the pass flow.

olive/auto_optimizer/__init__.py
xiaoyu-work: @devang-ml what does `opt_level` in Olive mean? Does it mean different pass flow?

devang-ml: yes

olive/evaluator/metric_config.py
```diff
@@ -100,6 +102,21 @@ def check_type(cls, v):
     def check_value(cls, v, values):
         if "type" not in values:
             raise ValueError("Invalid type")
-        if values["type"] in ["min-improvement", "max-degradation"] and v < 0:
-            raise ValueError(f"Value must be positive for type {values['type']}")
+        if (
+            values["type"]
+            in ["min-improvement", "max-degradation", "percent-min-improvement", "percent-max-degradation"]
```
justinchuby, line 107:
```suggestion
            in {"min-improvement", "max-degradation", "percent-min-improvement", "percent-max-degradation"}
```

nit: just flying by: use a set for membership checks

olive/hardware/accelerator.py
```diff
@@ -129,3 +133,78 @@ def infer_accelerators_from_execution_provider(execution_provider: List[str]):
...
+    accelerators: List[str] = target.accelerators
+    if accelerators is None:
+        inferred_accelerators = AcceleratorLookup.infer_accelerators_from_execution_provider(execution_providers)
+        if not inferred_accelerators:
+            logger.warning("Cannot infer the accelerators from the target system. Use CPU as default.")
+            accelerators = ["CPU"]
+        else:
+            logger.debug(
+                f"Use inferred accelerators {inferred_accelerators} "
+                f"from given execution providers {execution_providers}."
```
justinchuby, line 164: nit: use the `%` formatting for logs: https://docs.astral.sh/ruff/rules/logging-f-string/

olive/evaluator/metric_config.py
justinchuby: https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/use-set-for-membership.html

olive/auto_optimizer/config_template/precision_aware_pass_flows.yaml
trajepl: That is useless config from before's implementation, will remove it.

olive/hardware/accelerator.py
trajepl: Thanks! Seems olive doesn't catch this kinda linter warnings. 
I updated with your suggestion. Will send separated PR to enable these linter.

olive/evaluator/metric_config.py

olive/auto_optimizer/template_mapping.py
trajepl: image
It is hard to use flatten list before to express the hierarchical relationship rules. I updated with above structure and:
**search available pass flows with validation with pass capacity.**  That means:
1. we can skip some of the passes if there is no available passes.
2. for those level has only one passes like`[OnnxConversion]`, it is required that can not be skipped.
3. the search is done by bfs.

docs/source/tutorials/configure_auto_optimizer.rst
```diff
@@ -0,0 +1,263 @@
...
+Config Fields:
+
+    1. opt_level[int]: default 0, to look up built-in passes template.
+
+    2. disable_auto_optimizer[bool]: default False.
+        - If set to True, Auto Optimizer will be disabled and user need to provide passes combination manually.
+
+    3. precision[str]: default "fp32".
+        - The precision of output model. It can be "placeholder", "fp32", "fp16" and "int8" for now. "placeholder" means the precision of output model is not specified and will be determined by above optimization factors.
+
```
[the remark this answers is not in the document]
trajepl: Updated.

olive/auto_optimizer/__init__.py
trajepl: I suppose that should not be the common use case. Let us think this, for most of the users who leverage olive to optimize model, it would be simpler user just provide model, then the auto optimizer will begin to search.

olive/auto_optimizer/__init__.py
trajepl: Updated the comments. Or do you strongly suggest to update the variable name (opt_level -> ? )?

olive/auto_optimizer/__init__.py
trajepl: Updated with logics of: 
> If pass list is provided by the user then run user provided pass list by default. Ignore user provided pass list only if auto opt config is also provided and auto opt is not disabled explicitly.

> if pass list is not provided and run auto opt unless auto opt is explicitly disabled in the auto opt config.

> If user only wants to evaluate input model then he should not provide pass list and ( disable auto opt or not provide auto opt config)

docs/source/tutorials/configure_auto_optimizer.rst
```diff
@@ -0,0 +1,263 @@
...
+
+With the help of Auto Optimizer, for most popular models like bert, deberta and etc., user:
+
+1. DOES NOT need to provide the passes combination manually.
+2. DOES NOT need to write redundant passes configs.
+
+At the same time, user still has the flexibility to specify the accelerator, execution providers and expected precision of output models.
+
+Currently, **auto optimizer supports onnx optimization stack only**. For openvino and snpe, user still need to provide passes combination manually by now.
```
xiaoyu-work, line 27: Nit: we may need to clarify that openvino is specifically for OpenVINO IR model. ONNXRuntime still have OpenVINO EP (but we don't add this EP to auto optimizer?)

xiaoyu-work:
Do we have any plan to support multi EP with auto optimizer?

xiaoyu-work (dismissed):

olive/auto_optimizer/template_mapping.py
```diff
@@ -0,0 +1,85 @@
...
+
+    available_passes_tree = get_available_passes_by_opt_level(opt_level)
+    passes_cap = get_pass_capability()
+    pass_flows = []
+
+    # given available_passes_tree is [a] -> [b, c] -> [d, e, f], generate all possible pass flows
+    # [a, b, d], [a, b, e], [a, b, f], [a, c, d], [a, c, e], [a, c, f]
+    # as we need to step over some intermediate passes, we cannot use len(pass_flow_candidate) to
+    # indicate the current pass level, instead, we use the length of available_passes_tree to indicate
+
```
xiaoyu-work, line 43: Nit: extra blank line

olive/auto_optimizer/template_mapping.py
```diff
@@ -0,0 +1,85 @@
...
+    # item in pass tack is (pass, depth, pass_flow_candidate)
```
xiaoyu-work, line 44: Typo: pass stack?

docs/source/tutorials/configure_auto_optimizer.rst
devang-ml: We should add OpenVINO EP but it can be done as a follow up PR.

docs/source/tutorials/configure_auto_optimizer.rst
```diff
@@ -0,0 +1,263 @@
+.. _how_to_configure_auto_optimizer:
+
+[BETA] How To Configure Auto Optimizer
+=======================================
+
```
devang-ml, line 5: No need to have "[BETA]" prefix.

docs/source/tutorials/configure_auto_optimizer.rst
```diff
@@ -0,0 +1,263 @@
...
+4. target evaluation metric: accuracy, latency and etc.
+
+All above information(is called "optimization factors" in this doc) is provided by user through a configuration file now, then run by:
+
+.. code-block::
+
+    python -m olive.workflows.run --config <config_file>.json
+
+
+With the help of Auto Optimizer, for most popular models like bert, deberta and etc., user:
```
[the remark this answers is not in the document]
devang-ml: We can just drop the clarification ".... for most popular models ..."

olive/auto_optimizer/config_template/pass_capability.yaml
```diff
@@ -0,0 +1,55 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
```
devang-ml, line 5: Please add a comment describing how this yaml is used and what needs to be done to list a new pass here.

devang-ml (commented):
After the nitpicks and comments in the code are updated, this looks good to me. Thanks!

docs/source/tutorials/configure_auto_optimizer.rst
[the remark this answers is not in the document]
trajepl: Removed the confused description. Also we target auto opt is used for unknow models and the models which Olive has no customized configs. 
For the `single component model` you mentioned, I think auto opt should work fine for multi-component models after the changes from Hitesh. That change may integrate optimum stack into existing onnx conversion pass and merge multi-components into one. We can wait that then enable it in auto opt.
Thanks!

olive/auto_optimizer/template_mapping.py
trajepl: Removed.

docs/source/tutorials/configure_auto_optimizer.rst
trajepl: Removed.

olive/auto_optimizer/config_template/pass_capability.yaml
trajepl: Added.

trajepl:
> Do we have any plan to support multi EP with auto optimizer?

Thanks! We really have that in planning. From my understanding, the multi-eps should work if we enhance the pass capability properly unless there are potential bugs for mult-eps.  But I still have a few questions for how to distribute multi-eps in different target device and collect the results back.
Let us do that in next step.

docs/source/tutorials/configure_auto_optimizer.rst
trajepl: Updated to tell the diff between openvino ep and raw openvino IR.

xiaoyu-work (approved):

PR #785, opened 2023-12-06
Rename HFModelLoadingArgs to HFFromPretrainedArgs

xiaoyu-work:
## Describe your changes

Rename `HFModelLoadingArgs` to `HFFromPretrainedArgs`.

###
User interface change:
In model config, the `model_loading_args` should be `from_pretrained_args` instead:
```
"input_model":{
    "type": "PyTorchModel",
    "config": {
        "hf_config": {
            "model_name": "microsoft/phi-1_5",
            "task": "text-generation",
            "model_loading_args" -> "from_pretrained_args": {
                "trust_remote_code": true
            }
        }
    }
}
```

## Checklist before requesting a review
- [ ] Add unit tests for this change.
- [ ] Make sure all tests can pass.
- [ ] Update documents if necessary.
- [ ] Lint and apply fixes to your code by running `lintrunner -a`
- [x] Is this a user-facing change? If yes, give a description of this change to be included in the release notes.

PR #805, opened 2023-12-11
Remove some hardcoded use cases and organize hf_config/hf_utils

shaahji:
## Describe your changes

Remove some hardcoded use cases and organize hf_config/hf_utils
* Move static function in HfConfigMixin to ht_utils
* Add new function to get the hf model type
* Removed hardcoded reference to setup/restore Llama layers

### DIFF
--- a/olive/model/config/hf_config.py
+++ b/olive/model/config/hf_config.py
@@ -134,7 +134,8 @@ class HfFromPretrainedArgs(ConfigWithExtraArgs):
             )
             return v
 
-    def get_loading_args(self):
+    def get_loading_args(self) -> Dict[str, Any]:
+        """Return all args in a dict with types expected by `from_pretrained`."""
         loading_args = {}
         # copy args that can be directly copied
         direct_copy_args = ["device_map", "max_memory"]
@@ -156,24 +157,26 @@ class HfFromPretrainedArgs(ConfigWithExtraArgs):
         return loading_args
 
     def get_torch_dtype(self):
+        """Return the torch dtype to load the model under. It is either a torch.dtype or 'auto'."""
         v = self.torch_dtype
         if isinstance(v, str) and v != "auto":
             v = resolve_torch_dtype(v)
         return v
 
     def get_quantization_config(self):
+        """Return the quantization config to use. It is either None or a config class."""
         if not self.quantization_method or not self.quantization_config:
             return None
         return self.dict_to_quantization_config(self.quantization_method, self.quantization_config)
 
     @staticmethod
-    def device_to_str(device):
+    def device_to_str(device) -> str:
         if isinstance(device, torch.device):
             device = str(device)
         return device
 
     @staticmethod
-    def dict_to_quantization_config(quantization_method, config_dict):
+    def dict_to_quantization_config(quantization_method: str, config_dict: Dict[str, Any]):
         method_to_class_name = {"bitsandbytes": "BitsAndBytesConfig", "gptq": "GPTQConfig"}
         method_to_min_version = {
             # bitsandbytes exists from 4.27.0, but 4bit quantization is only supported from 4.30.0
@@ -245,3 +248,13 @@ class HfConfig(ConfigBase):
             if not v and not values.get("task", None):
                 raise ValueError("Either task or model_class must be specified")
         return v
+
+    def get_loading_args_from_pretrained(self) -> Dict[str, Any]:
+        """Return all args from from_pretrained_args in a dict with types expected by `from_pretrained`."""
+        return self.from_pretrained_args.get_loading_args() if self.from_pretrained_args else {}
+
+
+def get_model_type_from_hf_config(hf_config: HfConfig) -> str:
+    from olive.model.utils.hf_utils import get_hf_model_config
+
+    return get_hf_model_config(hf_config.model_name, hf_config.get_loading_args_from_pretrained()).model_type
--- a/olive/model/handler/mixin/hf_config.py
+++ b/olive/model/handler/mixin/hf_config.py
@@ -13,8 +13,8 @@ from olive.model.utils.hf_utils import (
     get_hf_model_config,
     get_hf_model_dummy_input,
     get_hf_model_io_config,
-    load_huggingface_model_from_model_class,
-    load_huggingface_model_from_task,
+    load_hf_model_from_model_class,
+    load_hf_model_from_task,
 )
 
 if TYPE_CHECKING:
@@ -24,7 +24,7 @@ logger = logging.getLogger(__name__)
 
 
 class HfConfigMixin:
-    """Provide the following Hugging Face model functionalites.
+    """Provide the following Hugging Face model functionalities.
 
         * loading huggingface model
         * getting huggingface model config
@@ -45,7 +45,7 @@ class HfConfigMixin:
         if self.hf_config is None:
             raise ValueError("HF model_config is not available")
 
-        return get_hf_model_config(self._get_model_path_or_name(), **self._get_loading_args())
+        return get_hf_model_config(self._get_model_path_or_name(), **self.hf_config.get_loading_args_from_pretrained())
 
     def get_hf_io_config(self):
         """Get Io config for the model."""
@@ -54,7 +54,7 @@ class HfConfigMixin:
                 self._get_model_path_or_name(),
                 self.hf_config.task,
                 self.hf_config.feature,
-                **self._get_loading_args(),
+                **self.hf_config.get_loading_args_from_pretrained(),
             )
         else:
             return None
@@ -94,14 +94,12 @@ class HfConfigMixin:
     def load_hf_model(self, model_path: str = None):
         """Load model from model_path or model_name."""
         model_name_or_path = model_path or self.hf_config.model_name
-        loading_args = self._get_loading_args()
+        loading_args = self.hf_config.get_loading_args_from_pretrained()
         logger.info(f"Loading Huggingface model from {model_name_or_path}")
         if self.hf_config.task:
-            model = load_huggingface_model_from_task(self.hf_config.task, model_name_or_path, **loading_args)
+            model = load_hf_model_from_task(self.hf_config.task, model_name_or_path, **loading_args)
         elif self.hf_config.model_class:
-            model = load_huggingface_model_from_model_class(
-                self.hf_config.model_class, model_name_or_path, **loading_args
-            )
+            model = load_hf_model_from_model_class(self.hf_config.model_class, model_name_or_path, **loading_args)
         else:
             raise ValueError("Either task or model_class must be specified")
 
@@ -113,7 +111,7 @@ class HfConfigMixin:
             self.model_path or self.hf_config.model_name,
             self.hf_config.task,
             self.hf_config.feature,
-            **self._get_loading_args(),
+            **self.hf_config.get_loading_args_from_pretrained(),
         )
 
     def is_model_loaded_from_hf_config(self) -> bool:
@@ -127,9 +125,6 @@ class HfConfigMixin:
             and self.hf_config
             and (self.hf_config.model_class or self.hf_config.task)
         )
-
-    def _get_loading_args(self):
-        return self.hf_config.from_pretrained_args.get_loading_args() if self.hf_config.from_pretrained_args else {}
 
     def _get_model_path_or_name(self):
         if self.model_file_format == ModelFileFormat.PYTORCH_MLFLOW_MODEL:
--- a/olive/model/utils/hf_utils.py
+++ b/olive/model/utils/hf_utils.py
@@ -15,7 +15,7 @@ from olive.model.utils.hf_mappings import FEATURE_TO_PEFT_TASK_TYPE, MODELS_TO_M
 logger = logging.getLogger(__name__)
 
 
-def load_huggingface_model_from_task(task: str, name: str, **kwargs):
+def load_hf_model_from_task(task: str, name: str, **kwargs):
     """Load huggingface model from task and name."""
     from transformers.pipelines import check_task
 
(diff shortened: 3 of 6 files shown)


olive/model/handler/mixin/hf_config.py
```diff
@@ -45,7 +44,7 @@ def get_hf_model_config(self):
         if self.hf_config is None:
             raise ValueError("HF model_config is not available")
 
-        return get_hf_model_config(self._get_model_path_or_name(), **HfConfigMixin._get_loading_args(self.hf_config))
+        return get_hf_model_config(self._get_model_path_or_name(), **get_loading_args(self.hf_config))
```
jambayk, line 47: The static methods are not needed and after talking with @guotuofeng, I already have dev branch that removed these changes https://github.com/microsoft/Olive/commit/b4c2700f0b3a3ca84768f1196a08b468c827a47a 

Just working on updating the unit tests to take care of this.

jambayk: PR with the above changes https://github.com/microsoft/Olive/pull/807

shaahji: I haven't removed them. Just moved them to utils. Could you push on top of this PR I.e. once this gets merged.

shaahji: Code rebased to include changes in PR #807

olive/model/handler/mixin/hf_config.py
```diff
@@ -91,29 +90,13 @@ def model_loader(_, model_component=model_component):
                 model_attributes=self.model_attributes,
             )
 
-    def load_hf_model(self, model_path: str = None):
```
jambayk, line 94: I think moving `get_loading_args` to `hf_utils` make sense but it is better to keep `load_hf_model` as a method here. 
It is only used by the pytorch model handler and is consistent with the other methods here. 

The conversion pass also doesn't use it anymore so a utils is unnecessary.

shaahji: Moved.

jambayk (approved):
Thanks!

olive/model/utils/hf_utils.py
```diff
@@ -10,12 +10,13 @@
 import transformers
 from transformers import AutoConfig, AutoModel, AutoTokenizer
 
+from olive.model.config.hf_config import HfConfig
```
guotuofeng, line 13: I prefer not import HfConfig to the utils. 

Basically, the hf_utils should not depends on HfConfig. Instead, it should only depends on transformers.

olive/model/utils/hf_utils.py
```diff
@@ -188,3 +189,11 @@ def get_model_max_length(model_name: str, fail_on_not_found=False) -> int:
             else:
                 logger.warning(not_found_msg)
                 return None
+
+
+def get_loading_args(hf_config: HfConfig):
```
guotuofeng, line 194: we need move this function into HfConfigMixin

olive/model/utils/hf_utils.py
```diff
@@ -188,3 +189,11 @@ def get_model_max_length(model_name: str, fail_on_not_found=False) -> int:
...
+    return hf_config.from_pretrained_args.get_loading_args() if hf_config.from_pretrained_args else {}
+
+
+def get_model_type(hf_config: HfConfig) -> str:
```
guotuofeng, line 198: we need decouple this function with HfConfig by:
1. use model_name,
2. use from_pretrained_args fetched from HfConfig in conversion

guotuofeng (changes requested):
Basically, the hf_utils need decouple with HfConfig

olive/model/utils/hf_utils.py
guotuofeng: something like:
```python
def get_model_type(model_name, loading_args) -> str:
```

shaahji: Doesn't that mean every site that calls this function will have to get the model_name and explicitly call to get loading args merely to pass it down to this function. Seems like a lot of unnecessary code duplication.

olive/model/utils/hf_utils.py
shaahji: The function will have to be static in a mixin. Mixins are meant to be subclassed and using it merely for assessing static functions in other files seems like bad use case. If the use case was only in the subclasses, I would agree but that's not the case here. This function does belong in utils.

guotuofeng: I'd like the HfConfig decouple with hf_utils. From the name, the HfConfig is concept of model. On another hand, hf_utils is collected utils function invoking the transformer package.

If we'd like to make these functions as common functions, my suggestion is to create another util above both HfConfig and existing hf_utils. 

Another candidate is we could move the get_loading_args to HfConfig class since it has not any dependencies to transformers. Then we only need define one function 
```
def get_model_type(model_name, loading_args) -> str:
```

How do you think? @jambayk

shaahji: Fair enough. Pulled the relevant functions into `hfconfig_utils.py`.

olive/model/utils/hfconfig_utils.py
```diff
@@ -0,0 +1,14 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+from olive.model.config.hf_config import HfConfig
+from olive.model.utils.hf_utils import get_hf_model_config
+
+
+def get_loading_args(hf_config: HfConfig):
```
jambayk, line 9: Like @guotuofeng suggested in an earlier comment, I think we can just make this directly a method of `HfConfig` like it used to be before the refactor https://github.com/microsoft/Olive/blob/e337a01f0e113ed063a2839337dd4ea54ce6840e/olive/model/hf_utils.py#L221

This way, `get_model_type` can be like any other `hf_utils` helper function like:
```
def get_model_type(model_name: str, **kwargs):
    ...
```

In the conversion pass, since we already have the hf_config object, it is not complicated calling something like
```
get_model_type(hf_config.model_name, **hf_config.get_loading_args)
```

jambayk: also might be better to rename `get_loading_args -> get_from_pretrained_args` since `loading_args` is an artifact from before this renaming PR https://github.com/microsoft/Olive/pull/785

jambayk: Even if you would like to keep the current signature of `get_model_type`, I think it is better to be directly in `hf_config.py` as a helper function named like `get_model_type_from_hf_config`. We already have something similar in https://github.com/microsoft/Olive/blob/ab0af2333e17e06d2245640f1759d1b7eed955f5/olive/model/config/io_config.py#L84. 

Putting it under `model.utils.hfconfig_utils` makes the hierarchy of the utils module confusing. Ideally, I think the utils should be at a lower level than the configs or handlers. So, they won't have any knowledge of the hfconfig class.

shaahji: Done. Removed hfconfig_utils.py & moved the functions in hf_config.py

olive/model/config/hf_config.py
```diff
@@ -245,3 +246,11 @@ def task_or_model_class_required(cls, v, values):
...
                 raise ValueError("Either task or model_class must be specified")
         return v
+
+
+def get_loading_args_from_pretrained(hf_config: HfConfig) -> Dict[str, Any]:
+    return hf_config.from_pretrained_args.get_loading_args() if hf_config.from_pretrained_args else {}
+
+
+def get_model_type(hf_config: HfConfig) -> str:
+    return get_hf_model_config(hf_config.model_name, **get_loading_args_from_pretrained(hf_config)).model_type
```
guotuofeng, line 256: I suppose @jambayk means to move this function into hf_utils.py with the following signature. 

def get_model_type(model_name: str, **kwargs):

olive/model/config/hf_config.py
```diff
@@ -245,3 +246,11 @@ def task_or_model_class_required(cls, v, values):
             if not v and not values.get("task", None):
                 raise ValueError("Either task or model_class must be specified")
         return v
+
+
+def get_loading_args_from_pretrained(hf_config: HfConfig) -> Dict[str, Any]:
```
guotuofeng, line 251: >get_loading_args_from_pretrained [](http://example.com/codeflow?start=4&length=32)

better to make it be method of HfConfig class.

shaahji: Both functions members of `HFConfig` class?

guotuofeng: just  get_from_pretrained_args in HfConfig is enough.

olive/model/config/hf_config.py
guotuofeng: anyway, you can keep it as is. it is not big deal.

olive/model/config/hf_config.py
shaahji:
```
class HfConfig:
....
    def get_loading_args_from_pretrained(self) -> Dict[str, Any]:
        return self.from_pretrained_args.get_loading_args() if self.from_pretrained_args else {}

    def get_model_type(self) -> str:
        return get_hf_model_config(self.model_name, **self.get_loading_args_from_pretrained()).model_type
```

olive/model/config/hf_config.py
guotuofeng: the major difference between this one with https://github.com/microsoft/Olive/blob/ab0af2333e17e06d2245640f1759d1b7eed955f5/olive/model/config/io_config.py#L84 @jambayk mentioned is:
the get_model_type need call transformers. that one @jambayk point to need not. 

That's the reason we suggest to move it into hf_utils.

olive/model/config/hf_config.py
guotuofeng:
```
class HfConfig:
     def get_from_pretrained_args(self) -> Dict[str, Any]:
        return self.from_pretrained_args.get_loading_args() if self.from_pretrained_args else {}
```

in hf_utils.py

```
def get_model_type(model_name: str, **kwargs):
    return get_hf_model_config(model_name, **kwargs).model_type
```

olive/model/config/hf_config.py
guotuofeng: the get_model_type only accept the model name and loading arguments, which need to be called in conversion.py
like:
```python
        hf_config = HfConfig(**model_config["hf_config"])
        model_type = get_model_type(hf_config.model_name, **hf_config.get_from_pretrained_args())
```

olive/model/config/hf_config.py
shaahji: Every calling site will have to duplicate code -

```
hf_config = HfConfig(**model_config["hf_config"])
model_type = get_model_type(hf_config.name, **hf_config.get_loading_args_from_pretrained())
```

Then why bother the function at all -

```
hf_config = HfConfig(**model_config["hf_config"])
model_type = get_hf_model_config(hf_config.name, **hf_config.get_loading_args_from_pretrained()).model_type
```

guotuofeng: I think it is ok since they are just call these functions/methods.

guotuofeng: by using the explicitly passing the name and args, we could decouple dependency of HfConfig and get_model_type.

shaahji: To avoid the circular dependency, HfConfig should be treated as a data object and shouldn't have any `smart` functions. All _logic_ based functions should be part of hf_utils which was my initial implementation. Adding functions in both hf_config and hf_utils is more likely to create circular dependency.

I think the restriction of not using HtConfig in hf_utils is limited. That should be relaxed, in fact, that should be encouraged.

The proposed `kwargs` makes code browsing hard since it doesn't explicitly say what these args are? Given that these args are actually *required* making them variable kwargs seems wrong.

shaahji: The decoupling is coming at the cost of code readability. My initial approach of adding the functions in hf_utils was better, in my opinion, since function arguments were explicit and hf_utils was above the `HfConfig` data class.

jambayk: What about this as a compromise https://github.com/microsoft/Olive/commit/cf60a73c9f2c0b8d769b604f3d0f04f839d9ca72? 

This way `olive.model.utils` is always the lowest level submodule under `olive.model` and we can avoid the circular import. utils should not import from eithe handler or config. That's how I personally arrange the hierarchy of utils submodules. 

We can also make this change later if you would like to merge your PR as is first.

guotuofeng: > The decoupling is coming at the cost of code readability. My initial approach of adding the functions in hf_utils was better, in my opinion, since function arguments were explicit and hf_utils was above the `HfConfig` data class.

Yes, you are correct. the cost is the code readabilities. Anyway, it is an open discussion. and I am ok to your changes and it is not a big issue if we import hf_utils in HfConfig.

olive/model/config/hf_config.py
shaahji: Resolving with introducing hf_utils dependency in hf_config

olive/model/config/hf_config.py
shaahji: Resolving with introducing hf_utils dependency in hf_config

olive/model/config/hf_config.py
```diff
@@ -245,3 +248,13 @@ def task_or_model_class_required(cls, v, values):
             if not v and not values.get("task", None):
                 raise ValueError("Either task or model_class must be specified")
         return v
+
+    def get_loading_args_from_pretrained(self) -> Dict[str, Any]:
```
guotuofeng, line 252: >get_loading_args_from_pretrained [](http://example.com/codeflow?start=8&length=32)

nit: will get_from_pretrained_args better?

guotuofeng (approved):

olive/model/config/hf_config.py
jambayk: Thanks!

PR #819, opened 2023-12-15
Update openvino api to 2023.2.0

xiaoyu-work:
## Describe your changes

Update openvino api to 2023.2.0


olive/passes/pass_config.py
```diff
@@ -38,6 +40,7 @@ class PassConfigParam(ConfigParam):
     """
 
     searchable_values: SearchParameter = None
+    allowed_values: Set = None
```
guotuofeng, line 43: what's the field for?

guotuofeng: I think it is a user-facing change?

xiaoyu-work: This is used to validate param value. It's not a user facing change as this validation is internal only.

guotuofeng: PassConfigParam is used as user-facing stuff.

olive/passes/openvino/quantization.py
```diff
@@ -65,39 +82,81 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
                 type_=Union[DataConfig, Dict],
                 description="Data config for calibration, required if dataloader_func is None.",
             ),
-            "metric_func": PassConfigParam(
-                type_=Union[Callable, str],
+            "model_type": PassConfigParam(
+                type_=str,
```
guotuofeng, line 86: should enum be better than allowed_values?

olive/passes/pass_config.py
xiaoyu-work: I don't think so. `PassConfigParam` is defined by developer to indicate the parameter's attributes. As for users, they only need to provide a dict.

olive/passes/openvino/quantization.py
xiaoyu-work: Do you mean defining allowed values as enum? `allowed_values` set is simpler than an enum with additional definition and validation.

olive/passes/openvino/conversion.py
```diff
@@ -19,31 +19,47 @@ class OpenVINOConversion(Pass):
...
-                    "Input can be set by passing a list of tuples. "
-                    "Each tuple should contain input name and optionally input type or input shape."
+                    "Set or override shapes for model inputs. "
+                    "It configures dynamic and static dimensions in model inputs "
+                    "depending on your inference requirements."
                 ),
             ),
-            "input_shape": PassConfigParam(
-                type_=List[int],
+            "example_input": PassConfigParam(
```
jambayk, line 30: since torch tensor or other inputs cannot be serialized, should this be made into a function or string name paramater instead. like for `dummy_input_func` or `dataloader_func`

olive/passes/openvino/conversion.py
```diff
@@ -19,31 +19,47 @@ class OpenVINOConversion(Pass):
...
+                    "https://docs.openvino.ai/2023.2/openvino_docs_OV_Converter_UG_Conversion_Options.html"
+                ),
+            ),
+            "output_model ": PassConfigParam(
+                type_=str,
+                default_value="ov_model",
+                required=False,
+                description=(
+                    "Name of the output OpenVINO model. "
+                    "The output model will be saved to the same directory as the input model."
```
jambayk, line 62: does look like this is the case based on the code? 

Also maybe a name like `output_file_name` or something would be clearer?

olive/passes/pass_config.py
```diff
@@ -4,7 +4,9 @@
 # --------------------------------------------------------------------------
 from enum import Enum
 from pathlib import Path
-from typing import Callable, Dict, Optional, Type, Union
+from typing import Callable, Dict, Optional, Set, Type, Union
+
+from pydantic import Field
```
jambayk, line 9: this needs to be imported from `olive.common.pydantic_v1` like in line 12 so that we get the correct `Field` from v1 api.

olive/passes/openvino/conversion.py
xiaoyu-work: This is `output_model` name is same as OpenVINO api parameter name: https://docs.openvino.ai/2023.2/openvino_docs_OV_Converter_UG_Conversion_Options.html, to keep the api parameter names unified. But I forgot to update the description here. Will update it.

olive/passes/openvino/quantization.py
guotuofeng: with enum you can only need definition. @jambayk, what's your thought?

guotuofeng: https://docs.pydantic.dev/2.0/usage/types/enums/

xiaoyu-work: Ah not aware that pydantic supports Enums here. This looks better than manual validation. Will update with Enum.

xiaoyu-work:
There is a bug for openvino example. I removed it from pipeline for now to unblock this change. Will add it back once the bug is fixed.

olive/passes/pass_config.py
```diff
@@ -38,6 +38,7 @@ class PassConfigParam(ConfigParam):
     """
 
     searchable_values: SearchParameter = None
+    allowed_values: Set = None
```
guotuofeng, line 41: it seems this field could be removed?

.azure_pipelines/olive-ci.yaml
```diff
@@ -119,9 +116,6 @@ jobs:
       bert_ptq_cpu_aml:
         exampleFolder: bert
         exampleName: bert_ptq_cpu_aml
-      cifar10_openvino_intel_hw:
-        exampleFolder: cifar10_openvino_intel_hw
-        exampleName: cifar10_openvino_intel_hw
```
guotuofeng, line 124: do we need enable the new test?

olive/evaluator/olive_evaluator.py
```diff
@@ -823,8 +822,7 @@ def _inference(
         device: Device = Device.CPU,
         execution_providers: Union[str, List[str]] = None,
     ) -> Tuple[OliveModelOutput, Any]:
-        session = model.prepare_session(inference_settings=self.get_inference_settings(metric), device=device)
-
+        session = model.prepare_session()
```
guotuofeng, line 825: >prepare_session [](http://example.com/codeflow?start=24&length=15)

why do we remove the device from the prepare_session?

olive/passes/pass_config.py
xiaoyu-work: Good catch. Will remove it.

olive/evaluator/olive_evaluator.py
xiaoyu-work: We enable session device here: https://github.com/microsoft/Olive/blob/main/olive/evaluator/olive_evaluator.py#L833. We could refactor this to move this to `prepare_session`

olive/evaluator/olive_evaluator.py
```diff
@@ -1017,14 +1015,13 @@ def _inference(
...
+        for input_data, label in dataloader:
+            session.infer({0: input_data})
+            result = session.get_output_tensor(0).data
             outputs = post_func(result) if post_func else result
-            if not isinstance(labels, list):
-                labels = [labels]  # noqa: PLW2901
-            preds.extend(outputs)
-            targets.extend(labels)
-            logits.extend(result)
+            preds.append(outputs)
```
jambayk, line 1022: is there a reason this is `append` and not `extend`. Is the data not in batches for openvino? Just one sample per iteration of the loop?

olive/passes/openvino/conversion.py
```diff
@@ -15,37 +15,52 @@
...
+                    "For PyTorch it can be torch.Tensor. "
+                    "For Tensorflow it can be tf.Tensor or numpy.ndarray. "
                 ),
             ),
-            "extra_config": PassConfigParam(
+            "compress_to_fp16": PassConfigParam(
+                type_=bool,
+                default_value=True,
+                required=False,
+                description=("Compress weights in output OpenVINO model to FP16. Default is True."),
```
jambayk, line 45: nit: the parenthesis here are redundant?

olive/passes/openvino/conversion.py
```diff
@@ -55,38 +70,51 @@ def _run_for_config(
...
-            model_input_param = "saved_model_dir"
         if model.framework == Framework.PYTORCH:
             input_model = model.load_model()
 
-        if config.get("extra_config") is None:
-            config["extra_config"] = {}
+        example_input = None
+        if config.get("example_input_func"):
+            example_input_func = config["example_input_func"]
+            if isinstance(example_input_func, str):
```
jambayk, line 88: nit: There is no need for the condition here. We can always directly use `example_input = self._user_module_loader.call_object(example_input_func)`. The `UserModuleLoader` object can call a function directly too https://github.com/microsoft/Olive/blob/main/olive/common/user_module_loader.py#L26

olive/passes/openvino/conversion.py
```diff
@@ -15,37 +15,52 @@
...
     """Converts PyTorch, ONNX or TensorFlow Model to OpenVino Model."""
 
+    _requires_user_script = True
+
     @staticmethod
     def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
         return {
             "input": PassConfigParam(
-                type_=List[Tuple],
+                type_=Union[Callable, str, List],
```
jambayk, line 24: not an issue but should we also make this type only as function or function name like we do in other places in Olive. 
The `List` case is only useful when someone want to use the python api but we don't have any use case for that except for the unit tests.

olive/evaluator/olive_evaluator.py
xiaoyu-work: Ah good catch! I made this change to test torchmetrics accuracy api feature but forgot to change it back. Will update this.

olive/passes/openvino/conversion.py
xiaoyu-work: Will do

olive/passes/openvino/conversion.py
xiaoyu-work: This is useful for those simpler model like our current example (even I removed this): https://github.com/microsoft/Olive/blob/main/examples/cifar10_openvino_intel_hw/config.json#L55

jambayk: Oh so, we can still provide the shapes as input?

xiaoyu-work: Yes, this is similar as onnx conversion, either input_shape (input) or example_input is needed.

test/unit_test/passes/openvino/test_openvino_conversion.py
```diff
@@ -13,7 +14,7 @@ def test_openvino_conversion_pass(tmp_path):
     # setup
     input_model = get_pytorch_model()
     dummy_input = get_pytorch_model_dummy_input(input_model)
-    openvino_conversion_config = {"extra_config": {"example_input": dummy_input}}
+    openvino_conversion_config = {"example_input": dummy_input}
```
jambayk, line 17: this got changed to `example_input_func`?

olive/passes/openvino/conversion.py
jambayk: My bad, this is still for input shapes. The name `input` was confusing but I guess, that's what the openvino api calls it too.

xiaoyu-work: Yeah this is how openvino api calls it. This is so confusing.

olive/passes/openvino/conversion.py
```diff
@@ -55,38 +70,45 @@ def _run_for_config(
...
-        if model.framework == Framework.TENSORFLOW and Path(model.model_path).is_dir():
-            model_input_param = "saved_model_dir"
         if model.framework == Framework.PYTORCH:
             input_model = model.load_model()
 
-        if config.get("extra_config") is None:
-            config["extra_config"] = {}
+        example_input = None
+        if config.get("example_input_func"):
+            example_input = self._user_module_loader.call_object(config["example_input_func"])
```
jambayk, line 87: Could it be better if we called the `example_input_func` with the `input_model` object? Like how we do for the pytorch dummy input func. That way the input func can choose to use information from the input model like `model_attributed` if it wants. Helps when we don't want to hard code values into the user scripts to make it more flexible.

jambayk: in the ut, we could then directly use `get_pytorch_model_dummy_input` as the `example_input_func` directly.

xiaoyu-work: I don't see any use case of this. Any example for this case?

jambayk: yeah 
https://github.com/microsoft/Olive/blob/main/examples/llama2/user_script.py#L22
https://github.com/microsoft/Olive/blob/main/examples/whisper/code/user_script.py#L145

xiaoyu-work: (Not remember if our team had such discussion before) I still don't want to encourage user to use OliveModelHandler in the user script (or any Olive internal implementation), as they have to have pre-knowledge of how OliveModelHandler looks like and how it works. For the llama2 example and whisper example, I can see it is possible to avoid using PyTorchModelHandler but those codes can still work after refactoring.

xiaoyu-work: If we do want to use some metadata from OliveModel, I prefer to specially expose a attribute like "metadata" or "extra_data" to user. So they don't need to read our source code of model handler but they can know they can get some data from this attribute from Olive model.

jambayk: Sure, we can discuss about it separately and just have this pass keep its current signature. 

The `model_attributes` are mean to function similar to the "metadata" you mention. We could also do something like allow the user to provide `func_kwargs` like we do for metrics https://github.com/microsoft/Olive/blob/bc8b3312251484986972cf53ac8d4b2f4ed2ca1f/examples/notebooks/llama2/config.json#L91

test/unit_test/passes/openvino/test_openvino_conversion.py
```diff
@@ -13,7 +14,7 @@ def test_openvino_conversion_pass(tmp_path):
     # setup
     input_model = get_pytorch_model()
     dummy_input = get_pytorch_model_dummy_input(input_model)
-    openvino_conversion_config = {"extra_config": {"example_input": dummy_input}}
+    openvino_conversion_config = {"example_input_func": dummy_input}
```
jambayk, line 17: This will fail because it's not a function.

olive/passes/openvino/conversion.py
xiaoyu-work: Yeah, we can especially expose some model_attributes for all our models.

jambayk (approved):

guotuofeng (approved):

PR #820, opened 2023-12-15
Fix perftuning to meet DLIS integration requirements

guotuofeng:
## Describe your changes

* Change the execution provider priority to use the parameters instead of that one in inference_settings. 
    Bug description: In DLIS scenarios, the inferenece_settings is something like:
{'execution_provider': [('MIGraphXExecutionProvider', {})]} and the provider_lists specified in perf_tuning is: ['ROCMExecutionProvider', 'MIGraphXExecutionProvider'].

    Originally, the inference_settings priority is higher than the execution_providers specified in arguments. As the result, when calling OnnxEvaluator.evaluate with the provider_list in baseline evaluation, the execution provider will be used as ["MIGraphXExecutionProvider"]. Later, the OnnxEvaluator.disable_ort_fallback(session, execution_providers) will check whether the underlying session's providers(MIGraphXExecutionProvider, CPUExecutionProvider) contains the execution_providers(ROCMExecutionProvider, MIGraphXExecutionProvider). In this case, the ROCMExecutionProvider is excluded when creating inference session. and the OliveEvaluationError exception will be raised.

    To fix this issue, we need move the check_ort_fallback into prepare_session in onnx model.
    perf tuning will always pass None as the execution provider to evaluate to make the logic clear

* Change perf tuning to always has execution provider info regardless of whether the best result is baseline or from perf tuning.
* Change perf tuning to accept both provider list and provider list with provider options.
* Add force_evaluate_other_eps to force run all other EPs if the current EP is not same with the additional EPs
* Change OnnxModel.prepare_session to accept both execution provider list and (execution provider, provider options) list
In this way, the EP/provider option handling logic is normalized.
* Add MIGraphXExecutionProvider to the accelerator map
* Fix the critical bug when creating inference session that doesn't respect provider options in ort_inference.py
* Add io_bind inference for accuracy since illegal memory access in cuda will be thrown if enable_cuda_graph. 
  If io_bind is not enabled and enable_cuda_graph is True, the following errors will be raised:

    onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : CUDA failure 700: an illegal memory access was encountered ; GPU=0 ; hostname=c93f1847c000000 ; file=/onnxruntime_src/onnxruntime/core/providers/cuda/cuda_graph.cc ; line=49 ; expr=cudaGraphLaunch(graph_exec_, stream_); 
    [E:onnxruntime:Default, cuda_call.cc:116 CudaCall] CUDA failure 700: an illegal memory access was encountered ; GPU=0 ; hostname=c93f1847c000000 ; file=/onnxruntime_src/onnxruntime/core/providers/cuda/cuda_execution_provider.cc ; line=286 ; expr=cudaStreamDestroy(stream_); 
    [E:onnxruntime:Default, cuda_call.cc:116 CudaCall] CUDNN failure 4: CUDNN_STATUS_INTERNAL_ERROR ; GPU=0 ; hostname=c93f1847c000000 ; file=/onnxruntime_src/onnxruntime/core/providers/cuda/cuda_execution_provider.cc ; line=181 ; expr=cudnnDestroy(cudnn_handle_); 

### DIFF
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -44,7 +44,7 @@ from olive.model import (
     SNPEModelHandler,
 )
 from olive.model.config.io_config import is_io_config_static
-from olive.model.utils.onnx_utils import CudaGraphHelper, prepare_io_bindings
+from olive.model.utils.onnx_utils import bind_input_data, bind_output_data, prepare_io_bindings
 from olive.snpe.data_loader import SNPECommonDataLoader, SNPEDataLoader
 
 logger = logging.getLogger(__name__)
@@ -440,19 +440,19 @@ class OnnxEvaluator(OliveEvaluator, framework=Framework.ONNX):
         io_bind = self.io_bind_enabled(metric, model.inference_settings)
         device = "cuda" if device == "gpu" else "cpu"
         if io_bind and device == "cuda":
-            input_and_output_shape = dict(zip(io_config["input_names"], io_config["input_shapes"]))
-            cuda_graph_helper = CudaGraphHelper(session, input_and_output_shape, device)
-            io_binding = cuda_graph_helper.io_binding
+            io_binding = session.io_binding()
 
         is_single_tensor_output = len(output_names) == 1
         for input_data, labels in dataloader:
             input_dict = OnnxEvaluator.format_input(input_data, io_config)
             if io_bind and device == "cuda":
-                cuda_graph_helper.update_inputs(input_dict)
+                use_fp16 = any(v.dtype == np.float16 for v in input_data.values())
+                bind_input_data(io_binding, input_dict, use_fp16, device)
+                bind_output_data(io_binding, session.get_outputs(), use_fp16, device)
                 io_binding.synchronize_inputs()
                 session.run_with_iobinding(io_binding)
                 io_binding.synchronize_outputs()
-                res = [cuda_graph_helper.get_output(output_name) for output_name in output_names]
+                res = [i.numpy() for i in io_binding.get_outputs()]
             else:
                 res = session.run(input_feed=input_dict, output_names=None)
             if is_single_tensor_output:
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -44,7 +44,6 @@ from olive.model import (
     SNPEModelHandler,
 )
 from olive.model.config.io_config import is_io_config_static
-from olive.model.utils.onnx_utils import check_ort_fallback
 from olive.snpe.data_loader import SNPECommonDataLoader, SNPEDataLoader
 
 logger = logging.getLogger(__name__)
@@ -468,8 +467,6 @@ class OnnxEvaluator(OliveEvaluator, framework=Framework.ONNX):
             execution_providers=execution_providers,
         )
 
-        check_ort_fallback(session, execution_providers)
-
         io_config = model.get_io_config()
 
         preds = []
@@ -531,8 +528,6 @@ class OnnxEvaluator(OliveEvaluator, framework=Framework.ONNX):
             device=device,
             execution_providers=execution_providers,
         )
-        check_ort_fallback(session, execution_providers)
-
         io_config = model.get_io_config()
 
         input_data, _ = next(iter(dataloader))
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -44,6 +44,7 @@ from olive.model import (
     SNPEModelHandler,
 )
 from olive.model.config.io_config import is_io_config_static
+from olive.model.utils.onnx_utils import bind_input_data, bind_output_data, prepare_io_bindings
 from olive.snpe.data_loader import SNPECommonDataLoader, SNPEDataLoader
 
 logger = logging.getLogger(__name__)
@@ -69,15 +70,27 @@ class OliveEvaluator(ABC):
     def __init__(self):
         pass
 
-    def get_inference_settings(self, metric: Metric) -> Dict[str, Any]:
+    @classmethod
+    def get_inference_settings(cls, metric: Metric) -> Dict[str, Any]:
         # user.config.inference_settings > model.inference_settings > default inference_settings
         # when user.config.inference_settings is None, the model.inference_settings
         # will be used in model.prepare_session(..)
         return (
-            metric.user_config.inference_settings.get(self.framework.lower())
+            metric.user_config.inference_settings.get(cls.framework.lower())
             if metric.user_config.inference_settings
             else None
         )
+
+    @classmethod
+    def io_bind_enabled(cls, metric: Metric) -> bool:
+        if metric.user_config.io_bind:
+            return True
+
+        inference_settings = cls.get_inference_settings(metric)
+        if inference_settings and inference_settings.get("io_bind"):
+            return True
+
+        return False
 
     @abstractmethod
     def _inference(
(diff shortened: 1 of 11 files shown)


olive/model/handler/onnx.py
```diff
@@ -74,28 +74,55 @@ def prepare_session(
...
+        else:
+            execution_providers = inference_settings.get("execution_provider")
+            provider_options = inference_settings.get("provider_options")
+
         if not execution_providers:
             execution_providers = self.get_default_execution_providers(device)
-        elif isinstance(execution_providers, str):
+            provider_options = None
+        elif isinstance(execution_providers, (str, tuple)):
             execution_providers = [execution_providers]
```
trajepl, line 108: Maybe more clear? 
```python

        if execution_providers:
            provider_options = None
        else:
            execution_providers = inference_settings.get("execution_provider")
            provider_options = inference_settings.get("provider_options")
            if not execution_providers:
                execution_providers = self.get_default_execution_providers(device)
                provider_options = None

        if isinstance(execution_providers, (str, tuple)):
            execution_providers = [execution_providers]


```
(BTW, can we assume that if the execution_providers is None from inference_settings, then `provider_options` must be None?)

olive/model/utils/onnx_utils.py
```diff
@@ -67,3 +71,111 @@ def get_onnx_file_path(model_path: str, onnx_file_name: Optional[str] = None) ->
...
     else:
         raise ValueError(f"No .onnx file found in the model folder {model_path}.")
+
+
+def check_and_normalize_provider_args(
+    providers: Sequence[Union[str, Tuple[str, Dict[Any, Any]]]],
+    provider_options: Sequence[Dict[Any, Any]],
+    available_provider_names: Sequence[str],
+):
+    """Validate the 'providers' and 'provider_options' arguments and returns a normalized version.
```
trajepl, line 81: Does `copy but not call from ort` because this function only exists after specific ort version?

test/unit_test/passes/onnx/test_perf_tuning.py
```diff
@@ -61,6 +65,113 @@ def test_ort_perf_tuning_with_customized_configs(mock_run, config):
...
+    if return_baseline:
+        assert len(acutal_eps) == 2
+        assert "enable_cuda_graph" not in acutal_eps[0][1]
+
+        assert acutal_eps[1] == "CPUExecutionProvider"
+        assert f"Best result({PERFTUNING_BASELINE}):" in caplog.text
+    else:
+        assert len(acutal_eps) == 1
+        if acutal_eps[0][0] == "CUDAExecutionProvider":
+            assert "enable_cuda_graph" in acutal_eps[0][1]
```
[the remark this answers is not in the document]
guotuofeng: I don't run evaluator. I just mock the evaluate.

test/unit_test/model/test_onnx_model.py
```diff
@@ -0,0 +1,37 @@
...
+        "session_options": {
+            "execution_mode": 0,
+            "graph_optimization_level": 99,
+            "extra_session_config": None,
+            "inter_op_num_threads": None,
+            "intra_op_num_threads": 1,
+        },
+    }
+
+    execution_providers = ["ROCMExecutionProvider", "MIGraphXExecutionProvider", "CUDAExecutionProvider"]
```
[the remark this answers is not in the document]
guotuofeng: It is just for UT

olive/model/utils/onnx_utils.py
guotuofeng: This function is not exposed. I copy here just avoid the signature changes breaking Olive in future.

olive/model/handler/onnx.py
guotuofeng: will do

trajepl (approved):

olive/model/handler/onnx.py
```diff
@@ -74,29 +74,58 @@ def prepare_session(
...
-            ]
+        # split the execution_providers and provider_options
+        execution_providers, provider_options = check_and_normalize_provider_args(
+            execution_providers, provider_options, ort.get_available_providers()
+        )
 
-        return get_ort_inference_session(self.model_path, inference_settings, self.use_ort_extensions)
+        if (device == Device.GPU) and (rank is not None):
+            for i, ep in enumerate(execution_providers):
+                if ep == "CUDAExecutionProvider" and not provider_options[i]:
```
jambayk, line 122: Wouldn't this condition return false if provider options is `{}`? Don't we want to provide the `device_id` even when it is `{}`?

jambayk: Actually, never mind. I misread the condition. This is true only when provider options is `{}`. So, we want to insert the device id if the user hasn't specified the provider options.

jambayk: maybe we could instead check for `provider_options.get(device_id) is not None` since for distributed models, we want to have device id always?

olive/model/handler/onnx.py
```diff
@@ -74,29 +74,58 @@ def prepare_session(
...
+        # Originally, the inference_settings priority is higher than the execution_providers specified in arguments.
+        # as the result, when calling OnnxEvaluator.evaluate with the provider_list in baseline evaluation,
+        # the execution provider will be used as ["MIGraphXExecutionProvider"].
+        # later, the check_ort_fallback(session, execution_providers) will check the
+        # whether the underlying session's providers(MIGraphXExecutionProvider, CPUExecutionProvider) contains the
+        # execution_providers(ROCMExecutionProvider, MIGraphXExecutionProvider). In this case, the
+        # ROCMExecutionProvider is excluded when creating inference session. and the OliveEvaluationError exception
+        # will be raised.
+        # To fix this issue, we need to make sure the execution_providers specified in arguments has higher priority
+        if execution_providers:
```
jambayk, line 98: Olive evaluator always provides execution provider https://github.com/microsoft/Olive/blob/3a774a75592cd4d12c5836dea19d6d462ccbaeac/olive/systems/local.py#L49 

So, wouldn't this ignore the perf tuned inference settings when evaluating the output model of the perf-tuning pass?

jambayk: Or it doesn't matter since only the `provider_options` is ignored in this case and we don't tune it. But what about settings like `trt_fp16_enable` and `enable_cuda_graph`? https://github.com/microsoft/Olive/blob/3a774a75592cd4d12c5836dea19d6d462ccbaeac/olive/passes/onnx/perf_tuning.py#L135 

Perhaps, it would be more complete if we resolve the `ep, provider_options` for both the user provided `ep` and model's inference settings. As use the inference setting's provider options for eps from user that don't have their own provider options?

olive/model/handler/onnx.py
guotuofeng: This is the original behavior
https://github.com/microsoft/Olive/blob/main/olive/model/handler/onnx.py#L94-L97

olive/model/handler/onnx.py
guotuofeng: fixed.

jambayk (approved):

jambayk (dismissed):

olive/evaluator/olive_evaluator.py
```diff
@@ -468,19 +430,53 @@ def _inference(
...
         preds = []
         targets = []
         logits = []
         logits_dict = collections.defaultdict(list)
         output_names = io_config["output_names"]
+        io_bind = self.io_bind_enabled(metric, model.inference_settings)
+        device = "cuda" if device == "gpu" else "cpu"
+        if io_bind and device == "cuda":
+            io_binding = session.io_binding()
+            input_data, _ = next(iter(dataloader))
```
trajepl, line 444: We may need to rewind the dataloader after `next(iter(dataloader))`.

Otherwise, the next `for input_data, labels in dataloader` will start from second one of the full data.

jambayk: I don't think that's needed? We expect the dataloader to be similar to torch dataloader. There is no concept of rewind like there is for the onnxruntime quantization calibration dataloader. 

For this dataloader, the expectation is that every new `iter(dataloader)` call on the dataloader starts the iteration from the start. 
The for loop is also an instance of a new `iter` over the dataloader.

trajepl: I am wrong.  No need to remind.

trajepl (approved):

PR #828, opened 2023-12-17
Implement ONNX decoder merge

shaahji:
## Describe your changes

Implement ONNX decoder merge

* Introducing logic to merge onnnx decoders
* OnnxConversion now returns a merged ONNX model conditionally.

### DIFF
--- a/olive/model/handler/pytorch.py
+++ b/olive/model/handler/pytorch.py
@@ -5,6 +5,7 @@
 import logging
 import os
 import tempfile
+from copy import deepcopy
 from pathlib import Path
 from typing import Any, Callable, ClassVar, Dict, List, Optional, Tuple, Union
 
@@ -16,7 +17,7 @@ from olive.common.user_module_loader import UserModuleLoader
 from olive.common.utils import copy_dir
 from olive.constants import Framework, ModelFileFormat
 from olive.hardware.accelerator import Device
-from olive.model.config import HfConfig, IoConfig
+from olive.model.config import HfComponent, HfConfig, IoConfig
 from olive.model.config.registry import model_handler_registry
 from olive.model.handler.base import OliveModelHandler
 from olive.model.handler.mixin import DummyInputsMixin, HfConfigMixin
@@ -152,6 +153,31 @@ class PyTorchModelHandler(OliveModelHandler, HfConfigMixin, DummyInputsMixin):
 
         return model
 
+    def get_component_model(self, component: HfComponent, rank: Optional[int] = None) -> "PyTorchModelHandler":
+        if component.component_func is None:
+            logger.debug("component_func is not provided, using hf_config to get component")
+            model_component = self.load_hf_model(self.model_path)
+        else:
+            user_module_loader = UserModuleLoader(self.model_script, self.script_dir)
+            model_component = user_module_loader.call_object(component.component_func, self)
+
+        # the second default parameter is to fix ruff b023:
+        # https://docs.astral.sh/ruff/rules/function-uses-loop-variable/
+        def model_loader(_, model_component=model_component):
+            return model_component
+
+        component_hf_config = deepcopy(self.hf_config).dict()
+        component_hf_config.pop("components", None)
+        return PyTorchModelHandler(
+            model_loader=model_loader,
+            io_config=component.io_config,
+            dummy_inputs_func=component.dummy_inputs_func,
+            model_script=self.model_script,
+            script_dir=self.script_dir,
+            hf_config=HfConfig.parse_obj(component_hf_config),
+            model_attributes=self.model_attributes,
+        )
+
     def prepare_session(
         self,
         inference_settings: Optional[Dict[str, Any]] = None,
@@ -251,7 +277,7 @@ class PyTorchModelHandler(OliveModelHandler, HfConfigMixin, DummyInputsMixin):
 
 
 @model_handler_registry("DistributedPyTorchModel")
-class DistributedPyTorchModelHandler(OliveModelHandler):
+class DistributedPyTorchModelHandler(OliveModelHandler, HfConfigMixin):
     resource_keys: Tuple[str, ...] = ("model_path", "script_dir", "model_script", "adapter_path")
     json_config_keys: Tuple[str, ...] = (
         "model_name_pattern",
@@ -295,7 +321,7 @@ class DistributedPyTorchModelHandler(OliveModelHandler):
             validate_config(io_config, IoConfig).dict() if isinstance(io_config, (IoConfig, dict)) else io_config
         )
         self.dummy_inputs_func = dummy_inputs_func
-        self.hf_config = hf_config
+        self.hf_config = validate_config(hf_config, HfConfig) if hf_config else None
 
     @property
     def script_dir(self) -> str:
@@ -325,6 +351,22 @@ class DistributedPyTorchModelHandler(OliveModelHandler):
             io_config=self.io_config,
             dummy_inputs_func=self.dummy_inputs_func,
             hf_config=self.hf_config,
+            adapter_path=self.adapter_path,
+            model_attributes=self.model_attributes,
+        )
+
+    def get_component_model(self, component: HfComponent, rank: int = 0) -> PyTorchModelHandler:
+        # TODO(shaahji): Add support for 'HfComponent.component_func'
+        hf_config = deepcopy(self.hf_config).dict()
+        hf_config.pop("components", None)
+        return PyTorchModelHandler(
+            model_path=self.ranked_model_path(rank),
+            model_file_format=ModelFileFormat.PYTORCH_ENTIRE_MODEL,
+            model_script=self.model_script,
+            script_dir=self.script_dir,
+            io_config=component.io_config,
+            dummy_inputs_func=component.dummy_inputs_func,
+            hf_config=HfConfig.parse_obj(hf_config),
             adapter_path=self.adapter_path,
             model_attributes=self.model_attributes,
         )
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -30,6 +30,7 @@ from olive.model.config.hf_config import HfConfig, get_model_type_from_hf_config
 from olive.model.utils import resolve_onnx_path
 from olive.passes import Pass
 from olive.passes.onnx.common import get_external_data_config, model_proto_to_olive_model
+from olive.passes.onnx.merge_decoders import merge_decoders
 from olive.passes.pass_config import PassConfigParam
 
 logger = logging.getLogger(__name__)
@@ -90,6 +91,11 @@ class OnnxConversion(Pass):
                 required=False,
                 description="Number of parallel jobs. Defaulted to number of CPUs. Set it to 0 to disable.",
             ),
+            "merge_components": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                description="Whether to merge the converted components.",
+            ),
         }
         config.update(get_external_data_config())
         return config
@@ -112,6 +118,9 @@ class OnnxConversion(Pass):
                 model, data_root, config, output_model_path, device, torch_dtype
             )
 
+        if not model.hf_config or not model.hf_config.components:
+            return self._convert_model_on_device(model, data_root, config, output_model_path, device, torch_dtype)
+
         onnx_models = []
         component_names = []
         for component_name, component_model in model.get_hf_components():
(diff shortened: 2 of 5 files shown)


olive/model/handler/pytorch.py
```diff
@@ -152,6 +153,31 @@ def load_model(self, rank: int = None) -> torch.nn.Module:
...
+    def get_component_model(self, component: HfComponent, rank: Optional[int] = None) -> "PyTorchModelHandler":
+        if component.component_func is None:
+            logger.debug("component_func is not provided, using hf_config to get component")
+            model_component = self.load_hf_model(self.model_path)
+        else:
+            user_module_loader = UserModuleLoader(self.model_script, self.script_dir)
+            model_component = user_module_loader.call_object(component.component_func, self)
+
+        # the second default parameter is to fix ruff b023:
+        # https://docs.astral.sh/ruff/rules/function-uses-loop-variable/
```
guotuofeng, line 165: the functionality is already in https://github.com/microsoft/Olive/blob/main/olive/model/handler/mixin/hf_config.py#L62

docs/source/api/passes.rst
```diff
@@ -174,9 +174,3 @@ VitisAIQuantization
 OptimumConversion
 --------------------
 .. autoconfigclass:: olive.passes.OptimumConversion
-
-.. _optimum_merging:
-
-OptimumMerging
```
jambayk, line 180: I don't think we should remove this pass entirely. It can be renamed and use the new merge utils. 

For passes like transformers optimization, quantization, inc quantization, it is not clear if the final model after the following workflows will be the same:
- convert -> optimize each component -> merge
- convert + merge -> optimize merged model 

So I think it is better to keep the option to merge later open.

examples/directml/dolly_v2/config_dolly_v2.json
```diff
@@ -1,9 +1,11 @@
 {
     "input_model":{
-        "type": "OptimumModel",
```
jambayk, line 3: I agree that the current `OptimumModel` is redundant since it is just a pytorch model with hf_config. I was considering removing it too. 

But I think the `decoder_model` and `decoder_with_past` might be better as config parameters for `optimum_conversion`? In the context of a pytorch model, they don't make much sense.

docs/source/api/passes.rst
shaahji: `merge_decoders` logic is the same as what `OptimumMerge` is. Instead of it being a separate pass, it is merely a utility function. If need be we can create pass that just calls the `merge_decoders` function instead of using `optimum`.

jambayk: Yes, that's what I am suggesting too. Until we can confirm that merging first and then running the other passes later will produce the same model as merging later, I don't think we should change the llama or dolly examples to merge first.

examples/directml/dolly_v2/config_dolly_v2.json
shaahji: They are just _names_ for models. Didn't see a point in changing them. Mostly to avoid any failures upstream if there was something depending on these models being called that.

jambayk: Since optimummodel is being removed, any incompatibility failures would already be caused by it? 

So imo, if we are fixing this redundant model handler, it's better to fix the weird placement of the names too. The names in the pytorch model makes it seem like it has multiple components, but it doesn't. It's only used by the optimum conversion pass to select the onnx components.

examples/directml/llama_v2/config_llama_v2.json
```diff
@@ -895,9 +895,6 @@
                     "/norm/Mul"
                 ]
             }
-        },
```
jambayk, line 898: also, this example still needs a merging pass. This model is a composite model so it follows this flow `https://github.com/microsoft/Olive/blob/main/olive/passes/olive_pass.py#L357`. There will still be composite onnx model here.

examples/directml/dolly_v2/config_dolly_v2.json
shaahji: > Since optimummodel is being removed, any incompatibility failures would already be caused by it?

Not exactly because OptimumModel would have followed by OptimumMerge which expected the inputs named certain way. That second part has been replaced with `merge_decoders` which also expect the models be named the same way.

olive/model/utils/merge_decoders.py
jambayk, line 1: This file might be better under `olive.passes.onnx`?

shaahji: It's not a pass so shouldn't be under _passes_.

jambayk: it's a util used by passes. We do the same for other utils too 
https://github.com/microsoft/Olive/tree/main/olive/passes/onnx/vitis_ai
https://github.com/microsoft/Olive/blob/main/olive/passes/pytorch/qat_utils.py
etc

examples/directml/dolly_v2/config_dolly_v2.json
jambayk: What I meant is that any example that uses optimummodel or any optimum passes need to be updated anyways since they are all tied to each other. 
So, moving the components as a parameter for optimum conversion pass would not cause any additional breakage.

docs/source/api/passes.rst
xiaoyu-work: Yeah I agree with Jambay. We could create a new pass to have an option to merge models later.

olive/passes/onnx/optimum_conversion.py
```diff
@@ -63,7 +64,7 @@ def validate_search_point(
         return True
 
     def _run_for_config(
-        self, model: OptimumModelHandler, data_root: str, config: Dict[str, Any], output_model_path: str
```
xiaoyu-work, line 66: I'm kinda confused now, since we removed optimum model, what is the difference between the onnxconversion pass and optimum conversion pass?

examples/llama2/tensor_parallel_generate.py
```diff
@@ -0,0 +1,59 @@
+import os
```
xiaoyu-work, line 1: What is this used for?

olive/passes/onnx/optimum_conversion.py
shaahji: Certain models need patching to export and optimum does that. Basically, the usual `torch.onnx.export` would not succeed for these models. OnnxConversion uses the `torch.onnx.export` and OptimumConversion uses the same with some patchwork to make a model _exportable_.

docs/source/api/passes.rst
shaahji: Puled out the change to OptimumMerge & OptimumModelHandler. PR focus on introducing merge decoders for OnnxConversion only.

examples/directml/dolly_v2/config_dolly_v2.json
shaahji: Puled out the change to OptimumMerge & OptimumModelHandler. PR focus on introducing merge decoders for OnnxConversion only.

examples/directml/llama_v2/config_llama_v2.json
shaahji: Pulled out the change to OptimumMerge & OptimumModelHandler. PR focus on introducing merge decoders for OnnxConversion only.

examples/llama2/tensor_parallel_generate.py
shaahji: Pulled out the change to OptimumMerge & OptimumModelHandler. PR focus on introducing merge decoders for OnnxConversion only.

olive/passes/onnx/conversion.py
```diff
@@ -124,10 +133,11 @@ def _run_for_config(
             onnx_models.append(output_model_component)
             component_names.append(component_name)
 
-        if onnx_models:
-            return CompositeModelHandler(onnx_models, component_names)
-        else:
-            return self._convert_model_on_device(model, data_root, config, output_model_path, device, torch_dtype)
+        if config["merge_components"] and len(onnx_models) == 2:
+            merged_model = merge_decoders(onnx_models[0].model_path, onnx_models[1].model_path)
+            return model_proto_to_olive_model(merged_model, output_model_path, config)
```
jambayk, line 138: Should we delete the component sub directories since they are already merged? The components were also saved in the output model directory. 

Might also have to resolve the final `output_model_path` like at https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/conversion.py#L375. Otherwise, the onnx model will probably be saved without a `.onnx` suffix.

olive/passes/onnx/conversion.py
```diff
@@ -412,26 +422,64 @@ def _export_ranked_model(params):
...
+                            pytorch_model,
+                            dummy_inputs,
+                            io_config,
+                            pass_config,
+                            device,
+                            torch_dtype,
+                            tempdir,
+                        )
+
+                    ranked_component_model = model_proto_to_olive_model(
```
jambayk, line 445: if it's saved to `output_filepath`, wouldn't the components overwrite each other?

jambayk: Since, the components are going to be merge anyways, might be easier to just keep the component model protos in memory and call merge_decoder on them directly? That way, there is no need to save, load and clean up.

olive/passes/onnx/conversion.py
shaahji: I don't think we should delete the component models. The component models wouldn't end up in the output folder unless this is the last pass in the config. Even in that case, the component models aren't returned so they shouldn't be in the output folder.

Fixed the path resolve issue.

olive/passes/onnx/conversion.py
shaahji: No output_file is ranked model name of the format model_XX.onnx so each file is named differently i.e. based on the rank.

jambayk: I mean the components of the same rank. this code is using `output_filepath` for both components. So wouldn't the second component overwrite the second component?

shaahji: Wouldn't that be assuming that there are only two components, and they will be merged. There aren't use case yet but theoretically there could be more than two components and the result might be a _CompositeModel_. Of course, that isn't supported in this implementation but that's a potential scenario.

jambayk: we can ignore the keep modelproto suggestion. But there is still the issue of the components being saved to the same filepath.

shaahji: Fixed.

olive/passes/onnx/conversion.py
```diff
@@ -412,26 +422,53 @@ def _export_ranked_model(params):
...
-                    tempdir,
-                )
+            if input_model.hf_config and input_model.hf_config.components:
+                ranked_models = []
+                for _, component_model in input_model.get_hf_components(local_rank):
+                    dummy_inputs = component_model.get_dummy_inputs()
+                    io_config = None if pass_config["use_dynamo_exporter"] else component_model.get_io_config()
+                    pytorch_model = component_model.prepare_session(rank=local_rank)
+
+                    with torch.no_grad():
```
jambayk, line 433: not completely related to this PR, but we could decorate the `_export_pytorch_model` method with `@torch.no_grad()` like in https://github.com/microsoft/Olive/blob/bc8b3312251484986972cf53ac8d4b2f4ed2ca1f/olive/evaluator/olive_evaluator.py#L816. 

This way, it won't matter if we forgot to do the `with torch.no_grad():` like I did in `_convert_model_on_device`

olive/passes/onnx/conversion.py
jambayk: Thanks! for the more than two component case, it will now be on our future selves to sort out the filepaths haha.

jambayk (approved):

olive/passes/onnx/conversion.py

jambayk (approved):

olive/passes/onnx/conversion.py
```diff
@@ -27,9 +26,11 @@
 )
 from olive.model.config import IoConfig
 from olive.model.config.hf_config import HfConfig, get_model_type_from_hf_config
+from olive.model.handler.composite import CompositeModelHandler
```
jambayk, line 29: btw, is there a reason for importing this separately from the other handlers?

olive/passes/onnx/optimum_conversion.py
xiaoyu-work: Maybe out of the scope of this PR, but can we add patching feature to our onnx conversion pass?

olive/passes/onnx/conversion.py
shaahji: Merely an oversight. Fixed.

olive/passes/onnx/optimum_conversion.py
jambayk: I don't think it's straightforward since there are lots of model specific information and configs required for such features. Optimum actively maintains it. 

With optimum conversion, we just have to provide the model name, task/feature. They do the conversion args and everything automatically. 

even our automatic conversion args is not up-to-date since new models get added everytime and huggingface doesn't maintain the `transformers.onnx` module we rely on. I have been considering migrating that feature to use optimum configs underneath when I have the time to try it.

jambayk (approved):

PR #832, opened 2023-12-19
Example to use model.generate for distributed Llama2 model

shaahji:
## Describe your changes

Example to use model.generate for distributed Llama2 model
Also, a small edit to suppress warnings generated during export and inference.

### DIFF
--- a/examples/llama2/tensor_parallel_generate.py
+++ b/examples/llama2/tensor_parallel_generate.py
@@ -1,0 +1,80 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+import os
+from pathlib import Path
+
+import torch
+import torch.distributed as dist
+from mpi4py.MPI import COMM_WORLD
+from optimum.onnxruntime import ORTModelForCausalLM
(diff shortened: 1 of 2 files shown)


examples/llama2/tensor_parallel_generate.py
```diff
@@ -0,0 +1,80 @@
...
+
+model = ORTModelForCausalLM.from_pretrained(
+    model_path,
+    config=config,
+    generation_config=gen_config,
+    use_io_binding=True,
+    provider="CUDAExecutionProvider",
+    provider_options={"device_id": str(rank)},
+)
+outputs = model.generate(**inputs, max_new_tokens=256)
```
jambayk, line 76: Just out of curiosity, so the ranked models already have the communication ops built in and can do the network communication and synchronization automatically? That's why there is no need for commands like `dist.barrier()` for the forward pass, etc?

jambayk (approved):

examples/llama2/tensor_parallel_generate.py
shaahji: That's correct. The comm-ops have their own blocking mechanism.

jambayk: That's cool! thanks.

PR #836, opened 2023-12-19
Whisper: Expose missing options, Enable timestamps, Transformers opt in `gpu_int8`

jambayk:
## Describe your changes
- Expose missing beam search options `use_vocab_mask`, `use_prefix_vocab_mask` and `use_logits_processor` that were added in ORT 1.16 along with `use_forced_decoder_ids`.
- Expose `skip_special_tokens` option in whisper pre-post-processing
- `use_logits_processor` enables timestamps in the whisper output. The output text looks something like 
   ```
   "...<|3.00|> The ache of his overstrained eyes,<|6.00|><|6.00|> even the soaring arena around him with thousands of spectators,<|10.00|><|10.00|> which revealed he's not worth thinking about.<|13.00|>"
   ```
- Add transformers optimization pass to `gpu,int8` workflow. It was originally omitted due to some bug in ort but was corrected a while ago. I had forgotten to add it back since we never used the gpu workflow until the fp16 workflow was recent updated.

**Note:** The time stamp prediction does not work for the new `whisper-large-v3` model since the ort logitsprocessor hardcoded some token ids which are not the same in the v3 model.

The whisper example now supports generating timestamps along with the text.

### DIFF
--- a/examples/whisper/prepare_whisper_configs.py
+++ b/examples/whisper/prepare_whisper_configs.py
@@ -99,7 +99,7 @@ def main(raw_args=None):
 
     # multi-lingual support check
     if not version_1_16:
-        if args.multilingual and not version_1_16:
+        if args.multilingual:
             raise ValueError("Multi-lingual support is only supported in ORT >= 1.16.0")
         if args.enable_timestamps:
             raise ValueError("Enabling timestamps is only supported in ORT >= 1.16.0")
--- a/examples/whisper/prepare_whisper_configs.py
+++ b/examples/whisper/prepare_whisper_configs.py
@@ -30,7 +30,13 @@ SUPPORTED_WORKFLOWS = {
     ],
     ("gpu", "fp32"): ["conversion", "transformers_optimization", "insert_beam_search", "prepost"],
     ("gpu", "fp16"): ["conversion", "transformers_optimization", "mixed_precision", "insert_beam_search", "prepost"],
-    ("gpu", "int8"): ["conversion", "onnx_dynamic_quantization", "insert_beam_search", "prepost"],
+    ("gpu", "int8"): [
+        "conversion",
+        "transformers_optimization",
+        "onnx_dynamic_quantization",
+        "insert_beam_search",
+        "prepost",
+    ],
 }
 DEVICE_TO_EP = {
     "cpu": "CPUExecutionProvider",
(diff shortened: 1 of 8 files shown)


olive/passes/utils/whisper_prepost.py
```diff
@@ -14,6 +14,7 @@ def add_pre_post_processing_to_model(
     model_name: str,
     use_audio_decoder: bool = True,
     use_onnx_stft: bool = True,
+    skip_special_tokens: bool = True,
```
guotuofeng, line 17: I don't see the example use the parameter?

trajepl: https://github.com/microsoft/Olive/blob/dfb7f8e2038ec10fcefca39cd1b038a39b3ea74a/olive/passes/utils/whisper_prepost.py#L35

examples/whisper/prepare_whisper_configs.py
```diff
@@ -79,8 +98,16 @@ def main(raw_args=None):
     transformers_version_4_36 = version.parse(TransformersVersion) >= version.parse("4.36.0")
 
     # multi-lingual support check
-    if args.multilingual and not version_1_16:
-        raise ValueError("Multi-lingual support is only supported in ORT >= 1.16.0")
+    if not version_1_16:
+        if args.multilingual and not version_1_16:
```
xiaoyu-work, line 102: Duplicated version_1_16 check

examples/whisper/prepare_whisper_configs.py
```diff
@@ -93,14 +120,24 @@ def main(raw_args=None):
...
+        template_json["engine"]["evaluator"] = None
 
     # update multi-lingual support
     template_json["passes"]["insert_beam_search"]["config"]["use_forced_decoder_ids"] = args.multilingual
+    # update predict timestep
+    template_json["passes"]["insert_beam_search"]["config"]["use_logits_processor"] = args.enable_timestamps
+    # update no audio decoder
+    template_json["passes"]["prepost"]["config"]["tool_command_args"]["use_audio_decoder"] = not args.no_audio_decoder
+    # update atol
+    template_json["passes"]["mixed_precision"]["config"]["atol"] = args.atol
```
xiaoyu-work, line 140: Any reason move this config setup here? For some pass flows we don't have mixed precision pass.

olive/passes/utils/whisper_prepost.py
jambayk: Yes, the example doesn't use it. But since it's a configurable option like `use_onnx_stft`, I think exposing it is good.

examples/whisper/prepare_whisper_configs.py
jambayk: Thanks! Will remove this.

examples/whisper/prepare_whisper_configs.py
jambayk: We are updating the template config which has all passes. So, I moved all template related updates to the same place. In the for loop, we will only update the values related to `device` or `precision`.

examples/whisper/prepare_whisper_configs.py
jambayk: Updated.

examples/whisper/prepare_whisper_configs.py
xiaoyu-work: If so, should we also update the `Performs optimization pipeline:` part in README.md?

jambayk: For which part? The template json by itself doesn't work. We just use it to hold the pass config params and other settings.

guotuofeng (approved):

PR #846, opened 2023-12-27
Add mistral examples

guotuofeng:
## Describe your changes
Add mistral examples

### DIFF
--- a/examples/mistral/mistral_optimize.json
+++ b/examples/mistral/mistral_optimize.json
@@ -77,6 +77,11 @@
                     "bits": 4,
                     "algorithm": "GPTQ"
                 },
+                "recipes":{
+                    "gptq_args": {
+                        "accuracy_level": 4
+                    }
+                },
                 "dataloader_func": "calib_dataloader",
                 "calibration_sampling_size": [
                     8
--- a/examples/mistral/mistral_optimize.json
+++ b/examples/mistral/mistral_optimize.json
@@ -86,7 +86,7 @@
                 },
                 "recipes":{
                     "gptq_args": {
-                        "accuracy_level": 4
+                        "accuracy_level": 0
                     }
                 },
                 "dataloader_func": "calib_dataloader",
--- a/examples/mistral/mistral_optimize.json
+++ b/examples/mistral/mistral_optimize.json
@@ -50,12 +50,10 @@
         "convert": {
             "type": "OptimumConversion",
             "config": {
-                "no_post_process": false,
                 "target_opset": 14,
-                "save_as_external_data": true,
-                "all_tensors_to_one_file": true,
                 "extra_args": {
-                    "legacy": false
+                    "legacy": false,
+                    "no_post_process": false
                 }
             }
         },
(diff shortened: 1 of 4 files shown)


examples/mistral/user_script.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def __iter__(self):
+        length = len(self.dataset)
+        counter = 0
+
+        while counter < length:
+            while True:
+                i = random.randint(0, len(self.dataset) - 1)
+                trainenc = self.dataset[i]
+                if trainenc["input_ids"].shape[0] > self.seq_len:
```
TriLoo, line 53: Perhaps we can add a judgment that all input data does not meet the conditions to avoid an infinite loop in the while True.

TriLoo (approved):
reviewed user script and config

examples/mistral/user_script.py
guotuofeng: what's your suggestion to avoid the infinite loop?

examples/mistral/mistral_optimize.json
```diff
@@ -0,0 +1,128 @@
...
+    "input_model": {
+        "type": "PyTorchModel",
+        "config": {
+            "hf_config": {
+                "model_name": "mistralai/Mistral-7B-v0.1",
+                "model_class": "MistralForCausalLM"
+            }
+        }
+    },
+    "systems": {
```
jambayk, line 11: Is this required? I think it's cleaner if we just specify ep as `CPUExecutionProvider` under engine so that it automatically infers the system as cpu?

examples/mistral/mistral_optimize.json
```diff
@@ -0,0 +1,128 @@
...
+                }
+            ]
+        }
+    },
+    "passes": {
+        "convert": {
+            "type": "OptimumConversion",
+            "config": {
+                "target_opset": 14,
+                "extra_args": {
```
jambayk, line 54: Since we are using the default values for these options, this option is redundant? Is there a reason for adding it? To explicitly provide them?

examples/mistral/requirements.txt
```diff
@@ -0,0 +1,6 @@
+datasets
+git+https://github.com/intel/neural-compressor.git
+onnxruntime-extensions
+optimum
```
jambayk, line 4: should we add a minimum version for optimum here? Just to be safe?

examples/mistral/mistral_optimize.json
guotuofeng: I put it here for demo purpose in case of user want to run on GPU by simply change the accelerators to gpu.

examples/mistral/mistral_optimize.json
guotuofeng: yeah, just for user know that the underlying settings for optimum conversion. The explicitly set value is same with current version default values.

examples/mistral/requirements.txt
guotuofeng: will add the 1.14.1 as the minimal requirement

jambayk (approved):

Issue #847, opened 2023-12-27
Validation errors for HfConfig


yuwenzho:
**Describe the bug**

When running [open_llama_config.json](https://github.com/microsoft/Olive/blob/main/examples/open_llama/open_llama_config.json), `hf_config` will cause the validation errors:
```json
"input_model":{
        "type": "PyTorchModel",
        "config": {
            "hf_config": {
                "model_name": "openlm-research/open_llama_3b",
                "model_class": "LlamaForCausalLM",
                "components": [{ "name": "decoder_model" }, { "name": "decoder_with_past_model" }]
            }
        }
    }
```

```python
pydantic.v1.error_wrappers.ValidationError: 4 validation errors for HfConfig
components -> 0 -> io_config
  field required (type=value_error.missing)
components -> 0 -> dummy_inputs_func
  field required (type=value_error.missing)
components -> 1 -> io_config
  field required (type=value_error.missing)
components -> 1 -> dummy_inputs_func
  field required (type=value_error.missing)
```

**To Reproduce**

In example of [open_llama](https://github.com/microsoft/Olive/tree/main/examples/open_llama), run `python -m olive.workflows.run --config open_llama_config.json`

**Other information**
 - OS: Linux
 - Olive version: main
 - ONNXRuntime package and version: [e.g. onnxruntime: 1.16.3]

**Additional context**

guotuofeng:
@jambayk, @trajepl, does the io_config/dummy_inputs_func in component need to be mandatory?

guotuofeng:
This is a regression coming from https://github.com/microsoft/Olive/pull/833, we will fix it

jambayk:
@yuwenzho I have created a PR with a potential fix https://github.com/microsoft/Olive/pull/851. Could you try this branch to see if everything works as expected? I tried running the example on my dev machine but the quantization takes too long.

guotuofeng:
> @yuwenzho I have created a PR with a potential fix #851. Could you try this branch to see if everything works as expected? I tried running the example on my dev machine but the quantization takes too long.

could we enable the similar UT?

yuwenzho:
> @yuwenzho I have created a PR with a potential fix #851. Could you try this branch to see if everything works as expected? I tried running the example on my dev machine but the quantization takes too long.

#851 looks good to me. Thanks for the fix.

PR #851, opened 2024-01-03
OptimumConversion: `components` option, version check

jambayk:
## Describe your changes
The `OptimumConversion` pass adds the following config parameters:
- `components`: #833 deprecated the `OptimumModelHandler` but uses a pytorch model with components. However, the components in the pytorch model handler doesn't make sense since conceptually, the pytorch model has no components and also causes #847. They are only used by the optimum conversion pass. So, making this a config parameter of the pass makes more sense.

Some other changes:
- Check the version of optimum and log a warning message about the different behavior of optimum for <1.14.0. 
- `legacy` and `no_post_process` must be passed as part of `extra_args`. The examples are still on legacy model but could be updated in the future to the latest single model.
- all external data config parameters are removed since they are ignored anyways. We save directly using the exporter.

Fixes:
#847 

OptimumConversion pass:
- Provide parameter `components` if the user wants to export only some models such as `decoder_model` and `decoder_with_past_model`.
- Uses the default exporter args and behavior of the underlying optimum version. For versions 1.14.0+, this means `legacy=False` and `no_post_process=False`. User must provide them using `extra_args` if legacy behavior is desired.

### DIFF
--- a/olive/passes/onnx/optimum_conversion.py
+++ b/olive/passes/onnx/optimum_conversion.py
@@ -98,17 +98,14 @@ class OptimumConversion(Pass):
         # TODO(anyone): consider using a temporary directory to export the model and then save the relevant components
         export_optimum_model(model.model_path or hf_config.model_name, output_model_path, **extra_args)
 
-        output_model_path = Path(output_model_path)
-
         # check the exported components
-        exported_models = [name.stem for name in output_model_path.iterdir() if name.suffix == ".onnx"]
-        logger.debug(f"Exported models: {exported_models}")
+        exported_models = [name.stem for name in Path(output_model_path).iterdir() if name.suffix == ".onnx"]
         if config["components"]:
             assert all(
                 component in exported_models for component in config["components"]
             ), f"Components {config['components']} are not exported. Only {exported_models} are exported."
         components = config["components"] or exported_models
-        logger.debug(f"Components to export: {components}")
+        logger.debug(f"Exported models are: {exported_models}. Returning components: {components}.")
 
         # if there is only one component, return it directly
         if len(components) == 1:
--- a/olive/passes/onnx/optimum_conversion.py
+++ b/olive/passes/onnx/optimum_conversion.py
@@ -12,7 +12,6 @@ import onnx
 
 from olive.hardware.accelerator import AcceleratorSpec
 from olive.model import CompositeModelHandler, ONNXModelHandler, PyTorchModelHandler
-from olive.model.config.hf_config import HfConfig
 from olive.model.utils import resolve_onnx_path
 from olive.passes import Pass
 from olive.passes.onnx.common import get_external_data_config, model_proto_to_olive_model
@@ -22,7 +21,7 @@ logger = logging.getLogger(__name__)
 
 
 class OptimumConversion(Pass):
-    """Convert a Optimum model to ONNX model using the Optimum export function."""
+    """Convert a Hugging Face PyTorch model to ONNX model using the Optimum export function."""
 
     _requires_user_script = True
 
@@ -85,6 +84,9 @@ class OptimumConversion(Pass):
     def _run_for_config(
         self, model: PyTorchModelHandler, data_root: str, config: Dict[str, Any], output_model_path: str
     ) -> Union[ONNXModelHandler, CompositeModelHandler]:
+        if not model.hf_config:
+            raise ValueError("OptimumConversion pass only supports PyTorchModelHandler with hf_config.")
+
         from optimum import version as optimum_version
         from optimum.exporters.onnx import main_export as export_optimum_model
         from packaging import version
@@ -99,9 +101,8 @@ class OptimumConversion(Pass):
                 "legacy": config["legacy"],
             }
         )
-        hf_config = deepcopy(model.hf_config) or HfConfig()
-        if hf_config.from_pretrained_args:
-            extra_args["trust_remote_code"] = hf_config.from_pretrained_args.trust_remote_code
+        if model.hf_config.from_pretrained_args:
+            extra_args["trust_remote_code"] = model.hf_config.from_pretrained_args.trust_remote_code
         if version.parse(optimum_version.__version__) < version.parse("1.14.0"):
             if not config["legacy"]:
                 raise ValueError(
@@ -112,7 +113,7 @@ class OptimumConversion(Pass):
 
         with tempfile.TemporaryDirectory() as temp_dir:
             export_optimum_model(
-                model.model_path or hf_config.model_name,
+                model.model_path or model.hf_config.model_name,
                 temp_dir,
                 **extra_args,
             )
--- a/olive/passes/onnx/optimum_conversion.py
+++ b/olive/passes/onnx/optimum_conversion.py
@@ -12,6 +12,7 @@ import onnx
 
 from olive.hardware.accelerator import AcceleratorSpec
 from olive.model import CompositeModelHandler, ONNXModelHandler, PyTorchModelHandler
+from olive.model.config.hf_config import HfConfig
 from olive.model.utils import resolve_onnx_path
 from olive.passes import Pass
 from olive.passes.onnx.common import get_external_data_config, model_proto_to_olive_model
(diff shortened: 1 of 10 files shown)


olive/passes/onnx/optimum_conversion.py
```diff
@@ -42,6 +52,14 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
                 default_value=True,
                 description="Whether to skip post-processing the exported model.",
             ),
+            "legacy": PassConfigParam(
```
guotuofeng, line 55: In my PR for mistral, I pass the legacy in extra_args. So, one question is should we expose the legacy explicitly? 

One more thing I hit when testing is the no_post_process is related to legacy. When I set legacy=True and no_post_process=False, the exception will be raised. you can see https://github.com/microsoft/Olive/blob/3037a4f8cf51f6a559279f8bd20a81a5bcae0f98/olive/passes/onnx/optimum_conversion.py#L95 for more info. 

On anther hand, it seems the default value for no_post_process is True, but the its default value is False in optimum. Need we make them consistent?

test/unit_test/utils.py
```diff
@@ -89,15 +89,6 @@ def get_optimum_model_by_model_path():
     )
```
guotuofeng, line 89: are there any existing UT to cover the failure like [https://github.com/microsoft/Olive/issues/847](https://github.com/microsoft/Olive/issues/847)? 

Basically, the deprecation PR should cover this kind of error before merging. 

So, I suggest add UT for the failure case.

olive/passes/onnx/optimum_conversion.py
```diff
@@ -66,46 +84,66 @@ def validate_search_point(
     def _run_for_config(
         self, model: PyTorchModelHandler, data_root: str, config: Dict[str, Any], output_model_path: str
     ) -> Union[ONNXModelHandler, CompositeModelHandler]:
+        if not model.hf_config:
```
guotuofeng, line 87: >hf_config [](http://example.com/codeflow?start=21&length=9)

It seems the hf_config.task or hf_config.model_class is not needed by optimum conversion. 

[https://github.com/microsoft/Olive/blob/3037a4f8cf51f6a559279f8bd20a81a5bcae0f98/examples/mistral/mistral_optimize.json#L7](https://github.com/microsoft/Olive/blob/3037a4f8cf51f6a559279f8bd20a81a5bcae0f98/examples/mistral/mistral_optimize.json#L7)

So, are there any better way to make them optional?

test/unit_test/utils.py
jambayk: The error in the issue was because the deprecation PR incorrectly used the components to make this Pass work. I said the same in a separate PR before it was split into the deprecation PR https://github.com/microsoft/Olive/pull/828#discussion_r1430593966 

Now, this error is not relevant to the optimum pass. It is expected behavior for components to require the missing params. I don't know what unit test we can make here.

olive/passes/onnx/optimum_conversion.py
jambayk: It is required if we need to load the input pytorch model and evaluate it. So I think, it is better to keep it.

test/unit_test/utils.py
guotuofeng: I mean the CI for deprecation PR should fail to block its merge in this case.

olive/passes/onnx/optimum_conversion.py
jambayk: If we don't expose the `legacy` option, the pass would behave differently for optimum 1.14.0+ and before when the user doesn't provide the legacy value themselves. 

For the `no_post_process` and `legacy` incompatibility, we will need to look into this more. maybe in the source code.

olive/passes/onnx/optimum_conversion.py
guotuofeng, line 87: > if not model.hf_config [](http://example.com/codeflow?start=7&length=23)

this check is not needed?

olive/passes/onnx/optimum_conversion.py
jambayk: Moved to `extra_args` as discussed offline.

test/unit_test/utils.py
jambayk: Added ut for components use case for optimum conversion pass. The validation check for pytorch model is unrelated to this PR.

olive/passes/onnx/optimum_conversion.py
jambayk: removed as discussed offline

olive/passes/onnx/optimum_conversion.py
```diff
@@ -70,42 +76,64 @@ def _run_for_config(
...
+            if "legacy" in extra_args:
+                logger.warning(
+                    "`legacy` option is set in the extra_args, but it is ignored because you are using optimum<1.14.0."
+                )
+                del extra_args["legacy"]
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+            export_optimum_model(
+                model.model_path or hf_config.model_name,
+                temp_dir,
```
guotuofeng, line 107: >temp_dir [](http://example.com/codeflow?start=16&length=8)

do we need cache the decoder/decoder_past intermediate model for troubleshooting purpose?

olive/passes/onnx/optimum_conversion.py
```diff
@@ -70,42 +76,64 @@ def _run_for_config(
...
+                    component in exported_models for component in config["components"]
+                ), f"Components {config['components']} are not exported. Only {exported_models} are exported."
+            components = config["components"] or exported_models
+            logger.debug(f"Components to export: {components}")
+
+            # if there is only one component, return it directly
+            if len(components) == 1:
+                output_model_path = resolve_onnx_path(output_model_path)
+                model_proto = onnx.load(Path(temp_dir) / f"{components[0]}.onnx")
+                return model_proto_to_olive_model(model_proto, output_model_path, config)
```
guotuofeng, line 124: Is it the copy better for large models?

olive/passes/onnx/optimum_conversion.py
```diff
@@ -70,42 +76,64 @@ def _run_for_config(
...
+
+            # if there are multiple components, return a composite model
+            model_components = []
+            model_component_names = []
+            for component_name in components:
+                # save the component model to a sub-directory
+                component_output_path = resolve_onnx_path(Path(output_model_path).with_suffix("") / component_name)
+                component_proto = onnx.load(Path(temp_dir) / f"{component_name}.onnx")
```
guotuofeng, line 132: >onnx.load(Path(temp_dir) / f"{component_name}.onnx") [](http://example.com/codeflow?start=34&length=52)

if the model is large(in mistral case, it is 28GB), then the load/resave will be time consuming.

olive/passes/onnx/optimum_conversion.py
jambayk: removed tempfolder as discussed offline.

olive/passes/onnx/optimum_conversion.py
```diff
@@ -70,41 +69,64 @@ def _run_for_config(
...
+                del extra_args["legacy"]
+
+        # export directly to the output path
+        # TODO(anyone): consider using a temporary directory to export the model and then save the relevant components
+        export_optimum_model(model.model_path or hf_config.model_name, output_model_path, **extra_args)
+
+        output_model_path = Path(output_model_path)
+
+        # check the exported components
+        exported_models = [name.stem for name in output_model_path.iterdir() if name.suffix == ".onnx"]
```
guotuofeng, line 104: >suffix [](http://example.com/codeflow?start=85&length=6)

need it ignore the case?

olive/passes/onnx/optimum_conversion.py
jambayk: removed

olive/passes/onnx/optimum_conversion.py
jambayk: meaning?

olive/passes/onnx/optimum_conversion.py
```diff
@@ -70,41 +69,64 @@ def _run_for_config(
...
+        logger.debug(f"Exported models: {exported_models}")
```
guotuofeng, line 105: needless log since it is duplicated with line 111

guotuofeng (dismissed):

olive/passes/onnx/optimum_conversion.py
guotuofeng: name.suffix.lower() == ".onnx"?

olive/passes/onnx/optimum_conversion.py
jambayk: Combined with line 111.

guotuofeng (approved):

Issue #852, opened 2024-01-03
Is this pass flow possible for Stable Diffusion?: OrtTransformersOptimization → IncDynamicQuantization or IncStaticQuantization


lshqqytiger:
**Describe the bug and context**
I'm trying to quantize an optimized Stable Diffusion model.
I got to know that `IncDynamicQuantization` has less reduction in inference speed than `OnnxDynamicQuantization`.
But I'm getting `IndexError` during UNet quantization pass.
The error belongs to `neural-compressor`, but except for the optimization pass, it works normally, so I think this would be a compatibility issue with `OrtTransformersOptimization`.

**To Reproduce**
1. Build and install `neural-compressor` from [source](https://github.com/intel/neural-compressor/tree/ly/fix_ort). https://github.com/intel/neural-compressor/pull/1512
2. Set passes and run olive.

*neural-compressor from pip will work with text encoder, unet, and vae encoder, but vae decoder throws an error.

**Expected behavior**
UNet should be quantized.

**Olive config**
provider: `DmlExecutionProvider`
pass flow: `["optimize", "inc_quantize"]`

text encoder passes:
```json
"optimize": {
  "type": "OrtTransformersOptimization",
  "disable_search": true,
  "config": {
    "model_type": "clip",
    "opt_level": 0,
    "float16": true,
    "use_gpu": true,
    "keep_io_types": false,
    "optimization_options": {
      "enable_gelu": true,
      "enable_layer_norm": true,
      "enable_attention": true,
      "use_multi_head_attention": true,
      "enable_skip_layer_norm": false,
      "enable_embed_layer_norm": true,
      "enable_bias_skip_layer_norm": false,
      "enable_bias_gelu": true,
      "enable_gelu_approximation": false,
      "enable_qordered_matmul": false,
      "enable_shape_inference": true,
      "enable_gemm_fast_gelu": false,
      "enable_nhwc_conv": false,
      "enable_group_norm": true,
      "enable_bias_splitgelu": false,
      "enable_packed_qkv": true,
      "enable_packed_kv": true,
      "enable_bias_add": false,
      "group_norm_channels_last": false
    },
    "force_fp32_ops": ["RandomNormalLike"],
    "force_fp16_inputs": { "GroupNorm": [0, 1, 2] }
  }
},
"inc_quantize": {
  "type": "IncDynamicQuantization",
  "disable_search": true,
  "config": {
    "save_as_external_data": false,
    "all_tensors_to_one_file": true
  }
}
```

[...]

PR #857, opened 2024-01-04
Fix weight loading of large model in INC quantzation pass

yuwenzho:
## Describe your changes

Fix weight loading of large model in inc quantzation pass. Reload weight for model > 2GB to prevent missing weight files.

## (Optional) Issue link

#852


yuwenzho:
@jambayk I noticed a TODO here [inc_quantization.py#L347](https://github.com/yuwenzho/Olive/blob/main/olive/passes/onnx/inc_quantization.py#L347), mentioning that there will be an error of missing external data file. I suspect it is also because the weight is not reloaded. I can check it.

guotuofeng (approved):

PR #860, opened 2024-01-08
🐌 SNPE env and example fix

trajepl:
## Describe your changes

There are a few issues for current SNPE examples:
1. The libxx.so under customized `python3.6` need to be put visible for system for SNPE cmd running
    - put `str(Path(f"{snpe_root}/python36-env/lib"))` under `LD_LIBRARY_PATH` can fix.
2. SNPE evaluation extent the logits wrongly, 
    - fixed in this PR.
3. We did not support tensorflow model evaluation.
    - update examples without input model evaluation.
4. Wrong metrics config for inception examples
5. Importing issue for vgg, tell that in README.md to let user update it manually as a workaround.
6. Python3.6 seems a bit outdated for latest SNPE updates. The vgg example failed to be run. Updated the python env to 3.8 can fix.

### DIFF
--- a/examples/snpe/inception_snpe_qualcomm_npu/README.md
+++ b/examples/snpe/inception_snpe_qualcomm_npu/README.md
@@ -13,8 +13,11 @@ Download the SNPE SDK zip following [instructions from Qualcomm](https://develop
 Unzip the file and set the unzipped directory path as environment variable `SNPE_ROOT`
 
 ### Configure SNPE
-```
-python -m olive.snpe.configure
+```sh
+# in general, python 3.8 is recommended
+python -m olive.snpe.configure --py_version 3.8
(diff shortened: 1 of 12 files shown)


jambayk (approved):
Thanks for the fixes. The lib and bin paths worked fine previously so I am not sure if it's something to do with the new snpe version or our environment differences. But it doesn't hurt to add extra paths.

guotuofeng:
It seems the shellcheck fail, do we need fix it?

guotuofeng (changes requested):
do we need fix the shellcheck failure?

trajepl:
> do we need fix the shellcheck failure?

Thanks! Added the Shebang info for shell script.

guotuofeng (approved):

PR #861, opened 2024-01-09
Add `LoftQ` pass

jambayk:
## Describe your changes
This PR adds a new pass `LoftQ` which is based on the paper https://arxiv.org/abs/2310.08659. 
It adds on the qlora fine tuning approach by initializing the quantization and lora adapters together. This "alleviates the discrepancy between the quantized and full-precision model and significantly improves generalization in downstream tasks". 

Since most of the fine-tuning steps are the same as `QLoRA`, the common code has been abstracted into a new base class called `QLoRABase`. `LoftQ` is a separate pass and not an option for `QLoRA` so that we don't conflate it too much and make the configuration confusing for the user. They can just switch the pass type instead.

 
Introduced a new pass `LoftQ` which performs model fine-tuning using the LoftQ initialization proposed in https://arxiv.org/abs/2310.08659.

### DIFF
--- a/examples/open_llama/open_llama_loftq_tinycodes.json
+++ b/examples/open_llama/open_llama_loftq_tinycodes.json
@@ -1,0 +1,74 @@
+{
+    "input_model":{
+        "type": "PyTorchModel",
+        "config": {
+            "hf_config": {
+                "model_name": "openlm-research/open_llama_7b_v2",
+                "task": "text-generation"
+            }
+        }
+    },
+    "data_configs": {
+        "tiny_codes_train": {
+            "name": "tiny_codes_train",
+            "type": "HuggingfaceContainer",
+            "user_script": "lora_user_script.py",
+            "components": {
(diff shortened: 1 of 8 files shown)


examples/open_llama/open_llama_loftq_tinycodes.json
```diff
@@ -0,0 +1,74 @@
...
+                        "corpus_strategy": "join",
+                        "text_template": "### Question: {prompt} \n### Answer: {response}",
+                        "source_max_len": 1024
+                    }
+                }
+            }
+        }
+    },
+    "passes": {
+        "qlora": {
```
xiaoyu-work, line 39: loftq?

jambayk: oh yeah. thanks for the catch. updated.

xiaoyu-work (approved):

PR #864, opened 2024-01-09
add phi2 model conversion

guotuofeng:
## Describe your changes
add phi2 model example to convert the microsoft/phi2 model. 


examples/phi2/phi2_optimize.json
```diff
@@ -0,0 +1,103 @@
...
+                "dataloader_func": "create_dataloader",
+                "batch_size": 2,
+                "enable_profiling": false
+            }
+        }
+    },
+    "pass_flows": [
+        [
+            "convert"
+        ]
```
devang-ml, line 89: Is there a reason to not add perf_tuning in the flow?

examples/phi2/phi2_optimize.json
```diff
@@ -0,0 +1,103 @@
...
+    },
+    "passes": {
+        "convert_optimum": {
+            "type": "OptimumConversion",
+            "config": {
+                "target_opset": 14,
+                "extra_args": {
+                    "legacy": false,
+                    "no_post_process": false,
+                    "trust_remote_code": true
```
jambayk, line 64: "trust_remote_code" should not be here. It's already inherited from the model from_pretrained_args. https://github.com/microsoft/Olive/blob/a013c6057d7b1245e6e048eb8d227be3e2962a84/olive/passes/onnx/optimum_conversion.py#L82

Even if you provided it, it's overwritten by line 82.

examples/phi2/phi2_optimize.json
guotuofeng: you are right, the perf_tuning does works when I do benchmark. will add it.
On the other hand, I'd like to add the transformer optimization when ORT support phi2.

examples/phi2/phi2_optimize.json
guotuofeng: it seems the extra_args will take precedence over the pretrained args. I add it here just for explicit purpose. the optimum conversion doesn't used from the time being. I will remove it for simplicity.

examples/phi2/phi2_optimize.json
```diff
@@ -0,0 +1,103 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+            ]
+        }
+    },
+    "passes": {
+        "convert_optimum": {
```
trajepl, line 57: Both of optimum and onnx conversion work here? Should we remove this? Seems you did not use that in pass flows.

trajepl: If you want to keep it, should we set the external saving config as following?
"all_tensors_to_one_file": true

guotuofeng: the two configs are removed already for optimum conversion in https://github.com/microsoft/Olive/pull/851

jambayk (approved):

examples/phi2/readme.md
```diff
@@ -0,0 +1,12 @@
+## Prerequisites
+* einops
+
+## Usage
+```bash
+python -m olive.workflows.run --config phi2_optimize.json
+```
+
+## Limitations
+From the time being, the official [phi2](https://huggingface.co/microsoft/phi-2) model could not be exported to ONNX by using the official model code. Therefore, we need patch the forward method to do preprocessing and postprocessing for the past_key_values arguments. When the official model could be exported to ONNX, we will remove this patch.
```
jambayk, line 10: This is not needed anymore?

guotuofeng: good catch

jambayk (approved):

PR #870, opened 2024-01-11
Add support for existing aml environment for aml system

xiaoyu-work:
## Describe your changes

Add support for existing aml environment for aml system. Users can specify env name, version and label to for their Environment.

### DIFF
--- a/olive/systems/common.py
+++ b/olive/systems/common.py
@@ -37,6 +37,12 @@ class AzureMLDockerConfig(ConfigBase):
             return None
 
 
+class AzureMLEnvironmentConfig(ConfigBase):
+    name: str
+    version: Optional[str] = None
+    label: Optional[str] = None
+
+
 class LocalDockerConfig(ConfigBase):
     image_name: str
     base_image: Optional[str] = None
--- a/docs/source/tutorials/configure_systems.rst
+++ b/docs/source/tutorials/configure_systems.rst
@@ -88,11 +88,43 @@ System Configuration
             )
             aml_system = AzureMLSystem(
                 aml_compute="cpu-cluster",
-                aml_docker_config={
-                    "base_image": "mcr.microsoft.com/azureml/openmpi4.1.0-ubuntu20.04",
-                    "conda_file_path": "conda.yaml"
-                }
+                aml_docker_config=docker_config
             )
+
+If you provide a :code:`aml_docker_config`, Olive will create a new Azure ML Environment using the :code:`aml_docker_config` configuration.
+Alternatively, you can provide an existing Azure ML Environment using :code:`aml_environment_config`:
+
+.. tabs::
+    .. tab:: Config JSON
+
+        .. code-block:: json
+
+            {
+                "type": "AzureML",
+                "config": {
+                    "aml_compute": "cpu-cluster",
+                    "aml_environment_config": {
+                        "name": "myenv",
+                        "version": "1"
+                    }
+                }
+            }
+
+    .. tab:: Python Class
+
+        .. code-block:: python
+
+            from olive.systems.azureml import AzureMLDockerConfig, AzureMLSystem
+
+            aml_environment_config = AzureMLEnvironmentConfig(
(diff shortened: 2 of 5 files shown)


olive/systems/common.py
```diff
@@ -37,6 +37,12 @@ def _get_abspath(cls, v):
             return None
 
 
+class AzureMLEnvironmentConfig(ConfigBase):
```
guotuofeng, line 40: is it better to merge this class with AzureMLDockerConfig by renaming it to AzureMLEnvironmentConfig?

guotuofeng: Basically, the two classes' behavior stand for get/create environment used by aml.

guotuofeng: @jambayk, what's your opinion?

xiaoyu-work: That makes sense however it will be confusing between get and create. `aml_docker_config` currently will create an anonymous environment, when creating a new environment, you can also pass same arguments as `get` like name/version/label. This reminds me that we could improve `create` feature that user can reuse the previous creating environment by also accepting name/version/label for `create` operation.

jambayk: I think since this has two different types of environments:
1. annonymous environment created with dockerconfig
2. pre-existing environment that we reference using name,version,label

I think it is better to have separate config for each. Otherwise combining them makes it harder for the user to understand the difference between the two use cases since all the config parameters will be together even though they don't have any common parameters.

xiaoyu-work: We should improve our logic in either way:

- If we want to keep the new creating environment anonymous, we may need to make sure we delete it after this run (we don't do this now). And additionally, supporting using existing Environment by specifying its name/version like what this PR is doing.
- If not, that would be a new feature in Olive: Olive supports creating a long-lasting Azure ML Environment without deleting it after Olive run, with this, user can also provide name/version to this new environment and Olive creates a new env with name/version so they can use it in the future.

@devang-ml what do you think?

jambayk: I don't think there is a need to delete the anonymous environment or name the environment. From what I understand about aml environments, the environments are hashed by the contents of the environment (dockerfile, etc) and backed by an acr docker image. 
This allows multiple aml runs to use the same cached environment (docker image in their backend) as long as they have the same environment config. You can see of the environments used by our ci tests is using an anonymous environment that was build months ago. 
image

I don't think deleting an "environment" actually deletes any of the underlying data. Probably just the reference to it (name, label, version). If it actually deletes it, then we would have to build the environment from scratch everytime which is time consuming. The caching behavior of aml is one of its advantages.

tldr: I think naming an environment is just creating a reference to underlying hashed + cached data (build environment, environment docker image). so there is no need for us to worry about naming and managing environment ourselves. 
I think it is enough to allow user to specify their own named environments if they want.

guotuofeng: I agree with @jambayk that we don't need to delete the anonymous envs. This comments can be treated as open discussion since I am not sure which approach is better. Maybe we can wait for comments from @devang-ml

xiaoyu-work: There was one issue happened to me before: if specifying Olive repo in dockerfile like "git+https://github.com/microsoft/Olive#egg=olive-ai", even there will be new version and building the environment again, it will always use old main branch commit to build the environment (I guess somehow AML cache system used old build cache), and did not fetch the latest commit for Olive main branch. I don't see any user reporting this but this could be a potential issue.

guotuofeng: This is because AML build environment using the conda.yaml. and it doesn't check whether the external link changes.

devang-ml: Conceptually, 'create' and 'get' has their own purpose. 

'get' should fail if the requested env does not exist. 
It seems here 'create' uses caches automatically for anonymous envs and there is no need to delete anonymous envs. 

So, keep these two configs separate is fine with me.

jambayk (approved):

PR #875, opened 2024-01-12
Merge OV and ORT stable diffusion examples

xiaoyu-work:
## Describe your changes

Merge OV and ORT stable diffusion examples


xiaoyu-work:
Hi @jstoecker @PatriceVignola  , I refactored the directml stable diffusion example in this PR. Since I don't have a direcml machine, can you please help me manually test if this change doesn't break anything from directml side?

examples/directml/stable_diffusion/ort_optimization_util.py
```diff
@@ -0,0 +1,158 @@
...
+        # Not necessary, but helps DML EP further optimize runtime performance.
+        # batch_size is doubled for sample & hidden state because of classifier free guidance:
+        # https://github.com/huggingface/diffusers/blob/46c52f9b9607e6ecb29c782c052aea313e6487b7/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion.py#L672
+        sess_options.add_free_dimension_override_by_name("unet_sample_batch", hidden_batch_size)
+        sess_options.add_free_dimension_override_by_name("unet_sample_channels", 4)
+        sess_options.add_free_dimension_override_by_name("unet_sample_height", image_size // 8)
+        sess_options.add_free_dimension_override_by_name("unet_sample_width", image_size // 8)
+        sess_options.add_free_dimension_override_by_name("unet_time_batch", 1)
+        sess_options.add_free_dimension_override_by_name("unet_hidden_batch", hidden_batch_size)
+        sess_options.add_free_dimension_override_by_name("unet_hidden_sequence", 77)
```
PatriceVignola, line 149: This code is needed for inferencing, not optimizing.

xiaoyu-work: This pipeline will only be used for inference.

jstoecker (commented):
It's a bit confusing to mix both directml and openvino requirements and scripts under a single sample that's called "directml/stable_diffusion". It would make sense if the purpose of the sample was to illustrate optimizing for different backends, but instead it feels like the goal is to leverage some of the visualization/GUI code in the directml sample.

My preference as a user would be to have simpler and more focused samples rather than larger samples with logic that may not be relevant to what I'm interested in (this applies equally to people that want to use DML or OV). What's the motivation for this change?

olive/model/config/model_config.py
```diff
@@ -58,7 +58,7 @@ class ModelConfig(ConfigBase):
             ]
     """
 
-    type: str  # noqa: A003
```
guotuofeng, line 61: these changes should be in another separated PR. I just create another one.

xiaoyu-work:
> It's a bit confusing to mix both directml and openvino requirements and scripts under a single sample that's called "directml/stable_diffusion". It would make sense if the purpose of the sample was to illustrate optimizing for different backends, but instead it feels like the goal is to leverage some of the visualization/GUI code in the directml sample.
> 
> My preference as a user would be to have simpler and more focused samples rather than larger samples with logic that may not be relevant to what I'm interested in (this applies equally to people that want to use DML or OV). What's the motivation for this change?

Since we already have CUDA provider also under `directml/stable_diffusion`, we decided to add OpenVINO as well here. The reason we merge these two examples is to demonstrate the ability of Olive that it can optimize the same model using multiple optimization tools at the same time (or in a same config file). So we merge the OpenVINOConversion pass here. 

I also have the concerns that all of these backends/tools examples are under `directml/stable_diffusion`. We would like to rename this to a more meaningful example name. But since there are a couple of external hyperlinks to this folder, we don't do this change for now.

@devang-ml do you want to provide more contexts here?

jstoecker:
> > It's a bit confusing to mix both directml and openvino requirements and scripts under a single sample that's called "directml/stable_diffusion". It would make sense if the purpose of the sample was to illustrate optimizing for different backends, but instead it feels like the goal is to leverage some of the visualization/GUI code in the directml sample.
> > My preference as a user would be to have simpler and more focused samples rather than larger samples with logic that may not be relevant to what I'm interested in (this applies equally to people that want to use DML or OV). What's the motivation for this change?
> 
> Since we already have CUDA provider also under `directml/stable_diffusion`, we decided to add OpenVINO as well here. The reason we merge these two examples is to demonstrate the ability of Olive that it can optimize the same model using multiple optimization tools at the same time (or in a same config file). So we merge the OpenVINOConversion pass here.
> 
> I also have the concerns that all of these backends/tools examples are under `directml/stable_diffusion`. We would like to rename this to a more meaningful example name. But since there are a couple of external hyperlinks to this folder, we don't do this change for now.
> 
> @devang-ml do you want to provide more contexts here?

Ah, I did not see that CUDA support was already merged into this sample in December. Ultimately, it would be good to know what the intent is for Olive samples:

1. Samples that are specialized and tested to work for specific backends
2. Samples that work for multiple backends (either documented, or up to the user to figure out which backends work)
3. A mix of 1 and 2 above

Earlier on I figured we were going in the directinon of 3 (mix of generic/specialized samples), but It sounds like there is a desire to converge all samples to be more generic. If that's the case, I don't want to block this change. You bring up a good point about the hyperlinks to the current path; one possible solution is to leave a README that redirects to the source to the common "examples/stable_diffusion" path. Moving forward, though, I'd like to understand the strategy for samples overall and whether we can keep backend-specific examples or if we should have them outside the olive repo itself.

xiaoyu-work:
> > > It's a bit confusing to mix both directml and openvino requirements and scripts under a single sample that's called "directml/stable_diffusion". It would make sense if the purpose of the sample was to illustrate optimizing for different backends, but instead it feels like the goal is to leverage some of the visualization/GUI code in the directml sample.
> > > My preference as a user would be to have simpler and more focused samples rather than larger samples with logic that may not be relevant to what I'm interested in (this applies equally to people that want to use DML or OV). What's the motivation for this change?
> > 
> > 
> > Since we already have CUDA provider also under `directml/stable_diffusion`, we decided to add OpenVINO as well here. The reason we merge these two examples is to demonstrate the ability of Olive that it can optimize the same model using multiple optimization tools at the same time (or in a same config file). So we merge the OpenVINOConversion pass here.
> > I also have the concerns that all of these backends/tools examples are under `directml/stable_diffusion`. We would like to rename this to a more meaningful example name. But since there are a couple of external hyperlinks to this folder, we don't do this change for now.
> > @devang-ml do you want to provide more contexts here?
> 
> Ah, I did not see that CUDA support was already merged into this sample in December. Ultimately, it would be good to know what the intent is for Olive samples:
> 
> 1. Samples that are specialized and tested to work for specific backends
> 2. Samples that work for multiple backends (either documented, or up to the user to figure out which backends work)
> 3. A mix of 1 and 2 above
> 
> Earlier on I figured we were going in the directinon of 3 (mix of generic/specialized samples), but It sounds like there is a desire to converge all samples to be more generic. If that's the case, I don't want to block this change. You bring up a good point about the hyperlinks to the current path; one possible solution is to leave a README that redirects to the source to the common "examples/stable_diffusion" path. Moving forward, though, I'd like to understand the strategy for samples overall and whether we can keep backend-specific examples or if we should have them outside the olive repo itself.
> 
> Also, I verified the sample still runs on DML. However, the performance is quite a bit slower since the default image size for SD 1.5 changes to 768x768. @PatriceVignola: we used to have a mapping of model to default resolution, and SD 1.5 was using 512x512 earlier. Should we put this back into the sample?

Thanks for the check! However, I didn't make any change about image size. The image size was changed in this commit: https://github.com/microsoft/Olive/commit/d29ac9006365a12c8f6e8fb9150aca9786897190#diff-e6446394c42aab85951bc2672ea0c22dafcfdb65c97cc55841698c382271a841. Maybe @PatriceVignola has more context. I'm okay with either default size.

PatriceVignola:
The image_size can easily be specified at runtime by the caller with `--image_size`. We can change it back to 512 if we want to, it doesn't really matter to me.

But the performance regression is not only due to this. I don't have time to get to the bottom of it today, but when I ran the example yesterday with 512x512 images, it was slower than I was expected (we should have more than 50 it/s on an RTX 4090 for example, but I was getting less than 40). I'll be able to look into it more on Tuesday.

PatriceVignola:
I re-ran some perf testing from a clean slate and I don't see any regressions. LGTM!

@xiaoyu-work feel free to change the default `--image_size` in the script back to 512 if you want, it probably makes more since since SD 1.15 is the default model, as opposed to SD 2.0.

xiaoyu-work:
> I re-ran some perf testing from a clean slate and I don't see any regressions. LGTM!
> 
> @xiaoyu-work feel free to change the default `--image_size` in the script back to 512 if you want, it probably makes more since since SD 1.15 is the default model, as opposed to SD 2.0.

Thanks for the testing! I've updated the image size back to 512.

PatriceVignola (approved):
:shipit:

jstoecker (approved):

PR #876, opened 2024-01-12
Rename the cache in DirectML's LLaMA 2 example from cache to past_key_values

PatriceVignola:
A recent change to make the cache work for phi-2 (https://github.com/microsoft/Olive/pull/863) broke the DirectML LLaMA 2 example because the cache was named differently.


examples/directml/llama_v2/config_llama_v2.json
```diff
@@ -13,207 +13,207 @@
...
+                                "past_key_values_out.15.value",
+                                "past_key_values_out.16.key",
+                                "past_key_values_out.16.value",
+                                "past_key_values_out.17.key",
+                                "past_key_values_out.17.value",
+                                "past_key_values_out.18.key",
+                                "past_key_values_out.18.value",
+                                "past_key_values_out.19.key",
+                                "past_key_values_out.19.value",
+                                "past_key_values_out.20.key",
```
guotuofeng, line 124: instead of changing all those stuff, is it better to using a custom function like https://github.com/microsoft/Olive/blob/main/examples/llama2/user_script.py#L153 or https://github.com/microsoft/Olive/blob/461419f3fa4f5802914a49b8786430e9c5d85772/examples/phi2/phi2_optimize.json#L7 to avoid manually changing it whenever there are ay schema updates in future?

guotuofeng: just some two cents. You can decide whether do the change or not.

examples/directml/llama_v2/config_llama_v2.json
```diff
@@ -13,207 +13,207 @@
...
+                                "past_key_values_out.26.key",
+                                "past_key_values_out.26.value",
+                                "past_key_values_out.27.key",
+                                "past_key_values_out.27.value",
+                                "past_key_values_out.28.key",
+                                "past_key_values_out.28.value",
+                                "past_key_values_out.29.key",
+                                "past_key_values_out.29.value",
+                                "past_key_values_out.30.key",
+                                "past_key_values_out.30.value",
```
guotuofeng, line 145: nit: for the output, we use present.[index].key/value in other examples.

guotuofeng: For example, https://github.com/huggingface/transformers/blob/651408a077f842e76e75bfc7d02b8ac38eeb6480/src/transformers/models/bart/configuration_bart.py#L240

do we need unify them?

examples/directml/llama_v2/config_llama_v2.json
PatriceVignola: This whole example will be replaced by another one soon-ish, but I'll keep this in mind for the other example.

guotuofeng (approved):

jambayk:
@PatriceVignola can you push an empty commit or rebase to retrigger the GitHub actions with the ruff linter fix?

PR #882, opened 2024-01-16
Add models rank json file

xiaoyu-work:
## Describe your changes

Add models rank json file. This file ranks all output models from different EPs. This json file includes `model_config` and `metrics`. User can find model path and EP info from `model_config`.

### DIFF
--- a/olive/engine/packaging/packaging_generator.py
+++ b/olive/engine/packaging/packaging_generator.py
@@ -79,16 +79,18 @@ def _package_models_rank(tempdir, footprints: Dict["AcceleratorSpec", "Footprint
         ),
         reverse=True,
     )
+    rank = 1
+    model_info_list = []
+    for node in sorted_nodes:
+        model_info = {
+            "rank": rank,
+            "model_config": node.model_config,
+            "metrics": node.metrics.value.to_json() if node.metrics else None,
+        }
+        model_info_list.append(model_info)
+        rank += 1
     with (tempdir / "models_rank.json").open("w") as f:
-        rank = 1
-        for node in sorted_nodes:
-            model_info = {
-                "rank": rank,
-                "model_config": node.model_config,
-                "metrics": node.metrics.value.to_json() if node.metrics else None,
-            }
-            f.write(json.dumps(model_info) + "\n")
-            rank += 1
+        f.write(json.dumps(model_info_list))
 
 
 def _package_sample_code(cur_path, tempdir):
--- a/olive/engine/packaging/packaging_generator.py
+++ b/olive/engine/packaging/packaging_generator.py
@@ -54,7 +54,7 @@ def _generate_zipfile_output(
     with tempfile.TemporaryDirectory() as temp_dir:
         tempdir = Path(temp_dir)
         _package_sample_code(cur_path, tempdir)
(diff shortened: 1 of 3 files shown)


olive/engine/packaging/packaging_generator.py
```diff
@@ -67,6 +69,28 @@ def _generate_zipfile_output(
...
+        itertools.chain.from_iterable(f.nodes.values() for f in footprints.values()),
+        key=lambda x: tuple(
+            x.metrics.value[metric].value if x.metrics.cmp_direction[metric] == 1 else -x.metrics.value[metric].value
+            for metric in metrics_dict
+        ),
+        reverse=True,
+    )
+    with (tempdir / "models_rank.json").open("w") as f:
+        rank = 1
+        for node in sorted_nodes:
```
guotuofeng, line 84: is it better to write once if we put all these model info into a list?

guotuofeng (changes requested):
The result json might not valid?

olive/engine/packaging/packaging_generator.py
guotuofeng: when I try to load the following json file, the following errors will be raised.
image


```json
{
    "a": "b"
}
{
    "a": "b"
}
{
    "a": "b"
}
```

xiaoyu-work: Oh cause I don't expect user to check this file in their code, but they may want to check this file manually. So this file is not a valid json file. But you are right they may want to load this file in their code. I can update this.

test/unit_test/engine/packaging/test_packaging_generator.py
```diff
@@ -67,6 +67,7 @@ def test_generate_zipfile_artifacts(mock_sys_getsizeof, save_as_external_data, m
         zip_ref.extractall(output_dir)
     assert (output_dir / "SampleCode").exists()
     assert (output_dir / "CandidateModels").exists()
+    assert (output_dir / "models_rank.json").exists()
```
guotuofeng, line 70: better load the json to verify the format is correct?

olive/engine/packaging/packaging_generator.py
```diff
@@ -67,6 +69,30 @@ def _generate_zipfile_output(
         shutil.move(f"{packaging_config.name}.zip", output_dir / f"{packaging_config.name}.zip")
 
 
+def _package_models_rank(tempdir, footprints: Dict["AcceleratorSpec", "Footprint"]):
+    metrics_dict = next(iter(footprints.values())).objective_dict
```
guotuofeng, line 73: just curious the reason that we only get the first node metrics?

test/unit_test/engine/packaging/test_packaging_generator.py
xiaoyu-work: Added to test.

olive/engine/packaging/packaging_generator.py
xiaoyu-work: All nodes should have the same metrics dict so getting the metrics dict from any of those should be fine.

guotuofeng (approved):

PR #894, opened 2024-01-23
Add AML resource support for data config

xiaoyu-work:
## Describe your changes

Add AML resource support for data configs. This config works now:
```
    "data_configs": {
        "dataset_default": {
            "name": "dataset_default",
            "params_config": {
                "data_name": "name", 
                "data_dir": {
                    "type": "azureml_datastore",
                    "config": {
                        "azureml_client": {
                            "subscription_id": "<>",
                            "resource_group": "<>",
                            "workspace_name": "<>"
                        },
                        "datastore_name": "<datastore_name>",
                        "relative_path": "<path/data>"
                    }
                },
                "component_kwargs": {
                    "pre_process_data": {...}
                }
            }
        }
    }
```

This AML data can be used both in local and on AML.


olive/systems/azureml/aml_system.py
```diff
@@ -193,9 +193,6 @@ def _create_args_from_resource_path(self, rp: OLIVE_RESOURCE_ANNOTATIONS):
             return None
         asset_type = get_asset_type_from_resource_path(model_resource_path)
```
trajepl, line 195: Any reason to remove these lines? Then how to tell the assert type for aml job input?

olive/systems/azureml/aml_system.py
```diff
@@ -193,9 +193,6 @@ def _create_args_from_resource_path(self, rp: OLIVE_RESOURCE_ANNOTATIONS):
...
 
-        if model_resource_path.type == ResourceType.AzureMLDatastore:
```
trajepl, line 196: Any reason to remove these lines? Then how to tell the assert type for aml job input?

xiaoyu-work: It is duplicated as https://github.com/microsoft/Olive/blob/main/olive/systems/azureml/aml_system.py#L70-L71

olive/data/component/load_dataset.py
```diff
@@ -32,7 +34,8 @@ def huggingface_dataset(data_dir, data_name=None, subset=None, split="validation
     from datasets import load_dataset
 
     assert data_name is not None, "Please specify the data name"
-    return load_dataset(path=data_name, name=subset, split=split, data_files=data_files, **kwargs)
+    data_files = str(Path(data_dir)) if data_dir and Path(data_dir).is_file() else data_files
```
guotuofeng, line 37: >data_files = str(Path(data_dir)) if data_dir and Path(data_dir).is_file() else data_files [](http://example.com/codeflow?start=4&length=89)

why do we need this kinds of logic? I think we can simply pass both the data_dir and data_files to Huggingface.

xiaoyu-work: Sometime the data_dir actually a datafile, for instance:
```
                "data_dir": {
                    "type": "azureml_datastore",
                    "config": {
                        "datastore_name": "<datastore_name>",
                        "relative_path": "a_json_file.json"
                    }
                },
```
In this case, the `data_dir` will be a json file path. While `load_dataset` func `data_dir` param cannot recognize a file path to load the data.

guotuofeng: In this case, could we just pass the data directory downloaded from AzureML to Olive? the data_dir could be either files or dir is confusing to end user.

xiaoyu-work: What do you mean by this "pass the data directory downloaded from AzureML to Olive"?

guotuofeng: from the name, data_dir represent data directory. In azureml case, it should be the data directory contains the files.

guotuofeng:
could we add UT for this functionality?

olive/data/component/load_dataset.py
xiaoyu-work: So how can we handle the case that the data on AML is a datafile?

olive/data/container/data_container.py
```diff
@@ -30,6 +31,8 @@ def load_dataset(self, data_root_path: Optional[str] = None):
         """Run load dataset."""
         params_config = self.config.load_dataset_params
         data_dir = params_config.get("data_dir")
+        if data_dir:
```
guotuofeng, line 34: for job run in AML, we don't need download them. Instead, we can directly mount it?

olive/data/component/load_dataset.py
trajepl: 1. As Mike mentioned, we should not download the aml dataset and upload it again if the job run in aml.
2. If the data is on aml, but we need run olive locally, maybe user can download the data by themselves?
3. To avoid the misconception on data dir and files, can we directly leverage the `data_files` argument?

olive/data/container/data_container.py
xiaoyu-work: This would be a little bit tricky. Data container has no knowledge of which system is running, local or aml. Even pass itself doesn't know. Any better thought here?

xiaoyu-work: Updated

guotuofeng:
Please rebase to main branch which just enable pylint caused by bug in pyproject.toml.

olive/data/container/data_container.py
```diff
@@ -68,3 +67,10 @@ def get_first_batch(self, dataloader=None, data_root_path=None):
 
     def update_component(self):
         return None
+
+    def _update_params_config(self, params_config, data_root_path, key):
+        param = params_config.get(key)
+        if param:
+            param = create_resource_path(param).get_path()
```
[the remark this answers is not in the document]
xiaoyu-work: Yes, this will download data from aml if the data was on aml.

guotuofeng:
>     def evaluate_model(

does aml_evaluation_runner need the data_files or not? just curious why pass need handle but evaluate runner doesn't need.

---
Refers to: olive/systems/azureml/aml_system.py:574 in e0b486e. [](commit_id = e0b486ea3f6f6faee5bbbca2827fbd4cdceae5da, deletion_comment = False)

olive/data/container/data_container.py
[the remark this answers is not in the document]
guotuofeng: So, the download will not happen if the system is aml and data is in aml too. is my understanding correct?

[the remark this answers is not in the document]
xiaoyu-work: No the data will not be downloaded if the system is aml, as this part of logic will be run on aml compute.

guotuofeng:
could we add TODO for azureml eval runner for data files support?

guotuofeng (approved):

PR #925, opened 2024-02-02
🏂 Bert SNPE conversion & evaluation case

trajepl:
## Describe your changes
1. Add Bert SNPE conversion & evaluation case
2. Fix bugs in raw dataloder and snpe evaluator
3. Expose max_length for huggingface pre_process_data

image

### DIFF
--- a/examples/bert/npu/bert_snpe.json
+++ b/examples/bert/npu/bert_snpe.json
@@ -45,18 +45,6 @@
             "components": {
                 "post_process_data": {
                     "type": "snpe_post_process"
-                },
-                "dataloader": {
-                    "name": "huggingface_dataloader",
-                    "type": "huggingface_data_to_raw_dataloader",
-                    "params": {
-                        "io_config": {
-                            "input_names": ["input_ids", "attention_mask", "token_type_ids"],
-                            "input_types": ["int64", "int64", "int64"],
-                            "input_shapes": [[2, 128], [2, 128], [2, 128]],
-                            "output_names": ["logits"]
-                        }
-                    }
                 }
             },
             "params_config": {
--- a/examples/bert/npu/bert_snpe.json
+++ b/examples/bert/npu/bert_snpe.json
@@ -1,0 +1,111 @@
+{
+    "input_model":{
+        "type": "PyTorchModel",
+        "config": {
+            "hf_config": {
+                "model_name": "Intel/bert-base-uncased-mrpc",
+                "task": "text-classification",
+                "hf_config": {
+                    "data_name":"glue",
+                    "subset": "mrpc",
+                    "split": "validation",
+                    "input_cols": ["sentence1", "sentence2"],
+                    "label_cols": ["label"],
+                    "batch_size": 1
+                }
+            },
+            "io_config": {
+                "input_names": ["input_ids", "attention_mask", "token_type_ids"],
+                "input_types": ["int64", "int64", "int64"],
+                "input_shapes": [[2, 128], [2, 128], [2, 128]],
+                "output_names": ["logits"]
+            }
+        }
+    },
+    "evaluators": {
+        "snpe_evaluator": {
+            "metrics": [
+                {
+                    "name": "accuracy",
+                    "type": "accuracy",
+                    "backend": "huggingface_metrics",
+                    "data_config": "snpe_dataset",
+                    "user_config": {
+                        "inference_settings": {
+                            "snpe":{
+                                "return_numpy_results": true
+                            }
+                        }
+                    },
+                    "sub_types": [
+                        {"name": "accuracy", "priority": 1, "goal": {"type": "max-degradation", "value": 0.05}},
+                        {"name": "f1"}
+                    ]
+                }
+            ]
+        }
+    },
+    "data_configs": {
+        "snpe_dataset": {
+            "name": "snpe_dataset",
+            "type": "HuggingfaceContainer",
+            "user_script": "user_script.py",
+            "components": {
+                "post_process_data": {
+                    "type": "snpe_post_process"
+                },
+                "dataloader": {
+                    "name": "huggingface_dataloader",
+                    "type": "huggingface_data_to_raw_dataloader",
+                    "params": {
+                        "io_config": {
+                            "input_names": ["input_ids", "attention_mask", "token_type_ids"],
+                            "input_types": ["int64", "int64", "int64"],
+                            "input_shapes": [[2, 128], [2, 128], [2, 128]],
+                            "output_names": ["logits"]
+                        }
+                    }
+                }
+            },
+            "params_config": {
+                "model_name": "Intel/bert-base-uncased-mrpc",
+                "task_type": "text-classification",
+                "batch_size": 2,
+                "data_name": "glue",
+                "input_cols": ["sentence1", "sentence2"],
+                "label_cols": ["label"],
+                "split": "validation",
+                "subset": "mrpc",
+                "component_kwargs": {
+                    "pre_process_data": {
+                        "max_length": 128,
+                        "padding": "max_length"
+                    }
+                }
+            }
+        }
+    },
+    "passes": {
+        "conversion": {
+            "type": "OnnxConversion",
+            "config": {
+                "target_opset": 13
+            }
+        },
+        "snpe_to_dlc": {
+            "type": "SNPEConversion",
+            "config": {
+                "input_names": ["input_ids", "attention_mask", "token_type_ids"],
+                "input_shapes": [[2, 128], [2, 128], [2, 128]],
+                "output_names": ["logits"]
+            }
+        }
+    },
+    "engine": {
+        "log_severity_level": 0,
+        "clean_cache": true,
+        "evaluator": "snpe_evaluator",
+        "evaluate_input_model": false,
+        "output_dir" : "models/bert_snpe"
+    }
+}
--- a/examples/bert/npu/bert_snpe.json
+++ b/examples/bert/npu/bert_snpe.json
@@ -49,7 +49,7 @@
             },
             "params_config": {
                 "model_name": "Intel/bert-base-uncased-mrpc",
-                "task_type": "text-classification",
+                "task": "text-classification",
                 "batch_size": 2,
                 "data_name": "glue",
                 "input_cols": ["sentence1", "sentence2"],
--- a/examples/bert/npu/bert_snpe.json
+++ b/examples/bert/npu/bert_snpe.json
@@ -4,15 +4,7 @@
         "config": {
             "hf_config": {
                 "model_name": "Intel/bert-base-uncased-mrpc",
-                "task": "text-classification",
-                "hf_config": {
-                    "data_name":"glue",
-                    "subset": "mrpc",
-                    "split": "validation",
-                    "input_cols": ["sentence1", "sentence2"],
-                    "label_cols": ["label"],
-                    "batch_size": 1
-                }
+                "task": "text-classification"
             },
             "io_config": {
                 "input_names": ["input_ids", "attention_mask", "token_type_ids"],
--- a/examples/bert/npu/user_script.py
+++ b/examples/bert/npu/user_script.py
@@ -5,7 +5,7 @@ from olive.data.registry import Registry
 def snpe_post_process(output_data, **kwargs):
     import torch
 
-    logits = torch.tensor(output_data["results"]["logits"])
+    logits = torch.tensor(output_data["logits"])
     _, preds = torch.max(logits, dim=-1)
 
     return preds
--- a/examples/bert/npu/user_script.py
+++ b/examples/bert/npu/user_script.py
@@ -1,12 +1,4 @@
-from olive.data.component.dataloader import default_dataloader
 from olive.data.registry import Registry
-from olive.platform_sdk.qualcomm.utils.data_loader import FileListCommonDataLoader
-
-
-@Registry.register_dataloader()
-def huggingface_data_to_raw_dataloader(dataset, batch_size, io_config, **kwargs):
-    data_loader = default_dataloader(dataset, batch_size=batch_size, **kwargs)
-    return FileListCommonDataLoader(data_loader, batch_size=batch_size, io_config=io_config, **kwargs)
 
 
 @Registry.register_post_process()
--- a/examples/bert/npu/user_script.py
+++ b/examples/bert/npu/user_script.py
@@ -1,0 +1,19 @@
+from olive.data.component.dataloader import default_dataloader
+from olive.data.registry import Registry
+from olive.platform_sdk.qualcomm.utils.data_loader import FileListCommonDataLoader
+
+
+@Registry.register_dataloader()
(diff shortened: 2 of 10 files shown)


examples/bert/npu/user_script.py
```diff
@@ -0,0 +1,19 @@
+from olive.data.component.dataloader import default_dataloader
+from olive.data.registry import Registry
+from olive.platform_sdk.qualcomm.utils.data_loader import FileListCommonDataLoader
+
+
+@Registry.register_dataloader()
+def huggingface_data_to_raw_dataloader(dataset, batch_size=1, io_config=None, **kwargs):
```
guotuofeng, line 7: no need the default batch_size since the caller will pass it. adding a default value will be confusing.

trajepl: Fixed.

examples/bert/npu/user_script.py
```diff
@@ -0,0 +1,19 @@
+from olive.data.component.dataloader import default_dataloader
+from olive.data.registry import Registry
+from olive.platform_sdk.qualcomm.utils.data_loader import FileListCommonDataLoader
+
+
+@Registry.register_dataloader()
+def huggingface_data_to_raw_dataloader(dataset, batch_size, io_config, **kwargs):
```
jambayk, line 7: Instead of doing this manually, maybe this is a good time to try doing it automatically in the evaluator like we do in the quantization pass 
https://github.com/microsoft/Olive/blob/main/olive/passes/snpe/quantization.py#L105

trajepl: We have that in SNPE already. Just remove the dataloader under user_scripts.

olive/evaluator/olive_evaluator.py
```diff
@@ -955,10 +955,10 @@ def _inference(
                 outputs = post_func(result)
             else:
                 raise ValueError("Post processing function is required for SNPE model")
-            preds.extend(outputs.tolist())
-            targets.extend(labels.tolist())
-            lg = result["results"].get("logits")
-            logits.extend(lg.to_list() if lg else [])
+            preds.extend(outputs.flatten().tolist())
```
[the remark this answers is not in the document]
guotuofeng: should the flatten in post_processing?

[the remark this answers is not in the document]
jambayk: That could work too but if the shape of output from the snpe inference is not as expected (batch size X sequence length X vocab), then I think it is worth looking at it first. Since we don't have any issue the pytorch and onnx evaluator with the same data.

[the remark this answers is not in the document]
jambayk: Looks like the output shape from the inference might be (batchsize X 1 X sequence length X vocab) since the original batch size becomes a part of the output shape. 

In that case, for this example we could flatten the first and second dimensions together in the post processing to make it simple. But in the long term, we could perhaps introduce a flag like `first_dim_is_batch` or something similar so that we concatenate along the first dim here https://github.com/microsoft/Olive/blob/main/olive/platform_sdk/qualcomm/snpe/tools/inference.py#L271 instead of stacking them.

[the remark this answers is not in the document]
trajepl: The accuracy metric only accept 1d-list as input. To achieve that, we can do that in `post_processing` or have a `flatten` by the bottom of evaluation. I chose the latter one in previous implementation.

As for why this is raised only in SNPE case, I think the root course is:
There are two kinds of `batch_size`:
1. from original data loader configs, let us call "batch_size"
image
2. from `FileListDataLoader(...).batch_size`, let us call "snpe_batch_size".

In SNPE case, let take `batch_size=2` as an example, 
1. the data loader firstly created by huggingface data loader`(2, sequence_length)`
2. then FileListCommonDataLoader save the numpy array  to local binary file where each file `(2, sequence_length)` contains the one iteration from data loader.
3. `snpe_batch_size` should always be 1(one file per iteration) for this kind of case, but currently, it is (None)[https://github.com/microsoft/Olive/blob/main/olive/evaluator/olive_evaluator.py#L1001C1-L1001C69]. The `None` or any numbers great than 1 for `snpe_batch_size` will expand the input shape to `[snpe_batch_size, batch_size, sequence_length]` that is why [stack](https://github.com/microsoft/Olive/blob/main/olive/platform_sdk/qualcomm/snpe/tools/inference.py#L271) here I think. It organizes the data shape into `[total_input_length/snpe_batch_size, *output_shape]`. For bert, `batch_size, classes_number` is the real output shape described in io_config.




For the flag of `first_dim_is_batch`, I think it might not work very well as that is the output from model where we cannot force user always know that.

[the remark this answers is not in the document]
jambayk: Maybe a post process function like:

```python
def snpe_post_process(output_data, **kwargs):
    import torch

    logits = torch.tensor(output_data["results"]["logits"])
    logits = logits.reshape(-1, *logits.shape[2:])
    _, preds = torch.max(logits, dim=-1)

    return preds
```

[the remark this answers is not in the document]
trajepl: Then how about the labels. :(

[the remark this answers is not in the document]
trajepl: I think it is needed. There are many case user will prepare the input list, they still need this feature. 
`[Jambayk]I added the option for a file list dataloader to have a batch size option so that we can call net run on multiple input at once`

But besides that, I am thinking, maybe we can avoid to expose the `snpe_batch_size` to users. Please help take a look this commit.
https://github.com/microsoft/Olive/pull/925/commits/551b2f359c0817fb66872522d08f4c1518b8a0f6
The idea is: user is only awareness of the single input file, olive can use `snpe_batch_size` internally run in parallel.
That means, for snpe/qnn model, if the input/output shape are specified by users, then in `post_process`, user only know the specified output shape.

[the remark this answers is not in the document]
trajepl: image

examples/bert/npu/bert_snpe.json
```diff
@@ -0,0 +1,91 @@
...
+            "name": "snpe_dataset",
+            "type": "HuggingfaceContainer",
+            "user_script": "user_script.py",
+            "components": {
+                "post_process_data": {
+                    "type": "snpe_post_process"
+                }
+            },
+            "params_config": {
+                "model_name": "Intel/bert-base-uncased-mrpc",
```
jambayk, line 51: nt: `model_name` and `task` will be auto inserted from the input model. + `task_type` is not a config param. We use `task`

trajepl: fixed.

test/unit_test/model/test_pytorch_model.py
```diff
@@ -33,7 +33,7 @@ def setup(self, tmpdir):
         self.tokenizer = transformers.AutoTokenizer.from_pretrained(self.architecture)
         self.input_text = "Today was an amazing day."
         self.hf_conf = {
-            "task_type": self.task,
+            "task": self.task,
```
jambayk, line 36: this is still `task_type` since it is being used by an external function.

olive/passes/pytorch/lora.py
```diff
@@ -454,14 +454,14 @@ def init_lora_adapters(
...
-        peft_task_type = get_peft_task_type_from_task(task, fail_on_not_found=True)
+        peft_task = get_peft_task_from_task(task, fail_on_not_found=True)
         lora_config = LoraConfig(
             r=config.lora_r,
             lora_alpha=config.lora_alpha,
             lora_dropout=config.lora_dropout,
             target_modules=target_modules,
             bias=config.bias,
-            task_type=peft_task_type,
+            task=peft_task,
```
jambayk, line 464: same. Still task type. I think we can just keep the task_type name and code the same. Just need to update the config json with the correct name.

jambayk: like in https://github.com/microsoft/Olive/pull/928

jambayk: even if we want to update the function names etc, we could do it a separate PR to avoid overloading this PR

olive/platform_sdk/qualcomm/snpe/tools/inference.py
```diff
@@ -303,7 +302,7 @@
 
     output_dict = {"latencies": latencies}
     if return_numpy_results:
-        output_dict["results"] = results
+        output_dict["results"] = [{k: results[k][i] for k in results} for i in range(len(results[output_names[0]]))]
```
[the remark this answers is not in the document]
jambayk: Would this pass the linter?

```python
[{k: v[i] for k, v in results.items()} for i in range(len(results[output_names[0]]))]
```

olive/passes/pytorch/lora.py
trajepl: I as just thinking it is about renaming, then updated it. You are right. Reverted the changes, let us update in following pr.

trajepl: Also saw your PR, so I just updated the bert_snpe.json.

olive/evaluator/olive_evaluator.py
```diff
@@ -998,7 +1009,9 @@ def _evaluate_raw_latency(
     def _prepare_dataloader(self, dataloader: Dataset, model: SNPEModelHandler) -> FileListDataLoader:
         if isinstance(dataloader, FileListDataLoader):
             return dataloader
-        return FileListCommonDataLoader(dataloader, model.io_config)
+        # batch_size=1 guarantees that the dataloader returns one input at a time. And the input has
+        # the same batch_size in original dataloader.
+        return FileListCommonDataLoader(dataloader, model.io_config, batch_size=1)
```
jambayk, line 1014: do we want to do this? I think using `None` might be better since we can finish the inference in one go? 
Or maybe provide an option to provide the file dataloader's batch size. Latency can provide 1 but accuracy could use `None`. 
You already implemented the logic to loop over the outputs anyways.

test/unit_test/model/test_pytorch_model.py
trajepl: Oh, True. I do not realize that the mlflow predictor need task_type. Could you help add a few comments in your PR?

trajepl: Anyway, as the CI takes longer time to run. We can add the comments later. Thanks.

jambayk: Already pushed the comment haha

olive/evaluator/olive_evaluator.py
trajepl: I will insist as 1 to respect the batch size from original data loader.  As in logics, we give the batch_size as 2 for example, but the real inference happens in one go. Sounds confused.

jambayk: We are respecting the original batch size since that's what makes a single input to the model. The outer batch size just defines how many of them we call the net-run on. I don't see any reason not to run as many as we can at once so that we don't have the overhead of calling net-run which has to load and call the model each time. I think the net-run loops over the inputs also so it is running them one by one.

Plus, this is an internal implementation of looping over the data. I don't think we expect the user to know this anyways. It also seems to fit with your earlier comment 
image

jambayk: for latency, I agree it should be 1 since we want to calculate it for one batch like we do for the other models.

guotuofeng: > I will insist as 1 to respect the batch size from original data loader. As in logics, we give the batch_size as 2 for example, but the real inference happens in one go. Sounds confused.

could we get the original data loader batch size? specify batch_size multiple times seems confused to us,

trajepl: Original data loader batch size cannot be assign file list data loader directly.
See this: https://github.com/microsoft/Olive/pull/925#discussion_r1477905900

Basically, file list data loader is trying to pack every iteration(batch_size, xxxx) in original data loader into a local binary file. Then user a text file(snpe_batch_size, xxx) to list the binary file as input.

Let us say here are 100 data, if original batch size is 2, then file list data loader will generate 50 binary files. 
Then if the file list batch size is 1, there will be 50 text file which contains 1 binary file path.
If the file list batch size is 10, there will be 5 text file which contains 10 binary file paths.

guotuofeng: I see. In this case, I think file_chunk_size is more appropriate than file_batch_size?

jambayk (approved):
Thanks for the changes

olive/evaluator/olive_evaluator.py
jambayk: yes, I think we can definitely rename this parameter both in the evaluator and the dataloader implementation to avoid confusion.

Could do it in this PR or in a follow up PR

trajepl: Upated.

olive/evaluator/olive_evaluator.py
```diff
@@ -1007,11 +1007,11 @@ def _evaluate_raw_latency(
         return results["latencies"]["total_inference_time"][warmup_num:]
 
     def _prepare_dataloader(
-        self, dataloader: Dataset, model: SNPEModelHandler, file_list_batch_size=None
+        self, dataloader: Dataset, model: SNPEModelHandler, file_chunk_size=None
     ) -> FileListDataLoader:
         if isinstance(dataloader, FileListDataLoader):
             return dataloader
-        return FileListCommonDataLoader(dataloader, model.io_config, batch_size=file_list_batch_size)
+        return FileListCommonDataLoader(dataloader, model.io_config, batch_size=file_chunk_size)
```
guotuofeng, line 1014: do we need rename the parameter batch_size of FileListCommonDataLoader to file_chunk_size?

trajepl: Maybe not. There are still many other cases which create the file list data loader directly. Let us say user prepare the local inputs text(like what we did for mobilenet/inception/vgg case), they still need the concept of batch_size to run batch inputs.

guotuofeng: In mobilenet/inception/vgg case, the batch_size is used as normal batch_size like dataset? I see the source code https://github.com/microsoft/Olive/blob/main/olive/platform_sdk/qualcomm/utils/data_loader.py#L73 use the batch size as chunk. No?

trajepl: For the file list data loader, the batch size always performs like `file chunk`.

I am thinking for Mobilenet case, when user create the the local file and file list data loader, then they begin to iterate the data loader. Then `batch_size` may looks better? As it is a common concept in any data loader.

trajepl: If you insist to choose the `file_chunk_size`, maybe I can do it in a separated PR? I do not know how many example configs need to be updated.

guotuofeng: does the parameter changes need to update the config json?

guotuofeng: I am just thinking the batch_size is a common term associated with dataloader. If we use same term but for different purpose, we'd better use a different name.

guotuofeng (approved):
Please split the batch_size and file_chunk_size in next PR.

PR #929, opened 2024-02-06
🏂 Inception linux snpe e2e tests

trajepl:
## Describe your changes
Inception linux snpe e2e tests

### DIFF
--- a/.azure_pipelines/job_templates/olive-example-template.yaml
+++ b/.azure_pipelines/job_templates/olive-example-template.yaml
@@ -49,6 +49,7 @@ jobs:
         scriptLocation: 'inlineScript'
         inlineScript: |
           python -m pip install pytest
+          python -m pip install azure-storage-blob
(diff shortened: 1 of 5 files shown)


.azure_pipelines/job_templates/olive-example-template.yaml
```diff
@@ -49,6 +49,7 @@ jobs:
         scriptLocation: 'inlineScript'
         inlineScript: |
           python -m pip install pytest
+          python -m pip install azure-storage-blob
```
guotuofeng, line 52: should this line be in requirements.txt?

trajepl: It is in requirements.txt before. But the azure-storage-blob is only used for e2e tests to download requirement sdk/models etc. When user run it locally, it is not needed. So I tried to move it here.

jambayk (approved):

PR #937, opened 2024-02-12
MobileNet: Update qnn ep quantization example

jambayk:
## Describe your changes
The MobileNet qnn ep quantization has been updated:
- Use `QNNPreprocess` Pass and `prepare_qnn_config`.
- Evaluation on QNN EP using an ARM64 python environment has been removed. The evaluation is out-of-date and hasn't been validated with the most recent ort-qnn package. This will be added back once the `PythonEnvironmentSystem` is updated as per #771 
  

olive/passes/onnx/quantization.py
```diff
@@ -217,6 +217,14 @@
             https://onnxruntime.ai/docs/performance/quantization.html for more details on data type selection
         """,
     ),
+    "prepare_qnn_config": PassConfigParam(
+        type_=bool,
```
devang-ml, line 221: This is only for QNN EP and it should be enabled by default.

jambayk: As discussed offline: exposing the param for all eps since the pass is accelerator agnostic so we don't have accelerator specific config parameters. QNN EP overrides the default values in `_default_config`.

devang-ml (approved):

PR #942, opened 2024-02-15
Add olive config file to AML job

xiaoyu-work:
## Describe your changes

Add olive config file to AML job

### DIFF
--- a/olive/systems/azureml/aml_system.py
+++ b/olive/systems/azureml/aml_system.py
@@ -278,6 +278,9 @@ class AzureMLSystem(OliveSystem):
         return args
 
     def _create_olive_config_file(self, olive_config: dict, tmp_dir: Path):
+        if olive_config is None:
+            return None
+
         olive_config_path = tmp_dir / "olive_config.json"
         with olive_config_path.open("w") as f:
             json.dump(olive_config, f, indent=4)
@@ -368,11 +371,15 @@ class AzureMLSystem(OliveSystem):
 
         # prepare code
         script_name = "aml_pass_runner.py"
-        olive_config_path = self._create_olive_config_file(self.olive_config, tmp_dir)
 
         cur_dir = Path(__file__).resolve().parent
         code_root = tmp_dir / "code"
-        code_files = [cur_dir / script_name, olive_config_path]
+        code_files = [cur_dir / script_name]
+
+        olive_config_path = self._create_olive_config_file(self.olive_config, tmp_dir)
+        if olive_config_path:
+            code_files.append(olive_config_path)
+
         self.copy_code(code_files, code_root)
 
         accelerator_info = {
@@ -667,11 +674,15 @@ class AzureMLSystem(OliveSystem):
         # prepare code
         script_name = "aml_evaluation_runner.py"
         tmp_dir.mkdir(parents=True, exist_ok=True)
-        olive_config_path = self._create_olive_config_file(self.olive_config, tmp_dir)
 
         cur_dir = Path(__file__).resolve().parent
         code_root = tmp_dir / "code"
-        code_files = [cur_dir / script_name, olive_config_path]
+        code_files = [cur_dir / script_name]
+
+        olive_config_path = self._create_olive_config_file(self.olive_config, tmp_dir)
+        if olive_config_path:
+            code_files.append(olive_config_path)
+
         self.copy_code(code_files, code_root)
 
         # prepare inputs
--- a/olive/systems/azureml/aml_system.py
+++ b/olive/systems/azureml/aml_system.py
@@ -76,6 +76,7 @@ def get_asset_type_from_resource_path(resource_path: ResourcePath):
 
 class AzureMLSystem(OliveSystem):
     system_type = SystemType.AzureML
+    olive_config = None
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


guotuofeng:
Would you add more descriptions for this PR? I'd like to know the scenarios that uses the olive_config in engine.

olive/systems/azureml/aml_system.py
```diff
@@ -654,6 +677,8 @@ def _create_metric_component(
         code_root = tmp_dir / "code"
         code_root.mkdir(parents=True, exist_ok=True)
         shutil.copy(str(code_file), str(code_root))
+        shutil.copy(str(olive_config_path), str(code_root))
```
guotuofeng, line 680: how the config file is used?

olive/workflows/run/run.py
```diff
@@ -153,6 +153,10 @@ def run_engine(config: RunConfig, data_root: str = None):
         config.engine.azureml_client_config = config.azureml_client
 
     engine = config.engine.create_engine()
+
+    # config file will be uploaded to AML job
```
guotuofeng, line 157: could we make this change to aml specific instead of change all the system and engine APIs?

guotuofeng (changes requested):
better not introduce one olive_config in engine and system level but only used in aml.

olive/workflows/run/run.py
xiaoyu-work: I don't think we can skip engine as host/target is only initialized when engine runs. I can try to set config file to aml system directly through engine without changing other systems.

guotuofeng: when creating system, we can set the config for AMLSystem

olive/workflows/run/run.py
```diff
@@ -153,6 +153,10 @@ def run_engine(config: RunConfig, data_root: str = None):
...
+    engine.set_olive_config(config.to_json())
```
guotuofeng, line 158: would AzureMLSystem.olive_config = config.json() accomplish similar functionality? In this way, we need change the engine anymore.

olive/systems/azureml/aml_system.py
```diff
@@ -657,9 +666,12 @@ def _create_metric_component(
 
         # prepare code
         script_name = "aml_evaluation_runner.py"
+        tmp_dir.mkdir(parents=True, exist_ok=True)
```
guotuofeng, line 669: nit: needless since the pass runner doesn't have this line?

xiaoyu-work: This is a different behavior from pass runner because actually this tmp_dir is [`tmp_dir / metric.name`](https://github.com/microsoft/Olive/blob/main/olive/systems/azureml/aml_system.py#L630) but we don't create them until `copy_code` is called.

olive/workflows/run/run.py
xiaoyu-work: This looks good to me! I'll update with this.

olive/systems/azureml/aml_system.py
```diff
@@ -276,6 +277,12 @@ def _create_model_args(self, model_json: dict, model_resource_paths: Dict[str, R
 
         return args
 
+    def _create_olive_config_file(self, olive_config: dict, tmp_dir: Path):
+        olive_config_path = tmp_dir / "olive_config.json"
+        with olive_config_path.open("w") as f:
```
guotuofeng, line 282: need check whether the olive_config is None

xiaoyu-work: Ideally olive_config should not be None for AML system. We want config for all aml jobs. But if there is any error it is None, this config json file will be null.

guotuofeng: The UT or integration test might be null

xiaoyu-work: If so, null would be expected right? What kind of check do you suggest here?

guotuofeng: If the olive_config is None, we could skip append the file to copy_files at line 375? so only cur_dir / script_name will be copied to code path?

xiaoyu-work: Updated.

olive/systems/azureml/aml_system.py
```diff
@@ -76,6 +76,7 @@ def get_asset_type_from_resource_path(resource_path: ResourcePath):
 
 class AzureMLSystem(OliveSystem):
     system_type = SystemType.AzureML
+    olive_config = None
```
jambayk, line 79: Not sure if creating this as a class attribute is a good idea since changing the value for one instance would change for other instances too. I think class attributes are generally immutable? Probably not an issue in Olive since we run workflows from the command line but still feels weird to use a class attribute here.

What about the aml system initializes with an `None` instance attribute `olive_config`. And the engine uses a method like `set_olive_config` to update the instance's value?

jambayk: There could in theory be two workflows running in parallel with aml systems. The second one would update the attribute for the first one too.

jambayk: I saw that @guotuofeng suggested this change. What is you thought on this?

guotuofeng: I'd like not change the OliveSystem interface to only support the scenario for AML. 

If you have concerns on class level, then we can move this kind of logic like what Xiaoyu previous changes. 
After creating the host/target, if it is aml then we can update there. The only issue with that approach is that the engine need hold the entire config even if it doesn't use it. 

Since we will have only one engine run at the same time, I give the suggest. 

As for two workflows, do you mean in one process there will be two different engine configs?

jambayk: I think it is okay for now since the hypothetical use case I described doesn't happen. If a process is running two workflows sequentially, then it's not an issue. Only becomes an issue if they run in parallel or the second one starts before the second one. 

For the sake of simplicity, I think the current idea is fine. We can revisit in the future if any issue arises.

guotuofeng (approved):

PR #944, opened 2024-02-16
Phi2 Optimization

zhangxiang1993:


examples/directml/llm/test_model.py
```diff
@@ -0,0 +1,101 @@
+from transformers import AutoTokenizer
```
PatriceVignola, line 1: This can be removed (keep it in a share somewhere if you want to reuse it layer on)

examples/directml/llm/run_phi2.py
```diff
@@ -0,0 +1,28 @@
+
```
PatriceVignola, line 1: This can be removed. We should be able to use the same inference script as the other models.

examples/directml/llm/phi2_hf.py
```diff
@@ -0,0 +1,49 @@
+import torch
```
PatriceVignola, line 1: This file should be removed.

examples/directml/llm/llm.py
```diff
@@ -23,6 +23,110 @@
...
+    "dense": "o_proj",
+}
+
+def map_key(origin_key):
+    for k, v in PHI_MAPPING.items():
+        if k in origin_key:
+            origin_key = origin_key.replace(k, v)
+    return origin_key
+
+def convert_weights(original_weights, mapping, config):
```
PatriceVignola, line 54: Could you move this to a file named `phi.py` and name it `convert_phi_weights`?

examples/directml/llm/llm.py
```diff
@@ -23,6 +23,110 @@
...
+    repo_id = f"{url.split('/')[3]}/{url.split('/')[4]}"
+    filename = f"{url.split('/')[-1]}"
+    hf_hub_download(
+        repo_id=repo_id,
+        filename=filename,
+        force_filename=root,
+        local_dir_use_symlinks=False,
+    )
+
+def load_phi2_checkpoint(checkpoint_path):
```
PatriceVignola, line 113: Why is this necessary? We should be able to automatically download the weights the same way we do for other LLMs, without having to interact and manually load the safetensors files.

PatriceVignola:
Make sure the `Lint / Python format (pull_request)` check passes.

PR #946, opened 2024-02-19
Refactor engine to use SystemConfig instead of OliveSystem to create the accelerator specs

guotuofeng:
## Describe your changes
* The PR just do code refactoring to facilitate the ep refactor work item to put it into system.
* In this PR, we deliberately delay the OliveSystem creation only when execution. When create accelerator spec, we will use system config to create the accelerator specs
* This PR also fix the multiple EPs evaluation bug when the host is local but target is DockerSystem. If the target is DockerSystem and host is local. the perftuning run pass need to be run in target instead of like other pass. From the time being, I choose a hack way to set the target for perf tuning pass. We need consider more better way to handle the case after this PR.
* For managed AzureMLSystem, we will first query the azureml env. If it doesn't exist, a new one will be created.
* simplify the logic of DockerSystem.
* Cleanup LocalDockerConfig.requirements_file_path since the DockerTargetUserConfig has already one
* unpin the ort version in dockerfiles used by multiple EP.
* refactor perftuning to uses PerfTuningRunner to host the underlying perf tuning logic.

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -891,7 +891,7 @@ class Engine:
                     start_time=run_cache.get("run_start_time", 0),
                     end_time=run_cache.get("run_end_time", 0),
                 )
-                logger.info(f"Loaded model from cache: {output_model_id} from f{self._run_cache_path}")
+                logger.info(f"Loaded model from cache: {output_model_id} from {self._run_cache_path}")
                 return output_model_config, output_model_id
 
         # new model id
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -922,7 +922,11 @@ class Engine:
 
             # For perf-tuning, we need the model evaluation happens in target even if it is a pass.
             # TODO(myguo): consider more better way to handle System doesn't support run pass like DockerSystem
-            if isinstance(p, OrtPerfTuning) and isinstance(host, LocalSystem):
+            if (
+                isinstance(p, OrtPerfTuning)
+                and isinstance(host, LocalSystem)
+                and not isinstance(self.target, LocalSystem)
+            ):
                 p.set_target(self.target)
 
             output_model_config = host.run_pass(p, input_model_config, data_root, output_model_path, pass_search_point)
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -113,6 +113,7 @@ class Engine:
         if self._config.clean_evaluation_cache:
             cache_utils.clean_evaluation_cache(cache_dir)
 
+        logger.info("Using cache directory: %s", cache_dir)
         self._model_cache_path, self._run_cache_path, self._evaluation_cache_path, _ = cache_utils.get_cache_sub_dirs(
             cache_dir
         )
@@ -890,7 +891,7 @@ class Engine:
                     start_time=run_cache.get("run_start_time", 0),
                     end_time=run_cache.get("run_end_time", 0),
                 )
-                logger.info(f"Loaded model from cache: {output_model_id}")
+                logger.info(f"Loaded model from cache: {output_model_id} from f{self._run_cache_path}")
                 return output_model_config, output_model_id
 
         # new model id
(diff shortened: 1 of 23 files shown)


olive/engine/engine.py
```diff
@@ -910,6 +918,13 @@ def _run_pass(
 
         run_start_time = datetime.now().timestamp()
         try:
+            from olive.passes.onnx.perf_tuning import OrtPerfTuning
+
+            # For perf-tuning, we need the model evaluation happens in target even if it is a pass.
+            # TODO(myguo): consider more better way to handle System doesn't support run pass like DockerSystem
+            if isinstance(p, OrtPerfTuning) and isinstance(host, LocalSystem):
```
trajepl, line 925: Is it possible to transfer the `target` system config into `OrtPerfTuning`'s config set? Then we can move the initialization of `target` to specific pass.

guotuofeng: this kind of logic is transparent to end user. If we choose expose it by default_config, user has to explicitly set the target. this is not the expected behavior.

olive/engine/engine.py
```diff
@@ -910,6 +919,17 @@ def _run_pass(
 
         run_start_time = datetime.now().timestamp()
         try:
+            from olive.passes.onnx.perf_tuning import OrtPerfTuning
+
+            # For perf-tuning, we need the model evaluation happens in target even if it is a pass.
+            # TODO(myguo): consider more better way to handle System doesn't support run pass like DockerSystem
+            if (
+                isinstance(p, OrtPerfTuning)
+                and isinstance(host, LocalSystem)
```
trajepl, line 928: Will there be the case: run pass in python evn system and run evaluation in aml system?

olive/engine/engine.py
```diff
@@ -1038,20 +1058,26 @@ def _get_prefix_output_name(output_name: str, accelerator_spec: "AcceleratorSpec
...
-        origin_target = self.target
-        origin_host = self.host
-        if origin_target.olive_managed_env:
-            self.target = create_new_system_with_cache(origin_target, accelerator_spec)
-        if origin_host.olive_managed_env:
-            self.host = create_new_system_with_cache(origin_host, accelerator_spec)
+        def create_system(config: "SystemConfig", accelerator_spec):
+            assert config, "System config is not provided"
+            if config.olive_managed_env:
+                logger.debug(f"Creating new system {config.type} with EP {accelerator_spec.execution_provider}")
```
trajepl, line 1064: Update the log to tell the system is from cache?

olive/systems/azureml/aml_system.py
```diff
@@ -89,29 +89,35 @@ def __init__(
         is_dev: bool = False,
         accelerators: List[str] = None,
         olive_managed_env: bool = False,
-        requirements_file: Union[Path, str] = None,
```
trajepl, line 92: Should we update python env system with corresponding system config as well?
https://github.com/microsoft/Olive/blob/main/olive/systems/python_environment/python_environment_system.py

olive/engine/engine.py
guotuofeng: I am not sure from the time being, I only enable the target from my existing knowledge.

olive/engine/engine.py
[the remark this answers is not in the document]
guotuofeng: will do

olive/systems/utils.py
```diff
@@ -64,11 +65,11 @@ def get_common_args(raw_args):
 
 
 @lru_cache(maxsize=8)
```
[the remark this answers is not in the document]
guotuofeng: I think Lei copy from other places. This doesn't matter since we always remove the system after we done.

olive/engine/engine.py
```diff
@@ -910,6 +919,17 @@ def _run_pass(
 
         run_start_time = datetime.now().timestamp()
         try:
+            from olive.passes.onnx.perf_tuning import OrtPerfTuning
+
+            # For perf-tuning, we need the model evaluation happens in target even if it is a pass.
```
jambayk, line 924: I agree this is an issue. I brought this up even during the initial design of Olive v2 that there are some passes like perf tuning which require to be run on the target. 

It is also similar for other passes like Inc quantization where they do latency evaluation. But that is a more complicated scenario than this.

guotuofeng: Yes, from the time being, it is a hack way to workaround current bug. we need consider more elegant way to fix this bug systematically.

trajepl (approved):

PR #947, opened 2024-02-20
Pass to generate ranked models compatible with tensor parallelism using orchard

shaahji:
## Describe your changes

Pass to generate ranked models compatible with tensor parallelism using orchard

### DIFF
--- a/examples/transformer/user_script.py
+++ b/examples/transformer/user_script.py
@@ -1,0 +1,25 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+import sys
+from pathlib import Path
+
+import torch
+
+wd = Path(__file__).parent.parent.parent.parent.resolve() / "orchard"
+sys.path.append(str(wd))
+
+
+def load_model(model_name):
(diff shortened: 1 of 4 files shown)


trajepl:
Could you help add the README.md to tensor parallelism example?

examples/transformer/user_script.py
```diff
@@ -0,0 +1,25 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+import sys
+from pathlib import Path
+
+import torch
+
+wd = Path(__file__).parent.parent.parent.parent.resolve() / "orchard"
```
jambayk, line 11: this path is specific to your work environment and not generalizable. 

I think it is better to give instructions in a README to tell where to clone to orchard repository from. And to set an environment variable or something to point to the cloned repo.
Ideally, I think orchard should be pip installable python package so that we can just import the package directly. But I see that the orchard repository currently doesn't have `__init__.py` files and `setup.py`.

trajepl, line 11: nit:  `wd = Path(__file__).parents[4].resolve() / "orchard"` may look more simple.

shaahji:
Abandoning it in favor of #981

PR #957, opened 2024-02-26
Add GPTQ quantization

Taka152:
## Describe your changes

Implemented GPTQ quantization utilizing Hugging Face Optimum and packed quant linear with ONNXRuntime-supported format to enable the latest 4-bit kernel (MatmulNBits), reducing quantization time to half an hour compared to several hours in INC.
Tested 4-bit quantization on winograde and humaneval with llama2 7b.
|Model|Winogrande|Humaneval
|---|---|---|
|Baseline|49.64|12.80
|GPTQ 4bit|49.17|12.20

All unittest passed.


olive/passes/pytorch/gptq.py
```diff
@@ -0,0 +1,161 @@
...
+        )
+
+        from unittest import mock
+
+        def get_onnx_quant_linear(*args, **kwargs):
+            return QuantLinearORT
+
+        # GPTQ Quantization in transformers use auto_gptq under the hood, so we
+        # replace QuantLinear in auto_gptq with QuantLinearORT for quant linear layer packing
+        with mock.patch("auto_gptq.utils.import_utils.dynamically_import_QuantLinear", get_onnx_quant_linear):
```
guotuofeng, line 139: why the production code has UT code?

Taka152: It is used to dynamic repalce 3rdlib func, I tried other method like subclass but they are not easy to implemented. This mock metho only replace the 3rdlib func in the with block, it will not influence other code.

olive/passes/pytorch/gptq.py
jambayk, line 1: Since this produces an onnx model as output, it should be under `passes.onnx`

olive/passes/pytorch/gptq.py
guotuofeng: It is not a good practice to use this way to replace the underlying function.

trajepl:
https://github.com/microsoft/Olive/commit/c093dd94628510a21e183ff0cf313f03124b13ab
Could you help also update the docs to tell this new pass? Please refer to above PR.

olive/passes/pytorch/gptq.py
jambayk, line 1: Could you also help add docs for the new pass. 

Files to updated would be:
- Docstring under the pass class here
- `docs/source/api/passes.rst`
- `features/passes/onnx.md 
Two PRs with docs for reference #861 #936

examples/llama2/README.md
```diff
@@ -15,9 +15,12 @@ Performs optimization pipeline:
 - CPU, INT4: *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model fp32 -> Onnx Block wise int4 Quantization*
 - GPU, FP16: *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model fp16 + Grouped Query Attention (optional)*
 - GPU, INT4: *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model fp16 + Grouped Query Attention (optional) -> Onnx Block wise int4 Quantization*
+- GPU, GPTQ INT4: *PyTorch Model -> GPTQ INT4 Onnx Model*
 
 **Note:** Group Query Attention is optional and can be enabled by passing `--use_gqa` flag to the script. It is only supported for GPU.
```
trajepl, line 20: Maybe we can merge this into:
```md
Note:
1. ....
2. .....
```

olive/passes/pytorch/gptq.py
```diff
@@ -0,0 +1,161 @@
...
+from olive.model import PyTorchModelHandler
+from olive.model.handler.onnx import ONNXModelHandler
+from olive.passes import Pass
+from olive.passes.onnx.common import get_external_data_config
+from olive.passes.pass_config import PassConfigParam
+from olive.passes.pytorch.gptq_utils import QuantLinearORT
+
+logger = logging.getLogger(__name__)
+
+ROUNDTRIP_CHECK = False
```
trajepl, line 20: Curious that where this flag is used?

olive/passes/pytorch/gptq.py
guotuofeng: could the following line works?
optimum.gptq.quantizer.dynamically_import_QuantLinear = get_onnx_quant_linear

Taka152: Tried several weeks ago, I'll try this again and maybe other ways to fix here.

olive/passes/pytorch/gptq.py
Taka152: Oh, It's a debug line and will be delete

olive/passes/pytorch/gptq.py
Taka152: Docs updated, and tested built htmls

Taka152: Docs updated, and tested built htmls

examples/llama2/requirements.txt
```diff
@@ -1,6 +1,7 @@
 datasets>=2.8.0
-git+https://github.com/huggingface/optimum.git
+optimum>=1.17.0
 onnx>=1.14.0
 protobuf==3.20.2
 torch -i https://download.pytorch.org/whl/nightly/cu118
 transformers>=4.33.2
+auto-gptq
```
jambayk, line 7: can you put this at the top of the list so that it is alphabetically sorted. Weird since the linter should have caught it, @guotuofeng?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,149 @@
...
+        )
+
+        def get_onnx_quant_linear(*args, **kwargs):
+            return QuantLinearORT
+
+        import auto_gptq
+
+        # GPTQ Quantization in transformers use auto_gptq under the hood, so we
+        # replace QuantLinear in auto_gptq with QuantLinearORT for quant linear layer packing
+        auto_gptq.utils.import_utils.dynamically_import_QuantLinear = get_onnx_quant_linear
```
jambayk, line 129: Perhaps a `try,finally` block would be good here like in https://github.com/microsoft/Olive/blob/9e7eba8c8d3fd500a0d0240210472c05169f17d2/olive/passes/onnx/conversion.py#L440? 
That way this function is not permanently replaced.

jambayk: We just keep a reference to the original value so that it is assigned back when our use is done.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,149 @@
...
+from typing import Any, Dict
+
+import torch
+
+from olive.hardware.accelerator import AcceleratorSpec
+from olive.model import PyTorchModelHandler
+from olive.model.handler.onnx import ONNXModelHandler
+from olive.passes import Pass
+from olive.passes.onnx.common import get_external_data_config
+from olive.passes.onnx.gptq_utils import QuantLinearORT
```
jambayk, line 15: nit: since this is a util, it is better to be delayed imported inside the `_run_for_config` method to be safe. Doesn't look like an issue here since it is only using torch and its submodules.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,149 @@
...
+                default_value=True,
+                description="Use GPU for quantization. Default value is True.",
+            ),
+            "export_optimization": PassConfigParam(
+                type_=str,
+                default_value=None,
+                description="Export optimization level. Default value is None.",
+            ),
+        }
+        config.update(get_external_data_config())
```
jambayk, line 98: since the model is being saved to file directly using the onnx exporter, these external data configs are not being used.

examples/llama2/requirements.txt
jambayk, line 7: Please also add the pass and it's dependencies (optimum, auto-gptq) to https://github.com/microsoft/Olive/blob/9e7eba8c8d3fd500a0d0240210472c05169f17d2/olive/workflows/run/run.py#L47

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,149 @@
...
+            ),
+            "dataset": PassConfigParam(
+                type_=str,
+                default_value="wikitext2",
+                description="Calibration dataset. Default value is wikitext2",
+            ),
+            "block_name_to_quantize": PassConfigParam(
+                type_=str,
+                default_value=None,
+                description="Block name to quantize. Default value is model.decoder.layers.",
```
xiaoyu-work, line 45: What is the default value here?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,149 @@
...
+
+        def get_onnx_quant_linear(*args, **kwargs):
+            return QuantLinearORT
+
+        import auto_gptq
+
+        # GPTQ Quantization in transformers use auto_gptq under the hood, so we
+        # replace QuantLinear in auto_gptq with QuantLinearORT for quant linear layer packing
+        auto_gptq.utils.import_utils.dynamically_import_QuantLinear = get_onnx_quant_linear
+        quantized_model = AutoModelForCausalLM.from_pretrained(
```
xiaoyu-work, line 130: Should we validate model_name before applying this pass?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,149 @@
...
+            model.hf_config.model_name,
+            quantization_config=gptq_config,
+            attn_implementation="eager",
+            torch_dtype=torch.half,
+        )
+
+        onnx_export_from_model(
```
xiaoyu-work, line 137: I don't think we should keep integrating optimum package func into Olive. Any specific reason we need to use this func?

examples/llama2/llama2.py
```diff
@@ -45,6 +46,12 @@ def get_args(raw_args):
         required=False,
         help="Whether to use GQA(grouped query attention) instead of MHA(multi-head attention). Only supported on gpu.",
     )
+    parser.add_argument(
+        "--use_gptq",
```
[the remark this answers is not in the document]
Taka152: I put the extra deps in this [requirements.txt](https://github.com/Taka152/Olive/blob/46b0316948921441524795c1e7f0fdec4877126c/examples/llama2/requirements.txt#L2). Is that better to put these to another requirements-gptq.txt?

olive/passes/onnx/gptq.py
Taka152: None is acceptable to pass in AutoModel.from_pretrained. I haven't use optimum quantizer directly for it requires this parameter.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,149 @@
...
+            ),
+            "desc_act": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                description="Use descriptive activation for quantization. Default value is False.",
+            ),
+            "sym": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                description="Symmetric quantization. Default value is True.",
```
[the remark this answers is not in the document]
Taka152: The default Value is False, for 4bit quant, the accuracy would be better to apply asymmetric quantization.

olive/passes/onnx/gptq.py
Taka152: Model path with hf config should be good to pass here, from_pretrained could validate this parameter and raise error?

olive/passes/onnx/gptq.py
Taka152: I tried conversion pass in olive followed by quantized pytorch model, but can't run text generation task like humaneval with expected accurary. Optimum export is used here to keep accuray combined with model.generate api.

examples/llama2/requirements.txt
trajepl: Linter will not catch it as of now. It was caught by pre-commit I guess. 
image

examples/llama2/llama2.py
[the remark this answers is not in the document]
xiaoyu-work: Yes, can you please have a separate requirements file for gptq? Similar as https://github.com/microsoft/Olive/blob/main/examples/llama2/requirements-qlora.txt. Please also update the README for this pip install.

olive/passes/onnx/gptq.py
xiaoyu-work: I mainly worry about the default values are different between actual default value and description. We should update either one to make them aligned.

olive/passes/onnx/gptq.py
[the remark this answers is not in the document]
xiaoyu-work: Same as `block_name_to_quantize`. Can we update the description here?

examples/llama2/requirements.txt
trajepl: image
Oh, sry, we also have that in linter runner. `lintrunner -a` can fix but somehow the checking passed even if the packages are not in order.

olive/passes/onnx/gptq.py
xiaoyu-work: If so, I would like to add a try catch block here to catch such error and log an accurate error message to user.

olive/passes/onnx/gptq.py
xiaoyu-work: I don't want to use our conversion pass here. What about the standard torch-onnx export api?

olive/passes/onnx/gptq.py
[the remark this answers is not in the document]
Taka152: Oh, Got it, it's just mistake to fix description when I change default values, thanks

olive/passes/onnx/gptq.py
Taka152: Didn't try, maybe not easy to do with torch onnx. As far as I know, it needs to export two graphs and merge with an if node to do prompt processing and token generation.

examples/llama2/requirements.txt
guotuofeng: > can you put this at the top of the list so that it is alphabetically sorted. Weird since the linter should have caught it, @guotuofeng?

I will let @Taka152 investigate why the lintrunner does't raise issue. If it is bug, we will contact the owner to fix.

olive/passes/onnx/gptq.py
jambayk: Yeah, it would not be straightforward to use torch.export since we need io_config and dummy input. 
One way could be to break this workflow into quantization (save torch model) and optimum conversion pass. but the torch model probably cannot be saved to disk because of the module replacement done here? 

If so, I think it's okay to use optimum export for now.

examples/llama2/requirements.txt
guotuofeng: this seems lintrunner-adapter's bug.

olive/passes/onnx/gptq.py
guotuofeng: @xiaoyu-work, https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/optimum_conversion.py#L69 use the similar manner to import Optimum functions.

@Taka152 , please add version check since the onnx_export_from_model only exist after Optimum 1.17

test/requirements-test.txt
```diff
@@ -17,7 +17,8 @@ nncf==2.7.0
 onnxconverter_common
 onnxruntime-extensions
 openvino==2023.2.0
-optimum
+optimum>=1.17.0
```
guotuofeng, line 20: sort the requirement file

olive/passes/onnx/gptq.py
xiaoyu-work: Yeah I believe one of our long term goals is to decoupling Olive from optimum. So eventually optimum conversion will be removed from Olive also. I'm okay with this for now but we should try to avoid using optimum package.

Taka152: @guotuofeng Can't use main_export because it don't support model with replaced module. It need to add onnx_export from_model func and parameters to optimum_conversion. I'll add a version check.

@jambayk I tried to breaking into two passes, it will bring changes in user_script.py to use dummpy inputs and dataloader without hf config and some other hackings to run evaluation for module replace pytorch model. Finally for a clean PR, I choose to use one pass.

Taka152: I could try other generation solution like onnxruntime-genai to decouple optimum step by step after this PR.

examples/llama2/llama2.py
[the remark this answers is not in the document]
Taka152: added a new requirements-gptq.txt

test/requirements-test.txt
Taka152: Done

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,165 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+import logging
+from typing import Any, Dict
+
+import torch
+from onnxruntime import __version__ as ort_version
```
trajepl, line 9: As we target to remove the dependencies of onnxruntime, please move it to proper function.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,165 @@
...
+        from optimum.exporters.onnx import onnx_export_from_model
+        from optimum.version import __version__ as optimum_version
+        from transformers import AutoModelForCausalLM, AutoTokenizer, GPTQConfig
+
+        from olive.passes.onnx.gptq_utils import QuantLinearORT
+
+        if version.parse(optimum_version) < version.parse("1.17.0"):
+            raise ValueError("Please use optimum>=1.17.0 for gptq quantization")
+        if version.parse(ort_version) < version.parse("1.17.0"):
+            raise ValueError("Please use onnxruntime-gpu>=1.17.0 for gptq quantization")
```
guotuofeng, line 112: if the pass is only available for GPU we need check the accelerator is GPU

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,165 @@
...
+            # GPTQ Quantization in transformers use auto_gptq under the hood, so we
+            # replace QuantLinear in auto_gptq with QuantLinearORT for quant linear layer packing
+            auto_gptq.utils.import_utils.dynamically_import_QuantLinear = get_onnx_quant_linear
+            quantized_model = AutoModelForCausalLM.from_pretrained(
+                model.hf_config.model_name,
+                quantization_config=gptq_config,
+                attn_implementation="eager",
+                torch_dtype=torch.half,
+            )
+        except ValueError:
```
guotuofeng, line 144: if it is only for logging purpose and re-raise, then the except is needless.

Taka152: Yeah, I think transformers will give accurate error info about the model_name_or_path parameter, the exception here to catch this error may not help much. Is that ok for your [previous suggestion](https://github.com/microsoft/Olive/pull/957#discussion_r1503563880)? @xiaoyu-work

olive/extra_dependencies.json
```diff
@@ -12,7 +12,7 @@
         "onnxruntime"
     ],
     "gpu": [
-        "onnxruntime-gpu"
+        "onnxruntime-gpu>=1.17.0"
```
jambayk, line 15: I don't think we should pin the version here. the version requirements are example/pass specific. 

Instead, you can just add the required packages for the new pass directly to `pass_to_extra` instead of `pass_to_extra_names`.

guotuofeng: yeah, better not to pin the version from production code.

olive/extra_dependencies.json
```diff
@@ -28,7 +28,7 @@
         "neural-compressor"
     ],
     "optimum": [
-        "optimum"
+        "optimum>=1.17.0"
```
guotuofeng, line 31: ditto

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,170 @@
...
+
+        from olive.passes.onnx.gptq_utils import QuantLinearORT
+
+        if version.parse(optimum_version) < version.parse("1.17.0"):
+            raise ValueError("Please use optimum>=1.17.0 for gptq quantization")
+        if version.parse(ort_version) < version.parse("1.17.0"):
+            raise ValueError("Please use onnxruntime-gpu>=1.17.0 for gptq quantization")
+        if (
+            self.accelerator_spec.accelerator_type != Device.GPU
+            or self.accelerator_spec.execution_provider != "CUDAExecutionProvider"
```
guotuofeng, line 115: do we need support tensorRT?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,170 @@
...
+                type_=bool,
+                default_value=False,
+                description="Use descriptive activation for quantization. Default value is False.",
+            ),
+            "sym": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                description="Symmetric quantization. Default value is False.",
+            ),
+            "gpu": PassConfigParam(
```
guotuofeng, line 86: could we use self.accelerator_spec.accelerator_type?

olive/passes/onnx/gptq.py
Taka152: I think it's not supported for now because we use our matmulnbits kernel in ort.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,170 @@
...
+                type_=int,
+                default_value=128,
+                description="number of samples in calibration dataset to apply quantization. Default value is 128",
+            ),
+            "bits": PassConfigParam(
+                type_=int,
+                default_value=4,
+                description="quantization bits. Default value is 4",
+            ),
+            "dataset": PassConfigParam(
```
guotuofeng, line 36: do we need unify with data_config? @trajepl ?

olive/passes/onnx/gptq.py
Taka152: Cause we only support cuda for now, I think this ConfigParam could be deleted?

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,308 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+import logging
+import math
+import os
+
+import torch
+import torch.nn as nn
+from torch.onnx.symbolic_helper import _get_tensor_dim_size, _get_tensor_sizes
```
guotuofeng, line 11: are there any public function? if it is private, the signature or name might be changed from different version.

guotuofeng (changes requested):
use accelerator instead of separated gpu

olive/passes/onnx/gptq.py
xiaoyu-work: I'd like to add some Olive specific error messages like `"{model.hf_config.model_name} is not supported. GptqQuantizer pass only supports auto-regressive language models."`

olive/passes/onnx/gptq_utils.py
Taka152: These func is used to remove `no shape inference` error when export, I didn't find other function, is there any better way to add shape info?

olive/passes/onnx/gptq.py
guotuofeng: for this kinds of ValueError or general exception, there is no way to figure out whether it is parameter issue or model is not supported since it is too general exception type.

On the other hand, only add logging and rethrown exception is needless since the caller will catch it and handle it.

Taka152:
I am tring to using PytorchModelHandler to suport more general input models, leaving the model check to this class. And also use olive data loader to support custom dataset, it may take one or two days. Thanks for your patience.

test/unit_test/passes/onnx/test_gptq.py
```diff
@@ -0,0 +1,75 @@
...
+@pytest.mark.skipif(
+    not torch.cuda.is_available(),
+    reason="gptq requires GPU.",
+)
+def test_gptq_default(tmp_path: Path):
+    # setup
+    input_model = PyTorchModelHandler(
+        hf_config={
+            "model_class": "OPTForCausalLM",
+            "model_name": "facebook/opt-125m",
```
jambayk, line 25: do you have an estimate for how long this test would take to run on this model? 
Could you try a smaller test model like `"hf-internal-testing/tiny-random-OPTForCausalLM"` like in https://github.com/microsoft/Olive/blob/9d1add1b94304c41ea3fdb0e118cee726c0c5819/test/unit_test/passes/onnx/test_optimum_conversion.py#L49

test/unit_test/passes/onnx/test_gptq.py
```diff
@@ -0,0 +1,75 @@
...
+from olive.model.handler.pytorch import PyTorchModelHandler
+from olive.passes.olive_pass import create_pass_from_dict
+from olive.passes.onnx import GptqQuantizer
+
+
+@pytest.mark.skipif(
+    not torch.cuda.is_available(),
+    reason="gptq requires GPU.",
+)
+def test_gptq_default(tmp_path: Path):
```
jambayk, line 20: this is a nit: since the test code is mostly the same, maybe the two cases can be parameterized using a flag like `dataset_source` or something?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,202 @@
...
+        if (
+            self.accelerator_spec.accelerator_type != Device.GPU
+            or self.accelerator_spec.execution_provider != "CUDAExecutionProvider"
+        ):
+            raise ValueError("Please use GPU and CUDAExecutionProvider to run gptq quantization.")
+
+        tokenizer = AutoTokenizer.from_pretrained(model.hf_config.model_name, use_fast=False)
+
+        dataset = None
+        if config["dataloader_func"]:
```
jambayk, line 143: thanks for integrating support for olive data config and custom dataloader!

test/unit_test/passes/onnx/test_gptq.py
Taka152: It takes about 200s for two tests on v100, I tried "hf-internal-testing/tiny-random-OPTForCausalLM" model but encounter errors may caused by too small size for int4 weight packing.

olive/passes/onnx/gptq.py
Taka152: Thanks for your great work, really like the huggingface dataset integration and custom dataset with different join strategy. It's smooth to use them.

test/unit_test/passes/onnx/test_gptq.py
jambayk: i see, I was wondering if the small model would be compatible or not. 
our gpu pool is v100 so 3 minutes is not too bad. 

Is there a min limit for the dataset size? Reducing this probably won't help much since the default dataset is still full length?

Taka152: Yeah, small models may not be compatible for some current implementation limitations when do 4bit quantization, and are not needed to be tested? Cause we are focusing on llm and CUDA for now, and I think 125m is really small in such a big model time. 🤣

The dataset could be smaller to just 1 sample, the test time only changes a few seconds.

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,308 @@
...
+
+logger = logging.getLogger(__name__)
```
guotuofeng, line 13: useless?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,202 @@
...
+
+from olive.common.config_utils import validate_config
+from olive.data.config import DataConfig
+from olive.hardware.accelerator import AcceleratorSpec, Device
+from olive.model import PyTorchModelHandler
+from olive.model.handler.onnx import ONNXModelHandler
+from olive.passes import Pass
+from olive.passes.pass_config import PassConfigParam
+
+logger = logging.getLogger(__name__)
```
guotuofeng, line 19: useless?

olive/passes/onnx/gptq_utils.py
guotuofeng: could we just copy these private function here? I am afraid the private function signature changes would break Olive in future.

olive/passes/onnx/gptq.py
guotuofeng: do we need update the type like [https://github.com/microsoft/Olive/blob/2374ab9b5d8a70829667f9ada09a94c7a3be76c2/olive/passes/pytorch/lora.py#L194](https://github.com/microsoft/Olive/blob/2374ab9b5d8a70829667f9ada09a94c7a3be76c2/olive/passes/pytorch/lora.py#L194)

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,199 @@
...
+                description="Export optimization level. Default value is None.",
+            ),
+            "data_config": PassConfigParam(
+                type_=Union[DataConfig, Dict],
+                default_value=None,
+                description="""
+                    Data config for quantization. Default value is None.
+                """,
+            ),
+            "dataloader_func": PassConfigParam(
```
guotuofeng, line 99: >dataloader_func [](http://example.com/codeflow?start=13&length=15)

the dataloader_func seems duplicate with dataset param?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,199 @@
...
+                type_=Union[Callable, str],
+                default_value=None,
+                description="""Function/function name to generate dataset for quantization.
+                The returned datasets is a list of tokenized data
+                (e.g. [{ 'input_ids': [ 1, 100, 15, ... ],'attention_mask': [ 1, 1, 1, ... ]},...]).
+                Default is None.
+                """,
```
guotuofeng, line 106: @trajepl, I suppose we suggest to use data_config. For new added pass, the dataloader_func is not need supported? I suppose data_config could accomplish similar functionality?

Taka152: I think dataloader_func is useful to pass more custom data without need to meet data_config. Cause sometimes user already have tokenized data.

olive/passes/onnx/gptq.py
Taka152: I'm not sure about this. In my usage, dataset, data_config and dataloader_func may serve different usage.
dataset is usefull when user only want try a few already tested open source datasets, they only just pass a string,.
data_config can use huggingface dataset with different predifined strategy and process to use.
dataloader_func can be passed by user's predefined dataset function in their training code to easily use this olive pass

olive/passes/onnx/gptq_utils.py
Taka152: I see, I will check here, or just delete these funcs, cause they just omit some ugly export warning.

olive/passes/onnx/gptq.py
trajepl: In this PR, it is fine to keep them all.
Could you help add a few TODO comments like:
[TODO](trajepl): consider to use data_config to implement the functionality of dataloader_func. 

I will try to refactor that.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,200 @@
...
+
+        from olive.passes.onnx.gptq_utils import QuantLinearORT
+
+        if version.parse(optimum_version) < version.parse("1.17.0"):
+            raise ValueError("Please use optimum>=1.17.0 for gptq quantization")
+        if version.parse(ort_version) < version.parse("1.17.0"):
+            raise ValueError("Please use onnxruntime-gpu>=1.17.0 for gptq quantization")
+        if (
+            self.accelerator_spec.accelerator_type != Device.GPU
+            or self.accelerator_spec.execution_provider != "CUDAExecutionProvider"
```
guotuofeng, line 134: move these validation logic into validate_search_point.
EP need not be checked since we are not using it.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,200 @@
...
+            return QuantLinearORT
+
+        import optimum
+
+        original = optimum.gptq.quantizer.dynamically_import_QuantLinear
+        try:
+            # GPTQ Quantization in transformers use optimum under the hood, so we
+            # replace QuantLinear in optimum with QuantLinearORT for quant linear layer packing
+            optimum.gptq.quantizer.dynamically_import_QuantLinear = get_onnx_quant_linear
+            quantized_model = quantizer.quantize_model(pytorch_model, tokenizer)
```
guotuofeng, line 178: >quantized_model = quantizer.quantize_model(pytorch_model, tokenizer) [](http://example.com/codeflow?start=12&length=68)

add comments about how the underlying device is used.
For e.g. force move the layer to cuda and when quantization finish, move it back to cpu by default.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,200 @@
...
+        finally:
+            optimum.gptq.quantizer.dynamically_import_QuantLinear = original
+
+        # save_pretrained in export will raise error when pass custom dataset, change this to str before save
+        if config["dataloader_func"]:
+            quantized_model.config.quantization_config["dataset"] = "olive_" + str(config["dataloader_func"])
+        elif config["data_config"]:
+            quantized_model.config.quantization_config["dataset"] = "olive_" + config["data_config"]["name"]
+
+        onnx_export_from_model(
```
guotuofeng, line 188: >onnx_export_from_model [](http://example.com/codeflow?start=8&length=22)

nit: add assert device == Device.GPU.

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,299 @@
...
+        zeros_cuda = (zeros_cuda.byte() & max_num_in_bits).int()
+        qzeros_cuda = torch.zeros(
+            (intzeros.shape[0], intzeros.shape[1] // 32 * self.bits), dtype=torch.int32, device=device
+        )
+        qzeros_cuda = qzeros_cuda.T.contiguous()
+        zeros_cuda = zeros_cuda.T.contiguous()
+        pack_on_row_fast_248bit(qzeros_cuda, zeros_cuda, self.bits)
+        self.qzeros = qzeros_cuda.T.contiguous().cpu()
+
+    def pack_on_device_for_even_bits(self, intweight_gpu, intzeros_T):
```
guotuofeng, line 248: >pack_on_device_for_even_bits [](http://example.com/codeflow?start=8&length=28)

make private method

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,299 @@
...
+        assert self.qzeros.shape == intzeros_pt.shape
+
+        self.scales = scales_pt.contiguous()
+        self.qweight = intweight_pt_t.contiguous().byte()
+        self.qzeros = intzeros_pt.contiguous().byte()
+
+    def accelerate_pack_on_device(self, linear, scales, zeros, g_idx=None, device="cuda"):
+        scales = scales.to(device)
+        zeros = zeros.to(device)
+        layer_weight = linear.weight.data.to(device)
```
guotuofeng, line 284: as we discussed, make these logic device agnostic by depends on the input tensor devices instead of explicitly to the device in the method.

olive/passes/onnx/gptq.py
```diff
@@ -113,28 +116,34 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
...
+        from optimum.version import __version__ as optimum_version
+
+        if version.parse(optimum_version) < version.parse("1.17.0"):
+            logger.info("Please use optimum>=1.17.0 for gptq quantization")
+            return False
+        if version.parse(ort_version) < version.parse("1.17.0"):
+            logger.info("Please use onnxruntime-gpu>=1.17.0 for gptq quantization")
+            return False
+        if self.accelerator_spec.accelerator_type != Device.GPU:
+            logger.info("Please use GPU and CUDAExecutionProvider to run gptq quantization.")
```
guotuofeng, line 132: remove EP?

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,273 @@
...
+
+        if self.orig_fp_weight is not None:
+            fw, _, _ = self._unpack()
+            assert (fw == self.orig_fp_weight).all()
+
+    def _pack_on_device_for_even_bits(self, intweight, intzeros_T):
+        self.act_order = self.g_idx[: self.groupsize // self.bits].sum().item() != 0
+        assert self.act_order is False, "please set the pack_model to others like [GPTQ, AUTO] and try again."
+        intzeros_pt = intzeros_T.T.byte()
+        scales_pt = self.scales.T.to(intweight.device)
```
guotuofeng, line 233: >scales_pt = self.scales.T.to(intweight.device) [](http://example.com/codeflow?start=8&length=46)

do we need the tensor.to from cpu to cuda?

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,273 @@
...
+        out = out.T.contiguous()
+    pack_tensor.copy_(out)
+
+
+def general_pack_on_row(pack_tensor, ori_int32_tensor, bits):
+    assert pack_tensor.shape[0] == ori_int32_tensor.shape[0] or pack_tensor.shape[1] == ori_int32_tensor.shape[1], ""
+    pack_tensor.mul_(0)
+    if bits in [2, 4, 8]:
+        return pack_on_row_fast_248bit(pack_tensor, ori_int32_tensor, bits)
+    return pack_on_row_fast_anybit(pack_tensor, ori_int32_tensor, bits)
```
guotuofeng, line 129: the two functions doesn't have return value.

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,273 @@
...
+            self.bias = None
+
+    def forward(self, x):
+        out = quant_linear_forward(
+            x, self.qweight, self.scales, self.qzeros, self.bits, self.groupsize, self.infeatures, self.outfeatures
+        )
+        return out + self.bias if self.bias is not None else out
+
+    def pack(self, linear, scales, zeros, g_idx=None):
+        return self._accelerate_pack_on_device(linear, scales, zeros, g_idx)
```
guotuofeng, line 169: >_accelerate_pack_on_device [](http://example.com/codeflow?start=20&length=26)

the method doesn't have return value.

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,273 @@
...
+
+        self.g_idx = g_idx.clone() if g_idx is not None else self.g_idx
+        intweight = self._quant_weight(layer_weight, scales, zeros, g_idx, need_transpose=False)
+
+        qzeros = zeros.t().contiguous()
+
+        if self.bits in [2, 4, 8]:
+            return self._pack_on_device_for_even_bits(intweight, qzeros)
+        else:
+            return self._pack_on_device_for_odd_bits(intweight, qzeros)
```
guotuofeng, line 273: no return value either

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,273 @@
...
+        )
+
+        qzeros = qzeros.T.contiguous()
+        zeros = zeros.T.contiguous()
+        general_pack_on_row(qzeros, zeros, self.bits)
+
+        self.qzeros = qzeros.T.contiguous()
+
+    # odd bits, 3,5,6,7
+    def _pack_on_device_for_odd_bits(self, intweight, intzeros):
```
guotuofeng, line 214: >device [](http://example.com/codeflow?start=17&length=6)

the name need device specific or not?

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,273 @@
...
+    wf1 = torch.arange(0, 32, 1).to(pack_tensor.device).view(1, 1, -1)
+    out = torch.bitwise_left_shift(out, wf1)
+    out = out.sum(dim=-1).int()
+
+    if need_transpose:
+        out = out.T.contiguous()
+    pack_tensor.copy_(out)
+
+
+def general_pack_on_row(pack_tensor, ori_int32_tensor, bits):
```
guotuofeng, line 124: >general_pack_on_row [](http://example.com/codeflow?start=4&length=19)

add documents for this function

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def _pack_on_row_fast_248bit(self, pack_tensor, ori_int_tensor, bits):
+        if pack_tensor.shape[0] == ori_int_tensor.shape[0]:
+            ori_int_tensor = ori_int_tensor.T
+            pack_tensor = pack_tensor.T
+        compress_ratio = 32 // bits
+        i = 0
+        row = 0
+        while row < pack_tensor.shape[0]:
+            if bits in [2, 4, 8]:
```
guotuofeng, line 179: >if bits in [2, 4, 8]: [](http://example.com/codeflow?start=12&length=21)

validate the arguments before the loop?

if bits not in  [2,4,8]:
    raise ValueError("Only 2,4,8 bits are supported.")

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,243 @@
...
+                for j in range(i, i + compress_ratio):
+                    pack_tensor[row:] |= ori_int_tensor[j::compress_ratio] << (bits * (j - i))
+                break
+            raise NotImplementedError("Only 2,4,8 bits are supported.")
+
+    def _pack_on_row_fast_anybit(self, pack_tensor, ori_int_tensor, bits):
```
guotuofeng, line 185: >self [](http://example.com/codeflow?start=33&length=4)

seems self not used? if so, better to make it static or extract as function

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,243 @@
...
+        scales_pt = scales_pt.reshape(-1)
+
+        assert self.qweight.shape == intweight_pt_t.shape
+        assert self.qzeros.shape == intzeros_pt.shape
+
+        self.scales = scales_pt.contiguous()
+        self.qweight = intweight_pt_t.contiguous().byte()
+        self.qzeros = intzeros_pt.contiguous().byte()
+
+    def _pack_on_row_fast_248bit(self, pack_tensor, ori_int_tensor, bits):
```
guotuofeng, line 171: >self [](http://example.com/codeflow?start=33&length=4)

ditto.

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,243 @@
...
+            ((intweight.shape[0] * self.bits + 31) // 32, intweight.shape[1]), dtype=torch.int32, device=device
+        )
+
+        self._general_pack_on_row(qweight, intweight, self.bits)
+        self.qweight = qweight
+
+        self._pack_qzeros_odd(intzeros)
+
+        if self.orig_fp_weight is not None:
+            fw, _, _ = self._unpack()
```
guotuofeng, line 224: >_unpack [](http://example.com/codeflow?start=28&length=7)

cannot find it.

olive/passes/onnx/gptq_utils.py
```diff
@@ -0,0 +1,243 @@
...
+    assert bits == 4, "Only 4 bits are supported."
+    return QuantLinearTorchFunction().apply(inputs, qweight, scales, qzeros, bits, groupsize, in_features, out_features)
+
+
+class QuantLinearORT(nn.Module):
+    # pylint: disable=W0201
+    def __init__(self, bits, groupsize, infeatures, outfeatures, bias, *args, **kwargs):
+        super().__init__()
+        if bits not in [2, 3, 4, 5, 6, 7, 8]:
+            raise NotImplementedError("Only 2,4,5,6,7,8 bits are supported.")
```
guotuofeng, line 74: >NotImplementedError [](http://example.com/codeflow?start=18&length=19)

basically, validate arguments raise ValueError is the best practice.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,213 @@
...
+logger = logging.getLogger(__name__)
+
+
+class GptqQuantizer(Pass):
+    """GPTQ quantization using Hugging Face Optimum and export model with onnxruntime optimized kernel."""
+
+    _requires_user_script = True
+
+    @staticmethod
+    def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
```
guotuofeng, line 28: please rebase from main the signature is changed to classmethod

guotuofeng: Please change the method to classmethod.

Taka152: This file should be move to pytorch/gptq.py, I've changed in the new file

olive/passes/onnx/optimum_conversion.py
```diff
@@ -48,18 +48,30 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
                 default_value=None,
                 description="Extra arguments to pass to the `optimum.exporters.onnx.main_export` function.",
             ),
+            "custom_model": PassConfigParam(
```
guotuofeng, line 51: >custom_model [](http://example.com/codeflow?start=13&length=12)

should we use another pass for the conversion? It seems like a different conversion strategy.

olive/passes/pytorch/gptq.py
```diff
@@ -0,0 +1,219 @@
...
+                type_=int,
+                default_value=4,
+                description="quantization bits. Default value is 4",
+            ),
+            "layers_block_name": PassConfigParam(
+                type_=str,
+                default_value="model.layers",
+                description="Block name to quantize. Default value is model.layers.",
+            ),
+            "outside_layer_modules ": PassConfigParam(
```
guotuofeng, line 44: >  [](http://example.com/codeflow?start=34&length=1)

extra space.

olive/passes/pytorch/gptq.py
```diff
@@ -0,0 +1,219 @@
...
+                type_=List[str],
+                default_value=None,
+                description="Names of other nn modules that in the same level as the transformer layer block. "
+                "Default value is None.",
+            ),
+            "inside_layer_modules  ": PassConfigParam(
```
guotuofeng, line 50: >   [](http://example.com/codeflow?start=33&length=2)

ditto.

olive/passes/pytorch/gptq.py
```diff
@@ -0,0 +1,219 @@
...
+            # will move each block(layer) to cuda before quantization and move back to cpu when finished.
+            quantized_model.quantize(dataset)
+        finally:
+            auto_gptq.modeling._utils.dynamically_import_QuantLinear = original  # pylint: disable=protected-access
+
+        quantized_model = quantized_model.model
+        assert self.accelerator_spec.accelerator_type == Device.GPU
+
+        output_model_path = Path(output_model_path).with_suffix(".pt")
+        torch.save(quantized_model, output_model_path)
```
guotuofeng, line 208: >torch.save(quantized_model, output_model_path) [](http://example.com/codeflow?start=8&length=46)

would the entire model need the pytorch code when loading?

olive/passes/onnx/optimum_conversion.py
Taka152: It seems has same args and similar func with [main_export](https://github.com/huggingface/optimum/blob/d03ab100206cb9f0e62167a36ee6997424bb9bb5/optimum/exporters/onnx/__main__.py#L47) but [onnx_export_from_model](https://github.com/huggingface/optimum/blob/d03ab100206cb9f0e62167a36ee6997424bb9bb5/optimum/exporters/onnx/convert.py#L901) accept a modified loaded model.

olive/passes/pytorch/gptq.py
Taka152: It needs custom module in olive to load

docs/source/features/passes/onnx.md
```diff
@@ -267,6 +267,15 @@ Olive consolidates the Vitis™ AI quantization into a single pass called VitisA
 
 Please refer to [VitisAIQuantization](vitis_ai_quantization) for more details about the pass and its config parameters.
 
+### Quantize with Huggingface Optimum
+Olive also integrates [Hugging Face Optimum](https://github.com/huggingface/optimum) for quantization.
+
+🤗 Optimum collaborated with AutoGPTQ library to provide a simple API that apply GPTQ quantization on language models. With GPTQ quantization, you can quantize your favorite language model to 8, 4, 3 or even 2 bits. This comes without a big drop of performance and with faster inference speed. This is supported by most GPU hardwares.
+
+Olive consolidates the Optimum gptq quantization into a single pass called GptqQuantizer which supports tune gptq quantization with hyperparameters for trade-off between accuracy and speed.
```
trajepl, line 275: Better to align the gptq and GPTQ, all uppercase or all lowercase

olive/passes/onnx/optimum_conversion.py
```diff
@@ -48,18 +48,30 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
...
             search_point = self.config_at_search_point(search_point or {})
 
         if search_point.get("fp16") and search_point.get("device") != "cuda":
             logger.info("OptimumConversion: fp16 is set to True, but device is not set to cuda.")
             return False
 
+        if search_point.get("custom_model") and version.parse(optimum_version.__version__) < version.parse("1.17.0"):
+            logger.info("Custom model export is only supported in Optimum version 1.17.0 or later.")
+            return False
+
```
trajepl, line 74: Maybe move this part of filter to _run_for_config?  Just ignore custome_model when optimum version is not expected. Then logger.warning to tell users.

Taka152:
@jambayk I've made some changes to remove optimum dependencies and use optimum_conversion pass to do export. GPTQ pass would save pytorch model like [TorchTRTConversion](https://github.com/microsoft/Olive/blob/2eb9f384db86bc0f1cf33b0b2c9cdbc2602d7078/olive/passes/pytorch/torch_trt_conversion.py#L183) pass. Could you help me review these changes?

olive/passes/pytorch/gptq.py
```diff
@@ -0,0 +1,219 @@
...
```
jambayk, line 208: This looks good to me. Just to be sure, the model when loading needs olive installed since it has modules that are defined in olive? 
Not an issue since we expect to convert the model to onnx after this pass anyways.

olive/passes/pytorch/gptq.py
```diff
@@ -0,0 +1,219 @@
...
+
+        model_config = model.to_json()["config"]
```
jambayk, line 210: We might to to delete `model_config["model_loader"] too since it takes precedence over the entire pytorch model format https://github.com/microsoft/Olive/blob/main/olive/model/handler/pytorch.py#L130.

olive/passes/pytorch/gptq.py
jambayk: I think this is a valid concern. Since the model is saved and loaded as an "entire model", the model load code won't have loaded the user script and imported it necessarily (especially for remote systems where pass runs and model evaluations don't happen in the same python process). 

If used loading the user script is enough, then we could move this line outside the `if` clause so that the userscript is always loaded https://github.com/microsoft/Olive/blob/2eb9f384db86bc0f1cf33b0b2c9cdbc2602d7078/olive/model/handler/pytorch.py#L131

jambayk: I just did some testing, I can't do `torch.save` on a model class that is defined in the user_script. It only works if the model class is defined in some external package like transformers.

jambayk: Looks like if we break this into two steps, we can only support models with hf_config like in the sparsegptconversion pass. 

If that is the case, maybe the previous process of just converting the model in this pass is better since it doesn't give much advantage.

olive/passes/onnx/optimum_conversion.py
guotuofeng: yes, that's the reason I say they are different loading strategy. @jambayk, do you think if it need to be in another separated pass instead of by using a hacky flag?

jambayk: Like I mentioned in this comment https://github.com/microsoft/Olive/pull/957#discussion_r1522208907, if the gptq pass is limited to hf models, then I think the previous method of quantizing and exporting the model in the same pass is good enough? 

There is not much to be gained if we cannot save a quantized custom model class.

jambayk: If want the keep the workflow of gptq quantized pytorch model -> export, I don't have strong opinions about using a flag in this pass or creating a new pass. Looks about the same to me.

Taka152: I think it should be not limited to hf modelsI made it to two steps only to meet a 1p customer's need, they have custom model and custom export and need gptq quantization. Is there better way to do that? Can I import the conversion code from OnnxConversion to do the custom export?

jambayk: If the `torch.save` method works for custom model, I think it is fine to make it two steps. It's just that in my local testing, I could not save a custom model I defined in a user script. 

For the optimum export question Mike brought up, I don't have opinions about a new optimum part is needed or not.

jambayk: If the custom model save doesn't work, I think the gptq quantization pass could maybe provide an option to use optimum exporter or torch.onnx exporter. The export code in the conversion pass is a static method so it can be called directly if needed https://github.com/microsoft/Olive/blob/2eb9f384db86bc0f1cf33b0b2c9cdbc2602d7078/olive/passes/onnx/conversion.py#L145. 

This does mean duplicate work for export to ideally the two step workflow makes more sense.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,254 @@
...
+            ),
+            "use_dynamo_exporter": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                description="Whether to use dynamo_export API to export ONNX model when export_optimum=False.",
+            ),
+            "target_opset": PassConfigParam(
+                type_=int,
+                default_value=13,
+                description="The version of the default (ai.onnx) opset to target when export_optimum=False.",
```
guotuofeng, line 140: do we need them?

Taka152: It is needed in config because the export func in conversion extract it from config https://github.com/microsoft/Olive/blob/9d780ea9c356c4fe0bf4eb8971c92f412b1f47fe/olive/passes/onnx/conversion.py#L262-L271

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,254 @@
...
+            auto_gptq.modeling._utils.dynamically_import_QuantLinear = original  # pylint: disable=protected-access
+
+        quantized_model = quantized_model.model
+        assert self.accelerator_spec.accelerator_type == Device.GPU
+
+        if config["export_optimum"]:
+            onnx_export_from_model(
+                model=quantized_model,
+                output=output_model_path,
+                device="cuda",
```
jambayk, line 237: "cuda" assumes that the user has enough gpu memory to run the conversion on the GPU but some models consume a large amount of memory. Is it possible to let this configurable like in conversion pass using `use_device` which defaults to cpu? or the quantized modules need to be on gpu?

olive/passes/onnx/gptq.py
```diff
@@ -149,17 +149,15 @@ def _run_for_config(
...
-        if config["export_optimum"]:
-            from optimum.version import __version__ as optimum_version
-
-            if version.parse(optimum_version) < version.parse("1.17.0"):
-                raise ValueError("Please use optimum>=1.17.0 for optimum export")
-            from optimum.exporters.onnx import onnx_export_from_model
+        if config["quantization_verbose"]:
+            auto_gptq_logger = logging.getLogger("auto_gptq")
+            auto_gptq_logger.addHandler(logging.StreamHandler(sys.stdout))
+            auto_gptq_logger.setLevel(logging.INFO)
```
guotuofeng, line 160: better config the logger in example code?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,229 @@
...
+            ),
+            "dataloader_func_kwargs": PassConfigParam(
+                type_=Dict[str, Any],
+                default_value=None,
+                description="Keyword arguments for dataloader_func. Default value is None.",
+            ),
+            "use_dynamo_exporter": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                description="Whether to use dynamo_export API to export ONNX model when export_optimum=False.",
```
jambayk, line 121: `export_optimum` doesn't exist anymore.

test/unit_test/passes/onnx/test_optimum_conversion.py
```diff
@@ -101,3 +101,16 @@ def test_optimum_configs(config, is_valid, tmp_path):
         assert p.validate_search_point(config, None)
         onnx_model = p.run(input_model, None, output_folder)
         assert Path(onnx_model.model_path).exists()
+
+
+def test_optimum_export_from_model(tmp_path):
```
jambayk, line 106: this is not relevant anymore.

test/unit_test/passes/onnx/test_gptq.py
```diff
@@ -0,0 +1,50 @@
...
+    p = create_pass_from_dict(
+        GptqQuantizer,
+        config,
+        disable_search=True,
+        accelerator_spec=AcceleratorSpec(accelerator_type=Device.GPU, execution_provider="CUDAExecutionProvider"),
+    )
+    gptq_out_folder = str(tmp_path / "gptq")
+
+    # execute
+    p.run(input_model, None, gptq_out_folder)
```
trajepl, line 50: Can we add a few assertion to check if the output is expected or not? Like the output model should exist and it's size should be less than the original model's size.

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,229 @@
...
+import tempfile
+from typing import Any, Callable, Dict, List, Union
+
+import torch
+
+from olive.common.config_utils import validate_config
+from olive.data.config import DataConfig
+from olive.hardware.accelerator import AcceleratorSpec, Device
+from olive.model import PyTorchModelHandler
+from olive.model.handler.onnx import ONNXModelHandler
```
guotuofeng, line 14: >ONNXModelHandler [](http://example.com/codeflow?start=37&length=16)

nit: could be merged with line 13

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,229 @@
...
+            "target_opset": PassConfigParam(
+                type_=int,
+                default_value=13,
+                description="The version of the default (ai.onnx) opset to target. Default value is 13.",
+            ),
+            "export_device": PassConfigParam(
+                type_=str,
+                default_value="cpu",
+                description="The device to export the model. Default value is 'cpu'.",
+            ),
```
guotuofeng, line 132: @jambayk , since the pass always be run in CUDA, do we still need this one?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,229 @@
...
+                description="number of samples in calibration dataset to apply quantization. Default value is 128",
+            ),
+            "bits": PassConfigParam(
+                type_=int,
+                default_value=4,
+                description="quantization bits. Default value is 4",
+            ),
+            "layers_block_name": PassConfigParam(
+                type_=str,
+                default_value="model.layers",
```
trajepl, line 42: Quick question: how user know which block they should use to quantize? It is a name stored in model? Or any fixed options user can choose?

olive/passes/onnx/gptq.py
Taka152: I guess jambay is worried about memory unsufficient when export. During quantization, model is loaded into gpu layer by layer, but when export it may consume more memory than that? And it may cause OOM?

olive/passes/onnx/gptq.py
Taka152: It's the name in model definition, and it's possible to have arbitrary name when custom model, so it maybe hard to provide options.  We also prefill these parameters for user for mainstream  models.
https://github.com/Taka152/Olive/blob/2361389e82bd8e468ce981603c26612c352bb267/olive/passes/onnx/gptq.py#L184-L197

olive/passes/onnx/gptq.py
guotuofeng: it seems the conversion only transfer the model and dummy inputs to CUDA in this case. Would this consume extra CUDA memories?

olive/passes/onnx/gptq.py
```diff
@@ -0,0 +1,229 @@
...
+            model_class = GPTQ_CAUSAL_LM_MODEL_MAP[model_type]
+            quantized_model = model_class(pytorch_model, False, quantize_config)
+        else:
+            quantized_model = BaseGPTQForCausalLM(pytorch_model, False, quantize_config)
+            if not (config["layers_block_name"] and config["outside_layer_modules"] and config["inside_layer_modules"]):
+                raise ValueError(
+                    "Can't get layers_block_name to quantize automatically, "
+                    "please set layers_block_name, outside_layer_modules and inside_layer_modules in config."
+                )
+            quantized_model.layers_block_name = config["layers_block_name"]
```
guotuofeng, line 195: do we need specify https://github.com/AutoGPTQ/AutoGPTQ/blob/896d8204bc89a7cfbda42bf3314e13cf4ce20b02/auto_gptq/modeling/_base.py#L201

olive/passes/onnx/gptq.py
Taka152: Yes, quantization only put one layer into cuda, but conversion may put whole model into cuda, so conversion may consume more memories than quantization, and user may succeed in quantization but fail in conversion.
auto_gptq:
https://github.com/AutoGPTQ/AutoGPTQ/blob/896d8204bc89a7cfbda42bf3314e13cf4ce20b02/auto_gptq/modeling/_base.py#L409-L415
conversion:
https://github.com/Taka152/Olive/blob/2361389e82bd8e468ce981603c26612c352bb267/olive/passes/onnx/conversion.py#L176

olive/passes/onnx/gptq.py
Taka152: I searched layer_type and it is only needed in from_pretrained() and from_quantized() which we will not call.
https://github.com/AutoGPTQ/AutoGPTQ/blob/896d8204bc89a7cfbda42bf3314e13cf4ce20b02/auto_gptq/modeling/_base.py#L764-L776
https://github.com/AutoGPTQ/AutoGPTQ/blob/896d8204bc89a7cfbda42bf3314e13cf4ce20b02/auto_gptq/modeling/_base.py#L1048-L1060

olive/passes/onnx/gptq.py
jambayk: yes, like @Taka152 said conversion on GPU takes more memory since it's not just putting input and model on GPU but also needs to do a forward pass. The forward pass, depending on the implementation can consume much more memory.

guotuofeng: could we try GPU firstly. If out of memory then fallback to CPU? or alternatively, just do conversion on CPU? in this case, avoid specify the convert device since user really doesn't care it?

Taka152: I think default arguments won't bring comfusion to user, and useful for advanced users, like extra_args in optimum_conversion.

OOM catching seems cool, and we could try it in another pr? Since other conversions can also benefit from it. Like user specify cuda but encounter OOM, we can fallback to cpu to do export. 

I'm ok with just do conversion on cpu since it seems most safest method.

guotuofeng: BTW, it seems the config name is different with that one used in conversion? https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/conversion.py#L75

Taka152: Yes, I changed that since user may think the use_device is used for quantization rather than export. But in conversion pass use_device is clear for export.

guotuofeng: Inc(device): https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/inc_quantization.py#L32
optimum(device): https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/optimum_conversion.py#L43
sparsegpt(compute_device): https://github.com/microsoft/Olive/blob/main/olive/passes/pytorch/sparsegpt.py#L75
snpe(target_device): https://github.com/microsoft/Olive/blob/main/olive/passes/snpe/snpe_to_onnx.py#L34
conversion(use_device).

@jambayk, @trajepl 
it seems the device names are not consistent. need we use the same name?

guotuofeng: please refer to https://aiinfra.visualstudio.com/Model%20optimization%20Toolkit/_sprints/taskboard/Model%20optimization%20Toolkit%20Team/Model%20optimization%20Toolkit/[Ge]%20Germanium

guotuofeng (approved):

olive/passes/onnx/gptq.py
Taka152: I added a link in argument description, it may help user to better set this args when using custom model.

trajepl (approved):

jambayk (approved):

PR #962, opened 2024-02-27
PythonEnvironmentSystem and IsolatedORTSystem, Mobilenet QNN EP eval

jambayk:
## Describe your changes
This PR is to support evaluation for QNN EP where the ort package cannot be installed in the same environment as Olive due to missing dependencies on Windows ARM.
Changes in this PR:
- Separate the current PythonEnvironmentSystem into two systems since it is overloaded and performs two different functions
   - Feature complete PythonEnvironmentSystem which requires Olive to be installed in the environment. Can run passes and evaluate models. 
   - IsolatedORTSystem which only requires onnxruntime to be installed in the environment. Can only evaluate onnx models.
- Move ort inference related utils from `models.utils.onnx_utils` to `common.ort_inference` to consolidate the code in one place.
- Abstract ort inference run and timing into helpers to avoid duplicate code. 
- Simplify the ep and provider options normalization code since there were some redundant steps. 
- Update mobilenet example with optional evaluation step.

Changes:
- `PythonEnvironmentSystem` is feature complete. It requires Olive to be installed in the environment. It can run passes and evaluate models.
- New `IsolatedORTSystem` introduced that only supports evaluation of ONNX models. It requires onnxruntime to be installed in the environment. Can be used to for packages like onnxruntime-qnn which can only be run on Windows ARM64 python environment.


olive/systems/ort_inference/ort_inference_system.py
```diff
@@ -0,0 +1,271 @@
...
+        self,
+        the_pass: "Pass",
+        model_config: "ModelConfig",
+        data_root: str,
+        output_model_path: str,
+        point: Optional[Dict[str, Any]] = None,
+    ) -> "ModelConfig":
+        """Run the pass on the model at a specific point in the search space."""
+        logger.warning("ORTInferenceSystem does not support running passes.")
+        raise NotImplementedError
```
guotuofeng, line 72: does it need run_pass?

olive/systems/utils/__init__.py
```diff
@@ -3,11 +3,19 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 from olive.systems.utils.arg_parser import get_common_args
-from olive.systems.utils.misc import create_new_system, create_new_system_with_cache, get_package_name_from_ep
+from olive.systems.utils.misc import (
+    create_new_environ,
+    create_new_system,
```
guotuofeng, line 8: the create_new_system doesn't need to be exposed except the testing

olive/systems/ort_inference/inference_runner.py
```diff
@@ -0,0 +1,97 @@
...
+import argparse
+import json
+import sys
+from pathlib import Path
+from typing import Dict
+
+import numpy as np
+
+# Add the common directory to the path
+ort_inference_utils_parent = Path(__file__).resolve().parents[2] / "common"
```
guotuofeng, line 14: do we need the two lines if we import from olive?

olive/systems/ort_inference/inference_runner.py
```diff
@@ -0,0 +1,97 @@
...
+    return dict(np.load(input_dir / batch_file, allow_pickle=True).items())
+
+
+def main(raw_args=None):
+    args = get_args(raw_args)
+    args.input_dir = Path(args.input_dir)
+    args.output_dir = Path(args.output_dir)
+
+    # load inference setting
+    with open(args.config_path, "rb") as f:
```
guotuofeng, line 44: args.config_path.open("rb") as f?

olive/systems/python_environment/evaluation_runner.py
```diff
@@ -0,0 +1,58 @@
...
+    from olive.systems.olive_system import OliveSystem
+
+
+def get_args(raw_args):
+    parser = argparse.ArgumentParser(description="Onnx model inference")
+
+    parser.add_argument("--model_config", type=str, required=True, help="Path to input model json file")
+    parser.add_argument("--metrics_config", type=str, required=True, help="Path to metrics json file")
+    parser.add_argument("--accelerator_config", type=str, required=True, help="Path to accelerator json file")
+    parser.add_argument("--data_root", type=str, required=False, help="Path to data root")
```
guotuofeng, line 26: needless?

olive/systems/ort_inference/inference_runner.py
jambayk: Since Olive cannot be installed in the environment, we have to import the ort_inference utils from file directly here.

That's why, the 'ort_inference.py' has the requirement that it can only import ort and its dependencies.

olive/systems/utils/__init__.py
jambayk: I didn't add the 'create_new_system'. It was already present.

jambayk: Do you mean 'create_new_environ' that I added?

test/unit_test/model/test_onnx_model.py
```diff
@@ -223,7 +222,7 @@ def test_model_prepare_session_with_unsupported_eps(get_available_providers_mock
     execution_providers = ["TensorrtExecutionProvider", "CUDAExecutionProvider"]
     # provider_options = [{}, {"device_id": "1"}]
     with pytest.raises(
-        OliveEvaluationError,
+        RuntimeError,
```
guotuofeng, line 225: what's the reason that change the exception type? the OliveEvaluationError will be ignored by engine but the RuntimeError will not

olive/systems/ort_inference/ort_inference_system.py
jambayk: Same as above, Olive cannot be installed in this environment. So we cannot run pass unless we fallback on the default local system running the engine.

Since we only expect to use this system for qnn ep, we could ignore this system for the 'run_on_target' feature that we discussed before? 
We don't expect to use perf tuning or inc for this ep target.

olive/systems/ort_inference/inference_runner.py
guotuofeng: then line 18 would fail?

olive/systems/utils/__init__.py
guotuofeng: my bad

olive/common/ort_inference.py
```diff
@@ -51,26 +71,313 @@ def get_ort_inference_session(
...
+    for ep in execution_providers:
+        if ep not in session_providers:
+            raise RuntimeError(
+                f"The onnxruntime fallback happens. {ep} is not in the session providers {session_providers}."
+                f" session._enable_fallback = {session._enable_fallback}"
+            )
+    session.disable_fallback()
+
+
+def run_inference(
```
jambayk, line 200: Will replace `run_inference` and `time_inference` with a InferenceSession wrapper with io_binding. Similar to https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/io_binding_helper.py#L211. 

The common code we want to abstract is the io_binding related code. `run_inference` is hacky with the requirements for data_generator and post_run parameters. We can do the looping in the evaluator.

olive/systems/ort_inference/inference_runner.py
jambayk: Oh good catch! It was supposed to import from `ort_inference` directly. 

I had a `if TYPE_CHECKING:` block for type hinting during dev but deleted the wrong import. Will fix this.

olive/systems/python_environment/evaluation_runner.py
jambayk: Thanks! Will remove this.

test/unit_test/model/test_onnx_model.py
jambayk: The error is raised by the `ort_inference` util module so it doesn't have access to this error class. 

I could catch this error and reraise with the correct error during `model.create_session` if that's better?

guotuofeng: yes, for onnx model, we need raise OliveEvaluationError. the only thing is how could know the RuntimeError is the EP fallback issue instead of other RuntimeError? do we need create a customized exception in that file?

jambayk: That sounds like a workable idea to me. I can do that if you are okay with it too.

jambayk: done!

olive/common/ort_inference.py
```diff
@@ -2,17 +2,43 @@
...
+import numpy as np
+
+if TYPE_CHECKING:
+    from onnxruntime import InferenceSession, IOBinding
+
 
 logger = logging.getLogger(__name__)
 
 
+class OrtSessionFallbackError(Exception):
```
guotuofeng, line 21: we could not inherit from OliveError?

olive/evaluator/olive_evaluator.py
```diff
@@ -415,81 +413,83 @@ def get_inference_settings(cls, metric: Metric, model: ONNXModelHandler) -> Dict
...
         if metric_inference_settings:
             inference_settings.update(metric_inference_settings)
 
         return inference_settings
 
-    def _inference(
-        self,
+
+class OnnxEvaluator(OliveEvaluator, OnnxEvaluatorMixin, framework=Framework.ONNX):
+    def __init__(self):
```
guotuofeng, line 424: needless?

olive/evaluator/olive_evaluator.py
```diff
@@ -415,81 +413,83 @@ def get_inference_settings(cls, metric: Metric, model: ONNXModelHandler) -> Dict
...
+        super().__init__()
+
+    @staticmethod
+    def get_session_wrapper(
         model: ONNXModelHandler,
         metric: Metric,
         dataloader: Dataset,
```
guotuofeng, line 431: Dataloader instead Dataset?

olive/common/ort_inference.py
jambayk: This file cannot import from Olive

olive/evaluator/olive_evaluator.py
jambayk: Which part? the `__init__`? I can remove it but kept it since the original code also had it.

olive/evaluator/olive_evaluator.py
jambayk: The original type hint was `Dataset`. I think that is right since we are expecting a `torch.utils.Dataset` here.

jambayk: Actually you are right, it is supposed to be `torch.utils.data.Dataloader` so the original type hints for this file was incorrect. I will open a separate PR to fix it first.

jambayk: https://github.com/microsoft/Olive/pull/969

olive/common/ort_inference.py
```diff
@@ -2,17 +2,43 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+# NOTE: Only onnxruntime and its dependencies can be imported here!!!
```
trajepl, line 5: Should we update this note? As the onnxruntime should not be dependency of Olive. Also you are using ort as TYPE_CHECKING

olive/common/ort_inference.py
```diff
@@ -51,26 +75,299 @@ def get_ort_inference_session(
...
+        ort.get_available_providers(),
+    )
+    for idx, provider in enumerate(providers):
+        if provider in ["CUDAExecutionProvider", "DmlExecutionProvider"] and device_id is not None:
+            provider_options[idx]["device_id"] = str(device_id)
+        elif provider == "QNNExecutionProvider":
             # add backend_path for QNNExecutionProvider
-            execution_provider[idx] = ("QNNExecutionProvider", {"backend_path": "QnnHtp.dll"})
-            break
+            provider_options[idx]["backend_path"] = "QnnHtp.dll"
```
trajepl, line 89: Not have to update in this PR.
We may need expose the backend_path to users? Also need to handle different suffix in different platform. Linux -> .so, Windows -> .dll.

Could you help add the TODO to tell this?

olive/common/ort_inference.py
```diff
@@ -51,26 +75,299 @@ def get_ort_inference_session(
...
-    for idx, ep in enumerate(execution_provider):
-        if ep == "QNNExecutionProvider":
+    # execution providers and provider options
+    providers, provider_options = check_and_normalize_provider_args(
+        inference_settings.get("execution_provider"),
+        inference_settings.get("provider_options"),
+        ort.get_available_providers(),
+    )
+    for idx, provider in enumerate(providers):
+        if provider in ["CUDAExecutionProvider", "DmlExecutionProvider"] and device_id is not None:
```
trajepl, line 85: Will Rocm/Migraphx also need device_id? I also considered to add the device id, but just found another simple solution by using `CUDA_VISIBLE_DEVICES` out of Olive.
image

If you have strong requirements to use `device_id`, could you help to update the related docs later?

olive/common/ort_inference.py
```diff
@@ -51,26 +75,299 @@ def get_ort_inference_session(
...
+        elif provider == "QNNExecutionProvider":
             # add backend_path for QNNExecutionProvider
-            execution_provider[idx] = ("QNNExecutionProvider", {"backend_path": "QnnHtp.dll"})
-            break
+            provider_options[idx]["backend_path"] = "QnnHtp.dll"
+    logger.debug(f"Normalized providers: {providers}, provider_options: {provider_options}")
 
     # dml specific settings
-    if len(execution_provider) >= 1 and execution_provider[0] == "DmlExecutionProvider":
+    if len(providers) >= 1 and providers[0] == "DmlExecutionProvider":
```
trajepl, line 93: What if the order of EP provided by user is not `DmlEP, CPUEP...`? How about:
```
if 'DmlExecutionProvider' in providers
```

olive/common/ort_inference.py
```diff
@@ -51,26 +75,299 @@ def get_ort_inference_session(
...
+
+
+def check_and_normalize_provider_args(
+    providers: Sequence[Union[str, Tuple[str, Dict[Any, Any]]]],
+    provider_options: Sequence[Dict[Any, Any]],
+    available_provider_names: Sequence[str],
+):
+    """Validate the 'providers' and 'provider_options' arguments and returns a normalized version.
+
+    :param providers: Optional sequence of providers in order of decreasing
```
trajepl, line 116: It is not related with this PR.
But how to guarantee the order of desc?

olive/common/ort_inference.py
```diff
@@ -51,26 +75,299 @@ def get_ort_inference_session(
...
+        self.shared_kv_buffer = shared_kv_buffer
+        self.use_fp16 = use_fp16
+        self.kv_cache_ortvalues = {} if (self.shared_kv_buffer and self.use_fp16) else None
+
+        self.io_binding = None
+        if self.io_bind:
+            self.io_binding = self.session.io_binding()
+            if self.shared_kv_buffer and self.use_fp16:
+                assert input_feed is not None, "input_feed is required when shared_kv_buffer and use_fp16 are True"
+                bind_input_data(
```
trajepl, line 228: Missed the "device_id"?

olive/evaluator/olive_evaluator.py
```diff
@@ -415,81 +413,83 @@ def get_inference_settings(cls, metric: Metric, model: ONNXModelHandler) -> Dict
...
-        post_func=None,
-        device: Device = Device.CPU,
-        execution_providers: Union[str, List[str]] = None,
-    ) -> Tuple[OliveModelOutput, Any]:
+        device: Device,
```
trajepl, line 432: Should we add `device_id`?
We may need to reconsider about the interface to expose `device_id` in which components.

olive/evaluator/olive_evaluator.py
```diff
@@ -415,81 +413,83 @@ def get_inference_settings(cls, metric: Metric, model: ONNXModelHandler) -> Dict
...
             inference_settings=inference_settings,
             device=device,
             execution_providers=execution_providers,
         )
 
+        # prepare for io binding
+        io_config = model.get_io_config()
+        io_bind = OnnxEvaluator.io_bind_enabled(metric, model.inference_settings)
+        shared_kv_buffer = metric.user_config.shared_kv_buffer
+        use_fp16 = any(v == "float16" for v in io_config["input_types"])
```
trajepl, line 448: Let us say, we have pass flow as converter -> transformer_optimization_fp16, but not enforce input as float32. 
Will the model io_config get updated to float16? Or inherit the input model's io_config?
 Will that lead to wrong `use_fp16`?

trajepl:
For ORTEnvironmentSystem and PythonEnvironmentSystem, what if in the case of snpe/qnn sdk? Will snpe/qnn sdk need a isolated python env?
In arm devices, it may need other SNPEEnvironmentSystem or QNNEnvironmentSystem in the future? Or is that possible to leverage the `ORTEnvironmentSystem` ?

olive/common/ort_inference.py
jambayk: It is fine to import ort under type checking since it doesn't make ort a required dependency. We do similar things in other places too https://github.com/microsoft/Olive/blob/683ed65a5c11c614759d186c607bcc26553afae3/olive/passes/onnx/transformer_optimization.py#L20 

This file does not make ort a dependency of olive. it is imported lazy when needed insided the functions here. I can update the note to say "imported in this file" instead of "imported here" to be clear.

olive/common/ort_inference.py
jambayk: I checked the qnn ep docs and since there is only windows arm package currently, their examples also only use htp backend (the other option is cpu backend but that's not interesting). 
The expectation for qnn ep is that they already have the qnn sdk path added to `PATH` so that it can find the backend library using the name.

olive/common/ort_inference.py
jambayk: Yeah, we use `CUDA_VISIBLE_DEVICES` for regular onnx model. This part is added for distributed onnx models where there are multiple models that need to be on their own device. 
This code change doesn't introduce a new behavior. I am only aggregating the logic in `ONNXModelHandler.prepare_session` into this module https://github.com/microsoft/Olive/blob/683ed65a5c11c614759d186c607bcc26553afae3/olive/model/handler/onnx.py#L104 

So, this is only used internally by Olive and is not really user facing. Most of the time, `device_id` here will be `None`. Same for `OrtInferenceSession` wrapper.

olive/common/ort_inference.py
jambayk: This is the pre-existing logic so I didn't want to change the behavior in the PR https://github.com/microsoft/Olive/blob/683ed65a5c11c614759d186c607bcc26553afae3/olive/common/ort_inference.py#L68

olive/common/ort_inference.py
jambayk: I think that is decided by the user if/when they provide their own inference settings. For olive workflows, we loop through the candidate EPs so this will only have one member. 

I think the only current use case with multiple providers in this list is for the rocm perf tuning with migraph and rocm eps, @guotuofeng?

olive/common/ort_inference.py
jambayk: I didn't add it here since the original code I abstracted didn't use device_id. but I can add it here for completeness if you think it's better.

olive/evaluator/olive_evaluator.py
jambayk: same as above

olive/evaluator/olive_evaluator.py
jambayk: Even before, we were checking the input types from the input feed which we expect to be in the same types as the model's io config. So the original logic has the same issues I think. 
We only use `fp16` along with `shared_kv_buffer` for gqa like use cases so I think it can only work if the io is fp16 anyways?

IIRC, you added this case to mimic the logic in the ort llama example https://github.com/microsoft/onnxruntime/blob/913bdc7306e11b65644f733861684a3a460e8db0/onnxruntime/python/tools/transformers/models/llama/llama_inputs.py#L291? 
It used to check for fp16 but now uses `use_gqa`. Maybe we could update this according in the future. In this PR, I don't change any existing behavior or logic. Only checking io_config instead of input_feed since they should technically be the same.

olive/common/ort_inference.py
jambayk: Update note to be clearer.

trajepl: Yes, I know the typing checking.  As you said `This file does not make ort a dependency of olive`. But the comment shows `onnxruntime can be imported here`. 

Yes, `in this file` is better. 
Thanks!

olive/common/ort_inference.py
trajepl: Ic, sounds like we can focus on windows by now as qnn ep is only available in that.

olive/common/ort_inference.py
trajepl: > So, this is only used internally by Olive and is not really user facing. Most of the time, device_id here will be None. Same for OrtInferenceSession wrapper.
Got it. Could you help also update the code comments to tell?

jambayk: sure, I will add some comments above this to describe the current use cases.

olive/evaluator/olive_evaluator.py
trajepl: Make sense.

olive/common/ort_inference.py
jambayk: Done!

olive/common/ort_inference.py
```diff
@@ -51,26 +80,299 @@
...
+    )
+    for idx, provider in enumerate(providers):
+        if provider in ["CUDAExecutionProvider", "DmlExecutionProvider"] and device_id is not None:
+            provider_options[idx]["device_id"] = str(device_id)
+        elif provider == "QNNExecutionProvider":
             # add backend_path for QNNExecutionProvider
-            execution_provider[idx] = ("QNNExecutionProvider", {"backend_path": "QnnHtp.dll"})
-            break
+            provider_options[idx]["backend_path"] = "QnnHtp.dll"
+    logger.debug(f"Normalized providers: {providers}, provider_options: {provider_options}")
```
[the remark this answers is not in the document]
jambayk: Resolved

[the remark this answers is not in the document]
jambayk: Resolved

olive/common/ort_inference.py
jambayk: Reraised as discussed offline.

olive/systems/system_config.py
```diff
@@ -72,18 +69,29 @@ def _validate_python_environment_path(cls, v):
         return v
 
 
+class PythonEnvironmentTargetUserConfig(CommonPythonEnvTargetUserConfig):
+    olive_managed_env: bool = False  # if True, the environment will be created and managed by Olive
+    requirements_file: Union[Path, str] = None  # path to the requirements.txt file
+
+
+class ORTEnvironmentTargetUserConfig(CommonPythonEnvTargetUserConfig):
```
guotuofeng, line 77: nit: ORTEnvironmentTargetUserConfig = CommonPythonEnvTargetUserConfig?

olive/systems/python_environment/python_environment_system.py
```diff
@@ -82,9 +68,40 @@ def __init__(
...
+            # write config jsons to files
+            args = []
+            for key, config_json in config_jsons.items():
+                config_json_path = tmp_dir_path / f"{key}.json"
+                with config_json_path.open("w") as f:
+                    json.dump(config_json, f, indent=4)
+                args.append(f"--{key} {config_json_path}")
+
+            # command to run
+            command = f"python {script_path} {' '.join(args)}"
```
guotuofeng, line 89: nit: I think using string list then join is similar to C# StringBuilder, which should have better performance than string concatenation

olive/systems/python_environment/pass_runner.py
```diff
@@ -4,41 +4,40 @@
...
-    with open(args.model_json_path) as f:
-        model_json = json.load(f)
-    with open(args.pass_json_path) as f:
-        pass_json = json.load(f)
+    set_tempdir(args.tempdir)
 
-    model = ModelConfig.from_json(model_json).create_model()
-    the_pass = FullPassConfig.from_json(pass_json).create_pass()
+    model = ModelConfig.parse_file(args.model_config).create_model()
+    the_pass = FullPassConfig.parse_file(args.pass_config).create_pass()
```
guotuofeng, line 33: I like it.

olive/systems/python_environment/python_environment_system.py
jambayk: Changed to use a string list for `command`

olive/systems/python_environment/pass_runner.py
jambayk: yeah, the `parse_file` method is very useful!

olive/systems/system_config.py
jambayk: Will keep it as it is for now to keep the inheritance structure as discussed offline.

guotuofeng (dismissed):

guotuofeng (approved):

docs/source/extending_olive/design.md
```diff
@@ -35,12 +35,13 @@ technique.
 
 The Evaluator **target** will ideally have the same specs as the desired deployment environment.
 
-Olive supports three main types of Systems:
+Olive supports four main types of Systems:
```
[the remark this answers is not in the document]
jambayk: ✋

trajepl (dismissed):

guotuofeng (changes requested):
like offline discuss, better change the ort env system to IsolatedORTSystem.

guotuofeng (approved):

PR #964, opened 2024-02-27
🏷️  Simplify run_subprocess log to avoid potential logging bug

trajepl:
## Describe your changes
1. Remove env from run_subprocess's debug logging to avoid potential l[ogging bug.](https://aiinfra.visualstudio.com/PublicPackages/_build/results?buildId=415082&view=logs&j=04f3f8bd-f6ad-546f-a3e7-a621004c3b55&t=a3ec9bd6-ecbf-5027-baca-9daffad9c3e0) The bug is caused by multitudinous path info which may trigger the UnicodeEncodeError.
2. Ignore the missed diag log files and just throw a error log, as the diag logs are used only for latencies measurement. And sometimes, the SNPEDiag_\d+.log cannot match the value of `runs`. For example:
    - if `runs` is 3, the diag log may be `SNPEDiag_0.log, SNPEDiag_3.log, SNPEDiag_4.log`. or  missed one of the log.

### DIFF
--- a/olive/common/utils.py
+++ b/olive/common/utils.py
@@ -37,7 +37,7 @@ def run_subprocess(cmd, env=None, cwd=None, check=False):
             f"Failed to run {cmd} with returncode {e.returncode}!",
             f"Stderr: {e.stderr.decode('utf-8')}",
             f"Stdout: {e.stdout.decode('utf-8')}",
-            f"Env: {str(env).encode(errors='ignore').decode()}",
+            f"Env: {env}",
         ]
         logger.error("\n".join(err_msg))  # noqa: TRY400
         raise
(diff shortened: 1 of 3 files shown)


olive/common/utils.py
```diff
@@ -19,8 +19,8 @@
 logger = logging.getLogger(__name__)
 
 
-def run_subprocess(cmd, env=None, cwd=None, check=False):  # pragma: no cover
-    logger.debug(f"Running command: {cmd} with env: {env}")
```
guotuofeng, line 23: encode the env to UTF-8 in case of error happens in line 40?

trajepl: Updated.

guotuofeng (dismissed):

guotuofeng (approved):

PR #979, opened 2024-03-01
Phi2 Generation example

trajepl:
## Describe your changes
1. Phi2 generation example run in linux and windows.
2. Update the way to match input's past_key_values in conversion passes with regex match.


olive/passes/onnx/conversion.py
```diff
@@ -195,13 +195,15 @@ def _export_pytorch_model(
 
             # handle dummy inputs for model with past, which has past_key_values
             # match input names in `past_key_values.(hidden_layer_num).(key|value)` pattern(llama2 case)
-            # or `past_key_values.(hidden_layer_num)` pattern (phi2 case)
+            # or `past_key_(hidden_layer_num)`| `past_value_(hidden_layer_num)` pattern (phi2 case)
+            import re
+
             for name, dm_input in dummy_inputs.items():
                 if name == "past_key_values" and isinstance(dm_input, list):
```
guotuofeng, line 202: does the name need to be updated?

guotuofeng:
Update how past_key_values is different between Windows and Linux in description for more clarification.

examples/phi2/__init__.py
```diff
@@ -0,0 +1,4 @@
+# -------------------------------------------------------------------------
```
guotuofeng, line 1: do we need this file?

examples/phi2/generate.py
```diff
@@ -0,0 +1,233 @@
...
+        self.use_fp16 = False
+        self.device = -1
+        self.use_buffer_share = False
+        self.packed_kv = False
+        self.use_step = False
+        self.torch_dtype = None
+        self.sess = None
+        self.tokenizer = None
+
+    def get_initial_inputs_and_outputs(self, encodings_dict):
```
guotuofeng, line 36: private method?

examples/phi2/generate.py
```diff
@@ -0,0 +1,233 @@
...
+                present = torch.zeros(present_shape, device=self.device, dtype=self.torch_dtype)
+                (
+                    outputs.update({present_key_name: present.contiguous(), present_value_name: present.contiguous()})
+                    if not self.packed_kv
+                    else outputs.update({f"present_{i}": present.contiguous()})
+                )
+
+        return inputs, outputs
+
+    def apply_io_binding(self, model: ort.InferenceSession, inputs: dict, outputs: dict):
```
guotuofeng, line 89: ditto?

examples/phi2/generate.py
trajepl: The file is copied from ort. https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/transformers/models/phi2/inference_example.py#L91

We may can keep what it is now.

guotuofeng: Please add comments about where the file is copied.

olive/passes/onnx/conversion.py
trajepl: This name is the argument from huggingface model class. I can add more comments to tell why the name should be `past_key_values`

E.g.:
1. phi2
image

2. mistral.
image

guotuofeng: just a little confusion what's the new name in Windows?

trajepl: Currently it is same in linux and windows. Previously, the differences was led by:
In windows, we cannot use dynamo which fallbacks to get_io_config function where we already set the input name. And the input name is different with the one in dynamo. 

Update the name in user_script.py can fix.
image

olive/passes/onnx/conversion.py
```diff
@@ -195,13 +195,17 @@ def _export_pytorch_model(
...
             for name, dm_input in dummy_inputs.items():
+                # the `past_key_values` is the argument name from huggingface model class
+                # which is independent of the kv-related variables in input list
                 if name == "past_key_values" and isinstance(dm_input, list):
-                    key_value_names = [f"{name}.{idx}" for idx in range(len(dm_input))]
+                    kv_name_regrex_strs = [rf"past_(.*)(.|_){idx}(.(key|value))?" for idx in range(len(dm_input))]
                     if all(
-                        any(key_value_name in input_name for input_name in input_names)
-                        for key_value_name in key_value_names
+                        any(re.match(kv_r_str, input_name) for input_name in input_names)
```
guotuofeng, line 207: regex building/match is pretty time consuming compared with simply string comparison.

trajepl: Updated. Also see comments:
```
        # In the case of dynamo exporting, user do not need to provide input names
        # for past_key_values, when exported to onnx, it will be expand to
        # `past_key_values.(hidden_layer_num).(key|value)` pattern(llama2 case)
        # or `past_(key|value)_(hidden_layer_num)` pattern (phi2 case)
        # after the expansion, user should provide the correct input names for inference
```

olive/passes/onnx/conversion.py
```diff
@@ -192,18 +192,33 @@ def _export_pytorch_model(
...
+                        [f"past_key_{idx}" for idx in range(len(dm_input))]
+                        + [f"past_value_{idx}" for idx in range(len(dm_input))]
+                    ]
+                    kv_name_groups += [
+                        [f"past_key_values.{idx}.key" for idx in range(len(dm_input))]
+                        + [f"past_key_values.{idx}.value" for idx in range(len(dm_input))]
+                    ]
+                    for kv_name_group in kv_name_groups:
+                        if all(any(kv_str in input_name for input_name in input_names) for kv_str in kv_name_group):
+                            dummy_input_keys.discard(name)
```
guotuofeng, line 221: could we make it a utils function?

trajepl: Updated by moving it into io_config.py

olive/model/config/io_config.py
```diff
@@ -87,3 +87,32 @@ def is_io_config_static(config: Union[IoConfig, Dict]):
...
+        ["past_key_<id>", "past_value_<id>"],
+        # can be used to match past_key_values.0.key, past_key_values.0.value...
+        # and past_key_values.0, past_key_values.0... at the same time
+        ["past_key_values.<id>"],
+    ]
+    possible_kv_names_group = [[] for _ in range(len(possible_kv_names_templates))]
+    for i in range(hidden_layer_num):
+        for j, kv_template in enumerate(possible_kv_names_templates):
+            for kv_str in kv_template:
+                possible_kv_names_group[j].append(kv_str.replace("<id>", str(i)))
```
guotuofeng, line 113:
```python
        for j, kv_template in enumerate(possible_kv_names_templates):
            for kv_str in kv_template:
                possible_kv_names_group[j].extend(kv_str.replace("<id>", str(i)) for i in range(hidden_layer_num))
```

guotuofeng: the  hidden_layer_num should be in inner loop since it is larger than the template size.

trajepl: It's nice! Thanks.

trajepl: Updated.

guotuofeng (approved):

PR #980, opened 2024-03-01
Add Mistral fp16 config

apsonawane:
## Describe your changes
Add float16 configuration to Mistral.


examples/mistral/mistral_fp16_optimize.json
```diff
@@ -0,0 +1,109 @@
...
+    "input_model": {
+        "type": "PyTorchModel",
+        "config": {
+            "hf_config": {
+                "model_name": "mistralai/Mistral-7B-v0.1",
+                "model_class": "MistralForCausalLM"
+            }
+        }
+    },
+    "systems": {
```
jambayk, line 11: nit: this `systems` and `engine.target` are not really needed since Olive can infer from the `CUDAExecutionProvider` that the system is local gpu.

jambayk (approved):
Thanks for adding this example!

devang-ml (changes requested):
Please add e2e example that uses the output model produced using mistral_fp16_optimize.json 

Thanks!

examples/mistral/readme.md
```diff
@@ -28,10 +28,15 @@ git config --system core.longpaths true
...
 
 ## Usage
+CPU:
 ```bash
 python -m olive.workflows.run --config mistral_optimize.json
 ```
 
+GPU:
+```bash
+python -m olive.workflows.run --config mistral_fp16_optimize.json
```
guotuofeng, line 38: do we need update the example doc https://microsoft.github.io/Olive/examples.html by adding the float16?

jambayk: I checked the doc and saw fp16 but turns out it was for phi2 and not mistral. 

Yeah, we should add list the cuda/fp16 setting.

devang-ml (approved):

examples/mistral/readme.md
jambayk: @apsonawane could you update `docs/source/examples.md` too?

examples/mistral/mistral.py
```diff
@@ -0,0 +1,98 @@
...
+            onnx_inputs[f"past_key_values.{i}.value"] = torch.rand(
+                1, num_heads // 1, past_seq_len, head_size, dtype=torch.float16
+            ).numpy()
+
+        model_path = optimized_model_dir / "model.onnx"
+
+        session = ort.InferenceSession(
+            model_path,
+            providers=["CUDAExecutionProvider"],
+            provider_options=[{"device_id": "7"}],
```
jambayk, line 90: Can you remove this provider options since it is specific to your environment? For users, they can choose the device by using `CUDA_VISIBLE_DEVICES=6 python mistral.py --inference` instead.

examples/mistral/mistral.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+        session.run(None, onnx_inputs)[0]
+
+        print("Inference test completed successfully!")
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
jambayk, line 98: nit: I think you can just run `main()` directly here. the sys.exit feel redundant.

examples/mistral/mistral.py
apsonawane: Should I add this in Readme?

examples/mistral/mistral.py
apsonawane: Done

examples/mistral/mistral.py
jambayk: Sure, I think that would be good too. Maybe as a note below the original instruction? if they would like to use a specific gpu, do this...

examples/mistral/readme.md
apsonawane: Done

jambayk (approved):
Thanks!

PR #991, opened 2024-03-06
👑 Add configs for qnn_preprocess from ort-nightly packages

trajepl:
## Describe your changes
Add configs for qnn_preprocess from ort-nightly packages which is helpful for the coming model tests.


olive/passes/onnx/qnn_preprocess.py
```diff
@@ -47,7 +99,18 @@ def _run_for_config(
...
+
+        modified = qnn_preprocess_model(
+            model.model_path,
+            output_model_path,
+            fuse_layernorm=config["fuse_layernorm"],
+            save_as_external_data=config["save_as_external_data"],
+            all_tensors_to_one_file=config["all_tensors_to_one_file"],
+            external_data_size_threshold=config["size_threshold"],
+            external_data_convert_attribute=config["convert_attribute"],
+            inputs_to_make_channel_last=config["inputs_to_make_channel_last"],
```
jambayk, line 111: Won't this cause key error for ort < 1.18? If we only want to support the nightly version and there is no use case for version 1.17, I think we should just make 1.18 the required version. 

Also, I don't think we should have `default_config` be conditional for ort version. We don't have any passes that do that + cache is agnostic to ort version currently. 
We could add the new options directly and ignore them with some logs for lower versions (if we want to keep support for lower version)

olive/passes/onnx/qnn_preprocess.py
```diff
@@ -47,7 +99,18 @@ def _run_for_config(
         from onnxruntime.quantization.execution_providers.qnn import qnn_preprocess_model
 
         output_model_path = resolve_onnx_path(output_model_path, Path(model.model_path).name)
-        modified = qnn_preprocess_model(model.model_path, output_model_path, **config)
+
+        modified = qnn_preprocess_model(
```
jambayk, line 103: This also has a `external_data_location` https://github.com/microsoft/onnxruntime/blob/e93a860819545ea64acfe36e19e2b954389d48bf/onnxruntime/python/tools/quantization/execution_providers/qnn/preprocess.py#L24 which is equivalent to `location` in https://github.com/microsoft/Olive/blob/fed7c66ec1cb44105e57ca86242c35d79b4080ff/olive/passes/onnx/common.py#L118. 

could you add a similar logic to decide the value so that the `external_data_config` behaves the same for all passes?

olive/passes/onnx/qnn_preprocess.py
guotuofeng: @jambayk, you are right. From the time being, the default_config doesn't support conditional.

olive/passes/onnx/qnn_preprocess.py
```diff
@@ -40,10 +85,42 @@ def _run_for_config(
...
+        }
+        if version.parse(OrtVersion) < version.parse("1.18.0"):
+            removed_config = [
+                "save_as_external_data",
+                "all_tensors_to_one_file",
+                "size_threshold",
+                "convert_attribute",
+                "inputs_to_make_channel_last",
+                "outputs_to_make_channel_last",
+            ]
```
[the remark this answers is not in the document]
jambayk: looks like `.keys()` is not required.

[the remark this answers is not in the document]
trajepl: Oh, yes. Just updated.

jambayk (approved):

PR #992, opened 2024-03-06
make _default_config classmethod so the child class could use super

guotuofeng:

### DIFF
--- a/olive/passes/pytorch/lora.py
+++ b/olive/passes/pytorch/lora.py
@@ -157,8 +157,8 @@ class LoRABase(Pass):
     # values from the input model will be ignored and new values will be set based on the pass config
     model_overwrites: ClassVar[tuple] = ("torch_dtype", "device_map")
 
-    @staticmethod
-    def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+    @classmethod
+    def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
         return {
             "use_ort_trainer": PassConfigParam(
                 type_=bool, default_value=False, description="Whether or not to use ORTTrainer."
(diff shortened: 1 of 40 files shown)


olive/passes/pytorch/lora.py
```diff
@@ -964,7 +964,7 @@ def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigPa
                 description="Number of LoftQ iterations.",
             ),
         }
-        config.update(QLoRABase._default_config(accelerator_spec))  # pylint: disable=protected-access
+        config.update(super()._default_config(accelerator_spec))
```
guotuofeng, line 967: @jambayk, this is a behavior change, could you help confirm?

jambayk: This should not have been a behavior change. But currently it is because I mistakenly made the parent of this class as `QLoRA` when it was supposed to be `QLoRABase`. Can you make that correct that as well? Thanks!

guotuofeng: with the new super(), its base class will be QLoRA

guotuofeng: > This should not have been a behavior change. But currently it is because I mistakenly made the parent of this class as `QLoRA` when it was supposed to be `QLoRABase`. Can you make that correct that as well? Thanks!

Or do you mean we need change the base class as QLoRABase instead of QLoRA?

jambayk: Yes, update the base class for `LoftQ` as `QLoRABase` since it was a mistake from my side. You can see that all the methods defined in `QLoRA` are overwritten in `LoftQ`.

guotuofeng: done

jambayk (approved):

PR #998, opened 2024-03-07
Unify accelerator and ep concept into system

guotuofeng:
## Describe your changes
* In this PR, the execution_providers is moved from engine to system config. typical config looks like
```json
    "systems": {
        "local_system": {
            "type": "LocalSystem",
            "config": {
                "accelerators": [
                    {
                        "device": "gpu",
                        "execution_providers": [
                            "CPUExecutionProvider"
                        ]
                    }
                ]
            }
        }
    },
  "engine": {
        "host": "local_system",
        "target": "local_system",
}

```
Please note the above config is optional if we only use local system with default execution providers. 
the device and execution_providers are both optional but they cannot be all None if accelerators are specified. 

* the `infer_accelerators_from_execution_providers` is changed to ignore the CPUExecutionProvider. As the result, the `["CPUExecutionProvider", "CUDAExecutionProvider"]` will be inferred as `gpu`


xiaoyu-work:
Since you are updating accelerator, can we also update the engine run API output? Like current output node is a dict where key is an [Accelerator object](https://github.com/microsoft/Olive/blob/main/olive/engine/engine.py#L285). When using Python API, it is difficult to get the output value directly for a specific device or ep for example, where you have to manually build an Accelerator object or for loop the whole result. One option is simplifying the key to a str `<device>-<ep>`, or other alternative solutions also fine.

guotuofeng:
> Since you are updating accelerator, can we also update the engine run API output? Like current output node is a dict where key is an [Accelerator object](https://github.com/microsoft/Olive/blob/main/olive/engine/engine.py#L285). When using Python API, it is difficult to get the output value directly for a specific device or ep for example, where you have to manually build an Accelerator object or for loop the whole result. One option is simplifying the key to a str `<device>-<ep>`, or other alternative solutions also fine.

The `device-ep` is a good suggestion to replace the `AcceleratorSpec`. One simple approach to switch to `device-ep` is to use something like `str(AcceleratorSpec)` instead of the original object. On the other hand, this is a separated improvement PR after current PR is merged. 

@jambayk, @trajepl, what's your opinion to switch to `device-ep` from the original `AcceleratorSpec`?

examples/bert/bert_cuda_gpu.json
```diff
@@ -76,7 +76,6 @@
             }
         },
         "evaluator": "common_evaluator",
```
trajepl, line 78: What will happend if we remove EP but no any system configs are given like this example? Seems it will use CPU and CPU EP, right?

Also similar questions in:
1. What if only `device` is given in accelerators. -> get_available_providers? 
2. What if only `execution_provider` is given in accel. -> infer device but ignore CPU as you did?
3. What if no `accelerators` config is given like in this example.

Could you help double check? Also update the related docs if needed.

docs/source/tutorials/configure_systems.rst
```diff
@@ -7,6 +7,19 @@ A system is the environment (OS, hardware spec, device platform, supported EP) t
...
+-------------------------
+For each **host** or **target**, it could specify multiple accelerators associated with it. Each accelerator could have the following attributes.
+
+* :code:`device`: The device type of the accelerator. It could be "cpu", "gpu", "npu", etc. Please refer to the API documentation for the full list of supported devices.
+* :code:`execution_providers`: The execution provider list that are supported by the accelerator. For e.g. ["CUDAExecutionProvider", "CPUExecutionProvider"].
+
+Note:
+
+- The accelerators for local system or python system is optional. If not provided, Olive will get the available execution providers installed in current local machine and infer its device.
+- If the accelerators are specified, the device is mandatory. The execution_providers is optional. If not provided, Olive will get the available execution providers associated with the device.
```
jambayk, line 20: from the PR description and this, it looks like the inferring device from ep is now removed? I think that was a useful feature since some of the information is redundant. if the ep is cuda, the device is always gpu. same for qnn ep and npu.

jambayk: also what is the difference between 
`[{device: gpu, ep=[cpuep, cudaep]}]` and `[{device:gpu, ep:[cpuep]}, {device:gpu, ep:[cudaep]}]`

examples/bert/bert_auto_opt_gpu.json
```diff
@@ -50,7 +66,8 @@
             }
         },
         "evaluator": "common_evaluator",
-        "execution_providers": ["CUDAExecutionProvider", "TensorrtExecutionProvider"],
```
jambayk, line 53: I think, we might have to print a warning if the config class sees this option. Since a user might use an old config and it would run without error.

jambayk: The no error run makes the config appear backward compatible but it is not.

docs/source/tutorials/configure_systems.rst
guotuofeng: Actually, I am not sure whether the infer device from EP is required feature after both device and ep is consolidated. From my understanding, I think ep is specified but device not is a little bit confusing. 

@devang-ml, how do you think?
@jambayk, for your second question, they are same after flatten to AcceleratorSpec.

xiaoyu-work: I have same question here. I can't remember the reason why we have both device and ep, but seems only one of them is needed. If user specified device, we can use all available eps for this device. If user specified eps, we can run Olive with those eps.

examples/bert/bert_auto_opt_gpu.json
guotuofeng: we should raise ValidationError if user specify the execution provider by legacy way.

guotuofeng: updated.

docs/source/tutorials/configure_systems.rst
devang-ml: A generic EP could support GPU and CPU both so we need device as well as EP.
In cases where there is a clear mapping between EP and the device (e.g. CudaEP) then we can make device optional as a convenience.

guotuofeng: updated.

.azure_pipelines/performance_check/configs/bert.json
```diff
@@ -78,7 +78,6 @@
         },
         "clean_cache": true,
         "evaluator": "common_evaluator",
-        "execution_providers": ["CPUExecutionProvider"],
```
jambayk, line 81: If we don't specify the ep, then it infers the ep from the local system, doesn't it? If this is the case, then if the user has installed onnxruntime-gpu, it would loop over all of the available eps even though that was not the intention. So, I don't think this change is equivalent to the previous config?

jambayk: same for similar changes in the other configs.

examples/bert/bert_cuda_gpu.json
```diff
@@ -76,7 +76,6 @@
...
-        "execution_providers": ["CUDAExecutionProvider"],
```
jambayk, line 79: target it not specified here. so, this would also loop over the installed eps.

examples/bert/bert_auto_opt_gpu.json
```diff
@@ -16,6 +16,22 @@
             }
         }
     },
+    "systems": {
+        "local_system": {
+            "type": "LocalSystem",
+            "config": {
+                "accelerators": [
+                    {
+                        "device": "gpu",
```
jambayk, line 25: Should we have some examples where the device is not provided if it is clear from the ep? Isn't that the motivation for the feature implemented in https://github.com/microsoft/Olive/pull/428 @trajepl?

.azure_pipelines/performance_check/configs/bert.json
guotuofeng: the system is configured in the run script.

guotuofeng: https://github.com/microsoft/Olive/blob/359fe22e3f8679de61ff3a56e0b9958c11479968/.azure_pipelines/performance_check/run_performance_check.py#L261

examples/bert/bert_cuda_gpu.json
guotuofeng: yes, it will use the installed the EPs.

examples/bert/bert_auto_opt_gpu.json
guotuofeng: Just remove the device attribute should be fine.

guotuofeng: see https://github.com/microsoft/Olive/blob/359fe22e3f8679de61ff3a56e0b9958c11479968/test/unit_test/hardware/test_accelerator.py#L76

examples/bert/bert_cuda_gpu.json
jambayk: for this example, we need a target then? Otherwise, the example is not the same as before.

examples/bert/bert_auto_opt_gpu.json
jambayk: yes. I was asking if we should have some examples that do that. Or keep all of the examples with complete configs.

docs/source/tutorials/configure_systems.rst
```diff
@@ -7,6 +7,19 @@ A system is the environment (OS, hardware spec, device platform, supported EP) t
...
+
+* :code:`device`: The device type of the accelerator. It could be "cpu", "gpu", "npu", etc. Please refer to the API documentation for the full list of supported devices.
+* :code:`execution_providers`: The execution provider list that are supported by the accelerator. For e.g. ["CUDAExecutionProvider", "CPUExecutionProvider"].
+
+Note:
+
+- The accelerators for local system or python system is optional. If not provided, Olive will get the available execution providers installed in current local machine and infer its device.
+- For local system or python system, either device or execution_providers is optional but not both if the accelerators are specified. If device or execution_providers is not provided, Olive will infer the device or execution_providers if possible.
+- For docker system and AzureML system, both device and execution_providers are mandatory. Otherwise, Olive will raise an error.
+
```
jambayk, line 22: Can you add some examples of different ways of specifying the accelerator configs.
(1. both given, 2. ep given, device inferrered, 3. device given, ep inferred, 4. none given, inferred from local environment, etc)?

docs/source/tutorials/configure_systems.rst
```diff
@@ -345,7 +382,7 @@ Ort Environment System
                 "type": "IsolatedORT",
                 "config": {
                     "python_environment_path": "/home/user/.virtualenvs/myenv/bin",
-                    "accelerators": ["cpu"]
+                    "accelerators": [{"device": "cpu"}]
```
jambayk, line 385: Can you also add ep in this example to make it consistent with the above systems? 
I also forgot to update the title at line 374. Can you correct that too? Thanks!

docs/source/tutorials/configure_systems.rst
guotuofeng: As I describe offline, there will be a followup work item https://aiinfra.visualstudio.com/Model%20optimization%20Toolkit/_workitems/edit/22491/ for how to use system. I will update then

docs/source/tutorials/configure_systems.rst
guotuofeng: will do

examples/bert/bert_cuda_gpu.json
guotuofeng: we can make the examples configure as simple as possible. This is suggested from @devang-ml. 

IMHO, basically, so long as the example could works. the config for hardware doesn't matter.

olive/hardware/accelerator.py
```diff
@@ -92,89 +92,192 @@ def filter_execution_providers(execution_providers, available_providers):
...
             return None
 
-        is_unique_inferring = True
-        accelerators = []
-        for idx, ep in enumerate(execution_provider):
-            accelerators.append([])
+        ep_to_accelerator = {}
+        for ep in execution_providers:
+            if ep == "CPUExecutionProvider":
+                # cannot infer device for CPUExecutionProvider since all ORT EP supports CPU
```
devang-ml, line 113: As a special case, If CPUExecutionProvider is explicitly listed alone (without any other EPs) then it should be sufficient to infer CPU.

jambayk: Mike told me he has added this as a special case already. It is here https://github.com/microsoft/Olive/blob/9c7dc9f7303b5527857ca8bd183a59beb601135e/olive/hardware/accelerator.py#L203

guotuofeng: Yeah, @jambayk is right. The logic is handled in the caller, If the EPs(after removing AzureExecutionProvider) is only one "CPUExecutionProvider", the inferred device will be cpu.

jambayk (approved):

PR #1003, opened 2024-03-08
Add `SpecialParamValue` to pass config param type, Dynamic quantization UT

jambayk:
## Describe your changes
This PR fixes a bug in the type of a pass config param where it can be `SpecialParamValue` but this enum was not added to the type. 
This case happens for `ONNXQuantization` pass where the values for `static_option_config` params are set to `SpecialParamValue.IGNORED` when `quant_mode` is `dynamic`. The bug was not caught earlier since all previous optional params were of type string, so it automatically passed the string validation. We also didn't have a dynamic quantization unit test. 

Added a unit test for dynamic quantization. However, there is a bug in ORT>=1.17.0 because of which the dynamic quantizer fails. so added a check for the failure for the affected versions.

### DIFF
--- a/olive/passes/onnx/quantization.py
+++ b/olive/passes/onnx/quantization.py
@@ -332,6 +332,9 @@ class OnnxQuantization(Pass):
             assert (
                 config["dataloader_func"] or config["data_config"]
             ), "dataloader_func or data_config is required for static quantization."
+            # whether to prepare qnn config
+            if config["prepare_qnn_config"] and version.parse(OrtVersion) < version.parse("1.17.0"):
+                raise OlivePassError("prepare_qnn_config is only supported for onnxruntime-qnn>=1.17.0")
 
         output_model_path = resolve_onnx_path(output_model_path, Path(model.model_path).name)
 
@@ -364,10 +367,6 @@ class OnnxQuantization(Pass):
             else:
                 logger.info("Already processed model for quantization, skipping preprocessing")
                 model = ONNXModelHandler(LocalFile({"path": preprocessed_temp_model_path}))
-
-        # whether to prepare qnn config
-        if run_config.get("prepare_qnn_config", False) and version.parse(OrtVersion) < version.parse("1.17.0"):
-            raise OlivePassError("prepare_qnn_config is only supported for onnxruntime-qnn>=1.17.0")
 
         # keys not needed for quantization
         to_delete = [
--- a/test/unit_test/passes/onnx/test_quantization.py
+++ b/test/unit_test/passes/onnx/test_quantization.py
@@ -33,7 +33,7 @@ def dummpy_dataloader_func(data_dir, batch_size, *args, **kwargs):
 
 
 @pytest.mark.parametrize("calibrate_method", ["MinMax", "Entropy", "Percentile"])
-def test_quantization(calibrate_method, tmp_path):
+def test_static_quantization(calibrate_method, tmp_path):
     input_model = get_onnx_model()
     config = {
         "quant_mode": "static",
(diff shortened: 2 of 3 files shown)


test/unit_test/passes/onnx/test_quantization.py
```diff
@@ -52,6 +52,20 @@ def test_quantization(calibrate_method, tmp_path):
     assert out is not None
 
 
+def test_dynamic_quantization(tmp_path):
```
guotuofeng, line 55: do we need failure case?

olive/passes/onnx/quantization.py
```diff
@@ -332,6 +332,9 @@ def _run_for_config(
             assert (
                 config["dataloader_func"] or config["data_config"]
             ), "dataloader_func or data_config is required for static quantization."
+            # whether to prepare qnn config
```
guotuofeng, line 335: could we move them into validate_search_point?

jambayk: Yeah, I can move it there. Although, I have been thinking that version check in validate_search_point doesn't work well for non local systems. Validate_search_point is run by the engine so the version of the engine machine might be different from the host/target. But we can address that later.

test/unit_test/passes/onnx/test_quantization.py
jambayk: I am not sure what a failure case would look like since the default config and search points are valid. The bug cannot be reproduced for a failure case since it is not directly related to this pass or mode.

olive/passes/onnx/quantization.py
jambayk: Actually, I think it is better inside the run itself. There is only one pass `transformers optimization` which does the version check in search point validation. But that is technically not a good way because of what i described above.

guotuofeng: could we add comments about the reason that this kind of logic doesn't belong to validate_search_point? actually, I don't know it doesn't work well with non-local system.

test/unit_test/passes/onnx/test_quantization.py
guotuofeng: so, I suppose the test case will catch the bug if the production code is reverted?

jambayk: yes, this test fails if I revert the param type fix.

olive/passes/onnx/quantization.py
```diff
@@ -332,6 +332,12 @@ def _run_for_config(
...
+            # we do the version check here and not in `_validate_search_point` since search point validation
```
guotuofeng, line 336: no underscore. `validate_search_point`

jambayk: Thanks! corrected

guotuofeng (approved):

test/unit_test/passes/onnx/test_quantization.py
```diff
@@ -52,6 +52,20 @@ def test_quantization(calibrate_method, tmp_path):
...
+    input_model = get_onnx_model()
+    config = {"quant_mode": "dynamic"}
+    p = create_pass_from_dict(OnnxQuantization, config, disable_search=True)
+
+    if version.parse(OrtVersion) >= version.parse("1.17.0"):
```
guotuofeng, line 60: do we need change it to 
> 1.16.0 and < 1.18.0 

In this case, the ort-nightly would fail

jambayk: update to check for 1.17.x

guotuofeng (approved):

PR #1011, opened 2024-03-13
fix windows shutil.which return None when passing full python path

guotuofeng:


olive/common/utils.py
```diff
@@ -26,10 +26,7 @@ def run_subprocess(cmd, env=None, cwd=None, check=False):
     windows = platform.system() == "Windows"
     if isinstance(cmd, str):
         cmd = shlex.split(cmd, posix=not windows)
```
trajepl, line 28: Seems, `posix=not windows` is not needed, either?

guotuofeng: I suppose it is needed.

trajepl (approved):

trajepl (approved):

PR #1013, opened 2024-03-13
🅰️ AutoOpt disable search when no evaluator

trajepl:
## Describe your changes
This PR is used for:
1. If there no passes given, enable AutoOpt be default.
2. If the evaluation config is given, in AutoOpt case, we should set disable_search as True.

### DIFF
--- a/olive/workflows/run/run.py
+++ b/olive/workflows/run/run.py
@@ -165,8 +165,10 @@ def run_engine(config: RunConfig, data_root: str = None):
 
         AzureMLSystem.olive_config = config.to_json()
 
-    no_evaluation = engine.evaluator_config is None and all(
-        pass_config.evaluator is None for pass_config in config.passes.values()
+    no_evaluation = (
+        engine.evaluator_config is None
+        and config.passes
+        and all(pass_config.evaluator is None for pass_config in config.passes.values())
     )
     accelerator_specs = create_accelerators(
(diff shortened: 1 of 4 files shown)


olive/workflows/run/run.py
```diff
@@ -165,8 +165,10 @@ def run_engine(config: RunConfig, data_root: str = None):
 
         AzureMLSystem.olive_config = config.to_json()
 
-    no_evaluation = engine.evaluator_config is None and all(
-        pass_config.evaluator is None for pass_config in config.passes.values()
+    no_evaluation = (
+        engine.evaluator_config is None
+        and config.passes
```
guotuofeng, line 170: when the pass is None?

trajepl: If there no passes given in input model config, we need to check `config.passes` since we need to enable auto opt. See the PR description item 1.

guotuofeng: but the no_evaluation only used by create_accelerators? how it related to auto opt?

trajepl: no_evaluation is not related with auto opt. I just triggered the issue when I enable auto opt.

trajepl: image
That is introduced in https://github.com/microsoft/Olive/pull/937

guotuofeng (approved):

Issue #1014, opened 2024-03-13
Whisper model converted via onnxruntime 1.17.1 won't work


khmyznikov:
**Describe the bug**
If you use Olive example passes for whisper-small model using onnxruntime-1.17.1, for example whisper_cpu_int8.json, 
model will not work. Just no output at all with no errors.

**To Reproduce**
- Use onnxruntime 1.17.1
- `python .\prepare_whisper_configs.py --model_name "openai/whisper-small"  --multilingual --enable_timestamps --package_model --skip_evaluation --no_audio_decoder`
- `python -m olive.workflows.run --config .\whisper_cpu_int8.json `
- `python test_transcription.py --config whisper_cpu_int8.json --predict_timestamps `

**Expected behavior**
Should have some output, on 1.16.3 everything works fine

**Other information**
 - OS: Windows
 - Olive version: 0.6.0
 - ONNXRuntime package and version: onnxruntime-gpu: 1.17.1

xiaoyu-work:
What error did you get? Can you try without `--predict_timestamps` if it works?

jambayk:
Hi, thanks for reporting this issue. I have opened a PR to fix this https://github.com/microsoft/Olive/pull/1016

jambayk:
Fix has been merged. Please reopen if it doesn't work on your side.

khmyznikov:
Still doesn't work for me... Have converted with latest Olive and still just silent no output.

PR #1016, opened 2024-03-13
Whisper: Support ORT 1.17.1

jambayk:
## Describe your changes
ORT 1.17.1 made some changes to the beam search ops. For version 1.17.1, the beam search op used will be `WhisperBeamSearch`. 

Fixes:
#1014

### DIFF
--- a/olive/passes/onnx/insert_beam_search.py
+++ b/olive/passes/onnx/insert_beam_search.py
@@ -21,7 +21,7 @@ logger = logging.getLogger(__name__)
 
 
 class InsertBeamSearch(Pass):
-    """Insert Beam Search Op."""
+    """Insert Beam Search Op. Uses WhisperBeamSearch op if ORT version >= 1.17.1, else uses BeamSearch op."""
 
     _accepts_composite_model = True
 
@@ -29,6 +29,13 @@ class InsertBeamSearch(Pass):
     def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
         use_gpu = accelerator_spec.accelerator_type == Device.GPU
         config = {
+            "model_name": PassConfigParam(
+                type_=str,
+                description=(
+                    "Name of the model. Such as openai/whisper-tiny. Will try to infer from model attributes if not"
+                    " provided."
+                ),
+            ),
             "no_repeat_ngram_size": PassConfigParam(
                 type_=int,
                 default_value=0,
@@ -65,6 +72,13 @@ class InsertBeamSearch(Pass):
                     " 1.16.0"
                 ),
             ),
+            "use_temperature": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                description=(
+                    "Use temperature as an extra graph input to the beam search op. Only supported in ORT >= 1.17.1"
+                ),
+            ),
             "fp16": PassConfigParam(
                 type_=bool,
                 default_value=False,
@@ -84,9 +98,20 @@ class InsertBeamSearch(Pass):
     ):
         from onnxruntime import __version__ as OrtVersion
         from onnxruntime.transformers.convert_generation import get_shared_initializers
+        from transformers import AutoTokenizer
 
         # version check
         version_1_16 = version.parse(OrtVersion) >= version.parse("1.16.0")
+        version_1_17_1 = version.parse(OrtVersion) >= version.parse("1.17.1")
+        # NOTE: will ignore cross qk related options for now
+
+        # get tokenizer
+        model_name = options["model_name"] or model_config.get("_name_or_path")
+        if not model_name:
+            raise ValueError("model_name is not provided and cannot be inferred from model attributes.")
+        if not options["model_name"]:
+            logger.info("Inferred model name from model attributes: %s", model_name)
+        tokenizer = AutoTokenizer.from_pretrained(model_name)
 
         # Chain two models (model_A and model_B) by inserting beam search op in between.
         model_A.graph.name = f"{model_A_name} subgraph"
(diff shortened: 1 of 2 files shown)


xiaoyu-work:
I'm not sure if this is the root cause, cause I'm able to run `test_transcription.py` with ORT 1.17.1 successfully without `--predict_timestamps`. Do you know the reason?

jambayk:
> I'm not sure if this is the root cause, cause I'm able to run `test_transcription.py` with ORT 1.17.1 successfully without `--predict_timestamps`. Do you know the reason?

@xiaoyu-work I am not entirely sure about the root cause but you can see in this PR https://github.com/microsoft/onnxruntime/pull/19509 that some token ids are not part of the model attributes and not inferred using offsets. This affects the logits processor which does the time stamp prediction. I think the lack of these values in the beam search node's attributes causes the issue.

xiaoyu-work:
> > I'm not sure if this is the root cause, cause I'm able to run `test_transcription.py` with ORT 1.17.1 successfully without `--predict_timestamps`. Do you know the reason?
> 
> @xiaoyu-work you can see in this PR [microsoft/onnxruntime#19509](https://github.com/microsoft/onnxruntime/pull/19509) that some token ids are now part of the model attributes and not inferred using offsets. This affects the logits processor which does the time stamp prediction. I think the lack of these values in the beam search node's attributes causes the issue.

Can you test if this change works?

jambayk:
> > > I'm not sure if this is the root cause, cause I'm able to run `test_transcription.py` with ORT 1.17.1 successfully without `--predict_timestamps`. Do you know the reason?
> > 
> > 
> > @xiaoyu-work you can see in this PR [microsoft/onnxruntime#19509](https://github.com/microsoft/onnxruntime/pull/19509) that some token ids are now part of the model attributes and not inferred using offsets. This affects the logits processor which does the time stamp prediction. I think the lack of these values in the beam search node's attributes causes the issue.
> 
> Can you test if this change works?

It's already tested.

xiaoyu-work (dismissed):

olive/passes/onnx/insert_beam_search.py
```diff
@@ -84,9 +98,20 @@ def chain_model(
...
+        version_1_17_1 = version.parse(OrtVersion) >= version.parse("1.17.1")
+        # NOTE: will ignore cross qk related options for now
+
+        # get tokenizer
+        model_name = options["model_name"] or model_config.get("_name_or_path")
+        if not model_name:
+            raise ValueError("model_name is not provided and cannot be inferred from model attributes.")
+        if not options["model_name"]:
+            logger.info("Inferred model name from model attributes: %s", model_name)
+        tokenizer = AutoTokenizer.from_pretrained(model_name)
```
xiaoyu-work, line 114: Nit: seems only ORT >= 1.17.1 needs load tokennizer. Add a condition check here can save unnecessary load.

jambayk: Thanks! updated to only get tokenizer when needed.

xiaoyu-work (dismissed):

olive/passes/onnx/insert_beam_search.py
```diff
@@ -21,14 +21,21 @@
...
-    """Insert Beam Search Op."""
+    """Insert Beam Search Op. Uses WhisperBeamSearch op if ORT version >= 1.17.1, else uses BeamSearch op."""
 
     _accepts_composite_model = True
 
     @classmethod
     def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
         use_gpu = accelerator_spec.accelerator_type == Device.GPU
         config = {
+            "model_name": PassConfigParam(
```
guotuofeng, line 32: could the model_name be get from model_config or hf_config?

jambayk: yes, you can see in the description that the model_name will be inferred from the model_config is not provided. I kept this here in case they user has a model with custom tokenization and want to use it.

guotuofeng: for custom tokenizer case, should we provide the function to load the tokenizer in user_script? customized tokenizer might not be supported by  AutoTokenizer.from_pretrained

guotuofeng: From my understanding, AutoTokenizer.from_pretrained will mostly be the model supported by Hugging Face. Is that the case?

jambayk: the tokenizer with the same. It is just that they might have changed the tokenizer config json when saving the model. so as long as they provide the correct model name or path here, it will be okay.

guotuofeng: so we will not support https://github.com/BlinkDL/ChatRWKV/blob/main/tokenizer/rwkv_tokenizer.py?

jambayk: i was not clear, I meant if they changed the vocabulary which would in turn change the tokenization.

jambayk: the use case is similar to `model_name` here https://github.com/microsoft/Olive/blob/main/olive/passes/utils/whisper_prepost.py#L14. This is to support locally save fine-tuned hf whisper models that might have updated vocab, etc. Not to support a custom model or tokenizer class.

guotuofeng: for whisper, I am suppose the model_name will not be used?

jambayk: updated to remove as discussed offline.

guotuofeng (approved):

PR #1017, opened 2024-03-14
use sys.executable instead of python

guotuofeng:
## Describe your changes
* according to the best practice https://docs.python.org/3/library/subprocess.html#popen-constructor, we should use sys.executable instead of python to specify the full path of executable.
* for other program, prefer specify the full path by calling `[shutil.which()](https://docs.python.org/3/library/shutil.html#shutil.which) to expand the full path of the program.
image

### DIFF
--- a/olive/systems/isolated_ort/isolated_ort_system.py
+++ b/olive/systems/isolated_ort/isolated_ort_system.py
@@ -5,6 +5,7 @@
 import collections
 import json
 import logging
+import sys
 import tempfile
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple, Union
@@ -128,7 +129,7 @@ class IsolatedORTEvaluator(OliveEvaluator, OnnxEvaluatorMixin, framework="ort_in
         output_dir: Union[str, Path],
     ):
         command = [
-            "python",
+            sys.executable,
             str(self.inference_runner_path),
             "--config_path",
             str(config_path),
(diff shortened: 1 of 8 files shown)


olive/systems/isolated_ort/isolated_ort_system.py
```diff
@@ -128,7 +129,7 @@ def _run_inference(
         output_dir: Union[str, Path],
     ):
         command = [
-            "python",
+            sys.executable,
```
jambayk, line 132: this cannot be sys.executable since it points to the python executable of the engine. We need the python executable in the environment.

olive/systems/python_environment/python_environment_system.py
```diff
@@ -78,7 +79,7 @@ def _run_command(self, script_path: Path, config_jsons: Dict[str, Any], **kwargs
             tmp_dir_path = Path(tmp_dir).resolve()
 
             # command to run
-            command = ["python", str(script_path)]
```
jambayk, line 81: same here

olive/systems/isolated_ort/isolated_ort_system.py
jambayk: if we want a full path, we could add an attribute in the init like `self.executable = shutil.which(...)` and use it.

guotuofeng: let's see whether the CI fail or not. If not fail, then it is a bug.

olive/systems/python_environment/python_environment_system.py
guotuofeng: good catch

olive/systems/isolated_ort/isolated_ort_system.py
jambayk: The ci will probably pass since the test environment has all of the packages installed. We only test for cpu package I think.

jambayk (approved):

PR #1018, opened 2024-03-14
📤 Output model config fix in run_no_search

trajepl:
## Describe your changes

The case of run_no_search, if the final model output is composite model, the model_config will be None which leads the error output packing.

```json
{
    "input_model":{
        "type": "PyTorchModel",
        "config": {
            "hf_config": {
                "model_name": "microsoft/trocr-base-handwritten",
                "task": "image-to-text"
            }
        }
    },
    "passes": {
        "conversion": {
            "type": "OptimumConversion"
        }
    },
    "engine": {
        "cache_dir": "cache",
        "packaging_config": {
            "type": "Zipfile", <---- should be removed
            "name": "OutputModels" <---- should be removed
        },
        "output_dir" : "models"
    }
}
```

Basically, Olive did not support pack composite model and we should add log to tell user when they encounter similar issues.

This PR is used to fix above issue.

### DIFF
--- a/olive/cache.py
+++ b/olive/cache.py
@@ -7,7 +7,7 @@ import logging
 import os
 import shutil
 from pathlib import Path
-from typing import Optional, Union
+from typing import Dict, Optional, Union
 
 from olive.common.config_utils import serialize_to_json
 from olive.common.utils import hash_dict
@@ -197,7 +197,7 @@ def save_model(
     output_name: Union[str, Path] = None,
     overwrite: bool = False,
     cache_dir: Union[str, Path] = ".olive-cache",
-):
+) -> Optional[Dict]:
     """Save a model from the cache to a given path."""
     # This function should probably be outside of the cache module
     # just to be safe, import lazily to avoid any future circular imports
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -459,6 +459,8 @@ class Engine:
                     overwrite=True,
                     cache_dir=self._config.cache_dir,
                 )
+                # it is not supported to save compositepytorchmodel/compositemodel again
+                # so the output_model_json could be None
                 output_models[pass_output_model_id] = output_model_json
 
             # save the evaluation results to output_dir
@@ -471,7 +473,7 @@ class Engine:
         fp_outputs = self.footprints[accelerator_spec].create_footprints_by_model_ids(output_model_ids)
         # update the output model config
         for model_id, model_config in output_models.items():
-            fp_outputs.nodes[model_id].model_config = model_config
+            fp_outputs.nodes[model_id].model_config = model_config or fp_outputs.nodes[model_id].model_config
 
         return fp_outputs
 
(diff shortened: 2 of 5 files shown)


olive/engine/engine.py
```diff
@@ -471,7 +473,7 @@ def run_no_search(
         fp_outputs = self.footprints[accelerator_spec].create_footprints_by_model_ids(output_model_ids)
         # update the output model config
         for model_id, model_config in output_models.items():
-            fp_outputs.nodes[model_id].model_config = model_config
+            fp_outputs.nodes[model_id].model_config = model_config or fp_outputs.nodes[model_id].model_config
```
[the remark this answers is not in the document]
trajepl: `fp_outputs' model_config` is the config of cached model. The `model_config` from output_models might be changed when the model are copied to output_dir.

guotuofeng:
It seems https://github.com/microsoft/Olive/pull/209 return None for composite model

trajepl:
> It seems #209 return None for composite model
Thanks, also added Jambay to review.

olive/cache.py
jambayk, line 1: The changes look good to me. I skipped composite model saving since they were always intermediate models at the time and I wasn't sure how I should structure the directory structure when saving the model again. 

This is also complicated for composite models created by optimum conversion since all model files are in the same directory when each component model is supposed to be in their own directory. So, it makes sense to just point them to the cached location.

jambayk (approved):

olive/engine/engine.py
guotuofeng, line 476: I prefer the following line even if it is same with above line, since
1, the logic is clear
2. no perf penalty. 
```
if model_config:
    fp_outputs.nodes[model_id].model_config = model_config
```

guotuofeng (dismissed):

olive/engine/engine.py
guotuofeng: https://stackoverflow.com/questions/8747740/assignment-with-or-in-python

jambayk: This is an interesting discussion. We do use `or` in our codebase to check for `None`. 

Both the current line and the proposed line do use the fact that `None` evaluates to False in a boolean expression. But I do see the benefit of point 3. Point 1 is more about if the developer is used to this type of logic or not.

trajepl: Updated.

guotuofeng (approved):

olive/engine/engine.py
trajepl: Shared from Mike:
https://stackoverflow.com/questions/21566106/what-does-an-x-y-or-z-assignment-do-in-python

For some easy/readable case, `or` looks pretty clean. I think there is a point looks good:
` final element should probably be a literal`

jambayk: The literal part makes sense if we are defining a new variable like 'x = y or z or default' where we don't know if at least one of x or y is "True"

For the current case, we know that the second argument is always "True" so I don't think it really applies here.

PR #1022, opened 2024-03-18
🔑  Update kv cache input type

trajepl:
## Describe your changes

1. For phi2 optimization, the transformer optimization keep_io_bytes is False. That means, the input/output of kv should be float16. Update the generation example to fix.
2. optimum conversion does not require ort >=1.18. Update the condition in this PR.

### DIFF
--- a/examples/phi2/generate.py
+++ b/examples/phi2/generate.py
@@ -31,7 +31,7 @@ class ORTGenerator:
         self.use_buffer_share = False
         self.packed_kv = False
         self.use_step = False
-        self.torch_dtype = None
+        self.kv_torch_dtype = None
         self.sess = None
         self.tokenizer = None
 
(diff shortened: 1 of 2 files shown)


examples/phi2/generate.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, decoder_path):
         self.use_buffer_share = False
         self.packed_kv = False
         self.use_step = False
-        self.torch_dtype = None
+        self.kv_torch_dtype = None
```
jambayk, line 34: it might be clearer if the value is set to `torch.float32` instead of `None`?

jambayk: Oh sorry, I didn't see line 141 where the value is set. Please ignore this.

jambayk (approved):

PR #1024, opened 2024-03-19
🆙 Update matmul 4bit for ort>=1.17.0

trajepl:
## Describe your changes
Update arguments of matmul 4bit for ort>=1.17.0
1. accuracy_level
2. algorithm

### DIFF
--- a/olive/passes/onnx/quantization.py
+++ b/olive/passes/onnx/quantization.py
@@ -594,6 +594,29 @@ class OnnxMatMul4Quantizer(Pass):
                 default_value=None,
                 description="List of node names to exclude from quantization.",
             ),
+            "accuracy_level": PassConfigParam(
+                type_=Any,
+                default_value=None,
+                description="Available from onnxruntime>=1.17.0 "
+                "The minimum accuracy level of input A, can be: 0(unset), 1(fp32), 2(fp16), 3(bf16), "
+                "or 4(int8) (default unset when 0 or None). It is used to control how input A is quantized or downcast "
+                "internally while doing computation, for example: 0 means input A will not be quantized "
+                "or downcast while doing computation. 4 means input A can be quantized with the same "
+                "block_size to int8 internally from type T1. "
+                "Refer to the MatMulNBits contrib op's 'accuracy_level' attribute for details "
+                "(https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftmatmulnbits).",
+            ),
+            "algorithm": PassConfigParam(
+                type_=str,
+                default_value=None,
+                description="Available from onnxruntime>=1.17.0 "
+                "Algorithm config for quantization, can be 'DEFAULT', 'HQQ' or 'RTN', 'GPTQ' "
+                "(default DEFAULT when not set). For 4b quantize a model with RTN or GPTQ algorithm. "
+                "Please refer to https://github.com/intel/neural-compressor/blob/master/docs/source/quantization_weight_only.md "  # noqa: E501
+                "for more details on weight only quantization using Intel® Neural Compressor. "
+                "For HQQ, please refer to onnxruntime for more details: "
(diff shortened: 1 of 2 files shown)


olive/passes/onnx/quantization.py
```diff
@@ -594,6 +594,29 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
                 default_value=None,
                 description="List of node names to exclude from quantization.",
             ),
+            "accuracy_level": PassConfigParam(
+                type_=Any,
```
xiaoyu-work, line 598: Should this be `int`?

olive/passes/onnx/quantization.py
```diff
@@ -594,6 +594,29 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
...
+                description="Available from onnxruntime>=1.17.0 "
+                "The minimum accuracy level of input A, can be: 0(unset), 1(fp32), 2(fp16), 3(bf16), "
+                "or 4(int8) (default unset when 0 or None). It is used to control how input A is quantized or downcast "
+                "internally while doing computation, for example: 0 means input A will not be quantized "
+                "or downcast while doing computation. 4 means input A can be quantized with the same "
+                "block_size to int8 internally from type T1. "
+                "Refer to the MatMulNBits contrib op's 'accuracy_level' attribute for details "
+                "(https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftmatmulnbits).",
+            ),
+            "algorithm": PassConfigParam(
```
xiaoyu-work, line 609: What if the input str value is not desired? Should we either add a value validation here, or having an enum class defined for validation for this value?

guotuofeng:
Any user requires this feature?

olive/passes/onnx/quantization.py
trajepl: Updated.

olive/passes/onnx/quantization.py
```diff
@@ -610,13 +642,57 @@ def _run_for_config(
...
+    if not v:
+        return v
+
+    if v not in (0, 1, 2, 3, 4):
+        raise ValueError(f"OnnxMatMul4Quantizer {field.name} must be 0(unset), 1(fp32), 2(fp16), 3(bf16) or 4(int8)")
+
+    return v
+
+
+def _validate_algorithm(v, values, field):
```
guotuofeng, line 683: the validate_algorithm seems has additional logic to verify neural compressor installation. Better to make sure only algo valiation logic in this function and move the validation inc logic somewhere else.

olive/passes/onnx/quantization.py
```diff
@@ -610,13 +642,58 @@ def _run_for_config(
 
         output_model_path = resolve_onnx_path(output_model_path, Path(model.model_path).name)
 
-        quant = MatMul4BitsQuantizer(
-            model.load_model(), config["block_size"], config["is_symmetric"], config["nodes_to_exclude"]
-        )
+        if version.parse(OrtVersion) >= version.parse("1.17.0"):
+            from onnxruntime.quantization.matmul_4bits_quantizer import WeightOnlyQuantConfig
+
+            if config["algorithm"] in ("RTN", "GPTQ"):
```
guotuofeng, line 648: I suppose the underlying ORT will do raise the ImportError and we might need not this check?

olive/passes/onnx/quantization.py
jambayk: saw the changes have already been made but just wanted to agree to Mikes point. Version checks and import checkers are runtime decisions to be run on the pass host. doing it during config validation or search point validation runs it on the engine's machine which is not necessarily the pass host.

olive/passes/onnx/quantization.py
trajepl: Yes, the import error will be raised in ort as well. But no extra info to tell which config needs inc package. 
Maybe we can keep it here to provide clear description once user choose the wrong config.

PR #1025, opened 2024-03-20
Remove export in gptq pass

Taka152:
## Describe your changes
The current gptq pass includes onnx export since there are some edge cases where models defined in user_script cannot be successfully saved and loaded.
This PR fixes this problem and removes export in gptq pass, it also enables other conversion methods and pytorch passes combined with gptq pass.


olive/model/handler/pytorch.py
```diff
@@ -127,8 +127,8 @@ def load_model(self, rank: int = None) -> torch.nn.Module:
             return self.model
 
         # Load special path or format model -> load model from hf config -> load normal path model
+        user_module_loader = UserModuleLoader(self.model_script, self.script_dir)
```
guotuofeng, line 130: add comments why need move the user_module_loader here.

olive/passes/pytorch/gptq.py
```diff
@@ -215,16 +197,17 @@ def get_onnx_quant_linear(*args, **kwargs):
...
 
-        converted_onnx_model = OnnxConversion._export_pytorch_model(  # pylint: disable=protected-access
-            quantized_model,
-            model.get_dummy_inputs(),
-            model.get_io_config(),
-            config,
-            config["export_device"],
-            None,
-            tempfile.tempdir,
+        output_model_path = Path(output_model_path).with_suffix(".pt")
```
guotuofeng, line 200: will the output_model_path be a directory or file name?

examples/llama2/user_script.py
```diff
@@ -60,7 +61,7 @@ def get_merged_sample_with_past_kv_inputs(
     elif model.hf_config is not None:
         config = model.get_hf_model_config() if model else LlamaConfig.from_pretrained(model_id)
     else:
-        config = model.model_attributes
+        config = Namespace(**model.model_attributes)
```
guotuofeng, line 64: add comments why need the namespace and what's the problem of original calls.

Taka152: Comments updated. The config variable is used via config.item, hf_config can be used via this but when custom models it is saved in model_attributes dict. I create a Namespace class here to use it via config.item rather than config["item"].

olive/passes/pytorch/gptq.py
guotuofeng: @jambayk, I am suppose the output_model_path is a directory? do we need append model file name?

Taka152: Maybe output_model_path / "model.pt" is better?

examples/llama2/user_script.py
trajepl: nit: https://docs.python.org/3/library/types.html#types.SimpleNamespace
there is a `SimpleNamespace` in python `types` could achieve the same thing.

olive/passes/pytorch/gptq.py
jambayk: `output_model_path` is neither a file or directory https://github.com/microsoft/Olive/blob/d5c489ed60c91ea3ea9c1a28a90c2fe400e86bb1/olive/engine/engine.py#L915 

So you can either just attach a `.pt` suffix and save to it directly or do something like https://github.com/microsoft/Olive/blob/d5c489ed60c91ea3ea9c1a28a90c2fe400e86bb1/olive/passes/qnn/conversion.py#L108

olive/passes/onnx/__init__.py
```diff
@@ -31,7 +30,6 @@
     "AppendPrePostProcessingOps",
     "DynamicToFixedShape",
     "GenAIModelExporter",
-    "GptqQuantizer",
```
jambayk, line 34: Please also update the docs to move it from `sources/features/passes/onnx` to `pytorch`.

olive/passes/pytorch/gptq.py
```diff
@@ -215,16 +197,17 @@ def get_onnx_quant_linear(*args, **kwargs):
...
-            config,
-            config["export_device"],
-            None,
-            tempfile.tempdir,
+        output_model_path = normalize_path_suffix(output_model_path, "model.pt")
+        torch.save(quantized_model, output_model_path)
+
+        model_config = model.to_json()["config"]
+        model_config["model_path"] = output_model_path
+        model_config.pop("model_loader")
```
trajepl, line 205: image
Can we guarantee `model_loader` always exists? Maybe `model_config.pop("model_loader", None)` is better?

trajepl: image
Seems for pytorch model handler, the `model_loader` exists.

Taka152: Yeah, default value is missed, it should not raise error if "model_loader" not setted

Taka152: Yeah, default value is missed, it should not raise error if "model_loader" not setted

trajepl (approved):

guotuofeng (approved):

PR #1027, opened 2024-03-20
LoRA: Don't save embedding weights

jambayk:
## Describe your changes
- Provide `save_embedding_layers=False` to saving the peft model. Otherwise, it defaults to "auto" which checks if the vocab size changed. For models with no pad token, we add a dummy token and increase the embedding size but these are not trained at all. The fine-tuned model can be used without these extra dummy weights and tokens so we don't save them.
- Update the `extra_keys` logic in quant config validation since some attributes were made private and causes incorrect warning log.

### DIFF
--- a/olive/passes/pytorch/lora.py
+++ b/olive/passes/pytorch/lora.py
@@ -671,7 +671,9 @@ class LoRABase(Pass):
         # save adapter weights
         adapter_path = Path(output_model_path) / "adapter"
         adapter_path.mkdir(parents=True, exist_ok=True)
-        model.save_pretrained(adapter_path)
+        # don't save embedding layers since only adapter weights are trained
+        # if we don't provide as False, it defaults to "auto" which checks if the vocab size changed
+        model.save_pretrained(adapter_path, save_embedding_layers=False)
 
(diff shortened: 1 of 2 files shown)


olive/passes/pytorch/lora.py
```diff
@@ -671,7 +671,9 @@ def train_and_save_new_model(
         # save adapter weights
         adapter_path = Path(output_model_path) / "adapter"
         adapter_path.mkdir(parents=True, exist_ok=True)
-        model.save_pretrained(adapter_path)
+        # don't save embedding layers since only adapter weights are trained
+        # if we don't provide as False, it defaults to "auto" which checks if the vocab size changed
+        model.save_pretrained(adapter_path, save_embedding_layers=False)
```
guotuofeng, line 676: does the save_embedding_layers need to be configurable by exposing it in the pass config?

jambayk: it doesn't. We only train the adapters. There are more details under the `smart_tokenizer_...` method.

trajepl (approved):

PR #1028, opened 2024-03-20
[Performance]: Delayed Python pass module load

shaahji:
### [Performance]: Delayed Python module load
Introducing olive_config.json to encapsulates all the different global environment properties (like pass module configuration) to setup olive. Also, merged extra_dependencies.json into olive_config.json.

User can provide a alternative configuration file with the command line argument '--package-config'.

Instead of loading all modules (specifically passes) at launch, delay the load until after the run config is parsed and load only
the ones that are explicitly required.

Moved a few hard-coded dependency management into the olive config.

### Release Note:
Added support for extending available Olive passes and configuring each via data file (olive_config.json). The configuration file is provided to Olive run with command line param `--package-config`.


olive/passes/onnx/vitis_ai/__init__.py
```diff
@@ -2,19 +2,3 @@
 # Copyright (C) 2023, Advanced Micro Devices, Inc. All rights reserved.
 # SPDX-License-Identifier: MIT
 #
-from onnxruntime.quantization.calibrate import CalibrationDataReader
```
jambayk, line 5: these are not passes and the vitis quantizer depends on this init file. I think it's okay to keep them here since the pass only imports then during run https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/vitis_ai_quantization.py#L266

olive/workflows/run/__main__.py
```diff
@@ -3,21 +3,39 @@
...
     parser.add_argument("--tempdir", type=str, help="Root directory for tempfile directories and files", required=False)
 
     args = parser.parse_args()
 
+    if "--config" in sys.argv:
+        print(  # noqa:T201
+            "WARNING: Option '--config' is deprecated and will be removed in future versions. "
+            "Use '--run-config' instead."
+        )
+    if "--data_root" in sys.argv:
```
jambayk, line 27: is there a reason for preferring `-` over `_`?

olive/workflows/run/config.py
```diff
@@ -24,6 +24,23 @@
 logger = logging.getLogger(__name__)
 
 
+class PassConfig(ConfigBase):
```
jambayk, line 27: Can you use a different name like `PassModuleConfig` or something since we already have similar named configs like `RunPassConfig`, `FullPassConfig`, `PassConfigParam`?

olive/olive_config.json
```diff
@@ -0,0 +1,243 @@
+{
+    "passes": {
+        "AppendPrePostProcessingOps": {
+            "module_path": "olive.passes.onnx.append_pre_post_processing_ops.AppendPrePostProcessingOps"
```
guotuofeng, line 4: the stringlized module full path is hard to maintain. User might forget to update the module path when it is changed.

How could we make sure the path here is updated whenever module path changes?

docs/source/getstarted/installation.md
```diff
@@ -62,5 +62,6 @@ pip install -e .
 ```
 
 ## Optional Dependencies
-Olive has optional dependencies that can be installed to enable additional features. Please refer to [extra dependencies](https://github.com/microsoft/Olive/blob/main/olive/extra_dependencies.json)
-for the list of extras and their dependencies.
+Olive has optional dependencies that can be installed to enable additional features. Please refer to
+[extra dependencies](https://github.com/microsoft/Olive/blob/main/olive/olive_config.json) for the list of extras
```
guotuofeng, line 66: need refer to your private branch. after the PR is merged, update to main.

guotuofeng:
> Instead of loading all modules (specifically passes) at launch, delay the load of modules until after the run config is parsed and load only the ones that are explicitly required.

What's the perf improvement compared with all passes are imported? From the time being, only the pass itself is imported and the dependencies only be imported until run. So, could I suppose the perf improvement is limit? What's the consideration of this PR?

guotuofeng (dismissed):
need to understand the consideration or background about why we need the delay load PR.

olive/workflows/run/__main__.py
```diff
@@ -3,21 +3,39 @@
...
+    parser.add_argument("--data-root", "--data_root", help="The data root path for optimization", required=False)
     parser.add_argument("--tempdir", type=str, help="Root directory for tempfile directories and files", required=False)
 
     args = parser.parse_args()
 
+    if "--config" in sys.argv:
+        print(  # noqa:T201
+            "WARNING: Option '--config' is deprecated and will be removed in future versions. "
+            "Use '--run-config' instead."
+        )
```
devang-ml, line 26: We don't want to deprecate --config.
Is there a reason to introduce new --run-config ?

olive/workflows/run/__main__.py
```diff
@@ -3,21 +3,39 @@
...
+from pathlib import Path
 
 from olive.common.utils import set_tempdir
 from olive.workflows import run
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser("Olive Workflow: Custom Run")
-    parser.add_argument("--config", type=str, help="Path to json config file", required=True)
+    parser.add_argument("--olive-config", type=str, help="Path to Olive (json) config file", required=False)
+    parser.add_argument("--run-config", "--config", type=str, help="Path to json config file", required=True)
```
devang-ml, line 15: Description and name should be clearer that to indicate that this optional .json file, meant for advanced users only, lists location of passes you want to use in your --config.

olive/olive_config.json
devang-ml: As I understand, the change in path would require a PR and the pass would stop working if the string is not updated.

olive/workflows/run/__main__.py
shaahji: I read about the convention to use '-' in the parameter which python converts it to '_' - justification being that it distinguishes the use cases and avoids accidental changes with operations like find/replace.

Unfortunately, I can't seem to find the actual source link again. :(

For the moment, both conventions are supported.

olive/workflows/run/config.py
shaahji: In my opinion, this is the most appropriate name, especially since the other is named with prefix 'Run'. I would argue that the other classes should be named with the same prefix because they relate to the 'run'.

Regardless, how about `PassPackageConfig` since files are considered package in Python (borrowing the convention from `importlib.import_module` API where the second parameter is named `package`.)

olive/olive_config.json
shaahji: > As I understand, the change in path would require a PR and the pass would stop working if the string is not updated.

That's correct.

olive/workflows/run/__main__.py
shaahji: Merely to explicitly differentiate between `olive-config` and `the other config`. In my opinion, think naming the other config explicitly i.e. prefixing with 'run-' adds value.

docs/source/getstarted/installation.md
shaahji: I don't follow your comment. Please explain!

olive/workflows/run/__main__.py
shaahji: Are you referring to using something else in place of 'olive-config'? I started off with calling it passes-config but then realized there are other scattered input files that can be merged into one and that can serve as a more *entire module config* rather than having to explain to user what each config does.

shaahji:
> What's the perf improvement number compared with all passes are imported? From the time being, only the pass itself is imported and the dependencies only be imported until run. So, could I suppose the perf improvement is limited? What's the consideration of this PR?

> need to understand the consideration or background about why we need the delay load PR.

I don't think one would notice a significant performance gain with this change. All that loading/initialization is one time thing and not a hot path to really show up in any profiles. 

Primary motivation for this change -
* This approach gives the user a way to add his/her own pass and drive all with data.
* Load/initialization time might be small today with the limited passes we have but over time as the list of passes grow this might become significant. This change attempt to preemptively circumvent that scenario. 
* Passes can now freely include the necessary imports at package level rather than having to be localized in functions and such. Since the packages aren't loaded until a pass needs them, those dependencies won't matter.

@devang-ml can add more.

olive/passes/onnx/vitis_ai/__init__.py
shaahji: Fixex.

docs/source/getstarted/installation.md
guotuofeng: before the PR is merged, the main/olive/olive_config.json will not exist in main branch. As the result, the linkcheck would fail.

trajepl: There is link checker in our CI pipeline and it will fail if the link is unreachable.https://github.com/microsoft/Olive/pull/358

olive/olive_config.json
```diff
@@ -0,0 +1,234 @@
...
```
trajepl, line 4: Not related to this PR:
This central passes config looks nice. If we think further, it might be a good place to put pass's attributes like:
1. which ep this pass can be run
2. which precision(float32/float16/int4) the passes can conduct.
3. which ort version it is supported
That might be helpful for auto optimization as well.
Like this kind of configs
https://github.com/microsoft/Olive/blob/main/olive/auto_optimizer/config_template/pass_capability.yaml

olive/olive_config.json
xiaoyu-work: We can raise a specific error if this path couldn't find the pass.

olive/workflows/run/__main__.py
xiaoyu-work: Most people will not use advanced `--olive-config` but still `--config`. I think keeping `--config` will make Olive as simple to use as possible.

xiaoyu-work:
Can you complete the 5th checklist:  add a description of this change to be included in the release notes.

docs/source/getstarted/installation.md
shaahji: So, what's the workaround? Split the PR into two submissions? Or merge the PR while overriding/ignoring the lint checker?

olive/workflows/run/__main__.py
shaahji: Added descriptive help text.

olive/workflows/run/__main__.py
```diff
@@ -3,21 +3,48 @@
...
     parser.add_argument("--setup", help="Whether run environment setup", action="store_true")
-    parser.add_argument("--data_root", help="The data root path for optimization", required=False)
+    parser.add_argument("--data-root", "--data_root", help="The data root path for optimization", required=False)
     parser.add_argument("--tempdir", type=str, help="Root directory for tempfile directories and files", required=False)
 
     args = parser.parse_args()
 
+    if "--config" in sys.argv:
+        print(  # noqa:T201
```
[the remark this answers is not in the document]
shaahji: To avoid having to import logger so soon in the launch process. 

Regardless, those messages are removed.

olive/olive_config.json
shaahji: >  If we think further, it might be a good place to put pass's attributes

Agreed. There is lot more that can be data driven allowing easier user configuration changes.

olive/systems/azureml/aml_pass_runner.py
jambayk, line 1: Please also update https://github.com/microsoft/Olive/blob/main/olive/systems/python_environment/pass_runner.py

shaahji: Done!

olive/olive_config.json
trajepl: > Don't we have passes that can be loaded only in specific environment like Azure or OpenVino?

Could you help explain more?

The UT function here I mean is like to verify all of `module_path` in  `olive_config.json` are right to the olive build-in passes. 
In my rough thoughts, we can just read the `olive_config.json` and start a looper to check if the `module_path` is right. 
I think it is quite simple, mostly? 

The UT I mean is just  for our build-in passes. Otherwise, if any typos of the `module_path` are checked into main, we have to start a new PR then wait CI and approval.

docs/source/getstarted/installation.md
trajepl: Previously, we split the PR into two submissions.

olive/olive_config.json
shaahji: Done!

docs/source/getstarted/installation.md
shaahji: Done! Removed docs changes from this PR and will follow up with another ones this is merged.

olive/systems/azureml/aml_pass_runner.py
```diff
@@ -21,6 +21,7 @@
 from olive.passes import FullPassConfig, Pass
 from olive.resource_path import create_resource_path
 from olive.systems.utils import get_common_args
+from olive.workflows.run.config import OliveModuleConfig
```
guotuofeng, line 24: did this import introduce circular import?

guotuofeng: need we move OlivePackageConfig into olive_pass.py or pass_config.py so that the module dependencies between workflow and system could be decoupled?

olive/workflows/run/__main__.py
guotuofeng: it seems the `--config` is renamed to `--run-config`? if this is the case, we need update all readme.md for all examples.

olive/workflows/run/__main__.py
guotuofeng: If we change the `--config`, all the examples' readme need to be updated. Otherwise, better to keep using the original parameter name.

olive/systems/azureml/aml_pass_runner.py
shaahji: I don't think this is an issue. All tests have passed successfully.

olive/workflows/run/__main__.py
shaahji: `--config` was marked deprecated not removed. So using `--config` was still acceptable but `--run-config` was preferred. Regardless, that change has been completed reverted. `--config` stays and the new config parameter is called `--package-config`.

olive/workflows/run/__main__.py
shaahji: `--run-config` is acceptable parameter in place of `--config`. New argument it explicitly named `--package-config`.

olive/systems/azureml/aml_pass_runner.py
guotuofeng: I understand there is no issue from the time being. Not sure some body would change the SystemConfig depends on OlivePackageConfig in future. If not, there will no issues. Otherwise, we need avoid the circular imports.

shaahji: Both _olive_pass.py_ and _pass_config.py_ seems like wrong place for `OlivePackageConfig`. How about creating a new config file specifically for this, say _olive/common/olive_config.py_. I could move the `PassModuleConfig` into _pass_config.py_.

guotuofeng: sure, I am ok with your suggestion.

shaahji: Done!

olive/common/package_config.py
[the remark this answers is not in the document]
shaahji: Done!

shaahji:
@guotuofeng Can you review the change please! Thx

setup.py
```diff
@@ -25,13 +25,13 @@ def get_version(rel_path):
...
     here = os.path.abspath(os.path.dirname(__file__))
     with open(os.path.join(here, rel_path)) as fp:
-        return json.load(fp)
+        return json.load(fp)["extra_dependencies"]
 
 
 # use techniques described at https://packaging.python.org/en/latest/guides/single-sourcing-package-version/
 # Don't use technique 6 since it needs extra dependencies.
 VERSION = get_version("olive/__init__.py")
-EXTRAS = get_extra_deps("olive/extra_dependencies.json")
```
guotuofeng, line 34: it seems this file could be removed?

shaahji: Yes, as part of a follow up PR to avoid docs build from failing. WIP: #1051

olive/package_config.py
```diff
@@ -0,0 +1,36 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
guotuofeng, line 2: the file location seems be different with the PR https://github.com/microsoft/Olive/pull/1051. which PR is the final one?

shaahji: #1051 is meant to be a follow up after this PR merges. I still have to rebase that PR.

guotuofeng:
the PR looks good to me except above comments. I will let @devang-ml for more comments.

devang-ml (approved):

PR #1030, opened 2024-03-21
🦅 Phi2 Fine tune example

trajepl:
## Describe your changes
1. add phi2 fine tune example
2. try to optimize fine tuned model
  - need set `torch_dtype` as `float32` to avoid the inconsistent weights shape(adapters bf16, but base model with fp32)
  - merge adaptors into base model to make dynamo conversion work.


examples/phi2/phi2.py
```diff
@@ -124,55 +132,64 @@ def main(raw_args=None):
         legacy_optimization_setting(template_json)
 
     # add pass flows
-    model_type = str(args.model_type)
-    template_json["pass_flows"] = SUPPORTED_WORKFLOWS[model_type]
-    if args.optimum_optimization:
-        # if args.model_type in ("cpu_int4", "cuda_int4"):
-        #     raise ValueError("Int4 optimization is not supported in phi2 optimum optimization")
+    pass_flows = [[]]
+    if args.finetune_method:
```
guotuofeng, line 136: chech whether it is qlora?

olive/passes/onnx/conversion.py
```diff
@@ -392,6 +392,9 @@ def _convert_model_on_device(
         """Convert a PyTorchModelHandler to an ONNXModelHandler."""
         # load the model
         pytorch_model, model_attributes = self._load_pytorch_model(model, device, torch_dtype)
+        if model.adapter_path:
+            # merge the adapter weights back to base model
+            pytorch_model = pytorch_model.merge_and_unload()
```
guotuofeng, line 397: do we need assert the model is peft?

trajepl: Updated.

examples/phi2/phi2.py
trajepl: Updated. But keep the finetune_method checking, and push the value of `finetune_method` to pass_flows. That is helpful once we introduce other new fine-tune way like Lora.

olive/passes/onnx/conversion.py
```diff
@@ -392,6 +392,9 @@ def _convert_model_on_device(
         """Convert a PyTorchModelHandler to an ONNXModelHandler."""
         # load the model
         pytorch_model, model_attributes = self._load_pytorch_model(model, device, torch_dtype)
+        if is_peft_model(pytorch_model):
```
jambayk, line 395: We don't want this to always be done. The llama qlora example depends on the adapaters not being merged. 

Can you make this an option?

jambayk (changes requested):
Please make the adapter merge optional.

olive/passes/onnx/conversion.py
trajepl: Done.

examples/phi2/README.md
```diff
@@ -35,7 +53,7 @@ Above commands will generate optimized models with given model_type and save the
 Besides, for better generation experience, this example also let use use [Optimum](https://huggingface.co/docs/optimum/v1.2.1/en/onnxruntime/modeling_ort) to generate optimized models.
 Then use can call `model.generate` easily to run inference with the optimized model.
 ```bash
-# optimum optimization
```
xiaoyu-work, line 38: If we don't support it, should we remove this?

examples/phi2/phi2.py
```diff
@@ -124,55 +132,64 @@ def main(raw_args=None):
...
+        pass_flows[0].append(args.finetune_method)
+        template_json["systems"]["local_system"]["config"]["accelerators"][0]["device"] = "gpu"
+        # torch fine tuning does not require execution provider, just set it to CUDAExecutionProvider
```
xiaoyu-work, line 139: I thought ep is not mandatorily required after Mike's PR got merged. Is it still required?

examples/phi2/phi2.py
```diff
@@ -59,9 +59,17 @@ def get_args(raw_args):
     parser.add_argument(
         "--model_type",
```
xiaoyu-work, line 60: What if `model_type` and `finetune_method` are None?

examples/phi2/phi2.py
jambayk: If we don't provide an EP, it would loop over the installed EPs which is not what we want.

xiaoyu-work: @jambayk will "loop over the installed eps" sill be executed even if we are not running any pass needed onnxruntime like lora/snpe/openvino?

jambayk: Yes, we haven't made the changes that would not do this for workflows with no ort targeting passes. 

Mike's PR didn't really change any behavior of the workflows. It only updated the configs to collect the  hardware/ep related options together.

examples/phi2/README.md
trajepl: For the model that did not be fine tuned, optimum works well.

examples/phi2/phi2.py
trajepl: Yes, but the qlora/snpe/openvino pass is EP agnostic, so even with "loop over the installed eps", the pass only run once.

examples/phi2/phi2.py
trajepl: fixed by raising error.

examples/phi2/phi2.py
jambayk: The cache does take care of the rerun but still, the looping behavior and multiple footprint/outputs for workflows with no onnx models is not ideal. That's what will be good to improve.

jambayk (approved):

PR #1033, opened 2024-03-25
🦝 Matmul RTN/HQQ/GPTQ quant configs

trajepl:
## Describe your changes
1. add `weight_only_quant_configs` to let user set WeightOnlyQuantConfig
2. Make OnnxMatMul4Quantizer `_requires_user_script` be True as `GPTQ` requires calibration dataset.
3. Extract the logics of `calibration data loader`
4. Add more UT to cover the new configs


olive/passes/onnx/quantization.py
```diff
@@ -636,16 +685,46 @@ def _run_for_config(
...
+                from onnxruntime.quantization.matmul_4bits_quantizer import (
+                    DefaultWeightOnlyQuantConfig,
+                    HQQWeightOnlyQuantConfig,
+                )
+
+                if config["algorithm"] == "DEFAULT":
+                    weight_only_quant_config_class = DefaultWeightOnlyQuantConfig
+                elif config["algorithm"] == "HQQ":
+                    weight_only_quant_config_class = HQQWeightOnlyQuantConfig
+            if version.parse(OrtVersion) < version.parse("1.18.0") and config["algorithm"] in ("HQQ", "DEFAULT"):
```
jambayk, line 723: I think this is equivalent to `elif config["algorithm"] in ....`?

trajepl: Tue and actually it is suggested by copilot as well.

At that time, I just want to make it clear then change it. :)

Anyway, I will change back and add comment to make it clear.

jambayk: Thanks. To me, the elif condition looks more clear. Otherwise, it seems like it is an unrelated condition. But it is nit

trajepl: Updated.

olive/passes/onnx/quantization.py
```diff
@@ -611,23 +618,65 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
             "algorithm": PassConfigParam(
                 type_=str,
                 default_value=None,
-                description="Available from onnxruntime>=1.17.0 "
-                "Algorithm config for quantization, can be 'DEFAULT', 'HQQ' or 'RTN', 'GPTQ' "
-                "(default DEFAULT when not set). For 4b quantize a model with RTN or GPTQ algorithm. "
-                "Please refer to https://github.com/intel/neural-compressor/blob/master/docs/source/quantization_weight_only.md "  # noqa: E501
+                description="'RTN' and 'GPTQ' are available from onnxruntime>=1.17.0 "
+                "'DEFAULT', 'HQQ' are available from onnxruntime>=1.18.0 (default DEFAULT when not set). "
```
xiaoyu-work, line 622: `default DEFAULT when not set` doesn't apply to our pass right? Cause if algorithm is None, we will skip weight only quant.

olive/passes/onnx/quantization.py
```diff
@@ -611,23 +618,65 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
...
                 "for more details on weight only quantization using Intel® Neural Compressor. "
                 "For HQQ, please refer to onnxruntime for more details: "
                 "https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/matmul_4bits_quantizer.py#L102C1-L126C25",
             ),
+            "weight_only_quant_configs": PassConfigParam(
+                type_=dict,
+                default_value=None,
+                description="""Available from onnxruntime>=1.17.0
+                The config is binding to the algorithm with following map:
+                1. "algorithm" is "DEFAULT", by default, the weight_only_quant_configs is:
```
xiaoyu-work, line 634: I'm kinda confused what is the default value for `algorithm`

olive/passes/onnx/quantization.py
```diff
@@ -611,23 +618,65 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
...
+                    "weight_only_quant_configs": {
+                        "block_size": 128,
+                        "is_symmetric": False,
+                        "accuracy_level": None
+                    }
+                    https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/matmul_4bits_quantizer.py#L129C1-L140C45
```
xiaoyu-work, line 640: If there is any doc link rather than main branch code link? Cause this link will point to wrong line if the ort code changes. Same as below. Or if there is no doc, we can use a permanent branch link rather main branch.

olive/passes/onnx/quantization.py
trajepl: Default value is `None` and the behavior is same with option `DEFAULT`.
Updated the docs. Could you help take a look? :)

olive/passes/onnx/quantization.py
trajepl: Good catch. Updated with commit changes.

olive/passes/onnx/quantization.py
trajepl: For `weight_only_quant_configs`, if it is None, the **default config** of given algorithm will be used.
image

xiaoyu-work: If `weight_only_quant_configs` and `algorithm` are both given, will the Olive ignore `algorithm` but use `weight_only_quant_configs`?

trajepl: Nope. `algorithm` is used to tell which kind of `weight_only_quant_configs` will be used. Otherwise, the quantization would fail.
If both given, and they are matched, it works. If both given, but they are not matched, would fail.

xiaoyu-work (approved):

PR #1041, opened 2024-03-28
♻️ Use system variable to disable olive env for qualcomm sdk

trajepl:
## Describe your changes
1. Previously, we have `use_olive_env` for qnn/snpe passes. That is used to disable the qualcomm environment variables detected by Olive, like SNPE_ROOT, QNN_SDK_ROOT, PATH and etc. It is enabled only for QNN before. In this PR, I tried to enable it both for QNN and SNPE.
2. Also I tried to remove this flag from pass config and enable it with the way of `system variable`. In this way, we can only focus the logics on qualcomm runner and keep qnn/snpe passes and utils clean. Otherwise, we have to add `use_olive_env` as a argument in every function which may call qnn/snpe sdk.
3. Update the mobilenet/inception example config to limited the EP has only one value. Otherwise, all available EPs will be picked and run over and over again.
4. Fix the bug of snpe quantization. `snpe-dlc-quantize` is only available for linux. `snpe-dlc-quant` is available for both windows and linux but it has less options than `snpe-dlc-quantize` like `--enable-htp`
5. Enable e2e test for qnn/snpe for `USE_OLIVE_ENV:["1", "0"]`


examples/test/test_qnn_tooklit.py
```diff
@@ -6,64 +6,88 @@
...
-        python_cmd = str(Path(os.environ["QNN_SDK_ROOT"]) / "olive-pyenv" / "bin" / "python")
-    install_cmd = [
-        python_cmd,
-        str(Path(os.environ["QNN_SDK_ROOT"]) / "bin" / "check-python-dependency"),
-    ]
-    run_subprocess(cmd=" ".join(install_cmd), check=True)
-    retry_func(run_subprocess, kwargs={"cmd": "python download_files.py", "check": True})
-    retry_func(run_subprocess, kwargs={"cmd": "python prepare_config.py --use_raw_qnn_sdk", "check": True})
+        if use_olive_env:
+            os.environ["USE_OLIVE_ENV"] = "1"
```
jambayk, line 48: not part of this PR but it might be nice to provide an option in the local system config to set environment variables like https://github.com/microsoft/Olive/blob/25eec4950f34e7db4d8f75a4b1ba8d6c858c80e5/olive/systems/python_environment/python_environment_system.py#L37 
This way, during the local system initialization, we can update `os.environ`. Also lets the user know they set the variable correctly.

olive/platform_sdk/qualcomm/snpe/tools/dev.py
```diff
@@ -186,10 +187,15 @@ def quantize_dlc(dlc_path: str, input_list: str, config: dict, output_file: str)
...
+    quant_cmd = "snpe-dlc-quantize"
+    if platform.system() == "Windows":
+        # snpe-dlc-quant is the Windows version of the quantizer tool
+        # and it does not support the --enable_htp flag
+        quant_cmd = "snpe-dlc-quant"
+    cmd = f"{quant_cmd} --input_dlc {dlc_path} --input_list {input_list} --output_dlc {output_file}"
     if config["use_enhanced_quantizer"]:
         cmd += " --use_enhanced_quantizer"
-    if config["enable_htp"]:
+    if config["enable_htp"] and platform.system() != "Windows":
```
jambayk, line 198: it might be better if we log a warning on windows if `enable_htp` is true?

examples/test/test_qnn_tooklit.py
trajepl: Yes, good suggestion. 👍
I think that before actually based on the purpose like: how to let user directly set the SNPE_ROOT or other required options in Olive config? User always expects the simplest way without any other efforts. Just config~
Let me add the TODO item.

jambayk (approved):

PR #1052, opened 2024-04-04
Transformer Compression using SliceGPT

shaahji:
## Transformer Compression using SliceGPT

Adding new pass to use SliceGPT compression technique to improve performance and reduce memory footprint.
Updated phi2 example with a new workflow that uses the implemented pass.

Release Note: New pass `SliceGPT` to compress transformer to improve performance and reduce memory footprint.


docs/source/features/passes/pytorch.md
```diff
@@ -183,6 +183,24 @@ This pass only supports Hugging Face transformers PyTorch models. Please refer t
 }
 ```
 
+## SliceGPT
+`SliceGPT` is post-training sparsification scheme that makes transformer networks smaller by applying orthogonal transformations to each transformer layer that leave the model unchanged, and then slicing off the least-significant rows and columns of the weight matrices. This results in speedups and a reduced memory footprint.
```
devang-ml, line 187: "leave the model unchanged" could be confusing and misleading to some people's opinion. Just reduces size of the weight metrices.

examples/phi2/phi2_slicegpt.json
```diff
@@ -0,0 +1,62 @@
+{
```
devang-ml, line 1: Instead of adding new .json file, could you just leverage template .json and update phi2.py to add "slicegpt" pass based on a command line argument?

docs/source/features/passes/pytorch.md
shaahji: Done!

examples/phi2/phi2_slicegpt.json
shaahji: Updated.

docs/source/features/passes/pytorch.md
```diff
@@ -183,6 +183,24 @@ This pass only supports Hugging Face transformers PyTorch models. Please refer t
 }
 ```
 
+## SliceGPT
+`SliceGPT` is post-training sparsification scheme that makes transformer networks smaller by applying orthogonal transformations to each transformer layer that reduces the model size by slicing off the least-significant rows and columns of the weight matrices. This results in speedups and a reduced memory footprint.
+
+Please refer to the original [paper](https://arxiv.org/abs/2401.15024) for more details on the algorithm and expected results for different models, sparsities and datasets.
+
+This pass only supports Hugging Face transformers PyTorch models. Please refer to [SparseGPT](sparsegpt) for more details on the types of transformers models supported.
```
devang-ml, line 191: Why do you need to refer to SparseGPT here? What if the list of supported model differ from each other?

olive/passes/pytorch/slicegpt.py
```diff
@@ -0,0 +1,165 @@
...
+        self, model: PyTorchModelHandler, data_root: str, config: Dict[str, Any], output_model_path: str
+    ) -> PyTorchModelHandler:
+        if sys.version_info < (3, 10):
+            raise ValueError("SliceGPT requires python3.10 or higher")
+
+        from slicegpt import layernorm_fusion, rotate
+        from slicegpt.data_utils import get_dataset, prepare_dataloader
+        from slicegpt.hf_utils import get_model_and_tokenizer
+        from slicegpt.slicing_scheduler import ConstSlicingScheduler
+
```
devang-ml, line 98: Could we move these imports at top since we have lazy loading of passes now in place?

docs/source/features/passes/pytorch.md
shaahji: That should have been SliceGPT .. typo error. Fixing it.

shaahji: Fixed.

olive/passes/pytorch/slicegpt.py
shaahji: I had to revert this change and put the imports back in function local scope.

This pass requires Python >= 3.10 and CI currently uses v3.8 so none of the tests nor the example can be tested on CI. Specific test to check whether or not all the passes can be imported fails with these imports in global/file scope because the requirements file specifically exclude installing SliceGPT for a python version less than the required 3.10. So the tests fails with missing slicegpt import.

devang-ml (dismissed):

examples/phi2/phi2.py
```diff
@@ -142,11 +148,16 @@ def main(raw_args=None):
     pass_flows = [[]]
     if args.finetune_method:
         pass_flows[0].append(args.finetune_method)
-        template_json["systems"]["local_system"]["config"]["accelerators"][0]["device"] = "gpu"
         # torch fine tuning does not require execution provider, just set it to CUDAExecutionProvider
-        template_json["systems"]["local_system"]["config"]["accelerators"][0]["execution_providers"] = [
-            "CUDAExecutionProvider"
-        ]
+        update_accelerator(template_json, "gpu")
+    if args.slicegpt:
```
jambayk, line 153: should we make line 153 and 161 `elif` since 149,153,161 appear to be mutually exclusive? 
@trajepl

examples/phi2/phi2.py
```diff
@@ -159,19 +170,16 @@ def main(raw_args=None):
...
+
         if "cpu" in model_type:
-            # no need to set device for CPU since default it is CPU
-            template_json["systems"]["local_system"]["config"]["accelerators"][0]["execution_providers"] = [
-                "CPUExecutionProvider"
-            ]
-    if args.optimum_optimization or (args.finetune_method and not args.model_type):
+            update_accelerator(template_json, "cpu")
+
+    if args.optimum_optimization or (args.finetune_method and not args.model_type) or args.slicegpt:
```
jambayk, line 178: could be simplified as `args.optimum_optimization or not args.model_type`? @trajepl

examples/phi2/phi2.py
```diff
@@ -112,10 +118,10 @@ def get_output_model_path(footprints):
 
 def main(raw_args=None):
     args = get_args(raw_args)
-    if not args.model_type and not args.finetune_method:
-        raise ValueError("Please specify either model_type or finetune_method")
+    if not args.model_type and not args.finetune_method and not args.slicegpt:
```
jambayk, line 121: I think we should update the condition here to make sure only one of these is provided at one time since they are supposed to be mutually exclusive. 

This way, we can use the mutually exclusive fact to simplify the conditions in the following code. Like I described in my other comments.

jambayk: Acutually never mind. model_type and finetune_method are compatable with one another.

examples/phi2/phi2.py
jambayk: Please ignore. They are not mutually exclusive.

devang-ml (approved):

olive/passes/pytorch/slicegpt.py
```diff
@@ -0,0 +1,165 @@
...
+    See https://arxiv.org/pdf/2401.15024.pdf for more details on the algorithm.
+
+    This pass only supports PyTorchModelHandler with hf_config.
+    """
+
+    @staticmethod
+    def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        return {
+            "calibration_data_config": PassConfigParam(
```
jambayk, line 35: looks like only the data_name is used from this config. Like we discussed offline previously, wouldn't it be simpler to just have string param called `dataset_name` or something? 

The data config is not used at all but having it here makes the config complicated and also makes it look like any data config is supported.

shaahji: I will follow up on this once I have a discussion with Devang. I think a full data_config keeps the option open for future extension but again it can always be changed in future. Let me follow up after discussion.

PR #1058, opened 2024-04-08
Add fast_mode support

xiaoyu-work:
## Describe your changes

_This is only a draft PR._

Add fast_mode support.

For large models, there is a pain point for Olive for now. We constantly save and load model again between pass and pass and between pass and evaluator, which could waste a bunch of time. So we introduce `fast_mode` for Olive workflow.

`fast_mode` will:

- Pass loaded model between pass and pass or between pass and evaluator.
- No cache for intermediate models.
- Save final model.
- Only support local system.
- require latest ort nightly package >= 1.18.0.dev20240405004

Cons:

- Search could be slower than running workflow with cache. Depends on model size and search algorithm.

`fast_mode` should be used for large model + no search for best running time.

### DIFF
--- a/olive/model/config/model_config.py
+++ b/olive/model/config/model_config.py
@@ -2,6 +2,8 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+from typing import Any
+
 from olive.common.config_utils import ConfigBase
 from olive.common.pydantic_v1 import validator
 from olive.model.config.registry import get_model_handler, is_valid_model_type
@@ -88,6 +90,8 @@ class ModelConfig(ConfigBase):
 
     type: str
     config: dict
+    # For enable_fast_mode only. Otherwise, it will be None.
+    loaded_model: Any = None
 
     @validator("type")
     def validate_type(cls, v):
@@ -107,3 +111,7 @@ class ModelConfig(ConfigBase):
     def create_model(self):
         cls = get_model_handler(self.type)
         return cls(**self.config)
+
+    def to_json(self, check_object: bool = False) -> dict:
+        self.loaded_model = None
+        return super().to_json(check_object)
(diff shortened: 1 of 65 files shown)


olive/model/config/model_config.py
```diff
@@ -88,6 +90,8 @@ class ModelConfig(ConfigBase):
 
     type: str
     config: dict
+    # For enable_fast_mode only. Otherwise, it will be None.
+    loaded_model: Any = None
```
xiaoyu-work, line 94: This is a compromise solution. Since we decided to return model config for pass run, we should somehow pass loaded model back to engine. Another option is changing returning model config back to model handler, where I tried in [another branch](https://github.com/microsoft/Olive/blob/xiaoyu/lmim/olive/engine/engine.py#L997). But that will break a bunch of current logics and tests.

olive/model/handler/onnx.py
```diff
@@ -64,9 +69,18 @@ def model_path(self) -> str:
         model_path = super().model_path
         return get_onnx_file_path(model_path, self.onnx_file_name) if model_path else None
 
-    def load_model(self, rank: int = None) -> ModelProto:
+    def load_model(self, rank: int = None, enable_fast_mode: bool = False) -> ModelProto:
```
xiaoyu-work, line 72: We can't add
```
if self.model is not None:
    return self.model
self.model = onnx.load(self.model_path)
return self.model
```
here like how PyTorchModelHandler. Some passes may reuse the same modelhandler for search and we don't want to reuse the same loaded model but need to load model again from new model path. It would be great if someone knows which pass is doing such search and we can see if we can update it.

olive/systems/local.py
```diff
@@ -25,11 +28,17 @@ def run_pass(
         data_root: str,
         output_model_path: str,
         point: Optional[Dict[str, Any]] = None,
+        enable_fast_mode: bool = False,
```
guotuofeng, line 31: from the PR description, it seems the fast mode only enabled in local system. If that's the case, it would be better to move it to the `__init__`, In this way, all other system will not need be changed

olive/model/config/model_config.py
guotuofeng: we could allow system accept OliveModelHandler or ModelConfig as input and output for the fast mode.

PR #1061, opened 2024-04-09
🧞 Fix genai exporter arguments bug and add Genai phi2 example

trajepl:
## Describe your changes
1. With the latest onnxruntime-genai, there are a few bugs:
  a. `precision` can only accept string value(int4, fp16, fp32) but not the 
   image
  b. `execution_provider` accept ep value but not device value
  c. the pass of `GenAIModelExporter` is not EP agnostic. For cuda int4, the input shape is changed to fp16, but for cpu int4, it is still fp32.

2. Add phi2 example.
 - CUDA: v100
image
- CPU:  Intel(R) Xeon(R) CPU E5-2698
image

4. Update genai generation example in README of llama2/phi2.

### DIFF
--- a/olive/passes/onnx/genai_model_exporter.py
+++ b/olive/passes/onnx/genai_model_exporter.py
@@ -40,6 +40,30 @@ class GenAIModelExporter(Pass):
             )
         }
 
+    def validate_search_point(
+        self, search_point: Dict[str, Any], accelerator_spec: AcceleratorSpec, with_fixed_value: bool = False
+    ) -> bool:
+        if with_fixed_value:
+            search_point = self.config_at_search_point(search_point or {})
+        precision = search_point.get("precision")
+        device = (
+            Device.CPU
+            if self.accelerator_spec.execution_provider
+            in AcceleratorLookup.get_execution_providers_for_device(Device.CPU)
+            else Device.GPU
+        )
+        if precision == self.Precision.FP16 and device == Device.CPU:
+            logger.info(
+                "FP16 is not supported on CPU. Valid precision + execution"
+                "provider combinations are: FP32 CPU, FP32 CUDA, FP16 CUDA, INT4 CPU, INT4 CUDA"
+            )
+            return False
+        return True
+
+    @staticmethod
+    def is_accelerator_agnostic(accelerator_spec: AcceleratorSpec) -> bool:
+        return False
+
     def _run_for_config(
         self, model: PyTorchModelHandler, data_root: str, config: Dict[str, Any], output_model_path: str
     ) -> ONNXModelHandler:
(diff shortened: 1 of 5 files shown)


shaahji (approved):

guotuofeng:
>         INT4 = "int4"

I think add `__str__` could fix the `str(Precision.FP32)` issue.

---
Refers to: olive/passes/onnx/genai_model_exporter.py:31 in b5f1e71. [](commit_id = b5f1e710e8bcb024aa1bb32f6e8ba37b2b40e559, deletion_comment = True)

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -54,17 +78,12 @@ def _run_for_config(
         output_model_filepath = Path(resolve_onnx_path(output_model_path))
 
         precision = config["precision"]
-        device = (
-            Device.CPU
+        execution_provider = (
+            "cpu"
             if self.accelerator_spec.execution_provider
```
guotuofeng, line 83: >self.accelerator_spec.execution_provider [](http://example.com/codeflow?start=15&length=40)

host_device or target_device?

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -40,6 +40,30 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
...
+            logger.info(
+                "FP16 is not supported on CPU. Valid precision + execution"
+                "provider combinations are: FP32 CPU, FP32 CUDA, FP16 CUDA, INT4 CPU, INT4 CUDA"
+            )
+            return False
+        return True
+
+    @staticmethod
+    def is_accelerator_agnostic(accelerator_spec: AcceleratorSpec) -> bool:
+        return False
```
guotuofeng, line 65: Basically, hardware dependent means need create InferenceSession and need run on target.

olive/passes/onnx/genai_model_exporter.py
trajepl: Pass is in any device. Model is in target device.

guotuofeng:
> > ```
> >     INT4 = "int4"
> > ```
> 
> I think add `__str__` could fix the `str(Precision.FP32)` issue.
> 
> Refers to: olive/passes/onnx/genai_model_exporter.py:31 in [b5f1e71](https://github.com/microsoft/Olive/commit/b5f1e710e8bcb024aa1bb32f6e8ba37b2b40e559). [](commit_id = [b5f1e71](https://github.com/microsoft/Olive/commit/b5f1e710e8bcb024aa1bb32f6e8ba37b2b40e559), deletion_comment = True)

Please make sure the search point is always Enum. If there are any possibility to use str for precision, we'd better add `__str__`

olive/passes/onnx/genai_model_exporter.py
jambayk: Yes, like Mike says, a search point can be invalid for a particular ep while the pass is accelerator agnostic.

Pass is only non-agnostic if the output of the pass changes based on the accelerator.

jambayk: This PR description #481 has more info on this.

jambayk: Actually, line 81 depends on the accelerator_spec so it satisfies the requirement.

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -79,8 +98,8 @@ def _run_for_config(
             model_name=str(model.model_path or model.hf_config.model_name),
             input_path="",  # empty string for now
             output_dir=str(output_model_filepath.parent),
-            precision=str(precision),
-            execution_provider=str(device),
+            precision=precision.value,
```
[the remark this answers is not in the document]
jambayk: Btw, this is just a nit.

Probably safer to pass as string either using the value of str with __str__ implemented since that's the expected type of the argument.

olive/passes/onnx/genai_model_exporter.py
trajepl: Seems, we need an extra pass config like `target_ep`. Then the `is_accelerator_agnostic` can be True.

olive/passes/onnx/genai_model_exporter.py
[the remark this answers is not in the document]
trajepl: True~

guotuofeng (approved):

PR #1063, opened 2024-04-09
Engine config refactoring

guotuofeng:
## Describe your changes
* simplify the create engine logic
* Cleanup engine config

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -52,7 +52,6 @@ class Engine:
         clean_evaluation_cache=False,
         plot_pareto_frontier=False,
         azureml_client_config=None,
-        **kwargs,
     ):
         self.no_search = False
         # default search strategy
@@ -103,7 +102,6 @@ class Engine:
         self.pass_flows_search_spaces = None
 
         self.footprints = defaultdict(Footprint)
-        assert not kwargs
 
         self._initialized = False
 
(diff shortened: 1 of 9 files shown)


olive/engine/engine.py
```diff
@@ -99,24 +103,24 @@ def __init__(
         self.pass_flows_search_spaces = None
 
         self.footprints = defaultdict(Footprint)
-        self.azureml_client_config = self._config.azureml_client_config
+        assert not kwargs
```
jambayk, line 106: since kwargs is not used and we don't allow them, isn't it equivalent to not having `**kwargs` in the init signature? 

Or is it added for future proofing?

guotuofeng: originally, I add it for check the additional unused parameters. I will remove it since passing redundant parameters and assert here has same result.

olive/engine/engine.py
```diff
@@ -43,52 +43,55 @@ class Engine:
 
     def __init__(
         self,
-        config: Union[Dict[str, Any], EngineConfig] = None,
-        search_strategy: Optional[SearchStrategy] = None,
-        host_config: Optional[SystemConfig] = None,
-        target_config: Optional[SystemConfig] = None,
-        evaluator_config: Optional["OliveEvaluatorConfig"] = None,
+        search_strategy: Optional[Union[Dict[str, Any], SearchStrategy]] = None,
```
jambayk, line 46: Could you add `SearchStrategyConfig` as another option here?

to make it consistent with the other parameters, we could also remove `SearchStrategy` from accepted type and also remove the `else` clause at line 65.

jambayk (approved):
Thanks!

PR #1064, opened 2024-04-09
`ExtractAdapters`: Extract lora adapters and use them as model inputs or external initializers

jambayk:
## Describe your changes
- Introduce new pass `ExtractAdapters` which extracts the lora adapters (float or static quantized) weights and saves them in a separate file. The model graph is also modified in one of the following ways:
  1. Adapters weights are set as external tensors pointing to a non-existent file. The onnx model is thus invalid by itself as it cannot be loaded. In order to create an inference session using this model, the adapter weights must be added to a sessions options object using `add_initializer` or `add_external_initializers`.
  2. Adapter weights are converted into model inputs. The onnx model is valid. During inference, the adapter weights must be provided as part of the inputs. We call them constant inputs here since these weights don't change between runs when using the one set of adapters.
- olive.scripts.export_adapters provided as a standalone script to directly export pre-existing adapter weights into the same format as `ExtractAdapters` pass.
- `OnnxDAG` utils added to perform onnx graph manipulations. The methods are generic and can be used in other new passes.
- `ONNXModelHandler` has two corresponding new attributes that must be names of files in the `model_path` directory:
  1. `external_initializers_file_name`
  2. `constant_inputs_file_name`
- Both `olive.common.ort_inference.get_ort_inference_session` and `OrtInferenceSession` can handle the external initializers and constant inputs.


olive/passes/onnx/conversion.py
```diff
@@ -319,73 +321,77 @@ def _load_pytorch_model(
                 - the onnx model must be quantized using OnnxBnb4Quantization pass after conversion
         Model attributes is None if the output model should inherit the model attributes from the input model.
         """
+        pytorch_model = None
+        model_attributes = deepcopy(model.model_attributes or {})
         if not model.is_model_loaded_from_hf_config() or not model.hf_config.from_pretrained_args:
             # if the model is not loaded from hf config, or the model loading args is not specified,
             # we can load the model directly
-            return model.load_model(), None
```
jambayk, line 325: only changes are: 
- code is indented
- `new_model_attibutes -> model_attributes`
- return is at the end

olive/common/ort_inference.py
```diff
@@ -52,6 +54,24 @@ def get_ort_inference_session(
         from onnxruntime_extensions import get_library_path
 
         sess_options.register_custom_ops_library(get_library_path())
+    if external_initializers_path:
+        from onnxruntime import OrtValue
+
+        # load external initializers
+        external_initializers = np.load(external_initializers_path)
```
guotuofeng, line 61: the load from file seems not be the responsibilities of this function. What about user want to directly pass the initializer object directly instead of using file?

olive/model/handler/onnx.py
```diff
@@ -32,7 +32,13 @@ class ONNXModelHandler(OliveModelHandler, OnnxEpValidateMixin, OnnxGraphMixin):
     the mixin class OnnxGraphMixin is used to support onnx graph operations.
     """
 
-    json_config_keys: Tuple[str, ...] = ("onnx_file_name", "inference_settings", "use_ort_extensions")
+    json_config_keys: Tuple[str, ...] = (
+        "onnx_file_name",
+        "inference_settings",
+        "use_ort_extensions",
+        "external_initializers_name",
```
guotuofeng, line 39: >external_initializers_name [](http://example.com/codeflow?start=9&length=26)

external_initializers_file_name?

olive/passes/onnx/common.py
```diff
@@ -166,15 +169,20 @@ def model_proto_to_olive_model(
...
     else:
         model_path = LocalFile({"path": output_model_path})
         onnx_file_name = None
 
-    olive_model = ONNXModelHandler(model_path=model_path, onnx_file_name=onnx_file_name)
+    olive_model = ONNXModelHandler(
+        model_path=model_path,
+        onnx_file_name=onnx_file_name,
+        external_initializers_name=external_initializers_name,
+        constant_inputs_name=constant_inputs_name,
```
guotuofeng, line 184: is it possible to merge these initializers into the onnx file?

olive/passes/onnx/extract_adapters.py
```diff
@@ -0,0 +1,336 @@
...
+        if not config["make_inputs"]:
+            output_model.model_attributes["external_initializers"] = weights_info
+        else:
+            output_model.model_attributes["constant_inputs"] = weights_info
+            if config["pack_inputs"]:
+                output_model.model_attributes["packed_inputs"] = packings
+        return output_model
+
+    @staticmethod
+    def remove_identity_nodes(dag: OnnxDAG):
```
guotuofeng, line 258: should it be method of OnnxDag?

olive/passes/onnx/extract_adapters.py
```diff
@@ -0,0 +1,336 @@
...
+                re.match(pattern, node_name) for pattern in lora_name_patterns
+            ):
+                # not a lora module
+                continue
+
+            # new name for the weight
+            new_weight_name = self._create_new_weight_name(node_name)
+
+            # original weight name
+            old_weight_name = dag.nodes[node_name].inputs[1]
```
guotuofeng, line 96: dag.nodes.items()?

olive/passes/onnx/extract_adapters.py
```diff
@@ -0,0 +1,336 @@
...
+        if not new_initializer.HasField("raw_data"):
+            new_initializer.raw_data = b""
+        set_external_data(new_initializer, location="dummy-location.bin")
+        # clear the data fields
+        new_initializer.ClearField("raw_data")
+        new_initializer.ClearField("float_data")
+        return new_initializer
+
+    @classmethod
+    def _create_empty_initializer(cls, dag: OnnxDAG, weights: Dict[str, "NDArray"], old_name: str, new_name: str):
```
guotuofeng, line 316: part of OnnxDag class?

olive/passes/onnx/extract_adapters.py
```diff
@@ -0,0 +1,336 @@
...
+        weight_name = old_name[1:] if old_name.startswith("/") else old_name
+        return (
+            weight_name.replace("/", ".")
+            .replace("default.", "lora_A.")
+            .replace("default_1.", "lora_B.")
+            .replace("MatMul", "weight")
+        )
+
+    @staticmethod
+    def _copy_initializer(old_initializer: onnx.TensorProto, new_name: str) -> onnx.TensorProto:
```
guotuofeng, line 296: ditto?

olive/passes/onnx/extract_adapters.py
```diff
@@ -0,0 +1,336 @@
...
+        # dictionary to store adapter weights
+        weights = {}
+        # keep track of float and quantized modules
+        float_modules = set()
+        quant_modules = set()
+
+        # nodes to remove at the end
+        nodes_to_remove = set()
+
+        for node_name in list(dag.nodes.keys()):
```
guotuofeng, line 85: seems the loop only related to DAG and could be moved to the DAG class?

examples/llama2/requirements-qlora.txt
```diff
@@ -1,6 +1,7 @@
 -r requirements.txt
 accelerate
 bitsandbytes
+git+https://github.com/jambayk/optimum.git@jambayk/constant-inputs
```
guotuofeng, line 4: would this branch merge into HF optimum?

olive/passes/onnx/onnx_dag.py
```diff
@@ -0,0 +1,473 @@
...
+        for i in graph.input:
+            ios[i.name] = OnnxIO(proto=i, source=SpecialInput.INPUT, graph_idx=graph_idx)
+        for o in graph.output:
+            ios[o.name] = OnnxIO(proto=o, destination=[SpecialOutput.OUTPUT], graph_idx=graph_idx)
+        for initializer in graph.initializer:
+            ios[initializer.name] = OnnxIO(proto=initializer, source=SpecialInput.INITIALIZER, graph_idx=graph_idx)
+        return ios
+
+    @classmethod
+    def process_node(
```
guotuofeng, line 116: private function?

olive/common/ort_inference.py
jambayk: I was thinking about this too. I will make the caller of this function responsible for loading the weights and pass a dict of numpy arrays to the function.

olive/model/handler/onnx.py
jambayk: sure, will change to `external_initializers_file_name` and `constant_inputs_file_name` to make it clear they are file names.

olive/passes/onnx/common.py
jambayk: It is not possible without manipulating the graph again. The purpose of these files is to have the adapter weights separately so that we can provide different adapter weights to the same base model. 
merging them would make the extraction useless.

olive/passes/onnx/extract_adapters.py
jambayk: Yes, I can make this a method of OnnxDAG since it only deals with the graph and can be reused for other passes too.

olive/passes/onnx/extract_adapters.py
jambayk: the DAG object is meant to be a utils to represent an onnx model as a dag. It will only handle simple logic like adding, removing or replacing nodes. 
For example, the autofusion pass will also use the dag https://github.com/microsoft/Olive/blob/8ff3da1ff0ddc0b276156527712ad7801d3546a1/olive/passes/onnx/auto_fusion.py#L75

The loop here is specific to this pass. It will be responsible for the logic.

olive/passes/onnx/extract_adapters.py
jambayk: I missed updating this. it is supposed to be `dag.get_node_inputs(node_name)[1]`. 

Users of `OnnxDAG` are supposed to use the public methods and not deal with the graph directly. So that they are agnostic to how the graph is represented internally.

olive/passes/onnx/extract_adapters.py
jambayk: This logic is specific to the pass.

olive/passes/onnx/extract_adapters.py
jambayk: specific to the pass.

examples/llama2/requirements-qlora.txt
jambayk: Our end goal is this method be the standard way to use different adapters in onnx models. So, I think we will probably be able to contribute this to optimum once the solution matures.

guotuofeng: if the private changes is not in the optimum, would this feature fail?

jambayk: Without these changes in my branch, we would not be able to use `ORTModelForCausalLM` for generation using this model. 
However, we are in the early stages of lora adapters deployment so the support for generation will be made available publicly (whether it is through optimum, onnxruntime genai library or custom model wrapper).

guotuofeng: thanks for the clarification.

olive/common/ort_inference.py
jambayk: Done.

olive/model/handler/onnx.py
jambayk: done.

olive/passes/onnx/extract_adapters.py
jambayk: done.

olive/passes/onnx/extract_adapters.py
jambayk: updated.

guotuofeng:
the external_initializers_name description need to be updated.

olive/common/ort_inference.py
```diff
@@ -52,6 +56,21 @@ def get_ort_inference_session(
...
+        initializer_values = []
+        for name, value in external_initializers.items():
+            initializer_names.append(name)
+            initializer_values.append(OrtValue.ortvalue_from_numpy(value))
+
+        # add external initializers to the session
+        sess_options.add_external_initializers(initializer_names, initializer_values)
+
+        # delete external initializers to free up memory
+        del external_initializers
```
guotuofeng, line 73: the del should be in callers?

jambayk: Oh, I forgot about this. Will remove it.

jambayk: Done.

olive/model/handler/onnx.py
```diff
@@ -51,19 +60,33 @@ def __init__(
...
         self.io_config = None
         self.graph = None
         self.all_graphs: Optional[List[GraphProto]] = None
 
-        # check for onnx file name since it will do validation
-        _ = self.model_path
+        # check for file names since it will automatically validate the paths
+        to_check = ["model_path", "external_initializers_path", "constant_inputs_path"]
+        for attr in to_check:
+            getattr(self, attr)
```
guotuofeng, line 73: >getattr(self, attr) [](http://example.com/codeflow?start=12&length=19)

when the error will be raised?

olive/passes/onnx/onnx_dag.py
```diff
@@ -0,0 +1,492 @@
...
+class OnnxIO(ConfigBase):
+    """ONNX input/output.
+
+    Behaves similar to labeled edges in a graph but can connect to multiple nodes.
+    """
+
+    source: str = None
+    destination: List[str] = Field(default_factory=list)
+    graph_idx: int
+    proto: Union[ValueInfoProto, TensorProto] = None
```
guotuofeng, line 55: >ValueInfoProto [](http://example.com/codeflow?start=17&length=14)

could the ProtoBuf type as the pydantic type hint?

olive/passes/onnx/onnx_dag.py
```diff
@@ -0,0 +1,492 @@
...
+    """ONNX node."""
+
+    op_type: str
+    inputs: List[str]
+    outputs: List[str]
+    graph_idx: int
+    proto: NodeProto  # reference to the node in the model graph
+
+
+class OnnxIO(ConfigBase):
```
guotuofeng, line 46: >ConfigBase [](http://example.com/codeflow?start=13&length=10)

need the two classes be configured using pydantic?

olive/model/handler/onnx.py
jambayk: Any of the errors from https://github.com/microsoft/Olive/blob/25a0b78949fe295fa0f7faabd30dc328c229f853/olive/model/utils/onnx_utils.py#L30 or https://github.com/microsoft/Olive/blob/25a0b78949fe295fa0f7faabd30dc328c229f853/olive/model/utils/onnx_utils.py#L66

This is to make sure the values are provided correctly. if external data, initializers or constant inputs, they should be files inside `model_path` which should be folder.

olive/passes/onnx/onnx_dag.py
jambayk: ConfigBase behaves like a dataclass. So, I am using it since I like the syntax and it's behaviors.

guotuofeng: could we directly use dataclass?

olive/passes/onnx/onnx_dag.py
jambayk: Yes, it works.

guotuofeng: Please add comments that it doesn't support serialization.

jambayk: Done.

olive/passes/onnx/onnx_dag.py
jambayk: as discussed offline. I will keep it as configbase but added comment about serialization to avoid potential confusion.

guotuofeng:
LGTM, I will let other guys put comments.

olive/passes/onnx/extract_adapters.py
```diff
@@ -0,0 +1,317 @@
...
+        # create a new initializer
+        new_initializer = onnx.TensorProto()
+        # copy the old initializer
+        new_initializer.CopyFrom(old_initializer)
+        # set the new name
+        new_initializer.name = new_name
+        # raw_data is required for set_external_data
+        if not new_initializer.HasField("raw_data"):
+            new_initializer.raw_data = b""
+        set_external_data(new_initializer, location="dummy-location.bin")
```
trajepl, line 290: Will this external data be stored under the path user running olive?
Seems it is `dummy` stuff? Where will we remove it?

olive/passes/onnx/extract_adapters.py
```diff
@@ -0,0 +1,317 @@
...
+    external files. Inference session must be created by first loading the adapter weights using
+    SessionOptions.add_external_initializers.
+
+    If make_inputs is True, the adapter weights are inputs to the model and must be provided during inference.
+    """
+
+    @classmethod
+    def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        return {
+            "make_inputs": PassConfigParam(
```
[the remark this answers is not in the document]
trajepl: Seems the external related config would be enabled always, right?

olive/passes/onnx/extract_adapters.py
jambayk: This is actually just a place holder. Since the raw_data field is cleared, during model save, it doesn't save anything at this location. https://github.com/onnx/onnx/blob/main/onnx/external_data_helper.py#L304

This is why the description of this PR says the onnx model produced by extract as initializers is invalid because it points to non-existent file.
This is okay since we only intend to use this model using ort inference session.

examples/llama2/README.md
```diff
@@ -46,13 +47,76 @@ Supported languages are Python, TypeScript, JavaScript, Ruby, Julia, Rust, C++,
...
+
+The workflow above extracts the lora adapters from the fine-tuned model and converts them into inputs for the model. This way, you can use adapters for the same base model with different tasks.
+Pre-existing adapters can be exported directly using the following command:
+```bash
+# change the adapter_path to the path of the adapter you want to export
+# ensure that the target modules are the same as those of the above fine-tuned model
+python -m olive.scripts.export_adapters --adapter_path Mikael110/llama-2-7b-guanaco-qlora --dtype float16 --pack_weights --output_path models/guanaco_fp16_packed.npz
+```
+
+Snippet below shows an example runs of the generated fine-tuned model using two different adapters.
```
devang-ml, line 60: As a follow up PR, we want to move this snippet into a working Jypter notebook.

olive/passes/onnx/onnx_dag.py
```diff
@@ -0,0 +1,496 @@
+# -------------------------------------------------------------------------
```
devang-ml, line 1: nitpick: I'd make this part of ONNX utils.

olive/scripts/export_adapters.py
```diff
@@ -0,0 +1,100 @@
+import argparse
```
devang-ml, line 1: We need to cover this script in Olive docs.

devang-ml (approved):

PR #1066, opened 2024-04-10
🍵 Use system's temp dir to set genai exporter cache folder

trajepl:
## Describe your changes

Use system's temp dir to set genai exporter cache folder as in windows, the length of cache folder may exceed file path max_length limitation.
https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry

### DIFF
--- a/olive/passes/onnx/genai_model_exporter.py
+++ b/olive/passes/onnx/genai_model_exporter.py
@@ -95,7 +95,7 @@ class GenAIModelExporter(Pass):
             cache_dir = os.environ.get("TRANSFORMERS_CACHE", None)
         if not cache_dir:
             # please do not clean up the cache dir as it may contain model's data
-            os.environ["HF_HOME"] = str(Path(tempfile.gettempdir()) / "hf_cache")
+            cache_dir = str(Path(tempfile.gettempdir()) / "hf_cache")
 
         # currently we only support regular hf models so we can pass the name_or_path directly to model_name
         # could also check if it is a locally saved model and pass the path to input_path but it is not necessary
--- a/olive/passes/onnx/genai_model_exporter.py
+++ b/olive/passes/onnx/genai_model_exporter.py
@@ -6,6 +6,7 @@
 # --------------------------------------------------------------------------
 import logging
 import os
+import tempfile
 from enum import Enum
 from pathlib import Path
 from typing import Any, Dict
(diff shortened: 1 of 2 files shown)


olive/passes/onnx/genai_model_exporter.py
```diff
@@ -93,7 +94,8 @@ def _run_for_config(
         if not cache_dir:
             cache_dir = os.environ.get("TRANSFORMERS_CACHE", None)
         if not cache_dir:
-            cache_dir = output_model_filepath.parent / "genai_cache_dir"
+            # please do not clean up the cache dir as it may contain model's data
+            os.environ["HF_HOME"] = str(Path(tempfile.gettempdir()) / "hf_cache")
```
guotuofeng, line 98: should this be part of responsibility of user code? genai logic should not change the HF_HOME

guotuofeng (changes requested):
HF_HOME should not be changed in genai pass.

olive/passes/onnx/genai_model_exporter.py
trajepl: It should be `cache_dir`, updated.

EmmaNingMS (approved):

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -93,7 +94,8 @@ def _run_for_config(
         if not cache_dir:
             cache_dir = os.environ.get("TRANSFORMERS_CACHE", None)
         if not cache_dir:
-            cache_dir = output_model_filepath.parent / "genai_cache_dir"
+            # please do not clean up the cache dir as it may contain model's data
+            cache_dir = str(Path(tempfile.gettempdir()) / "hf_cache")
```
guotuofeng, line 98: how could user know the cache folder after the pass done since it is a temp dir?

guotuofeng: or need we use the output_model_path as cache_dir?

trajepl: image
I just tried to run model build scripts and find that the `cache_dir` is used to:
1. store the downloaded huggingface model snapshot which plays the same role with huggingface's cache folder.
2. besides, model builder will use this cache dir to generate temporal onnx model weights which will be removed after the conversion.

So after the conversion, the cache folder only has the context of #1. User do not need to know where is the cache dir.

guotuofeng (approved):

PR #1070, opened 2024-04-10
Restrict SliceGPT pass to only supported datasets

shaahji:
## Restrict SliceGPT pass to only supported datasets

SliceGPT supports only a limited number of datasets for calibration purposes. However, the Olive/SliceGPT pass' use of DataConfig to accept that user parameter gave the impression that any/all datasets are legal.

Removed use of DataConfig and instead accept a dataset name (from among the list of supported datasets) to resolve the ambiguity.

### DIFF
--- a/olive/passes/pytorch/slicegpt.py
+++ b/olive/passes/pytorch/slicegpt.py
@@ -6,12 +6,11 @@
 import json
 import logging
 import sys
-from typing import Any, Dict, Union
+from typing import Any, Dict
 
 import torch
 
 from olive.constants import ModelFileFormat
-from olive.data.config import DataConfig
 from olive.hardware.accelerator import AcceleratorSpec
 from olive.model import PyTorchModelHandler
 from olive.model.utils.path_utils import normalize_path_suffix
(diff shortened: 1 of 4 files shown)


olive/passes/pytorch/slicegpt.py
```diff
@@ -32,8 +31,8 @@ class SliceGPT(Pass):
     @staticmethod
     def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
         return {
-            "calibration_data_config": PassConfigParam(
-                type_=Union[DataConfig, Dict],
+            "calibration_dataset": PassConfigParam(
+                type_=str,
                 required=True,
                 description=("Data config for Dataset to calibrate and calculate perplexity on."),
```
jambayk, line 37: Can you add the supported datasets in the description?

devang-ml:
Let's first understand the root cause of this SliceGPT limitation. Is it just because original author didn't try other datasets?

olive/passes/pytorch/slicegpt.py
shaahji: Done!

nailimixaM:
> Let's first understand the root cause of this SliceGPT limitation. Is it just because original author didn't try other datasets?

That's right, we just used a few common datasets but there's no limitation on what dataset can be used.

shaahji:
Not required since the slicegpt implementation should be able to handle other datasets which aren't part of the restricted list. Need to update the logic to make it generic and not depend on the restricted list.

PR #1072, opened 2024-04-11
Make the accelerator EP optional for non-onnx pass

guotuofeng:
## Describe your changes
* Originally, the execution_providers in accelerator spec is required regardless whether it is used or not. Even if user doesn't specify them, the local system will infer and get the available EPs to autofill the EPs. This is painful for passes that doesn't use the EP. 
* In this PR, if there is no pass that belongs to onnx techniques, we will not infer the EP based on the device. 
* On the other hand, if the device is not specified, we will choose cpu by default.

### DIFF
--- a/olive/hardware/accelerator.py
+++ b/olive/hardware/accelerator.py
@@ -76,7 +76,7 @@ class AcceleratorLookup:
 
     @staticmethod
     def get_execution_providers_for_device_by_available_providers(device: Device, available_providers):
-        eps_per_device = [*DEVICE_TO_EXECUTION_PROVIDERS.get(device), "CPUExecutionProvider"]
+        eps_per_device = AcceleratorLookup.get_managed_supported_execution_providers(device)
         return AcceleratorLookup.get_execution_providers(eps_per_device, available_providers)
 
     @staticmethod
@@ -284,9 +284,10 @@ def normalize_accelerators(
         ep_not_supported = []
         for accelerator in system_config.config.accelerators:
             device = Device(accelerator.device.lower())
+            eps_per_device = AcceleratorLookup.get_managed_supported_execution_providers(device)
 
             if system_config.olive_managed_env:
-                available_eps = AcceleratorLookup.get_managed_supported_execution_providers(device)
+                available_eps = eps_per_device
             elif (
                 system_config.type in (SystemType.Local, SystemType.PythonEnvironment, SystemType.IsolatedORT)
                 and not skip_supported_eps_check
@@ -295,7 +296,8 @@ def normalize_accelerators(
                 # target is only used for evaluation
                 available_eps = system_supported_eps
             else:
-                available_eps = accelerator.execution_providers
+                eps = AcceleratorLookup.filter_execution_providers(accelerator.execution_providers, eps_per_device)
+                available_eps = eps or ["CPUExecutionProvider"]
 
             supported_eps = AcceleratorLookup.get_execution_providers_for_device_by_available_providers(
                 device, available_eps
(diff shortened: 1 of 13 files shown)


olive/hardware/accelerator.py
```diff
@@ -168,7 +171,65 @@ def infer_single_device_from_execution_providers(execution_providers: List[str])
         return inferred_devices[0]
 
 
-def normalize_accelerators(system_config: "SystemConfig", skip_supported_eps_check: bool = True) -> "SystemConfig":
+def fill_device(system_config: "SystemConfig"):
```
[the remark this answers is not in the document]
guotuofeng: since there is no way to specify the default device. I think current name is fine.

olive/systems/utils/misc.py
```diff
@@ -83,6 +83,7 @@ def create_managed_system(system_config: "SystemConfig", accelerator: "Accelerat
     elif system_config.type == SystemType.Docker:
         from olive.systems.docker import DockerSystem
 
+        assert accelerator.execution_provider, "Execution provider must be specified for Docker system"
```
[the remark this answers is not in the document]
guotuofeng: this is for managed system.

olive/workflows/run/run.py
```diff
@@ -147,20 +165,28 @@ def run_engine(package_config: OlivePackageConfig, run_config: RunConfig, data_r
 
         AzureMLSystem.olive_config = run_config.to_json()
 
+    auto_optimizer_enabled = (
+        not run_config.passes
+        and run_config.auto_optimizer_config is not None
+        and not run_config.auto_optimizer_config.disable_auto_optimizer
+    )
+    if auto_optimizer_enabled:
+        is_ep_required = True
```
trajepl, line 174: This is ok for me now. But in future, if user want to specify the output model to non-onnx model, we can update this part of logics.

olive/hardware/accelerator.py
```diff
@@ -33,16 +33,19 @@ class AcceleratorSpec:
...
     num_cores: int = None
 
     def __str__(self) -> str:
-        # remove the suffix "ExecutionProvider", len("ExecutionProvider") = 17
-        ep = self.execution_provider[:-17] or self.execution_provider
-        return f"{str(self.accelerator_type).lower()}-{ep.lower()}"
+        if self.execution_provider:
+            # remove the suffix "ExecutionProvider", len("ExecutionProvider") = 17
+            ep = self.execution_provider[:-17]
+            return f"{str(self.accelerator_type).lower()}-{ep.lower()}"
```
trajepl, line 46: Could we append `ep`? Make the output folder be like: cpu_cpuep_xx, gpu_cudaep_xxx

Emma also mentioned this before, when she tried to find the output model, she got confused when saw cpu_cpu_xx, gpu_cpu_xxx.

Appending `ep` may be better?

trajepl:
Could you help add TODO item to the example config? For example, llama-qlora.json
image

We can try to update the example config in a separated PR.

guotuofeng:
> Could you help add TODO item to the example config? For example, llama-qlora.json image
> 
> We can try to update the example config in a separated PR.

json could not add comments

olive/hardware/accelerator.py
guotuofeng: I tried to add "ep" suffix but it seems all the current cache folder structure is impacted. I will stick to existing behavior to avoid feature broken.

trajepl (approved):

olive/hardware/accelerator.py
```diff
@@ -171,199 +171,192 @@ def infer_single_device_from_execution_providers(execution_providers: List[str])
...
-                if filtered_eps:
-                    logger.warning(
-                        "The following execution providers are filtered: %s. "
-                        "Please raise issue in Olive site since it might be a bug. ",
-                        ",".join(filtered_eps),
+class AcceleratorNormalizer:
+    """Fill the device, execution providers and check the compatibility between the device and execution providers."""
+
+    def __init__(
+        self, system_config: "SystemConfig", skip_supported_eps_check: bool = True, is_ep_required: bool = True
```
trajepl, line 178: Is it possible to set `is_ep_required` as False by default? Since in olive/hardware level, it does not know the concept of EP which is a onnxruntime only config.

guotuofeng: make it default true or false doesn't matter since EP is part of hardware when running against ort.

xiaoyu-work:
Can you also remove the EP section from our examples for those which don't need EP?

trajepl (approved):

PR #1073, opened 2024-04-11
onnxruntime-genai based metadata and packaging

shaahji:
## onnxruntime-genai based metadata and packaging

* Added support for packaging models (and additional files) generated by the GenAIModelExporter. Also, updated the pass configuration to include search parameters that are forwarded to the generated genai_config file.
* Added support for carrying "additional files" from one pass to next. These files will end up in the generated models output folder and will be packaged.
* Two new packaging configuration options -
  * include_sample_code
  * inlcude_runtime_packages

### Release Note:
Packaging config include two new options -
* include_sample_code - whether or not to include sample code in zip archive. Defaults to true.
* include_runtime_packages - whether or not to include runtime packages (like onnxruntime) in zip archive. Defaults to true.

### DIFF
--- a/examples/llama2/README.md
+++ b/examples/llama2/README.md
@@ -114,8 +114,10 @@ For using ONNX runtime GenAI to optimize, follow build and installation instruct
 
 Run the following command to execute the workflow:
 ```bash
-python -m olive.workflows.run --config lamma2_genai.json
+python llama2_genai.py [--model_name <>] [--metadata_only]
 ```
+
+To generate metadata only for pre-exported onnx model, use the `--metadata_only` option.
 
 Snippet below shows an example run of generated llama2 model.
 ```python
--- a/examples/llama2/llama2_genai.py
+++ b/examples/llama2/llama2_genai.py
@@ -1,0 +1,60 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+import argparse
+import json
+
+from olive.common.utils import set_tempdir
+from olive.workflows import run as olive_run
+
+
+def get_args(raw_args):
+    parser = argparse.ArgumentParser(description="Llama2 optimization using Generative AI")
+    parser.add_argument(
+        "--model_name",
+        type=str,
+        default="meta-llama/Llama-2-7b-hf",
+        help="Model name, currently only supports llama2 7B/13B",
+    )
+    parser.add_argument(
+        "--metadata-only", action="store_true", required=False, help="Whether to use gpu for optimization."
+    )
+    parser.add_argument("--tempdir", type=str, help="Root directory for tempfile directories and files", required=False)
+
+    return parser.parse_args(raw_args)
+
+
+def main(raw_args=None):
+    args = get_args(raw_args)
+    model_name = args.model_name
+
+    # set tempdir
+    set_tempdir(args.tempdir)
+
+    input_template = "llama2_genai_template.json"
+    with open(input_template) as f:
+        template_json_str = f.read()
+
+    # update model name
+    template_json_str = template_json_str.replace("<model_name_placeholder>", model_name)
+    template_json = json.loads(template_json_str)
+
+    # add pass flows
+    if args.metadata_only:
+        template_json["pass_flows"] = [["conversion", "metadata"]]
+    else:
+        template_json["pass_flows"] = [["exporter", "perf_tuning"]]
+    template_json["engine"]["output_dir"] = f"models/{model_name}"
+
+    # dump config
+    output_template = "llama2_genai.json"
+    with open(output_template, "w") as f:
+        json.dump(template_json, f, indent=4)
+
+    olive_run(template_json)  # pylint: disable=not-callable
+
+
+if __name__ == "__main__":
+    main()
--- a/olive/engine/packaging/packaging_config.py
+++ b/olive/engine/packaging/packaging_config.py
@@ -48,6 +48,8 @@ class PackagingConfig(ConfigBase):
     type: PackagingType = PackagingType.Zipfile
     name: str = "OutputModels"
     config: CommonPackagingConfig = None
+    include_runtime_packages: bool = True
+    include_sample_code: bool = True
 
     @validator("config", pre=True, always=True)
     def _validate_config(cls, v, values):
--- a/olive/model/handler/mixin/hf_config.py
+++ b/olive/model/handler/mixin/hf_config.py
@@ -50,15 +50,7 @@ class HfConfigMixin:
 
         return get_hf_model_config(self._get_model_path_or_name(), **self.hf_config.get_loading_args_from_pretrained())
 
-    def save_hf_model_config(self, output_dir: str, **kwargs):
-        if self.hf_config is None:
-            raise ValueError("HF model_config is not available.")
-        if not Path(output_dir).is_dir():
-            raise ValueError("Expecting a directory as input.")
-
-        save_hf_model_config(self.get_hf_model_config(), output_dir, **kwargs)
-
-    def get_hf_model_generation_config(self):
+    def _get_hf_model_generation_config(self):
         if self.hf_config is None:
             raise ValueError("HF model_config is not available")
 
@@ -66,29 +58,13 @@ class HfConfigMixin:
             self._get_model_path_or_name(), **self.hf_config.get_loading_args_from_pretrained()
         )
 
-    def save_hf_model_generation_config(self, output_dir: str, **kwargs):
-        if self.hf_config is None:
-            raise ValueError("HF model_config is not available.")
-        if not Path(output_dir).is_dir():
-            raise ValueError("Expecting a directory as input.")
-
-        save_hf_model_generation_config(self.get_hf_model_generation_config(), output_dir, **kwargs)
-
-    def get_hf_model_tokenizer(self):
+    def _get_hf_model_tokenizer(self):
         if self.hf_config is None:
             raise ValueError("HF model_config is not available")
 
         return get_hf_model_tokenizer(
             self._get_model_path_or_name(), **self.hf_config.get_loading_args_from_pretrained()
         )
-
-    def save_hf_model_tokenizer(self, output_dir: str, **kwargs):
-        if self.hf_config is None:
-            raise ValueError("HF model_config is not available.")
-        if not Path(output_dir).is_dir():
-            raise ValueError("Expecting a directory as input.")
-
-        save_hf_model_tokenizer(self.get_hf_model_tokenizer(), output_dir, **kwargs)
 
     def save_metadata_for_token_generation(self, output_dir: str, **kwargs):
         if self.hf_config is None:
@@ -97,14 +73,14 @@ class HfConfigMixin:
             raise ValueError("Expecting a directory as input.")
 
         save_hf_model_config(self.get_hf_model_config(), output_dir, **kwargs)
-        save_hf_model_generation_config(self.get_hf_model_generation_config(), output_dir, **kwargs)
-        tokenizer_filepaths = save_hf_model_tokenizer(self.get_hf_model_tokenizer(), output_dir, **kwargs)
+        save_hf_model_generation_config(self._get_hf_model_generation_config(), output_dir, **kwargs)
+        tokenizer_filepaths = save_hf_model_tokenizer(self._get_hf_model_tokenizer(), output_dir, **kwargs)
 
         output_dir = Path(output_dir)
         return [
             str(output_dir / "config.json"),
             str(output_dir / "generation_config.json"),
-            *tokenizer_filepaths,
+            *[fp for fp in tokenizer_filepaths if Path(fp).exists()],
         ]
 
     def get_hf_io_config(self):
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -4,6 +4,7 @@
 # --------------------------------------------------------------------------
 import inspect
 import logging
+import shutil
 from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Any, Callable, ClassVar, Dict, Optional, Tuple, Type, Union, get_args
@@ -12,7 +13,7 @@ from olive.common.config_utils import ConfigBase, ParamCategory, validate_config
 from olive.common.user_module_loader import UserModuleLoader
 from olive.data.config import DataConfig
 from olive.hardware import DEFAULT_CPU_ACCELERATOR, AcceleratorSpec
-from olive.model import CompositeModelHandler, DistributedOnnxModelHandler, OliveModelHandler
+from olive.model import CompositeModelHandler, DistributedOnnxModelHandler, OliveModelHandler, ONNXModelHandler
 from olive.passes.pass_config import (
     PassConfigBase,
     PassConfigParam,
@@ -189,7 +190,10 @@ class Pass(ABC):
             for rank in range(model.num_ranks):
                 input_ranked_model = model.load_model(rank)
                 ranked_output_path = Path(output_model_path).with_suffix("") / model.ranked_model_name(rank)
-                self._run_for_config(input_ranked_model, data_root, config, str(ranked_output_path))
+                output_ranked_model = self._run_for_config(
+                    input_ranked_model, data_root, config, str(ranked_output_path)
+                )
+                Pass._carry_forward_additional_files(input_ranked_model, output_ranked_model)
 
             output_model = DistributedOnnxModelHandler(
                 model_path=str(Path(output_model_path).with_suffix("")),
@@ -211,13 +215,61 @@ class Pass(ABC):
                 )
                 components.append(output_model_component)
                 component_names.append(component_name)
+                Pass._carry_forward_additional_files(component_model, output_model_component)
             output_model = CompositeModelHandler(components, component_names)
         else:
             output_model = self._run_for_config(model, data_root, config, output_model_path)
+            Pass._carry_forward_additional_files(model, output_model)
+
         # assumption: the model attributes from passes, if any, are more important than
         # the input model attributes, we should not update/extend anymore outside of the pass run
         output_model.model_attributes = output_model.model_attributes or model.model_attributes
         return output_model
+
+    @staticmethod
+    def _carry_forward_additional_files(input_model: OliveModelHandler, output_model: OliveModelHandler):
+        # NOTE: Can't use model.model_path because that always gets resolved to a filepath.
+        # We need the directory path here.
+        input_model_path = input_model.get_resource("model_path")
+        if not input_model_path:
+            return
+
+        input_model_path = Path(input_model_path)
+        if not input_model_path.is_dir():
+            return
+
+        input_model_attributes = input_model.model_attributes or {}
+        input_model_additional_files = set(input_model_attributes.get("additional_files", []))
+        if not input_model_additional_files:
+            return
+
+        output_model_path = Path(output_model.get_resource("model_path"))
+        if not output_model_path.is_dir():
+            if isinstance(output_model, ONNXModelHandler):
+                # change the "model_path" resource to the parent directory of the model file
+                output_model.set_resource("model_path", output_model_path.parent)
+                output_model.onnx_file_name = output_model_path.name
+                output_model_path = output_model_path.parent
+            else:
+                logger.warning("Expecting the output model to be in a directory but found a file.")
+                return
+
+        output_model_attributes = output_model.model_attributes or {}
+        output_model_additional_files = set(output_model_attributes.get("additional_files", []))
+
+        for filepath in input_model_additional_files:
+            input_filepath = Path(filepath)
+
+            # Make sure we don't overwrite an existing file in the output's directory.
+            # The follow up pass could have *potentially* generated a file with the same name.
+            output_filepath = output_model_path / input_filepath.name
+            if not output_filepath.exists():
+                # TODO(team): Use symlinks instead of copying the files.
+                output_model_additional_files.add(str(output_filepath))
+                shutil.copy(str(input_filepath), str(output_filepath))
+
+        output_model_attributes["additional_files"] = list(output_model_additional_files)
+        output_model.model_attributes = output_model_attributes
 
     def serialize_config(self, config: Dict[str, Any], check_object: bool = False) -> str:
         """Serialize the configuration."""
--- a/olive/passes/onnx/genai_model_exporter.py
+++ b/olive/passes/onnx/genai_model_exporter.py
@@ -4,12 +4,14 @@
 # --------------------------------------------------------------------------
 # Export a PyTorch model using the onnxruntime-genai package.
 # --------------------------------------------------------------------------
+import copy
+import json
 import logging
 import os
 import tempfile
 from enum import Enum
 from pathlib import Path
-from typing import Any, Dict
+from typing import Any, Dict, Union
 
 from olive.hardware.accelerator import AcceleratorLookup, AcceleratorSpec, Device
 from olive.model import ONNXModelHandler, PyTorchModelHandler
@@ -41,7 +43,16 @@ class GenAIModelExporter(Pass):
                 type_=GenAIModelExporter.Precision,
                 required=True,
                 description="Precision of model.",
-            )
+            ),
+            "metadata_only": PassConfigParam(
+                type_=bool,
+                default_value=False,
+                required=False,
+                description="Whether to export the model or generate required metadata only.",
+            ),
+            "search": PassConfigParam(
+                type_=Dict[str, Any], required=False, description="Search options to use for generate loop."
+            ),
         }
 
     def validate_search_point(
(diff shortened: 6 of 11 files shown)


examples/llama2/llama2_genai.json
```diff
@@ -47,7 +47,53 @@
             }
         }
     },
+    "evaluators": {
+        "common_evaluator": {
```
devang-ml, line 51: Why do you need to add evaluators and enable search_strategy for this example?

olive/engine/packaging/packaging_config.py
```diff
@@ -35,10 +36,15 @@ class AzureMLModelsPackagingConfig(CommonPackagingConfig):
...
+class OrtGenAIPackagingConfig(ZipfilePackagingConfig):
+    pass
+
+
 _type_to_config = {
     PackagingType.Zipfile: ZipfilePackagingConfig,
     PackagingType.AzureMLModels: AzureMLModelsPackagingConfig,
     PackagingType.AzureMLData: AzureMLDataPackagingConfig,
+    PackagingType.OrtGenAI: OrtGenAIPackagingConfig,
 }
```
devang-ml, line 48: Do we need a separate packaging type?
PackagingType.AzureMLModels and  PackagingType.Zipfile should also be used for LLMs that use generate() from onnxruntime-genai repo.

examples/llama2/llama2_genai.json
shaahji: No we don't. Pushed up a bad commit.

olive/engine/packaging/packaging_config.py
shaahji: No we don't. Pushed up a bad commit.

olive/passes/olive_pass.py
```diff
@@ -211,14 +215,52 @@ def run(
...
+        output_model_attributes = output_model.model_attributes or {}
+        output_model_additional_files = set(output_model_attributes.get("additional_files", []))
+
+        for filename in input_model_additional_files:
+            output_model_additional_files.add(filename)
+            output_filepath = output_model_output_path / filename
+
+            if not output_filepath.exists():
+                input_filepath = input_model_output_path / filename
+                shutil.copy(input_filepath, output_filepath)
```
jambayk, line 259: Can you add a TODO here to investigate using symlink here?

shaahji: Use of symlink is a wider issue not specific to this use case only. I think the issue deserves to be on our work item board. Will add a comment here as well.

olive/passes/olive_pass.py
```diff
@@ -211,14 +215,52 @@ def run(
...
+        if not input_model_output_path.is_dir():
+            return
+
+        input_model_attributes = input_model.model_attributes or {}
+        input_model_additional_files = set(input_model_attributes.get("additional_files", []))
+        if not input_model_additional_files:
+            return
+
+        output_model_output_path = Path(output_model.get_resource("model_path"))
+        if not output_model_output_path.is_dir():
```
jambayk, line 247: I think this check might fail if the optimization somehow made an onnx model smaller and it was not saved with external data. meaning model path might be file. 

it might be safer to have a different condition for onnx model like:
```python
if not output_model_output_path.is_dir() and isinstance(output_model, ONNXModelHandler):
    # change the "model_path" resource to the parent directory of the model file
    output_model.set_resource("model_path", output_model_output_path.parent)
    output_model.onnx_file_name = output_model_output_path.name
    output_model_output_path = output_model_output_path.parent
```

olive/passes/olive_pass.py
```diff
@@ -211,14 +215,52 @@ def run(
...
         # assumption: the model attributes from passes, if any, are more important than
         # the input model attributes, we should not update/extend anymore outside of the pass run
         output_model.model_attributes = output_model.model_attributes or model.model_attributes
         return output_model
 
+    @staticmethod
+    def _carry_forward_additional_files(input_model: OliveModelHandler, output_model: OliveModelHandler):
+        # NOTE: Can't use model.model_path because that always gets resolved to a filepath.
+        # We need the directory path here.
+        input_model_output_path = input_model.get_resource("model_path")
```
jambayk, line 233: nit: is there are reason the variable names are `input_model_output_path` and `output_model_output_path`? The `_output` part could be removed?

olive/passes/olive_pass.py
```diff
@@ -211,14 +215,52 @@ def run(
...
             output_model = self._run_for_config(model, data_root, config, output_model_path)
+            Pass._carry_forward_additional_files(model, output_model)
+
         # assumption: the model attributes from passes, if any, are more important than
         # the input model attributes, we should not update/extend anymore outside of the pass run
         output_model.model_attributes = output_model.model_attributes or model.model_attributes
         return output_model
 
+    @staticmethod
+    def _carry_forward_additional_files(input_model: OliveModelHandler, output_model: OliveModelHandler):
```
jambayk, line 230: Please add some unit tests for this method.

olive/passes/olive_pass.py
shaahji: Your solution would pull OnnxModelHandler dependency into Pass. So, had to use the suffix as condition.

olive/passes/olive_pass.py
shaahji: Stripped.

olive/passes/olive_pass.py
shaahji: Done!

olive/passes/olive_pass.py
```diff
@@ -211,14 +215,58 @@ def run(
...
+        input_model_additional_files = set(input_model_attributes.get("additional_files", []))
+        if not input_model_additional_files:
+            return
+
+        output_model_path = Path(output_model.get_resource("model_path"))
+        if not output_model_path.is_dir() and (output_model_path.suffix == ".onnx"):
+            # change the "model_path" resource to the parent directory of the model file
+            output_model.set_resource("model_path", output_model_path.parent)
+            output_model.onnx_file_name = output_model_path.name
+            output_model_path = output_model_path.parent
```
jambayk, line 251: i think an elif to still check for `not output_model_path.is_dir()` is still required? The if else only handles onnx models. 
It could be a value error like before or a warning and return.

olive/passes/olive_pass.py
jambayk: I think that is fine? We already have dependency on other handler classes. 

Checking for `.onnx` suffix is also fine but in theory there could be an onnx model file without the `.onnx` suffix. This probably won't be the case since most of our passes use the `resolve_onnx_path` util.

examples/llama2/llama2_genai.json
```diff
@@ -28,7 +28,11 @@
         "exporter": {
             "type": "GenAIModelExporter",
             "config": {
-                "precision": "int4"
+                "precision": "int4",
```
trajepl, line 31: Do we need a separted pass which set 'metadata_only' as True to prepare the metadata in this example?

shaahji: If we want an example to showcase this feature then yes. I can modify the example and update the script accordingly.

olive/passes/olive_pass.py
shaahji: A base class depending on a subclass is architecturally wrong in a design. It's bad software practice. I would recommend basing the decision on the extension for now unless you strongly feel that "there could potentially be onnx models with a different extension and this check would fail".

jambayk: I don't understand why it is "architecturally wrong" for a base class of one class (Pass) depending on the subclass of another class (ModelHandler). 
Can you point me to some resources where this practice is not recommended?

shaahji: Never mind, I still had the old pass hierarchy in mind where OnnxXXXX pass where direct subclass of Pass. We changed that by making Pass and ModelHandler independent. Pass and OnnxModelHandler are no more in the same class hierarchy. My bad.

I will change that check to refer to the handler directly.

examples/llama2/llama2_genai.json
shaahji: Done!

olive/passes/olive_pass.py
shaahji: Done!

olive/passes/olive_pass.py
shaahji: Done!

olive/passes/onnx/conversion.py
```diff
@@ -113,6 +113,36 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
 
     def _run_for_config(
         self, model: PyTorchModelHandler, data_root: str, config: Dict[str, Any], output_model_path: str
+    ) -> Union[CompositeModelHandler, DistributedOnnxModelHandler, ONNXModelHandler]:
+        output_model = self._run_for_config_internal(model, data_root, config, output_model_path)
+
+        if isinstance(output_model, CompositeModelHandler):
```
guotuofeng, line 119: it seems this if/else check against different model type is already in pass.run? could we leverage that check and only add the save_* here for onnx model?

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -98,16 +121,58 @@ def _run_for_config(
...
+            "special_tokens_map.json",
+            "tokenizer.json",
+            "tokenizer.model",
+            "tokenizer_config.json",
+        ]
+        expected_filepaths = [output_model_filepath.parent / fn for fn in expected_filenames]
+
+        model_attributes = copy.deepcopy(model.model_attributes or {})
+        model_attributes["additional_files"] = additional_files = model_attributes.get("additional_files", [])
+        additional_files.extend(fp for fp in expected_filepaths if fp.exists())
```
guotuofeng, line 166: >additional_files [](http://example.com/codeflow?start=8&length=16)

not used?

olive/passes/olive_pass.py
```diff
@@ -211,14 +215,61 @@ def run(
...
+
+        output_model_path = Path(output_model.get_resource("model_path"))
+        if not output_model_path.is_dir():
+            if isinstance(output_model, ONNXModelHandler):
+                # change the "model_path" resource to the parent directory of the model file
+                output_model.set_resource("model_path", output_model_path.parent)
+                output_model.onnx_file_name = output_model_path.name
+                output_model_path = output_model_path.parent
+            else:
+                raise RuntimeError("Expecting the output model to be in a directory but found a file.")
```
jambayk, line 254: Can you make this a log warning for now and return? since other model handlers are out of scope now, there might unexpected failures in case we ran some workflow that went from say an onnx model handler to snpe model handler which is always a file.

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -98,16 +122,56 @@ def _run_for_config(
             # snapshots that can be reused for future runs
             cache_dir = str(Path(tempfile.gettempdir()) / "hf_cache")
 
-        # currently we only support regular hf models so we can pass the name_or_path directly to model_name
-        # could also check if it is a locally saved model and pass the path to input_path but it is not necessary
+        extra_args = {"filename": model.onnx_file_name}
```
jambayk, line 125: I think the filename should be `str(output_model_filepath.name)` like before for `metadata_only=False`?

jambayk: actually after line 102, just using str(output_model_file.name) would work for both.

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -98,16 +122,56 @@ def _run_for_config(
...
+        genai_config_filepath = str(output_model_filepath.parent / "genai_config.json")
+        with open(genai_config_filepath) as istrm:
+            genai_config = json.load(istrm)
+
+        genai_config["search"] = {**genai_config.get("search", {}), **config.get("search", {})}
+
+        with open(genai_config_filepath, "w") as ostrm:
+            json.dump(genai_config, ostrm, indent=4)
+
+        expected_filenames = [
```
jambayk, line 154: would it be possible to iterdir through the model dir and get the names? We can ignore `filename" and "{filename}.data" since they are part of the model itself https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/builder.py#L296 

but the rest would be part of the extra files.

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -69,25 +81,37 @@ def is_accelerator_agnostic(accelerator_spec: AcceleratorSpec) -> bool:
...
+                "GenAIModelExporter pass only supports exporting HF models i.e. PyTorchModelHandler "
+                "with hf_config and exporting the metadata only for pre-optimized onnx models."
             )
 
         Path(output_model_path).mkdir(parents=True, exist_ok=True)
-        output_model_filepath = Path(resolve_onnx_path(output_model_path))
+        output_model_filepath = (
+            Path(resolve_onnx_path(output_model_path))
+            if isinstance(model, PyTorchModelHandler)
+            else Path(resolve_onnx_path(output_model_path, model.onnx_file_name))
```
jambayk, line 105: I am a bit unsure about the else part. 
With this output_model_filepath, the metadata will be saved in a fresh directory. but the actual `model.onnx` and `model.onnx.data` files are in the model.get_resource("model_path") directory. 

since the input model is copied in this case, during packages the metadata would be missing since they are not in the model_path folder.

jambayk: I think our options are
- copy over preexisting files into the new `output_model_path` and create a new onnx model handler with this path
- generate the metadata directly into the input model's directory and return a copied handler with the additional file attributes.

olive/passes/onnx/conversion.py
shaahji: Branching in Pass.run is conditioned on input model and here its conditioned on output model.

olive/passes/olive_pass.py
shaahji: Done!

olive/passes/onnx/genai_model_exporter.py
shaahji: Done!

olive/passes/onnx/genai_model_exporter.py
shaahji: Done!

examples/llama2/llama2_genai.py
```diff
@@ -0,0 +1,60 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+import argparse
+import json
+
+import olive.workflows.run as olive_run
```
jambayk, line 9: can you import this as `from olive.workflows import run as olive_run` to avoid the codeql error?

shaahji: Done!

olive/passes/onnx/genai_model_exporter.py
jambayk: I don't think it does? 
Since the genai_config.json is created inside a different folder than the "model_path" resource of the output_model, I think the packaged model will be missing this file.

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -69,25 +81,37 @@ def is_accelerator_agnostic(accelerator_spec: AcceleratorSpec) -> bool:
...
+        data_root: str,
+        config: Dict[str, Any],
+        output_model_path: str,
     ) -> ONNXModelHandler:
         from onnxruntime_genai.models.builder import create_model
 
-        if not model.hf_config:
+        precision = config["precision"]
+        metadata_only = config["metadata_only"]
+
```
jambayk, line 94: could you add a check here that the handler type is pytorch when metadata_only=false and onnx when true? 

Otherwise, there are potential missing attribute errors like the second condition at line 95.

olive/passes/onnx/genai_model_exporter.py
shaahji: I verified both paths (Pytorch model as input + GenAIExporter and OnnxConversion + GenAIExporter with metadata). Both succeeded with all the necessary files in the packaged zip file.

olive/passes/olive_pass.py
```diff
@@ -211,14 +215,62 @@ def run(
...
+        output_model_additional_files = set(output_model_attributes.get("additional_files", []))
+
+        for filepath in input_model_additional_files:
+            input_filepath = Path(filepath)
+
+            # Make sure we don't overwrite an existing file in the output's directory.
+            # The follow up pass could have *potentially* generated a file with the same name.
+            output_filepath = output_model_path / input_filepath.name
+            if not output_filepath.exists():
+                # TODO(team): Use symlinks instead of copying the files.
```
guotuofeng, line 267: The shutil.copy default is follow_symlinks=True, do we still need the TODO?https://docs.python.org/3/library/shutil.html#shutil.copy

olive/passes/onnx/genai_model_exporter.py
shaahji: Attributes have defaults so shouldn't they still be valid in either context?

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -98,16 +121,49 @@ def _run_for_config(
...
+            **extra_args,
         )
 
-        return ONNXModelHandler(output_model_filepath.parent, onnx_file_name=output_model_filepath.name)
+        # Override default search options with ones from user config
+        genai_config_filepath = str(output_model_filepath.parent / "genai_config.json")
+        with open(genai_config_filepath) as istrm:
+            genai_config = json.load(istrm)
+
+        genai_config["search"] = {**genai_config.get("search", {}), **config.get("search", {})}
```
guotuofeng, line 148: if both genai_config and config have search with same key, which one will take precedence?

olive/passes/olive_pass.py
shaahji: At the moment, all additional files are pretty small. But that may not always be the case. If the files are generated in the first pass, the files will be carried forward to every pass that follows. The intention is to avoid this copying/cloning and use symlinks in each pass's output folder.

Whether copy follows symlinks or not doesn't make a difference here. Follow Symlinks in the context of copy is to say for a path /a/b/c/d, if b were to be a symlink, copy operation will still end up locating d correctly.

Not the same issue so the TODO stays.

olive/passes/onnx/genai_model_exporter.py
shaahji: Values from config will override because of the order.

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -98,16 +121,49 @@ def _run_for_config(
             # snapshots that can be reused for future runs
             cache_dir = str(Path(tempfile.gettempdir()) / "hf_cache")
 
-        # currently we only support regular hf models so we can pass the name_or_path directly to model_name
-        # could also check if it is a locally saved model and pass the path to input_path but it is not necessary
+        extra_args = {"filename": str(output_model_filepath.name)}
+        if metadata_only:
+            extra_args["config_only"] = True
+            model_path = None
+            input_path = str(model.get_resource("model_path"))
```
guotuofeng, line 128: what's the input path? from https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/builder.py#L1166, it seems the input path only accept *.gguf file?

olive/model/handler/mixin/hf_config.py
```diff
@@ -82,14 +66,6 @@ def get_hf_model_tokenizer(self):
...
 
-    def save_hf_model_tokenizer(self, output_dir: str, **kwargs):
-        if self.hf_config is None:
-            raise ValueError("HF model_config is not available.")
-        if not Path(output_dir).is_dir():
-            raise ValueError("Expecting a directory as input.")
-
-        save_hf_model_tokenizer(self.get_hf_model_tokenizer(), output_dir, **kwargs)
-
     def save_metadata_for_token_generation(self, output_dir: str, **kwargs):
```
guotuofeng, line 69: it seems only this save_metadata_for_token_generation need to be public and all other new added files can be private. In the mean  time, please remember update the document section in line 32

shaahji: Those are all generic get functions (get_hf_config, get_hf_generation_config, etc) with no internal bookkeeping to maintain. I don't see a point in making those private.  I did remove the save_xxx functions since they weren't used anymore.

olive/passes/onnx/genai_model_exporter.py
guotuofeng: it seems the else clause doesn't use the input_path at all. did you test it?

image

olive/model/handler/mixin/hf_config.py
guotuofeng: From the time being, all method in Olive model mixins doesn't start with underscore will be public. You can see https://github.com/microsoft/Olive/blob/main/olive/model/handler/mixin/hf_config.py#L162

olive/passes/onnx/genai_model_exporter.py
shaahji: Yes, I verified both modes - model builder and meta data. In examples/llama2 folder -

```bash
> python llama2_genai.py
> python llama2_genai.py --metadata_only
```

Output from first command -
```
Archive:  models/meta-llama/Llama-2-7b-hf/OutputModel.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2024-04-18 23:33   CandidateModels/
     3085  2024-04-18 23:33   models_rank.json
        0  2024-04-18 23:33   CandidateModels/gpu-cuda/
        0  2024-04-18 23:33   CandidateModels/gpu-cuda/BestCandidateModel_1/
     1458  2024-04-18 23:33   CandidateModels/gpu-cuda/BestCandidateModel_1/configurations.json
3980206080  2024-04-18 23:32   CandidateModels/gpu-cuda/BestCandidateModel_1/model.onnx.data
     5001  2024-04-18 23:33   CandidateModels/gpu-cuda/BestCandidateModel_1/model_info.json
   273942  2024-04-18 23:32   CandidateModels/gpu-cuda/BestCandidateModel_1/model.onnx
      348  2024-04-18 23:33   CandidateModels/gpu-cuda/BestCandidateModel_1/inference_config.json
  1842767  2024-04-18 23:32   CandidateModels/gpu-cuda/BestCandidateModel_1/tokenizer.json
     1601  2024-04-18 23:32   CandidateModels/gpu-cuda/BestCandidateModel_1/genai_config.json
      414  2024-04-18 23:32   CandidateModels/gpu-cuda/BestCandidateModel_1/special_tokens_map.json
      918  2024-04-18 23:32   CandidateModels/gpu-cuda/BestCandidateModel_1/tokenizer_config.json
   499723  2024-04-18 23:32   CandidateModels/gpu-cuda/BestCandidateModel_1/tokenizer.model
---------                     -------
3982835337                     14 files
```

Output from second command -
```
Archive:  models/meta-llama/Llama-2-7b-hf/OutputModel.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        0  2024-04-19 00:23   CandidateModels/
     3201  2024-04-19 00:23   models_rank.json
        0  2024-04-19 00:23   CandidateModels/gpu-cuda/
        0  2024-04-19 00:23   CandidateModels/gpu-cuda/BestCandidateModel_1/
      895  2024-04-19 00:23   CandidateModels/gpu-cuda/BestCandidateModel_1/configurations.json
      684  2024-04-19 00:22   CandidateModels/gpu-cuda/BestCandidateModel_1/config.json
26953662464  2024-04-19 00:22   CandidateModels/gpu-cuda/BestCandidateModel_1/model.onnx.data
     4903  2024-04-19 00:23   CandidateModels/gpu-cuda/BestCandidateModel_1/model_info.json
135161487  2024-04-19 00:22   CandidateModels/gpu-cuda/BestCandidateModel_1/model.onnx
        2  2024-04-19 00:23   CandidateModels/gpu-cuda/BestCandidateModel_1/inference_config.json
  1842767  2024-04-19 00:22   CandidateModels/gpu-cuda/BestCandidateModel_1/tokenizer.json
     1601  2024-04-19 00:23   CandidateModels/gpu-cuda/BestCandidateModel_1/genai_config.json
      183  2024-04-19 00:22   CandidateModels/gpu-cuda/BestCandidateModel_1/generation_config.json
      414  2024-04-19 00:22   CandidateModels/gpu-cuda/BestCandidateModel_1/special_tokens_map.json
      918  2024-04-19 00:22   CandidateModels/gpu-cuda/BestCandidateModel_1/tokenizer_config.json
   499723  2024-04-19 00:22   CandidateModels/gpu-cuda/BestCandidateModel_1/tokenizer.model
---------                     -------
27091179242                     16 files
```

olive/passes/onnx/genai_model_exporter.py
jambayk: The current changes to cache.py were not present when I made this comment (maybe got lost in the rebase and force push).

With the changes it seems okay now. But I think we will probably have issues in the future if we want to use the generative api directly on the cached model due to the model.onnx and genai_config.json files being in different folders in cache (you could verify this by checking your Olive cache)

But that can be fixed later. 

I think we have to make some decisions on whether the extra file paths should be members of the model_path directory or they can be under other directories. But that can wait after this PR.

olive/passes/onnx/genai_model_exporter.py
jambayk: .hf_config is only an attribute for pytorch model handler. It will raise attribute error if you called it on an onnx handler

olive/passes/onnx/genai_model_exporter.py
jambayk: The else clause actually uses the input_path indirectly through self.model_name_or_path https://github.com/microsoft/onnxruntime-genai/blob/main/src/python/py/models/builder.py#L1695

olive/passes/onnx/genai_model_exporter.py
shaahji: Yes, evaluation from cache folder will be an issue. In my opinion, the right solution here is to use symlinks and each pass should have ALL its output under one root folder. Passes like the ORT perf tuning generate no output of its own but provides the previous pass's output as the output folder. That's kind of hack. If we had symlinks then the output folder can still be its own with symlinks into the previous passes output folder.

Sure deserves a revisit.

olive/passes/onnx/genai_model_exporter.py
```diff
@@ -98,16 +121,49 @@ def _run_for_config(
...
+        genai_config_filepath = str(output_model_filepath.parent / "genai_config.json")
+        with open(genai_config_filepath) as istrm:
+            genai_config = json.load(istrm)
+
+        genai_config["search"] = {**genai_config.get("search", {}), **config.get("search", {})}
+
+        with open(genai_config_filepath, "w") as ostrm:
+            json.dump(genai_config, ostrm, indent=4)
+
+        filepaths_to_ignore = {str(output_model_filepath), str(output_model_filepath.name) + ".data"}
```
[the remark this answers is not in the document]
shaahji: Fixed.

olive/model/handler/mixin/hf_config.py
shaahji: Done!

olive/passes/onnx/genai_model_exporter.py
shaahji: Done!

examples/llama2/README.md
```diff
@@ -114,9 +114,11 @@ For using ONNX runtime GenAI to optimize, follow build and installation instruct
 
 Run the following command to execute the workflow:
 ```bash
-python -m olive.workflows.run --config lamma2_genai.json
+python llama2_genai.py [--model_name <>] [--metadata_only]
 ```
 
+To generate metadata only for pre-exported onnx model, use the `--metadata_only` option.
```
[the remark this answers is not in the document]
shaahji: Yes!

jambayk (approved):
The changes look good to me (with the note that we will revisit the folder structure later). Please feel free to merge if Mike has no other comments.

PR #1076, opened 2024-04-11
Update `skip_supported_eps_check`, Fix Mobilenet QNN EP example

jambayk:
## Describe your changes
- Added doc and comment for `skip_supported_eps_check`.
- The value provided to `skip_supported_eps_check` is updated so that we also check that there are no passes that need to run on target.
- Mobilenet QNN EP example is fixed so that it only runs for qnn ep. #998 had made the example incorrect when running with no evaluation. Since no target was provided, it would run for all installed eps but it is only supposed to run for qnn ep.
- Added test for the example with no evaluation. Evaluation requires ARM64 Windows which we don't have in our agent pools.

### DIFF
--- a/examples/mobilenet/mobilenet_qnn_ep.py
+++ b/examples/mobilenet/mobilenet_qnn_ep.py
@@ -61,7 +61,7 @@ def main(raw_args=None):
         del config["systems"]["local_system"]
     else:
         # set target to local_system
-        # it is QNN EP as the target EP even though local environment has
+        # QNN EP is the target EP even though local environment has
         # cpu package installed
         # this is okay since there is no evaluation done in this case
         config["engine"]["target"] = "local_system"
--- a/examples/mobilenet/mobilenet_qnn_ep.py
+++ b/examples/mobilenet/mobilenet_qnn_ep.py
@@ -51,18 +51,24 @@ def main(raw_args=None):
                 "<evaluator>": "common_evaluator",
             }
         )
-        # Set the environment variables
+        # replace template variables with actual values
         config_str = json.dumps(config)
         for name, value in template_args.items():
             config_str = config_str.replace(name, value)
         config = json.loads(config_str)
         config_name += "_eval"
+        # delete unnecessary fields
+        del config["systems"]["local_system"]
     else:
+        # set target to local_system
+        # it is QNN EP as the target EP even though local environment has
+        # cpu package installed
+        # this is okay since there is no evaluation done in this case
+        config["engine"]["target"] = "local_system"
         # delete unnecessary fields
         del (
-            config["systems"],
+            config["systems"]["qnn_ep_env"],
             config["evaluators"],
-            config["engine"]["target"],
             config["engine"]["evaluator"],
             config["engine"]["evaluate_input_model"],
         )
@@ -72,10 +78,11 @@ def main(raw_args=None):
         with open(f"{config_name}.json", "w") as f:
             json.dump(config, f, indent=4)
         print(f"Config file {config_name}.json is generated")  # noqa: T201
+        return config
     else:
         if not args.skip_data_download:
             download_files()
-        olive_run(config)  # pylint: disable=not-callable
+        return olive_run(config)  # pylint: disable=not-callable
 
 
 if __name__ == "__main__":
--- a/examples/mobilenet/mobilenet_qnn_ep_template.json
+++ b/examples/mobilenet/mobilenet_qnn_ep_template.json
@@ -6,13 +6,24 @@
         }
     },
     "systems": {
+        "local_system": {
+            "type": "LocalSystem",
+            "config": {
+                "accelerators": [
+                    {
+                        "execution_providers": [
+                            "QNNExecutionProvider"
+                        ]
+                    }
+                ]
+            }
+        },
         "qnn_ep_env": {
             "type": "IsolatedORT",
             "config": {
                 "python_environment_path": "<qnn_env_path>",
                 "accelerators": [
                     {
-                        "device": "npu",
                         "execution_providers": [
                             "QNNExecutionProvider"
                         ]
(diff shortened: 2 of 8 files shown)


examples/mobilenet/mobilenet_qnn_ep_template.json
```diff
@@ -6,13 +6,24 @@
...
+                ]
+            }
+        },
         "qnn_ep_env": {
             "type": "IsolatedORT",
             "config": {
                 "python_environment_path": "<qnn_env_path>",
                 "accelerators": [
                     {
-                        "device": "npu",
```
jambayk, line 15: Removed since it is automatically inferred.

examples/mobilenet/mobilenet_qnn_ep_template.json
```diff
@@ -6,13 +6,24 @@
         }
     },
     "systems": {
+        "local_system": {
+            "type": "LocalSystem",
+            "config": {
+                "accelerators": [
+                    {
+                        "execution_providers": [
+                            "QNNExecutionProvider"
```
guotuofeng, line 15: it seems the qnn ep is installed in both local and python system. what would happens if we change skip_supported_eps_check=False or remove this flag?

guotuofeng: I suppose the flag is only used if the target ep is not installed with the related ORT?

olive/hardware/accelerator.py
```diff
@@ -263,10 +267,15 @@ def normalize_accelerators(system_config: "SystemConfig", skip_supported_eps_che
             system_config.type in (SystemType.Local, SystemType.PythonEnvironment, SystemType.IsolatedORT)
             and not skip_supported_eps_check
         ):
-            # don't need to check the supported execution providers if there is no evaluation
-            # target is only used for evaluation
+            # skip_supported_eps_check is True here
+            # target is used so we need to check that the system supported eps are compatible with the accelerators
```
guotuofeng, line 271: for qnn scenario, better to mention that the local system is cpu ORT but we specify qnn ep like we discussed offline. 

From the current change, I have no idea that the local system use cpu ort.

examples/mobilenet/mobilenet_qnn_ep.py
```diff
@@ -51,18 +51,24 @@ def main(raw_args=None):
...
+        # replace template variables with actual values
         config_str = json.dumps(config)
         for name, value in template_args.items():
             config_str = config_str.replace(name, value)
         config = json.loads(config_str)
         config_name += "_eval"
+        # delete unnecessary fields
+        del config["systems"]["local_system"]
     else:
+        # set target to local_system
```
guotuofeng, line 63: please copy this comments into the create_accelerators as well.

guotuofeng (dismissed):

olive/workflows/run/run.py
```diff
@@ -147,12 +147,24 @@ def run_engine(package_config: OlivePackageConfig, run_config: RunConfig, data_r
 
         AzureMLSystem.olive_config = run_config.to_json()
 
-    no_evaluation = (
+    # Register passes since we need to know whether they need to run on target
+    for pass_config in (run_config.passes or {}).values():
+        logger.debug("Registering pass %s", pass_config.type)
+        package_config.import_pass_module(pass_config.type)
+
+    target_not_used = (
```
guotuofeng, line 155: do we need only import passes if user specify pass_flows like https://github.com/microsoft/Olive/pull/1072/files#diff-e2990fcd6e151efbd0dd8ccd6817c2cdcdca63a2aea18404d8368df6a93a178aR129?

jambayk: In theory we only need to import passes that are being used by the workflows. 

So if no pass_flow give, it uses all passes. If pass_flows give, it uses the union of the flows.

jambayk: updated.

olive/workflows/run/run.py
```diff
@@ -147,12 +147,26 @@ def run_engine(package_config: OlivePackageConfig, run_config: RunConfig, data_r
 
         AzureMLSystem.olive_config = run_config.to_json()
 
-    no_evaluation = (
+    # Register passes since we need to know whether they need to run on target
+    for pass_config in get_used_passes(run_config):
```
guotuofeng, line 151: https://stackoverflow.com/questions/1271320/resetting-generator-object-in-python

Instead of invoking multiple times of the get_used_passes, we can
1. change it to list so that all subsequent call to get_used_passes will be removed.
2. or use tee like the link said.

jambayk: used list. thanks!

olive/workflows/run/run.py
```diff
@@ -147,12 +147,26 @@ def run_engine(package_config: OlivePackageConfig, run_config: RunConfig, data_r
...
+        logger.debug("Registering pass %s", pass_config.type)
+        package_config.import_pass_module(pass_config.type)
+
+    # check if target is not used
+    target_not_used = (
+        # no evaluator given (also implies no search)
```
guotuofeng, line 158:
```python
def get_used_passes(run_config: RunConfig) -> Generator["RunPassConfig", None, None]:
    if run_config.pass_flows:
        passes = set()
        for pass_flow in run_config.pass_flows:
            for pass_name in pass_flow:
                if run_config.passes[pass_name].type not in passes:
                    passes.add(run_config.passes[pass_name].type)
                    yield run_config.passes[pass_name]
    elif run_config.passes:
        yield from run_config.passes.values()


def is_target_not_used(run_config: RunConfig) -> bool:
    passes = list(get_used_passes(run_config))

    target_not_used = (
        # no evaluator given (also implies no search)
        run_config.engine.evaluator is None
        # no pass specific evaluator
        and all(
            pass_config.evaluator is None and Pass.registry[pass_config.type.lower()].run_on_target is False
            for pass_config in passes
        )
    )
    return target_not_used

```

jambayk: Already pushed similar changes.

guotuofeng (approved):

PR #1078, opened 2024-04-12
fix partial codel scan issues

guotuofeng:

### DIFF
--- a/olive/strategy/search_space.py
+++ b/olive/strategy/search_space.py
@@ -72,11 +72,14 @@ class SearchSpace:
             options = param.get_support(parent_vals)
         elif isinstance(param, Categorical):
             options = param.get_support()
-        for option in options:
-            if option == SpecialParamValue.INVALID:
-                continue
-            search_point[space_name][param_name] = option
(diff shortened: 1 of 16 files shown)


olive/strategy/search_space.py
```diff
@@ -72,11 +72,14 @@ def _iterate_util(
             options = param.get_support(parent_vals)
         elif isinstance(param, Categorical):
             options = param.get_support()
-        for option in options:
-            if option == SpecialParamValue.INVALID:
-                continue
-            search_point[space_name][param_name] = option
-            yield from self._iterate_util(full_iter_order, search_point, index + 1)
+        else:
```
jambayk, line 75: raise ValueError is better here since we only have Categorical and  Conditional as the valid types.

jambayk: just a nit since the else clause should never be triggered in real life scenarios.

guotuofeng: change to return. In this case, the generator return [].

trajepl (approved):

PR #1080, opened 2024-04-15
🅰️ Auto Awq in Olive with QuantLinear which provides the capbility for onnx conversion

trajepl:
## Describe your changes

1. Enable AutoAwq in Olive -> AwqQuantizer where we do not only call AutoAwq to quantize model, but also use customized QuantLinear to make it is possible to convert the quantized torch model to onnx.
**Background**: AutoAwq quantization will rewrite the quantization utils with their own code which may not be able to be converted to onnx.
3. Add facebook/opt-125m example with AwqQuantizer


olive/passes/pytorch/ort_quant_utils.py
```diff
@@ -148,6 +148,7 @@ def __init__(self, bits, groupsize, infeatures, outfeatures, bias, *args, **kwar
         self.orig_fp_weight = None
```
devang-ml, line 148: What is the motivation behind renaming this file?

trajepl: This quant utils can be leveraged by all the quantization method in QLLM, like AWQ, GPTQ, HQQ and etc.

Previous gptq utils sounds a bit narrow only to GPTQ method.

devang-ml: Then just use `qaunt_utils.py`.  Having ORT in name may confuse folks.

trajepl: Done.

guotuofeng:
it seems to be under pytorch passes but the title is OrtQuantLinear?

olive/passes/pytorch/gptq.py
```diff
@@ -126,7 +126,7 @@ def _run_for_config(
         from auto_gptq.modeling import BaseGPTQForCausalLM
         from auto_gptq.modeling.auto import GPTQ_CAUSAL_LM_MODEL_MAP
 
-        from olive.passes.pytorch.gptq_utils import QuantLinearORT
+        from olive.passes.pytorch.quant_utils import QuantLinearORT
```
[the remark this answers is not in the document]
trajepl: The quant_utils is wrote with Pytorch.Module but it is customized for onnx exporter.

trajepl:
> it seems to be under pytorch passes but the title is OrtQuantLinear?

Updated title.

examples/opt_125m/user_script.py
```diff
@@ -0,0 +1,242 @@
...
+import numpy as np
+import torch
+from transformers import AutoConfig
+
+from olive.constants import Framework
+
+# pylint: disable=broad-exception-raised, redefined-outer-name
+
+model_id = "facebook/opt-125m"
+config = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
```
guotuofeng, line 18: do we need add readme? and update the example link in docs?

olive/passes/pytorch/awq.py
```diff
@@ -0,0 +1,224 @@
...
+from olive.model import PyTorchModelHandler
+from olive.model.utils.path_utils import normalize_path_suffix
+from olive.passes import Pass
+from olive.passes.pass_config import PassConfigParam
+
+logger = logging.getLogger(__name__)
+
+
+class AwqQuantizer(Pass):
+    """AWQ quantization."""
```
devang-ml, line 22: Please rename the class to be AutoAWQQuantizer. There could be another implementation of AWQ.

olive/passes/pytorch/gptq.py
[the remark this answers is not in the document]
devang-ml: Does it the packing need custom ORT Operator at runtime? If the exported graph is a standard ONNX graph (which does not need custom ORT Operator), then ORT in name is not clear.

olive/passes/pytorch/awq.py
```diff
@@ -0,0 +1,224 @@
...
+                    "calib_data": config["data_config"].params_config.get("data_name"),
+                    "split": config["data_config"].params_config.get("split"),
+                    "text_column": config["data_config"].params_config.get("input_cols"),
+                }
+            )
+
+        quantizer = (
+            self._pack_model_for_onnx_conversion(config)
+            if config["pack_model_for_onnx_conversion"]
+            else AutoAwqQuantizer
```
devang-ml, line 148: Please add a comment discussing how the packing is done differently to make it suitable for onnx conversion.

trajepl: Done with: 
```
pack_model_for_onnx_conversion is a flag to switch between the two quantizers
1. PyAutoAWQQuantizer is a quantizer implemented by autoawq package which is used by default
for quantizing the model. But it does not work with ONNX conversion since there are some
operations that are not supported by ONNX. So, we have to pack the model for ONNX conversion
2. That is why we have another quantizer method self._pack_model_for_onnx_conversion(config) to
return OrtAutoAWQQuantizer which is used for ONNX conversion.
```

olive/passes/pytorch/awq.py
trajepl: Done.

examples/opt_125m/user_script.py
trajepl: Added readme. And will update the example link once this PR is merge.

olive/passes/pytorch/gptq.py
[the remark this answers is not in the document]
trajepl: Yes, it packed. But to align the implementations(the customer `ORT Operator` you mentioned is actually customized torch module), I already remove the `ort_xxx`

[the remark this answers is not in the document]
trajepl: Updated.

guotuofeng (approved):

PR #1083, opened 2024-04-16
ExtractAdapters: Support int4 models and external data config

jambayk:
## Describe your changes
- `ExtractAdapters` pass supports int4 quantized models. 
- Previously output models were always saved with external data. External data config options are exposed to the user.
- `export_adapters` script also supports quantizing the weights to int4.

### DIFF
--- a/olive/passes/onnx/extract_adapters.py
+++ b/olive/passes/onnx/extract_adapters.py
@@ -6,7 +6,7 @@ import logging
 import re
 from copy import deepcopy
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, List
+from typing import TYPE_CHECKING, Any, Dict, List, Set, Tuple
 
 import numpy as np
 import onnx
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/extract_adapters.py
```diff
@@ -37,7 +37,7 @@ class ExtractAdapters(Pass):
 
     @classmethod
     def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
-        return {
+        config = {
```
trajepl, line 40: I did not see whether to run matmul 4 bits in this pass? The matmul 4 bits quantization is run before this pass?

jambayk: As discussed offline, this pass expects the in4 quantization pass to be run before it if the user wants an in4 quantized model.

trajepl (approved):

PR #1087, opened 2024-04-17
🦙 Llama2 optimization with multi-ep managed env

trajepl:
## Describe your changes
1. Llama2 optimization with multi-ep
2. Add info-level log to tell the system creation time.
3. Use device and ep but not only device to determine the default value of `use_gpu`.


examples/llama2/notebook/llama2_multiep/user_script.py
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+
+# -----------------------------------------------------------------------------
+# Metric Data Loader
+# -----------------------------------------------------------------------------
+
+
+class RandomDataLoader:
+    def __init__(
```
devang-ml, line 186: We should just provide BandomdataWithKVCache as built-in data-config so that we do not need to keep copying this in user script in every example.

trajepl: Indeed, I have that in mind also there is work item to track it. Let us finish the build-in kv cache data config later.

guotuofeng (approved):

PR #1088, opened 2024-04-17
⚠️  Fix warning: Overwriting attribute in super-class or sub-class

trajepl:
## Describe your changes
 Fix warning: Overwriting attribute in super-class or sub-class
https://github.com/microsoft/Olive/security/code-scanning/10681

### DIFF
--- a/olive/model/handler/base.py
+++ b/olive/model/handler/base.py
@@ -2,6 +2,7 @@ import logging
 from abc import ABC, abstractmethod
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
+from olive.common.config_utils import validate_config
 from olive.constants import Framework, ModelFileFormat
 from olive.hardware.accelerator import Device
 from olive.model.config import IoConfig
@@ -43,7 +44,9 @@ class OliveModelHandler(ABC, ResourceMixin, IoConfigMixin, JsonMixin, CompositeM
         self.model_file_format = model_file_format
         self.composite_parent = None
         self.model_attributes = model_attributes
-        self.io_config = io_config
+        self.io_config = (
+            validate_config(io_config, IoConfig).dict() if isinstance(io_config, (IoConfig, dict)) else io_config
+        )
 
         # store resource paths
         self.resource_paths: Dict[str, str] = {}
--- a/olive/model/handler/base.py
+++ b/olive/model/handler/base.py
@@ -45,9 +45,7 @@ class OliveModelHandler(ABC, ResourceMixin, IoConfigMixin, JsonMixin, CompositeM
         self.composite_parent = None
         self.model_attributes = model_attributes
         self.io_config = (
-            validate_config(io_config, IoConfig).dict(exclude_none=True)
-            if isinstance(io_config, (IoConfig, dict))
-            else io_config
+            validate_config(io_config, IoConfig).dict() if isinstance(io_config, (IoConfig, dict)) else io_config
         )
 
         # store resource paths
--- a/olive/model/handler/base.py
+++ b/olive/model/handler/base.py
@@ -45,7 +45,9 @@ class OliveModelHandler(ABC, ResourceMixin, IoConfigMixin, JsonMixin, CompositeM
         self.composite_parent = None
         self.model_attributes = model_attributes
         self.io_config = (
-            validate_config(io_config, IoConfig).dict() if isinstance(io_config, (IoConfig, dict)) else io_config
+            validate_config(io_config, IoConfig).dict(exclude_none=True)
+            if isinstance(io_config, (IoConfig, dict))
+            else io_config
         )
 
         # store resource paths
--- a/olive/model/handler/base.py
+++ b/olive/model/handler/base.py
@@ -1,9 +1,10 @@
 import logging
 from abc import ABC, abstractmethod
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 from olive.constants import Framework, ModelFileFormat
 from olive.hardware.accelerator import Device
+from olive.model.config import IoConfig
 from olive.model.handler.mixin import CompositeMixin, IoConfigMixin, JsonMixin, ResourceMixin
 from olive.resource_path import OLIVE_RESOURCE_ANNOTATIONS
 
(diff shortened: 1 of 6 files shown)


guotuofeng:
actually, I am suspicious we really need the fix or we can safely ignore them.

guotuofeng:
https://github.com/microsoft/Olive/blob/380cbe5c17947ae6857f92466cf285fa546cbd0f/olive/model/handler/onnx.py#L154

olive/model/handler/base.py
```diff
@@ -36,12 +37,13 @@ def __init__(
...
         model_path: OLIVE_RESOURCE_ANNOTATIONS = None,
         model_attributes: Optional[Dict[str, Any]] = None,
+        io_config: Union[Dict[str, Any], IoConfig, str, Callable] = None,
     ):
         self.framework = framework
         self.model_file_format = model_file_format
         self.composite_parent = None
         self.model_attributes = model_attributes
-        self.io_config = None
+        self.io_config = io_config
```
jambayk, line 46: I think the simplest solution would be to just remove `self.io_config` in the base class and remove the IoConfigMixin. 

We can either add this mixin to the subclasses that have self.io_config but don't implement `get_io_config`, or implement the method directly in them.

jambayk: Nevermind, io_config is used in almost all sub classes so this is not efficient either.

olive/model/handler/pytorch.py
```diff
@@ -315,21 +315,22 @@ def __init__(
         adapter_path: OLIVE_RESOURCE_ANNOTATIONS = None,
         model_attributes: Optional[Dict[str, Any]] = None,
     ):
+        validate_io_config = (
+            validate_config(io_config, IoConfig).dict() if isinstance(io_config, (IoConfig, dict)) else io_config
+        )
```
guotuofeng, line 320: it seems we can move this call into base class?

jambayk: that would work too. but we would need to update some other snpe related code to get a dict version of `io_config` since all snpe related code expect dicts.

olive/model/handler/pytorch.py
```diff
@@ -72,21 +72,26 @@ def __init__(
...
+        validated_io_config = (
+            validate_config(io_config, IoConfig).dict() if isinstance(io_config, (IoConfig, dict)) else io_config
         )
-        self.add_resources(locals())
-
         self.hf_config = None
+        new_model_attributes = None
         if hf_config:
             self.hf_config = validate_config(hf_config, HfConfig)
             hf_model_config = self.get_hf_model_config().to_dict()
```
guotuofeng, line 82: >self [](http://example.com/codeflow?start=30&length=4)

could not call self.method before call init

olive/model/handler/pytorch.py
trajepl: the `self.io_config` is always a dict in previous logics no matter the input is dict or `IoConfig`. Maybe no need to override the get_io_config I think.
`validate_config(io_config, IoConfig).dict()`

olive/model/handler/pytorch.py
```diff
@@ -72,11 +72,12 @@ def __init__(
 
         self.model_loader = model_loader
         self.model = None
+
         super().__init__(
             framework=Framework.PYTORCH,
             model_file_format=model_file_format,
             model_path=model_path,
-            model_attributes=model_attributes,
```
guotuofeng, line 79: forget to pass model_attributes

guotuofeng:
>         config["config"].update(self.io_config)

need get dict here

---
Refers to: olive/model/handler/snpe.py:61 in 5fdbc20. [](commit_id = 5fdbc209dc059a800e22de25af42022a7112283b, deletion_comment = False)

olive/model/handler/pytorch.py
trajepl: fixed. 😅

olive/model/handler/base.py
```diff
@@ -36,12 +38,15 @@ def __init__(
...
         model_attributes: Optional[Dict[str, Any]] = None,
+        io_config: Union[Dict[str, Any], IoConfig, str, Callable] = None,
     ):
         self.framework = framework
         self.model_file_format = model_file_format
         self.composite_parent = None
         self.model_attributes = model_attributes
-        self.io_config = None
+        self.io_config = (
+            validate_config(io_config, IoConfig).dict() if isinstance(io_config, (IoConfig, dict)) else io_config
```
guotuofeng, line 48: exclude_none=True

olive/model/handler/pytorch.py
jambayk: oh yeah, I missed that part. makes it simple then.

olive/model/handler/snpe.py
```diff
@@ -58,7 +58,8 @@ def prepare_session(
 
     def to_json(self, check_object: bool = False):
         config = super().to_json(check_object)
-        config["config"].update(self.io_config)
+        # only update the key in io_config where the value is not None
+        config["config"].update({key: value for key, value in config["config"].items() if value})
```
guotuofeng, line 62: bug here.

jambayk: I think the similar issue with extra keys will happen here too https://github.com/microsoft/Olive/blob/d068cf963e0ec1c5d4ecf9b87c2a1e118fe2303f/olive/passes/snpe/quantization.py#L109. 

Maybe the `get_io_config` method for this handler can be implemented to only return the 4 valid key, values? 

and then we can just call `get_io_config()` for this handler instead of accessing `io_config` directly?

guotuofeng: IMHO, the line to exclude the key here is needless. or at least, it is not the right place to exclude keys. A better way is in the snpe.__init__ to remove the unexpected keys instead of base on value.

BTW, the io_config is missing after this line change.

trajepl: Then seems original logics for snpe is clear. Maybe we can just ignore this warning for snpe.
image

olive/model/handler/snpe.py
```diff
@@ -31,12 +31,20 @@ def __init__(
...
+                "output_shapes": output_shapes,
+            },
         )
-        self.io_config = {
-            "input_names": input_names,
-            "input_shapes": input_shapes,
-            "output_names": output_names,
-            "output_shapes": output_shapes,
+
+    def get_io_config(self):
```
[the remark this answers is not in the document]
trajepl: I also tried to exclude the None value. But the io_config is not only from user's input. We also have the way of `hf_config` to pull io_config automatically where we cannot avoid None unless we manually skip the None value.

Compared with exclude_none, keeping SNPE's input/output names/shapes looks more clear.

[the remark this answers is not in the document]
guotuofeng: or we need make io_config as private variable?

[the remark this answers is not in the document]
trajepl: It looks good me. We can ensure the only entry to access the io_config is `get_io_config()`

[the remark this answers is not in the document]
guotuofeng: https://github.com/microsoft/Olive/pull/1091

jambayk:
This PR can be close now that the other one is merged?

jambayk:
closing since #1091 is merged.

PR #1089, opened 2024-04-17
Write configs as part of conversion pass

shaahji:
## Write configs as part of conversion pass

For generative models using GenAI model builder, when using ONNX model as input configs (config.json, generation_config.json & tokenizer) are required inputs. So, write out these files as outputs from the conversion pass.

### DIFF
--- a/olive/model/handler/mixin/hf_config.py
+++ b/olive/model/handler/mixin/hf_config.py
@@ -3,15 +3,21 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 import logging
+from pathlib import Path
 from typing import TYPE_CHECKING, Generator, Optional, Tuple
 
 from olive.constants import ModelFileFormat
 from olive.model.utils.hf_utils import (
     get_hf_model_config,
     get_hf_model_dummy_input,
+    get_hf_model_generation_config,
     get_hf_model_io_config,
+    get_hf_model_tokenizer,
     load_hf_model_from_model_class,
     load_hf_model_from_task,
+    save_hf_model_config,
+    save_hf_model_generation_config,
+    save_hf_model_tokenizer,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/conversion.py
```diff
@@ -113,6 +113,37 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
 
     def _run_for_config(
         self, model: PyTorchModelHandler, data_root: str, config: Dict[str, Any], output_model_path: str
+    ) -> Union[CompositeModelHandler, DistributedOnnxModelHandler, ONNXModelHandler]:
+        output_model = self._run_for_config_internal(model, data_root, config, output_model_path)
+
+        if model.hf_config:
+            if isinstance(output_model, CompositeModelHandler):
```
guotuofeng, line 120: seems these isinstance check already exist in Pass.run. Could we just call save_hf_* simply here?

shaahji: Pass has branches based on input model. Here's the branching is based on output model.

jambayk (approved):
Can you abstract some of the common code like file -> parent_dir in the follow up PR?

olive/model/handler/mixin/hf_config.py
```diff
@@ -44,6 +50,63 @@ def get_hf_model_config(self):
 
         return get_hf_model_config(self._get_model_path_or_name(), **self.hf_config.get_loading_args_from_pretrained())
 
+    def save_hf_model_config(self, output_dir: str, **kwargs):
```
guotuofeng, line 53: it seems those new added method is only used in the mixin class? if that's the case, we could make them private by adding prefix "_" to make it clear.

olive/model/handler/mixin/hf_config.py
```diff
@@ -44,6 +50,63 @@ def get_hf_model_config(self):
...
+
+    def save_hf_model_tokenizer(self, output_dir: str, **kwargs):
+        if self.hf_config is None:
+            raise ValueError("HF model_config is not available.")
+        if not Path(output_dir).is_dir():
+            raise ValueError("Expecting a directory as input.")
+
+        save_hf_model_tokenizer(self.get_hf_model_tokenizer(), output_dir, **kwargs)
+
+    def save_metadata_for_token_generation(self, output_dir: str, **kwargs):
```
guotuofeng, line 93: it seems only this method need to be public?

olive/model/handler/mixin/hf_config.py
shaahji: All save_XXX functions removed in follow up PR.

olive/model/handler/mixin/hf_config.py
shaahji: All save_XXX functions removed in follow up PR.

PR #1091, opened 2024-04-18
Unify io_config public method and public attribute

guotuofeng:
## Describe your changes
* Unify io_config public method and public attribute
* Fix qnn bug which is not tested never.


olive/model/handler/snpe.py
```diff
@@ -54,11 +61,11 @@ def prepare_session(
         if device == Device.NPU:
             device = SNPEDevice.DSP
         session_options.device = device
-        return SNPEInferenceSession(self.model_path, self.io_config, session_options)
+        return SNPEInferenceSession(self.model_path, self._io_config, session_options)
```
jambayk, line 64: it is supposed to self.io_config both here and at line 68. Since it only expects the four keys.

olive/model/handler/qnn.py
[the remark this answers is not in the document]
jambayk: Also the behavior of the pass changes here. previously it was only looking at the value of `self.io_config`, but now it is looking at the value of the old `self.get_io_config()` which will also not be None for an onnx model. 
This also triggered the bug described above.

olive/model/handler/snpe.py
jambayk: This also works since it the `exclude_none=True` is used but I think `self.io_config` is more correct to use here.

olive/model/handler/base.py
```diff
@@ -36,12 +38,17 @@ def __init__(
...
         model_attributes: Optional[Dict[str, Any]] = None,
+        io_config: Union[Dict[str, Any], "IoConfig", str, Callable] = None,
     ):
         self.framework = framework
         self.model_file_format = model_file_format
         self.composite_parent = None
         self.model_attributes = model_attributes
-        self.io_config = None
+        self._io_config = (
+            validate_config(io_config, IoConfig).dict(exclude_none=True)
```
trajepl, line 48: This test test_pytorch_model:test_func_io_config will fail.

olive/model/handler/qnn.py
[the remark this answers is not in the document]
trajepl: True. It is a bug. Fixed. 
For qnn sdk, the configs in io_config are not required and I guess they will parse the onnx graph somehow.
image

olive/model/handler/base.py
guotuofeng: actually, the test_func_io_config could pass in my local machine

olive/model/handler/snpe.py
guotuofeng: good catch

olive/model/handler/qnn.py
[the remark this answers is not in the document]
jambayk: with the current change, we will always pass the `--input_dims` argument. 
This change is probably fine since we are just doing the model parsing for qnn-sdk ourselves?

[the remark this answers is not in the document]
trajepl: Currently for mobilenet, it worked. But I am not for other cases. Sometimes the input may contain symbolic value.

I will test it for more models and upate it if needed.

olive/model/handler/snpe.py
guotuofeng: updated.

olive/model/handler/qnn.py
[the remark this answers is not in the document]
jambayk: sounds good to me. I think qnn-sdk only generates a model with static shape currently? So, if the converer accepts  a model has dynamic shape, the converter tool probably needs the user to provide fixed values to override the dynamic shapes? 

In which case, we would expect the over to provide `input_dims` to the pass config and won't reach this condition.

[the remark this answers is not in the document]
trajepl: >  I think qnn-sdk only generates a model with static shape currently?

Oh, true. We have a pass to convert dynamic shape into static shape.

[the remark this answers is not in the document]
jambayk: i remember you said the qnn-sdk tool works similar to snpe-sdk tool. If qnn-sdk converter only accepts models with static shapes then, this is not a concern for us. 

could you add a TODO here to check if the qnn tool only accepts static shape model or (dynamic shaped model + dim override)? 
Using this we can do some asserts here later.

olive/passes/qnn/conversion.py
```diff
@@ -79,38 +79,46 @@ def _run_for_config(
...
+        if input_dims:
+            for i in input_dims:
+                cmd_list.extend(["--input_dim", *i])
+        if out_nodes:
+            for o in out_nodes:
+                cmd_list.extend(["--out_node", o])
+        if config["extra_args"]:
+            cmd_list.extend(config["extra_args"].split())
+
+        runner.run(cmd_list)
```
jambayk, line 123: the runner doesn't support a list yet https://github.com/microsoft/Olive/blob/d068cf963e0ec1c5d4ecf9b87c2a1e118fe2303f/olive/platform_sdk/qualcomm/runner.py#L49.

we could convert str command to list around line 48 and then make everything else after it operate on the list command?

jambayk: I do agree with the TODO to make all commands list. Since the shlex with posix and not posix deals with "'str'" and paths differently. 

I also encountered similar issues before when using single quotes.

Accepting only lists takes away the responsibility from the subprocess runner to parse it.

guotuofeng: agree

trajepl: Oh, it is True. Let me upate the resolve part to let it accept list as well to return the value of the promise.

olive/common/utils.py
```diff
@@ -26,6 +26,9 @@ def run_subprocess(cmd, env=None, cwd=None, check=False):
     assert isinstance(cmd, (str, list)), f"cmd must be a string or a list, got {type(cmd)}."
     windows = platform.system() == "Windows"
     if isinstance(cmd, str):
+        # The cmd may contain the folder/file path from windows. This kind of path is like: C:\\User\\xxx\\...
+        # in posix mode, the shlex.split will split the path into ['C:Userxxx...'], which is not correct.
+        # in non-posix mode, the shlex.split will split the path into ['C:\\User\\xxx\\...'], which is correct.
```
jambayk, line 31: can you also add a comment warning about how `"str_0 'str_1'"`  is `['str_0', 'str_1']` in posix and `['str_0', "'str_1'"]` in non-posix? In such cases, providing a list is recommended to avoid unexpected behavior.

olive/model/handler/qnn.py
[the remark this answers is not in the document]
trajepl: For onnx, static is fine. But qnn can convert torch/tensorflow models. I think snpe is same. We need keep the interface to let user specify the input dim.

olive/passes/qnn/conversion.py
guotuofeng: should we avoid using shlex.split by favor string.split?

trajepl: Updated and keep the input to `_resolve_cmd` is string as of know. Also added the TODO: use list instead of string to avoid shlex.split error in non-posix mode

olive/common/utils.py
trajepl: Added

olive/passes/qnn/conversion.py
JKinx: do you mean line 68 here? https://github.com/microsoft/Olive/blob/d068cf963e0ec1c5d4ecf9b87c2a1e118fe2303f/olive/platform_sdk/qualcomm/runner.py#L68 

yes, I think that would be safer. Since we know the commands here of the type `command --arg ...`, string split will work. 
shlex.split is only useful if we are using some complex bash scripts that need the args to split correctly.

jambayk: do you mean line 68 here? https://github.com/microsoft/Olive/blob/d068cf963e0ec1c5d4ecf9b87c2a1e118fe2303f/olive/platform_sdk/qualcomm/runner.py#L68 

yes, I think that would be safer. Since we know the commands here of the type `command --arg ...`, string split will work. 
shlex.split is only useful if we are using some complex bash scripts that need the args to split correctly.

trajepl: image
The qnn/conversion. Just ensure our ci can pass and try to use List in separated PR.

olive/passes/qnn/conversion.py
```diff
@@ -79,38 +79,46 @@ def _run_for_config(
...
-            " ".join([f"--out_node {o}" for o in out_nodes]) if out_nodes else "",
-            config["extra_args"] or "",
+            *converter_program,
+            "--input_network",
+            model.model_path,
+            "--output_path",
+            output_model_path,
         ]
+        if input_dims:
+            for i in input_dims:
```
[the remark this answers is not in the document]
guotuofeng: seems we miss UT?

olive/platform_sdk/qualcomm/runner.py
```diff
@@ -43,32 +44,37 @@ def __init__(
...
+            logger.debug("Resolving command %s on Windows.", cmd_list)
             cmd_dir = Path(self.sdk_env.sdk_root_path) / "bin" / self.sdk_env.target_arch
-            cmd_name = cmd.split(" ")[0]
+            cmd_name = cmd_list[0]
             cmd_full_path = cmd_dir / cmd_name
             if cmd_full_path.with_suffix(".exe").exists():
                 cmd_full_path = cmd_full_path.with_suffix(".exe")
             if cmd_full_path.exists():
-                cmd = str(cmd_full_path) + cmd[len(cmd_name) :]
+                cmd_list.insert(0, str(cmd_full_path))
```
jambayk, line 64: This needs to be cmd_list[0]=... instead of insert. We are directly replacing the first command with full path

guotuofeng: good catch

olive/passes/qnn/conversion.py
guotuofeng: just change the _resolve_cmd to support both string and list

jambayk (approved):

olive/model/handler/snpe.py
```diff
@@ -31,13 +31,20 @@ def __init__(
...
-            "input_shapes": input_shapes,
-            "output_names": output_names,
-            "output_shapes": output_shapes,
-        }
+
+    @property
+    def io_config(self) -> Dict[str, Any]:
+        assert self._io_config, "SNPEModelHandler: io_config is not set"
+
+        keys = {"input_names", "input_shapes", "output_names", "output_shapes"}
```
xiaoyu-work, line 46: Any reason we need this?

PR #1100, opened 2024-04-22
HFConfig: don't pass kwargs to get tokenizer

jambayk:
## Describe your changes
When creating a tokenizer using `AutoTokenizer.from_pretrained`, it saves the kwargs and tries to serialize them during `save_pretrained`. The model loading args contains kwargs that are not related to tokenizer and cannot be serialized like torch.dtype and BitsAndBytesConfig. 
Don't pass any kwargs when getting tokenizer. If needed, we can filter the keys in the future.

### DIFF
--- a/olive/model/handler/mixin/hf_config.py
+++ b/olive/model/handler/mixin/hf_config.py
@@ -62,9 +62,9 @@ class HfConfigMixin:
         if self.hf_config is None:
             raise ValueError("HF model_config is not available")
 
-        return get_hf_model_tokenizer(
-            self._get_model_path_or_name(), **self.hf_config.get_loading_args_from_pretrained()
-        )
+        # don't provide loading args for tokenizer since it tries to serialize all kwargs
+        # TODO(anyone): only provide relevant kwargs, no use case for now to provide kwargs
+        return get_hf_model_tokenizer(self._get_model_path_or_name())
 
     def save_metadata_for_token_generation(self, output_dir: str, **kwargs):
         if self.hf_config is None:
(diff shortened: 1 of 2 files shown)


devang-ml (dismissed):

olive/model/handler/mixin/hf_config.py
```diff
@@ -62,9 +62,9 @@ def _get_hf_model_tokenizer(self):
         if self.hf_config is None:
             raise ValueError("HF model_config is not available")
 
-        return get_hf_model_tokenizer(
-            self._get_model_path_or_name(), **self.hf_config.get_loading_args_from_pretrained()
-        )
+        # don't provide loading args for tokenizer since it tries to serialize all kwargs
+        # TODO(anyone): only provide relevant kwargs, no use case for now to provide kwargs
+        return get_hf_model_tokenizer(self._get_model_path_or_name())
```
shaahji, line 67: Instead of removing the parameters here, we should pass kwargs to this function and not pass any parameters from call in `save_metadata_for_token_generation`. That way the API will stay generic and allow passing parameters in future without any change.

jambayk: I changed this here to keep the signature consistent with other similar methods. Since, it is a private method, I think it's okay? 

If we need to pass parameters in the future, we will probably have to handle `get_loading_args_from_pretrained` somewhere anyways.

shaahji: kwargs are optional and keyed so that wouldn't have changed the API. Anyways, there is work in future one way or other.

jambayk: sure! pushed a commit to accept kwargs

shaahji (approved):

shaahji (approved):

PR #1111, opened 2024-04-25
Remove name ambiguity in dataset configuration

shaahji:
## Remove name ambiguity in dataset configuration
Dataset configuration i.e. data_configs are represeneted as objects in config and so have two name entries -
  - the key of the object itself
  - DataConfig::name field

Use cases around the code use either/or and expect both these names to be identifical with no enforcement during validation. Removing the duplication by making the data_configs a list of objects so the key isn't required anymore. Updated the code to use DataConfig::name for all look-ups.

Note: Pending document update.

### DIFF
--- a/docs/source/features/huggingface_model_optimization.md
+++ b/docs/source/features/huggingface_model_optimization.md
@@ -194,24 +194,22 @@ Pass config:
 
 Example: datasets in `data_configs`:
 ```json
-"data_configs": {
-    "oasst1_train": {
-        "name": "oasst1",
-        "type": "HuggingfaceContainer",
-        "params_config": {
-            "data_name": "timdettmers/openassistant-guanaco",
-            "split": "train",
-            "component_kwargs": {
-                "pre_process_data": {
-                    "text_cols": ["text"],
-                    "corpus_strategy": "line-by-line",
-                    "source_max_len": 512,
-                    "pad_to_max_len": false
-                }
+"data_configs": [{
+    "name": "oasst1_train",
+    "type": "HuggingfaceContainer",
+    "params_config": {
+        "data_name": "timdettmers/openassistant-guanaco",
+        "split": "train",
+        "component_kwargs": {
+            "pre_process_data": {
+                "text_cols": ["text"],
+                "corpus_strategy": "line-by-line",
+                "source_max_len": 512,
+                "pad_to_max_len": false
             }
         }
     }
-}
+}]
 ```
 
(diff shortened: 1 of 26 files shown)


docs/source/tutorials/configure_data.rst
```diff
@@ -9,10 +9,10 @@ The configuration of Olive data config is positioned under Olive run config with
 
 .. code-block::
 
-    "data_configs": {
-        "dataset_1": {...},
-        "dataset_2": {...},
-    }
+    "data_configs": [
+        { "name": "dataset_1", ...},
+        { "name": "dataset_2", ...},
```
guotuofeng, line 14: @trajep, should we keep the key or keep the name?

docs/source/features/huggingface_model_optimization.md
```diff
@@ -194,24 +194,22 @@ Pass config:
...
-            "split": "train",
-            "component_kwargs": {
-                "pre_process_data": {
-                    "text_cols": ["text"],
-                    "corpus_strategy": "line-by-line",
-                    "source_max_len": 512,
-                    "pad_to_max_len": false
-                }
+"data_configs": [{
+    "name": "oasst1_train",
```
guotuofeng, line 198: using the name instead of key would have the possibility that there are multiple dataset have the same name. we need dedup them in our code. 

For me, it seems we could leverage key

shaahji: > dedup them in our code

Yes, but this happens once during validation. Not necessarily a hot code path.

Having DataConfig::name makes it easier to identify object when debugging/printing. Having the key doesn't provide any such advantage.

guotuofeng: @jambayk, @trajepl, what's your opinion?

jambayk: Hitesh and I previously discussed the two options. I was originally for removing the name field and keeping the dict key. But the name for debugging purpose sounds useful too. 

I think it the decision depends on whether the debugging usefulness outweights the benefits of keeping the keys (1. consistent with other fields like systems, evaluators, etc, 2.guaranteed unique keys 3. direct index into data_configs)

guotuofeng: Thanks @jambayk, I will let you and @trajepl to make decision since I am ok with either approach.

trajepl (approved):

PR #1112, opened 2024-04-25
Additional ModelBuilder pass options

shaahji:
## Additional ModelBuilder pass options

Adding the following new options for model builder pass
+ int4_block_size
+ int4_accuracy_level
+ exclude_embeds
+ exclude_lm_head
+ enable_cuda_graph

### DIFF
--- a/olive/passes/onnx/model_builder.py
+++ b/olive/passes/onnx/model_builder.py
@@ -36,6 +36,15 @@ class ModelBuilder(Pass):
         def __str__(self) -> str:
             return self.value
 
+    class AccuracyLevel(int, Enum):
+        fp32 = 1
+        fp16 = 2
+        bf16 = 3
+        int8 = 4
+
+        def __str__(self) -> str:
+            return str(self.value)
+
     @classmethod
     def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
         return {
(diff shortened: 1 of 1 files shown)


olive/passes/onnx/model_builder.py
```diff
@@ -137,6 +178,19 @@ def _run_for_config(
             model_path = str(model.hf_config.model_name)
             input_path = ""
 
+        if config.get("int4_block_size"):
+            if int(config["int4_block_size"]) not in [16, 32, 64, 128, 256]:
+                raise RuntimeError("Invalid int4_block_size. Accepted values: 16/32/64/128/256.")
+            extra_args["int4_block_size"] = config["int4_block_size"]
+        if "int4_accuracy_level" in config:
```
jambayk, line 185: I don't think this and later if checks are necessary. The pass run always gets a full config so the keys will always be present.

jambayk: Actually it applies to the Boolean parameters.

For the accuracy and block size, the check should be for None since the keys are always present.

shaahji: Done!

olive/passes/onnx/model_builder.py
```diff
@@ -137,6 +178,18 @@ def _run_for_config(
             model_path = str(model.hf_config.model_name)
             input_path = ""
 
+        if config.get("int4_block_size"):
+            if int(config["int4_block_size"]) not in [16, 32, 64, 128, 256]:
+                raise RuntimeError("Invalid int4_block_size. Accepted values: 16/32/64/128/256.")
```
guotuofeng, line 183: ValueError?

olive/passes/onnx/model_builder.py
```diff
@@ -36,6 +36,15 @@ class Precision(str, Enum):
         def __str__(self) -> str:
             return self.value
 
+    class AccuracyLevel(int, Enum):
+        fp32 = 1
+        fp16 = 2
+        bf16 = 3
+        int8 = 4
+
+        def __str__(self) -> str:
```
guotuofeng, line 45: do we need the `__str__` method since it is int enum?

shaahji: Prints a simpler representation rather than the full name.

olive/passes/onnx/model_builder.py
shaahji: Done!

guotuofeng (approved):

jambayk (approved):

PR #1115, opened 2024-04-26
Olive kv_cahe_config under io_config

trajepl:
## Describe your changes
1. Add the option of `kv_cache_config` in `io_config`:
  - kv_cache_config: None or False, will not enable kv cache
  - kv_cahce_config: True or dict, will generate related kv_cache input
    - Suggest to set `True` for huggingface transformer models
    - Suggest to set dict for non-huggingface models
2. Update phi2/llama2 example by removing io_config/dummy_inputs_func from user_scirpts and config json.

Will wait for review, then to update the documents. 


olive/model/handler/pytorch.py
```diff
@@ -281,6 +291,52 @@
...
+        dummy_inputs = self.past_key_values_input_filter_hook(dummy_inputs)
+        io_config = IoConfig.parse_obj(self.io_config)
+        unused_keys = set()
+        if io_config.kv_cache_config and not dummy_inputs.get(past_kv_names):
+            torch_past_key_values = []
+            k_inputs = io_config.kv_cache_config.get_ort_past_key_names()
+            v_inputs = io_config.kv_cache_config.get_ort_past_value_names()
+            for k_input, v_input in zip(k_inputs, v_inputs):
+                if k_input not in dummy_inputs or v_input not in dummy_inputs:
+                    raise ValueError("Cannot find past key-value pair for %s and %s in dummy inputs.", k_input, v_input)
```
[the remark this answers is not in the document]
guotuofeng: good catch

examples/llama2/llama2_qlora.json
```diff
@@ -2,9 +2,40 @@
...
+            "io_config": {
+                "input_names": [
+                    "input_ids",
+                    "attention_mask"
+                ],
+                "output_names": [
+                    "logits"
+                ],
+                "input_shapes": [
+                    [
```
devang-ml, line 14: nitpick: Would it make sense to collapse certain key values like this one on a single line. Sometime this gives novice user an impression that they need to provide complicated input.

olive/model/config/io_config.py
```diff
@@ -80,6 +88,72 @@ def check_string_to_int_dim_params(cls, v):
                 raise ValueError(f"Invalid string_to_int_dim_params: {dim_param}. Must be castable to int.") from None
         return v
 
+    def get_seq_len(self):
+        for idx, name in enumerate(self.input_names):
+            if name == "input_ids":
+                return self.input_shapes[idx][1]
```
guotuofeng, line 94: need we check if it is None?

examples/llama2/llama2_qlora.json
trajepl: I used json formatter to get the `complicated` one.

Just updated with collapsed json.

olive/model/config/io_config.py
trajepl: Updated.

PR #1118, opened 2024-04-27
Fix handling of extra_args for ModelBuilder pass

shaahji:
## Fix handling of extra_args for ModelBuilder pass

GenAI model builder's extra_args works based on existence of a key rather then the value of the key itself. So, Olive cannot pass a key with a value of true/false. The key has to explicitly be added/removed based on the value of the key itself in the user config.

### DIFF
--- a/olive/passes/onnx/model_builder.py
+++ b/olive/passes/onnx/model_builder.py
@@ -186,9 +186,14 @@ class ModelBuilder(Pass):
         if config.get("int4_accuracy_level"):
             extra_args["int4_accuracy_level"] = config["int4_accuracy_level"].value
 
-        extra_args["exclude_embeds"] = config["exclude_embeds"]
-        extra_args["exclude_lm_head"] = config["exclude_lm_head"]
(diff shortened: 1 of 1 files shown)


jambayk (approved):

olive/passes/onnx/model_builder.py
```diff
@@ -186,9 +186,14 @@ def _run_for_config(
...
-        extra_args["exclude_embeds"] = config["exclude_embeds"]
-        extra_args["exclude_lm_head"] = config["exclude_lm_head"]
-        extra_args["enable_cuda_graph"] = "1" if config["enable_cuda_graph"] else "0"
+        if config["exclude_embeds"]:
+            extra_args["exclude_embeds"] = config["exclude_embeds"]
+
+        if config["exclude_lm_head"]:
+            extra_args["exclude_lm_head"] = config["exclude_lm_head"]
+
+        if config["enable_cuda_graph"] is not None:
```
jambayk, line 195: nit: this condition is not needed since the value is always a bool and this condition is always true.

PR #1121, opened 2024-04-28
🌵  Olive kv_cache_config under io_config

trajepl:
## Describe your changes

1. Add the option of kv_cache_config in io_config:
    - kv_cache_config: None or False, will not enable kv cache
    - kv_cahce_config: True or dict, will generate related kv_cache input based on existed io_config and model attributes.
      - Suggest to set True for huggingface transformer models
      - Suggest to set dict for non-huggingface models
2. Update phi2/llama2 example by removing io_config/dummy_inputs_func from user_scirpts and config json.


olive/model/config/io_config.py
```diff
@@ -35,6 +40,10 @@ class IoConfig(ConfigBase):
     # We use a workaround here: first use dim_param like "1" to represent the dimension, and then
     # convert it to int in the onnx model.
     string_to_int_dim_params: List[str] = None
+    # if False, skip kv_cache input
+    # if True, use default KVCacheConfig
+    # if KVCacheConfig, use the provided KVCacheConfig
+    kv_cache_config: Union[bool, KVCacheConfig] = False
```
guotuofeng, line 46: kv_cache better since it might be bool or config? the kv_cache_config seems always be config?

examples/llama2/notebook/llama2_multiep/config_gpu.json
```diff
@@ -2,9 +2,17 @@
...
+                "input_names": ["input_ids", "attention_mask"],
+                "output_names": ["logits"],
+                "input_shapes": [[2,8], [2,40]],
+                "input_types": ["int32", "int32"],
+                "dynamic_axes": {
+                    "input_ids": {"0": "batch_size", "1": "sequence_length"},
+                    "attention_mask": {"0": "batch_size", "1": "sequence_length"}
+                },
+                "kv_cache_config": true
+            },
```
guotuofeng, line 15: seems extract the logic in user_script.py into json? would it complicate the user scenario?  @devang-ml

guotuofeng: could we only provide configs to kv_cache and keep using the original configs?

examples/llama2/llama2_qlora.json
```diff
@@ -2,9 +2,17 @@
...
+            "io_config": {
+                "input_names": ["input_ids", "attention_mask"],
+                "output_names": ["logits"],
+                "input_shapes": [[2,8], [2,40]],
+                "input_types": ["int32", "int32"],
+                "dynamic_axes": {
+                    "input_ids": {"0": "batch_size", "1": "sequence_length"},
+                    "attention_mask": {"0": "batch_size", "1": "sequence_length"}
+                },
+                "kv_cache_config": true
```
guotuofeng, line 14: >kv_cache_config [](http://example.com/codeflow?start=17&length=15)

better have a way to specify the kv_cache if user would like to use io_config function.

olive/model/handler/pytorch.py
```diff
@@ -246,22 +253,25 @@ def get_user_io_config(self, io_config: Union[Dict[str, Any], IoConfig, str, Cal
...
         if isinstance(io_config, dict):
-            # io_config is provided
-            return io_config
-
-        if isinstance(io_config, IoConfig):
-            # io_config is an IoConfig
-            return io_config.dict()
-
-        if isinstance(io_config, (str, Callable)):
+            io_config_obj = IoConfig.parse_obj(io_config)
```
guotuofeng, line 258: >io_config_obj = IoConfig.parse_obj(io_config) [](http://example.com/codeflow?start=12&length=45)

need not the conversion to object and then back to dict for performance sake.

olive/model/handler/pytorch.py
```diff
@@ -284,6 +294,56 @@ def io_config(self) -> Dict[str, Any]:
...
+            ):
+                dummy_input_keys.discard(name)
+
+        unused_keys = dummy_input_keys - set(io_config.get("input_names"))
+
+        if unused_keys:
+            logger.debug("Removing unused dummy inputs: %s", unused_keys)
+            dummy_inputs = exclude_keys(dummy_inputs, unused_keys)
+        return dummy_inputs
+
```
guotuofeng, line 346: seems could be moved to dummy_input and io_config mixin?

guotuofeng: the two method here seems only depends on io_config dict, which means the new logic could be easily moved to other utils or mixin specific to pytorch kv_cache. Not sure these logic could be reused by onnx model

olive/model/utils/hf_mappings.py
```diff
@@ -73,6 +73,7 @@
 # there is the priority order: NUM_HEADS_NAMES[0] and HIDDEN_SIZE_NAMES[0] are the first choice
 # which means user can override the value in config file
 NUM_HEADS_NAMES = ("num_heads", "num_attention_heads", "n_head", "n_heads", "encoder_attention_heads")
+NUM_HIDDEN_LAYER_NAMES = ("num_hidden_layers", "num_layers", "n_layer", "n_layers")
```
guotuofeng, line 76: >NUM_HIDDEN_LAYER_NAMES [](http://example.com/codeflow?start=0&length=22)

could we let each model specify the hidden/head names instead of speculate by Olive using the central place?

examples/llama2/llama2_qlora.json
trajepl: I think user can directly implement the kv_cache in their function. The way of io_config function still exists and works.

olive/model/utils/hf_mappings.py
guotuofeng: in this way, Olive will not need maintain the head names/hidden layer names whenever there are new model released.

trajepl: We have that and expose the config to user in kv_cache_config if they are clear the detailed value. https://github.com/microsoft/Olive/pull/1121/files#diff-a9c02489c56cfee4594be95735d9edd61297573f46c68581ba6e38556020415bR20

guotuofeng (dismissed):
Better allow specify the kv cache if user would like keep using io_config function.

olive/model/config/io_config.py
```diff
@@ -80,6 +89,78 @@ def check_string_to_int_dim_params(cls, v):
...
+        return 1
+
+
+def complete_kv_cache_with_model_attributes(kv_cache_config, model_attributes):
+    if isinstance(kv_cache_config, bool) and kv_cache_config:
+        kv_cache_config = KVCacheConfig()
+    # get kv_cache_config from model config if not provided
+    # complete num_hidden_layers/num_attention_heads/hidden_size is not provided
+    # in io_config_obj.kv_cache_config
+    num_hidden_layers = find_first_matched_value(model_attributes, NUM_HIDDEN_LAYER_NAMES)
```
guotuofeng, line 127: I mean we need specify the name explicitily.

olive/model/utils/hf_mappings.py
guotuofeng: https://github.com/microsoft/Olive/pull/1121/files#r1582052278. This is what I mean, the first get value might not be correct.

examples/llama2/llama2_qlora.json
guotuofeng: So, it seems user would not be allowed to use the kv_cache config if he write function? what's your thought? @devang-ml

olive/model/utils/hf_mappings.py
guotuofeng: for example, we could choose the most populate name "num_hidden_layers" as default. for all other names like n_layer, user need specify the name or value by kv_cache config explicitly

olive/model/config/io_config.py
```diff
@@ -80,6 +89,78 @@ def check_string_to_int_dim_params(cls, v):
...
+        if not self.input_shapes or not self.input_names:
+            return 1
+        for idx, name in enumerate(self.input_names):
+            if name == "input_ids":
+                return self.input_shapes[idx][0]  # pylint: disable=E1136
+        # if no input_ids, return 1
+        return 1
+
+
+def complete_kv_cache_with_model_attributes(kv_cache_config, model_attributes):
```
guotuofeng, line 121: >kv_cache_config [](http://example.com/codeflow?start=44&length=15)

for example, we could pass the num_heads names as string or int value here. If num_heads is string, we could search from model_attributes before constructing the KVCacheConfig instance at line 123

olive/model/config/kv_cache_config.py
```diff
@@ -0,0 +1,113 @@
...
+    ort_past_value_name: str = "past_value_<id>"
+    ort_present_key_name: str = "present_key_<id>"
+    ort_present_value_name: str = "present_value_<id>"
+
+    # world_size is used for distributed model. If world_size > 1,
+    # the number of heads should be divisible by world_size
+    world_size: int = 1
+    num_hidden_layers: Optional[int] = None
+    num_attention_heads: Optional[int] = None
+    hidden_size: Optional[int] = None
```
guotuofeng, line 22: make them required.

olive/model/config/kv_cache_config.py
```diff
@@ -0,0 +1,113 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+from typing import Dict, Optional
+
+from olive.common.config_utils import ConfigBase
+from olive.common.pydantic_v1 import validator
+
+
+class KVCacheConfig(ConfigBase):
```
guotuofeng, line 11: >KVCacheConfig [](http://example.com/codeflow?start=6&length=13)

could onnx model leverage the config for evaluation?

olive/model/utils/hf_mappings.py
trajepl: I think value is enough, if user knows the name. he must know the value.

guotuofeng: If user specify the value, the value will exists in both config and model attributes, which has the possibility that they are different.

olive/model/handler/pytorch.py
trajepl: These 2 functions cannot be applied by Onnx model. Here is the though I added the hook function for dummy inputs generation:

Different models(pytorch, onnx) have different inputs which may reqire minor changes to adjust dummy input to feed into different models.

olive/model/utils/hf_mappings.py
trajepl: What kind of case they are different? For huggingface model attributes, it is pulled from huggingface hub. For the config, it should be same with the value in huggingface hub's config.json. 
If the values are different, I think it is user's fault? User should take the responsibility.

guotuofeng: as for difference, I mean config redundancy since there are multiple places to have the config

olive/model/config/kv_cache_config.py
trajepl: They are required to generate kv cache configs. But I may think they are not to specify by users.

If we make them required here, what if user only want to specify the world_size or other kv_config values in workflow config json? I think that will lead a pydantic parsing error as the auto-filling to these configs needs input model.

guotuofeng: I think the config is created manually using https://github.com/microsoft/Olive/pull/1121/files#diff-c1d0bb6a2b86e4e881354ced6194029e3da1ca1bf7da8baa1b381e7d63b8331fR123, where we could specify the argument explicitly

trajepl: > where we could specify the argument explicitly.
The line you pointed tells that when `kv_cache` is bool and is true, we create a empty KVCacheConfig. `kv_cache` still can be a dict to let workflows config to parse.

> From kv_cache point of view, the two values are required. If we want deduce, then the creator of kv_cache should handle the deduce logic.
True. I thought that too before. But for the deduction of these values, it needs model's huggingface config that is got during olive runtime. That means it is not easy to use pydantic validator to deduce the value.
Also on another hand, to users, I think these values are not required if they have already give the hugginface model name/class.

guotuofeng: the deduction of out of scope of pydantic validation. Instead, it is the model responsibility. We could manually check kv_cache PyTorch model. If it is bool, extract all the needed info from model attributes. When all the needed fields are ready, we can use them to construct the KvCacheConfig, In this way, we can split the pydantic validation logic and kv_cache deduction logic. At the same time, the kv_cache will only be type of KvCacheConfig instead of union with bool, which might confuse the end user. 

Basically, the deduction and pydantic validation is two different responsibility and we need split them.

olive/model/handler/pytorch.py
guotuofeng: could we move them to something like KvCacheMixin/PyKvCacheMixin since the two method only related to kv_cache so we can better organize the code? Originally, the PyTorch model only contains model loading stuff and mixin overrides method

olive/model/config/kv_cache_config.py
trajepl: Updated.

olive/model/config/io_config.py
```diff
@@ -80,6 +89,88 @@ def check_string_to_int_dim_params(cls, v):
...
+    num_attention_heads = find_first_matched_value(model_attributes, NUM_HEADS_NAMES)
+    hidden_size = find_first_matched_value(model_attributes, HIDDEN_SIZE_NAMES)
+    kv_cache_obj = None
+    if isinstance(kv_cache, bool) and kv_cache:
+        kv_cache_obj = KVCacheConfig(
+            num_hidden_layers=num_hidden_layers,
+            num_attention_heads=num_attention_heads,
+            hidden_size=hidden_size,
+        )
+    elif isinstance(kv_cache, Dict):
```
guotuofeng, line 132: dict instead of typing Dict?

olive/model/handler/pytorch.py
trajepl: Updated.

olive/model/config/io_config.py
trajepl: Updated.

examples/llama2/llama2_qlora.json
trajepl: Based our offline discussion,if user is writing io_config user function then it is fair to expect them to write their own kv_cache inputs in their scripts.

guotuofeng (approved):

PR #1124, opened 2024-04-29
fix codeql md5 error

guotuofeng:
## Describe your changes
* fix https://github.com/microsoft/Olive/security/code-scanning/10844

### DIFF
--- a/olive/common/utils.py
+++ b/olive/common/utils.py
@@ -56,13 +56,13 @@ def run_subprocess(cmd, env=None, cwd=None, check=False):
 
 
 def hash_string(string):  # pragma: no cover
-    md5_hash = hashlib.md5()
+    md5_hash = hashlib.sha256()
     md5_hash.update(string.encode())
     return md5_hash.hexdigest()
 
 
 def hash_io_stream(f):  # pragma: no cover
-    md5_hash = hashlib.md5()
+    md5_hash = hashlib.sha256()
     # Read and update hash in chunks of 4K
     for byte_block in iter(lambda: f.read(4096), b""):
         md5_hash.update(byte_block)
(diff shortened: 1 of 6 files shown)


olive/common/utils.py
```diff
@@ -56,13 +56,13 @@ def run_subprocess(cmd, env=None, cwd=None, check=False):
...
 def hash_string(string):  # pragma: no cover
-    md5_hash = hashlib.md5()
+    md5_hash = hashlib.sha256()
     md5_hash.update(string.encode())
     return md5_hash.hexdigest()
 
 
 def hash_io_stream(f):  # pragma: no cover
-    md5_hash = hashlib.md5()
+    md5_hash = hashlib.sha256()
```
guotuofeng, line 65: @jambayk, would this change introduce path length too long bug in Windows since it double the hash string length?

jambayk: I am not familiar with the windows path length issue. 

32 characters does seem long for a part of a file name. Since we are not using the hash for cryptography and only to generate a unique file name, I think we could truncate the hash and only use the first 8 hex characters or so? It looks like even 8 is unique enough to not have any clashes in practical use. 
We could truncate them in:
https://github.com/microsoft/Olive/blob/5d08fb7a424429ddc27f0311ffec6a43b5da5c2b/olive/engine/engine.py#L744
https://github.com/microsoft/Olive/blob/5d08fb7a424429ddc27f0311ffec6a43b5da5c2b/olive/engine/engine.py#L759

Ref:
https://stackoverflow.com/questions/49223904/how-unique-are-the-first-8-12-characters-of-sha256-hashes
https://developers.google.com/safe-browsing/v4/urls-hashing
https://github.com/openai/triton/blob/a454e546a895879184476bcbc9f45337c80e93ba/python/triton/tools/compile.py#L75

guotuofeng: make sense, will update

jambayk (approved):
Thanks!

PR #1128, opened 2024-04-30
Codeql: fix global variable warning

guotuofeng:
## Describe your changes
* Fix https://github.com/microsoft/Olive/security/code-scanning/10629

### DIFF
--- a/test/unit_test/common/test_retry.py
+++ b/test/unit_test/common/test_retry.py
@@ -9,6 +9,9 @@ from olive.common.utils import retry_func
 # pylint: disable=global-variable-undefined, used-before-assignment
 
 
+num_tries = 0
+
+
 def fail_with_key_error():
     global num_tries
     if num_tries == 0:
--- a/test/unit_test/common/test_retry.py
+++ b/test/unit_test/common/test_retry.py
@@ -6,7 +6,7 @@ import pytest
 
 from olive.common.utils import retry_func
 
-# pylint: disable=global-variable-undefined, used-before-assignment
+# pylint: disable=global-statement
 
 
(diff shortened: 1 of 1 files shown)


test/unit_test/common/test_retry.py
```diff
@@ -9,6 +9,9 @@
 # pylint: disable=global-variable-undefined, used-before-assignment
 
 
+num_tries = 0
```
jambayk, line 12: Since we are using num_tries to keep track of the number of calls, I found an alternative way of doing it

```python
def fail_with_key_error(lst):
    if len(lst) == 0:
        lst.append("dummy")
        raise KeyError("This is a key error")
    return True


@pytest.mark.parametrize("exceptions", [KeyError, (KeyError, ValueError), Exception])
def test_success(exceptions):
    input_lst = []
    assert retry_func(fail_with_key_error, args=[input_lst], max_tries=2, delay=1, exceptions=exceptions)
    assert len(input_lst) == 1
```

guotuofeng: I think it is ok to keep the global variable here for test only purpose.

guotuofeng: Personally, I like the global variable approach since it is simpler.

jambayk: Sure! I wasn’t sure if it would pass both the lint and codeql checks. But looks like it does https://github.com/microsoft/Olive/security/code-scanning?query=is%3Aopen+pr%3A1128

I agree, the global variable is better since it’s direct while the one I shared isn’t.

jambayk (approved):

PR #1131, opened 2024-05-01
Add sample code for GenAI model inferencing

shaahji:
## Add sample code for GenAI model inferencing

Also, update the packaging logic to include sample code and required runtime binaries.

NOTE: Native (both CPP & CS) binaries are not yet being published by the GenAI team and needs revisit once those binaries are available for download.


olive/engine/packaging/packaging_generator.py
```diff
@@ -565,6 +572,41 @@ def _generate_onnx_mlflow_model(model_dir: Path, inference_config: Dict):
     return mlflow_model_path
 
 
+def _package_model_builder_package(tempdir: Path):
```
devang-ml, line 575: nitpick : Please rename as _package_onnxruntime_genai_package

shaahji: Done!

.lintrunner.toml
jambayk, line 1: does adding the path to somewhere in https://github.com/microsoft/Olive/blob/main/pyproject.toml not work? Or there are too many different rules to ignore? 

I am not as familiar with the linter configurations, so we could revisit this after the PR is merged and Mike is back to if there is a way to ignore a path fully without having to add it to each lintrunner adapter.

shaahji: We can surely revisit the update to the linter configuration file. The current change seems to be the only thing that has worked so far. I tried quite a few different ways but they all work for one but fail for something else.

olive/engine/packaging/packaging_generator.py
```diff
@@ -296,11 +296,17 @@ def _package_candidate_models(
         tempdir = Path(temp_dir)
 
         if packaging_type == PackagingType.Zipfile:
+            best_node: FootprintNode = _get_best_candidate_node(pf_footprints, footprints)
+            model_attributes = best_node.model_config["config"].get("model_attributes") or {}
+
             if packaging_config.include_sample_code:
-                _package_sample_code(Path(__file__).parent, tempdir)
+                _package_sample_code(Path(__file__).parent, tempdir, model_attributes.get("is_generative", False))
```
xiaoyu-work, line 303: Nit: we can abstract `is_generative` to a separate method.

olive/engine/packaging/packaging_generator.py
```diff
@@ -565,7 +572,42 @@ def _generate_onnx_mlflow_model(model_dir: Path, inference_config: Dict):
     return mlflow_model_path
 
 
-def _package_onnxruntime_packages(tempdir: Path, pf_footprint: "Footprint"):
+def _package_onnxruntime_genai_runtime_dependencies(tempdir: Path):
+    # pylint: disable=not-an-iterable
+    installed_packages = [
+        pkg
+        for pkg in pkg_resources.working_set
+        if pkg.key.startswith("onnxruntime-genai") or pkg.project_name.startswith("onnxruntime-genai")
```
xiaoyu-work, line 580: Any difference between key and project_name?

olive/engine/packaging/packaging_generator.py
```diff
@@ -565,7 +572,42 @@ def _generate_onnx_mlflow_model(model_dir: Path, inference_config: Dict):
...
+        pkg_name = pkg.key if pkg.key.startswith("onnxruntime-genai") else pkg.project_name
+        download_command = DOWNLOAD_COMMAND_TEMPLATE.substitute(
+            package_name=pkg_name, version=pkg.version, python_download_path=python_download_path
+        )
+
+        try:
+            run_subprocess(download_command)
+        except Exception:
+            logger.exception(
+                "Failed to download %s package. Manually download & install the required package.", pkg_name
```
xiaoyu-work, line 603: Nit: `Please manually download & install the required package`, otherwise it doesn't sound like an action to the user lol

xiaoyu-work:
Can we also update unit tests for this?

olive/engine/packaging/packaging_generator.py
shaahji: I found some context here https://www.programcreek.com/python/?code=recipy%2Frecipy%2Frecipy-master%2Fintegration_test%2Fenvironment.py which gives the impression that they _could_ be different.

olive/engine/packaging/packaging_generator.py
shaahji: Done!

olive/engine/packaging/packaging_generator.py
shaahji: Done!

shaahji:
> Can we also update unit tests for this?

I will revisit the unit test once the runtime headers/binaries are available from genai repo. There is also some cleanup work in the packaging unit test to make packaging and sample code testing possible in CI. There is an open task on Dashboard from grab.

olive/engine/packaging/packaging_generator.py
xiaoyu-work: Not a big deal for now, we could be more consistent in our code.

xiaoyu-work (approved):

PR #1135, opened 2024-05-03
NV-ModelOPT INT4 quantization

riyadshairi979:

## Tests
- Unit test: `pytest test/unit_test/passes/onnx/test_nvmo_quantization.py`
- Example: `python -m olive.workflows.run --config bert_nvmo_ptq.json`

### DIFF
--- a/examples/bert/README.md
+++ b/examples/bert/README.md
@@ -5,6 +5,7 @@ This folder contains examples of BERT optimization using different workflows.
 - CPU: [Optimization with Intel® Neural Compressor PTQ](#bert-optimization-with-intel®-neural-compressor-ptq-on-cpu)
 - CPU: [Optimization with QAT Customized Training Loop](#bert-optimization-with-qat-customized-training-loop-on-cpu)
 - GPU: [Optimization with CUDA/TensorRT](#bert-optimization-with-cudatensorrt-on-gpu)
+- GPU: [Optimization with TensorRT-Model-Optimizer](#bert-optimization-with-tensorRT-model-optimizer-on-cpugpu)
 
 Go to [How to run](#how-to-run)
 
@@ -99,6 +100,15 @@ This workflow performs BERT optimization on GPU with CUDA/TensorRT. It performs 
 2. TensorRT: `TensorrtExecutionProvider`
     - *PyTorch Model -> Onnx Model -> ONNX Runtime performance tuning with trt_fp16_enable*
     Config file: [bert_trt_gpu.json](bert_trt_gpu.json)
+
+### BERT optimization with TensorRT-Model-Optimizer on CPU/GPU
+This workflow performs BERT post training quantization (PTQ) on CPU/GPU with TensorRT-Model-Optimizer. It performs the optimization pipeline:
+- *PyTorch Model -> Onnx Model -> Transformers Optimized Onnx Model -> TensorRT-Model-Optimizer Quantized Onnx Model*
+
+Deployment support for TensorRT-Model-Optimizer quantized models is coming soon in ORT, in the meantime try [TensorRT 10.x](https://github.com/NVIDIA/TensorRT/tree/v10.0.1).<br>
+Config file: [bert_nvmo_ptq.json](bert_nvmo_ptq.json)
+
+
 ## How to run
 ### Pip requirements
 Install the necessary python packages:
--- a/examples/bert/bert_nvmo_ptq.json
+++ b/examples/bert/bert_nvmo_ptq.json
@@ -1,0 +1,39 @@
+{
+    "input_model":{
+        "type": "PyTorchModel",
+        "config": {
+            "model_loader": "load_pytorch_origin_model",
+            "model_script": "user_script.py",
+            "io_config": {
+                "input_names": ["input_ids", "attention_mask", "token_type_ids"],
+                "input_shapes": [[1, 128], [1, 128], [1, 128]],
+                "input_types": ["int64", "int64", "int64"],
+                "output_names": ["output"]
+            }
+        }
+    },
+    "passes": {
+        "conversion": {
+            "type": "OnnxConversion",
+            "config": {
+                "target_opset": 13
+            }
+        },
+        "transformers_optimization": {
+            "type": "OrtTransformersOptimization",
+            "disable_search": true,
+            "config":{
+                "model_type": "bert"
+            }
+        },
+        "quantization": {
+            "type": "NVModelOptQuantization",
+            "config": {
+                "user_script": "nv_user_script.py",
+                "dataloader_func": "create_calibration_dataloader",
+                "precision": "int4",
+                "algorithm": "AWQ"
+            }
+        }
+    }
+}
--- a/examples/bert/bert_nvmo_ptq.json
+++ b/examples/bert/bert_nvmo_ptq.json
@@ -16,7 +16,7 @@
         "conversion": {
             "type": "OnnxConversion",
             "config": {
-                "target_opset": 13
+                "target_opset": 17
             }
         },
         "transformers_optimization": {
--- a/examples/bert/bert_nvmo_ptq.json
+++ b/examples/bert/bert_nvmo_ptq.json
@@ -23,7 +23,8 @@
             "type": "OrtTransformersOptimization",
             "disable_search": true,
             "config":{
-                "model_type": "bert"
+                "model_type": "bert",
+                "opt_level":0
             }
         },
         "quantization": {
--- a/examples/bert/nv_user_script.py
+++ b/examples/bert/nv_user_script.py
@@ -1,0 +1,30 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+import torch
+from datasets import load_dataset  # type: ignore[import]
+from datasets.utils import logging as datasets_logging  # type: ignore[import]
+from transformers import AutoTokenizer, BertModel  # type: ignore[import]
+
+datasets_logging.disable_progress_bar()
+datasets_logging.set_verbosity_error()
+
+
+def load_pytorch_origin_model(model_path):
+    model = BertModel.from_pretrained("bert-base-uncased")
+    model.eval()
+    return model
+
+
+def create_calibration_dataloader(data_dir, batch_size, calib_size, *args, **kwargs):
+    tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
+
+    def tokenization(example):
+        return tokenizer(example["text"], padding="max_length", max_length=128, truncation=True)
+
+    dataset = load_dataset("rotten_tomatoes", split="validation[:10%]")
+    dataset = dataset.map(tokenization, batched=True)
+    dataset.set_format(type="torch", columns=["input_ids", "token_type_ids", "attention_mask"])
+
+    return torch.utils.data.DataLoader(dataset.select(range(calib_size)), batch_size=batch_size, drop_last=True)
--- a/examples/bert/nv_user_script.py
+++ b/examples/bert/nv_user_script.py
@@ -11,13 +11,13 @@ datasets_logging.disable_progress_bar()
 datasets_logging.set_verbosity_error()
 
 
-def load_pytorch_origin_model(model_path):
+def load_pytorch_origin_model():
     model = BertModel.from_pretrained("bert-base-uncased")
     model.eval()
     return model
 
 
-def create_calibration_dataloader(data_dir, batch_size, calib_size, *args, **kwargs):
+def create_calibration_dataloader(batch_size, calib_size, *args, **kwargs):
     tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
 
     def tokenization(example):
--- a/olive/auto_optimizer/config_template/pass_capability.yaml
+++ b/olive/auto_optimizer/config_template/pass_capability.yaml
@@ -73,7 +73,9 @@ OrtPerfTuning:
 NVModelOptQuantization:
   EP:
     - CPU
+    - CUDA
   precision:
     - int4
   accelerator:
     - cpu
+    - gpu
--- a/olive/auto_optimizer/config_template/pass_capability.yaml
+++ b/olive/auto_optimizer/config_template/pass_capability.yaml
@@ -70,3 +70,10 @@ OrtPerfTuning:
   EP: null
   precision: null
   accelerator: null
+NVModelOptQuantization:
+  EP:
+    - CPU
+  precision:
+    - int4
+  accelerator:
+    - cpu
--- a/olive/auto_optimizer/config_template/pass_capability.yaml
+++ b/olive/auto_optimizer/config_template/pass_capability.yaml
@@ -70,12 +70,3 @@ OrtPerfTuning:
   EP: null
   precision: null
   accelerator: null
-NVModelOptQuantization:
-  EP:
-    - CPU
-    - CUDA
-  precision:
-    - int4
-  accelerator:
-    - cpu
-    - gpu
--- a/olive/olive_config.json
+++ b/olive/olive_config.json
@@ -240,7 +240,8 @@
             "onnxruntime-openvino"
         ],
         "nvmo": [
-            "nvidia-modelopt"
+            "nvidia-modelopt~=0.11.0",
+            "onnx-graphsurgeon"
         ]
     }
 }
--- a/olive/olive_config.json
+++ b/olive/olive_config.json
@@ -178,6 +178,12 @@
         },
         "SNPEtoONNXConversion": {
             "module_path": "olive.passes.snpe.snpe_to_onnx.SNPEtoONNXConversion"
+        },
+        "NVModelOptQuantization": {
+            "module_path": "olive.passes.onnx.nvmo_quantization.NVModelOptQuantization",
+            "extra_dependencies": [
+                "nvmo"
+            ]
         }
     },
     "extra_dependencies": {
@@ -232,6 +238,9 @@
             "onnxruntime-directml",
             "onnxruntime-gpu",
             "onnxruntime-openvino"
+        ],
+        "nvmo": [
+            "nvidia-modelopt"
         ]
     }
 }
--- a/olive/passes/onnx/nvmo_quantization.py
+++ b/olive/passes/onnx/nvmo_quantization.py
@@ -108,7 +108,7 @@ class NVModelOptQuantization(Pass):
             from modelopt.onnx.quantization.int4 import quantize_int4  # type: ignore[import]
         except ImportError as exc:
             raise ImportError(
-                "Please install `olive-ai[nvmo]` or `nvidia-modelopt` to use INT4 AWQ quantization!"
+                "Please install `olive-ai[nvmo]` or `nvidia-modelopt[onnx]` to use INT4 AWQ quantization!"
             ) from exc
 
         calib_dataloader = self._user_module_loader.call_object(
--- a/olive/passes/onnx/nvmo_quantization.py
+++ b/olive/passes/onnx/nvmo_quantization.py
@@ -2,51 +2,40 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+import logging
 from copy import deepcopy
+from enum import Enum
 from pathlib import Path
 from typing import Any, Callable, Dict, Union
 
-from olive.cache import get_local_path_from_root
 from olive.hardware.accelerator import AcceleratorSpec
 from olive.model import OliveModelHandler
 from olive.model.utils import resolve_onnx_path
 from olive.passes import Pass
 from olive.passes.onnx.common import model_proto_to_olive_model
 from olive.passes.pass_config import ParamCategory, PassConfigParam
-from olive.resource_path import OLIVE_RESOURCE_ANNOTATIONS
 from olive.strategy.search_parameter import Categorical
+
+logger = logging.getLogger(__name__)
+
 
 # static quantization specific config
 _dataloader_config = {
-    "data_dir": PassConfigParam(
-        type_=OLIVE_RESOURCE_ANNOTATIONS,
-        category=ParamCategory.DATA,
-        description="""
-            Path to the directory containing the dataset.
-            For local data, it is required if quant_mode is 'static' and dataloader_func is provided.
-        """,
-    ),
     "batch_size": PassConfigParam(
         type_=int,
         default_value=1,
-        description="""
-            Batch size for calibration, only used if dataloader_func is provided.
-        """,
+        description="Batch size for calibration.",
     ),
     "calib_size": PassConfigParam(
         type_=int,
         default_value=64,
-        description="""
-            Calibration data size, only used if dataloader_func is provided.
-        """,
+        description="Calibration data size.",
     ),
     "dataloader_func": PassConfigParam(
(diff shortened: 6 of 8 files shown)


olive/passes/onnx/nvmo_quantization.py
```diff
@@ -0,0 +1,118 @@
...
+    # set this to True if the pass has parameters that are functions or objects and the user script is required
+    # to import the module containing the function or object
+    _requires_user_script: bool = True
+
+    @classmethod
+    def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        config = {
+            "approach": PassConfigParam(
+                type_=str,
+                default_value="static",
```
jambayk, line 75: this default value doesn't match the searchable values. Also, the parameter might be better named as "precision"?

olive/passes/onnx/nvmo_quantization.py
```diff
@@ -0,0 +1,118 @@
...
+        description="""
+            Function/function name to generate dataloader for calibration,
+            required if quant_mode is 'static' and data_config is None.
+        """,
+    ),
+    "dataloader_func_kwargs": PassConfigParam(
+        type_=Dict[str, Any],
+        description="Keyword arguments for dataloader_func.",
+    ),
+    "data_config": PassConfigParam(
```
jambayk, line 54: since this param is not used in the pass, it can be removed

examples/bert/bert_nvmo_ptq.json
```diff
@@ -0,0 +1,50 @@
...
+            "config":{
+                "model_type": "bert"
+            }
+        },
+        "quantization": {
+            "type": "NVModelOptQuantization",
+            "config": {
+                "user_script": "nv_user_script.py",
+                "dataloader_func": "create_dataloader",
+                "metric": {
```
jambayk, line 34: "metric" is also not used as well.

test/unit_test/passes/onnx/test_nvmo_quantization.py
jambayk, line 1: Please also add the required package to https://github.com/microsoft/Olive/blob/main/test/requirements-test.txt.

olive/passes/onnx/nvmo_quantization.py
jambayk: Oh it looks like only int4 is support right now. If so, this field can be removed as well.

jambayk: "algoritm is not used as well

test/unit_test/passes/onnx/test_nvmo_quantization.py
jambayk: looks like the `nvidia-modelopt` is only available on windows, so I think we have to disable the unit test entirely for now. We currently don't have separate test requirements based on os.

riyadshairi979:
> @riyadshairi979 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree [company="Nvidia"]

riyadshairi979:
@microsoft-github-policy-service agree company="Nvidia"

test/unit_test/passes/onnx/test_nvmo_quantization.py
jambayk: @riyadshairi979 can you remove the package from the test requirements because it will fail to find the package on Linux. 
Instead please add a `@pytest.mark.skip(reason="nvidia-modelopt not available on Linux. Re-enable after introducing OS specific requirements")` instead of the current `skipif`

riyadshairi979: Actually [modelopt](https://pypi.org/project/nvidia-modelopt/) is going to be available for both windows and linux on next Monday/Tuesday. I will remove the `skipif` in next commit.

jambayk: Oh that's great! We won't be able to run the CI or merge it until then but I can review the logic.

olive/passes/onnx/nvmo_quantization.py
```diff
@@ -0,0 +1,116 @@
...
+        calib_dataloader = self._user_module_loader.call_object(
+            config["dataloader_func"],
+            data_dir,
+            config["batch_size"],
+            config["calib_size"],
+            model_path=model.model_path,
+            **(config["dataloader_func_kwargs"] or {}),
+        )
+
+        if config["precision"] != "int4" or config["algorithm"] not in ["RTN", "AWQ"]:
```
jambayk, line 108: will other precisions be supported in the future? If not, then we could just remove the "precision" config parameter.

Can you add this check as part of `validate_search_point` like in https://github.com/microsoft/Olive/blob/3120697d4e9d2dee5af0877d9cfc082f6a422dfb/olive/passes/onnx/optimum_conversion.py#L53

jambayk: It is also possible to use a `str,Enum` as type for the param which comes with automatic validation like in https://github.com/microsoft/Olive/blob/3120697d4e9d2dee5af0877d9cfc082f6a422dfb/olive/passes/onnx/model_builder.py#L71

test/unit_test/passes/onnx/test_nvmo_quantization.py
jambayk: oh that's great! I will wait until the release to trigger the PR validation pipelines then.

olive/passes/onnx/nvmo_quantization.py
riyadshairi979: Yes, other precisions will be supported as well. Looking into `validate_search_point `.

test/unit_test/passes/onnx/test_nvmo_quantization.py
```diff
@@ -0,0 +1,51 @@
...
+        data = get_pytorch_model_dummy_input()
+        try:
+            item = {"input": data.numpy()}
+            self.sample_counter -= 1
+            return item
+        except Exception:
+            return None
+
+
+def dummpy_dataloader_func(data_dir, batch_size, *args, **kwargs):
```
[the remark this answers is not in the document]
riyadshairi979: The `quantize_int4` API uses `CalibrationDataReader` but the other `quantize` API of modelopt uses iterable. We will make the change to make it congruent everywhere and compatible with Olive.

olive/passes/onnx/nvmo_quantization.py
```diff
@@ -0,0 +1,116 @@
...
+from olive.strategy.search_parameter import Categorical
+
+# static quantization specific config
+_dataloader_config = {
+    "data_dir": PassConfigParam(
+        type_=OLIVE_RESOURCE_ANNOTATIONS,
+        category=ParamCategory.DATA,
+        description="""
+            Path to the directory containing the dataset.
+            For local data, it is required if quant_mode is 'static' and dataloader_func is provided.
```
jambayk, line 26: Just some minor comments about the description of the config parameters. Please remove the references to `quant_mode` and `static`. 

If a calibration data loader is always needed, please add `required=True` under `dataloader_func` and update the descriptions of the other related params to remove the "only used if ..." part.

Can you also add a docstring under pass class to give a description of what this pass does? Thanks!

test/requirements-test.txt
```diff
@@ -16,6 +16,7 @@ git+https://github.com/microsoft/TransformerCompression.git ; python_version >=
 mlflow>=2.4.0
 neural-compressor
 nncf==2.7.0
+nvidia-modelopt
```
jambayk, line 19: Looks like onnx-graphsurgeon is also required. From the windows ci failure:
image

riyadshairi979: I have updated the requirement with correct package.
qq: we have a dependency on `onnxruntime~=1.16.3`, will that conflict with onnxruntime requirement for other packages/tools within olive?

jambayk: Oh, that is an issue. Our CI tests for the most recent stable release of ort. We have another pipeline that tests for the previous major release of ort but that one will soon migrate to 1.17.x since 1.18 is on the way.

riyadshairi979: How to see the status of the pipeline? The `int4` support that we added in this PR does not require strict `1.16.3`, it will work with latest ort too. Let see if pipeline passes.
We will update our ort dependency to latest version for other quantization formats like `fp8`, `int8` before supporting them in olive.

jambayk: okay, let me trigger the ci pipeline. you can see the status by clicking the jobs that show up. it takes you to another page that has links to the azure pipeline jobs.

jambayk: I see the the `[all]` extra pins the ort version so this will reinstall ort. Can we directly install onnx-graphsurgeon package so that the ort package is not reinstalled? 
image

riyadshairi979: I think it is doable, I have updated the requirements.
Bdw, I do not see `onnxruntime` as dependency in requirements-tests.txt, is it coming through core lib requirements?

jambayk: it is installed separately so that we can control the version being tested  https://github.com/microsoft/Olive/blob/63c0430fe3da033e086e1dfbe00d9bdc918ecca8/.azure_pipelines/job_templates/olive-test-template.yaml#L37

examples/bert/nv_user_script.py
```diff
@@ -0,0 +1,30 @@
...
+import torch
+from datasets import load_dataset  # type: ignore[import]
+from datasets.utils import logging as datasets_logging  # type: ignore[import]
+from transformers import AutoTokenizer, BertModel  # type: ignore[import]
+
+datasets_logging.disable_progress_bar()
+datasets_logging.set_verbosity_error()
+
+
+def load_pytorch_origin_model():
```
jambayk, line 14: have tested this example locally? the model_loader function is always called with `model_path` whether it is none or not

jambayk: you can empty the cache directory when rerunning after making code changes. This can also be done by adding `"clean_cache": true` under `"engine"` section during dev. otherwise, it will just reload from cache if the config json did not change.

riyadshairi979: I tested locally, it works.

riyadshairi979: Sorry, there was a formatting error. Please start the pipeline one more time.

olive/olive_config.json
```diff
@@ -232,6 +238,9 @@
             "onnxruntime-directml",
             "onnxruntime-gpu",
             "onnxruntime-openvino"
+        ],
+        "nvmo": [
+            "nvidia-modelopt"
```
jambayk, line 243: will graphsurgeon always be required for modelopt passes? perhaps we can add it here too?

riyadshairi979: Yes, it will always be required. I will make the change after the current pipeline ends/crashes.

riyadshairi979:
@jambayk By seeing the pipeline fails, just realized that the installed modelopt needs some fixes! We have to wait a day or two to get some of our int4 fixes from main branch to get into the release! Should we merge this PR by disabling the test or do you think we should wait?

jambayk:
> @jambayk By seeing the pipeline fails, just realized that the installed modelopt needs some fixes! We have to wait a day or two to get some of our int4 fixes from main branch to get into the release! Should we merge this PR by disabling the test or do you think we should wait?

@devang-ml @EmmaNingMS what do you think?

riyadshairi979:
Another question @jambayk: if modelopt adds dependency `onnxruntime-gpu` instead of `onnxruntime`, how it will differ in olive and what version it will install?

test/requirements-test.txt
jambayk, line 1: @riyadshairi979 replying to https://github.com/microsoft/Olive/pull/1135#issuecomment-2098555504 here so that we can have a thread without having to quote reply. 
> if modelopt adds dependency onnxruntime-gpu instead of onnxruntime, how it will differ in olive and what version it will install?

The behavior depends on whether the dependency is a direct requirement or an extra. 
- The easiest case would be an extra since we can just not install using the extra and directly consume whatever ort is already installed in the environment
- If it's an extra but ort-gpu is still needed for the pass to run, we could skip the test on the cpu ci and only run it on the gpu ci
- It gets complicated if it's a direct dependency that gets installed when we install the modelopt package. This is because ort is shipped as many different packages (onnxruntime, onnxruntime-gpu, ort-nightly, ort-nightly-gpu) but they all install into the same `onnxruntime` namespace. this means that if `onnxruntime` is already installed, `onnxruntime-gpu` doesn't uninstall `onnxruntime` and just adds it's files into the `onnxruntime` package folder overwriting files if necessary. This could lead to unexpected behavior so we usually uninstall all ort packages before installing a new one. Since in the ci, we test for a specific version of ort, if the modelopt packages installs ort-gpu, we should add it directly to the `requirements-test-gpu.txt`. Even then we would have to separate the requirements between the stable and nightly pipelines (which installs and tests ort-nighlty)

tldr; the ideal scenario is that onnxruntime(-gpu) is an optional dependency/extra for modelopt. If the pass requires a specific flavor and version of ort, we skip the tests as needed.

riyadshairi979: I see. So do you suggest we add only `onnxruntime` in our `modelopt` dependency and keep `onnxruntime-gpu` as optional?

olive/passes/onnx/nvmo_quantization.py
```diff
@@ -0,0 +1,129 @@
...
+        return True
+
+    def _run_for_config(
+        self, model: OliveModelHandler, data_root: str, config: Dict[str, Any], output_model_path: str
+    ) -> OliveModelHandler:
+        try:
+            from modelopt.onnx.quantization.int4 import quantize_int4  # type: ignore[import]
+        except ImportError as exc:
+            raise ImportError(
+                "Please install `olive-ai[nvmo]` or `nvidia-modelopt` to use INT4 AWQ quantization!"
```
kevalmorabia97, line 111:
```suggestion
                "Please install `olive-ai[nvmo]` or `nvidia-modelopt[onnx]` to use INT4 AWQ quantization!"
```

test/requirements-test.txt
riyadshairi979: `pip install nvidia-modelopt` will NOT install those onnx requirements.

jambayk: That will be fine then. The changes in this PR to `requirements-test.txt` is sufficient since the other `onnx` dependencies are already present.

test/requirements-test.txt
```diff
@@ -16,6 +16,8 @@ git+https://github.com/microsoft/TransformerCompression.git ; python_version >=
 mlflow>=2.4.0
 neural-compressor
 nncf==2.7.0
+nvidia-modelopt~=0.11.0
+onnx-graphsurgeon
```
kevalmorabia97, line 20:
```suggestion
nvidia-modelopt[onnx]~=0.11.0
onnx-graphsurgeon
```

test/requirements-test.txt
riyadshairi979: cool. do you want me to move the `onnx-graphsurgeon` to `olive_config.json` [here](https://github.com/microsoft/Olive/pull/1135/files#diff-f359471e7bc949029f5a4508182e76085dcc0acbb8db0a8c171ca19051210689R243)?

jambayk: yes. if graphsurgeon will always be required for this pass, adding it along with modelopt would be good.

jambayk: btw, is onnxruntime-extensions also required for the pass?

riyadshairi979: Not for INT4 support but for other quantization format we support plugins too, which requires `onnxruntime-extensions`.

riyadshairi979:
Please have another pass @all

test/requirements-test.txt
```diff
@@ -16,8 +16,7 @@ git+https://github.com/microsoft/TransformerCompression.git ; python_version >=
 mlflow>=2.4.0
 neural-compressor
 nncf==2.7.0
-nvidia-modelopt~=0.11.0
-onnx-graphsurgeon
+nvidia-modelopt[onnx]~=0.11.0
```
jambayk, line 19: can you change this to the previous `nvidia-modelopt~=0.11.0` and `onnx-graphsurgeon`. 
Like we discussed earlier, the `onnx` extra in the ci test would install the ort dependencies from this extra on top of what ort is already installed in the test environment.

olive/auto_optimizer/config_template/pass_capability.yaml
```diff
@@ -70,3 +70,12 @@ OrtPerfTuning:
   EP: null
   precision: null
   accelerator: null
+NVModelOptQuantization:
```
riyadshairi979, line 73: @jambayk how this capability is used? ModelOpt work both with CPU and CUDA EP. But I didn't understand the `accelerator` field.

test/requirements-test.txt
jambayk: This change cannot be made due to dependency clash between the ci's ort package requirements and the extras requirements. Please refer to the discussion above for more details https://github.com/microsoft/Olive/pull/1135#discussion_r1592736575

olive/auto_optimizer/config_template/pass_capability.yaml
jambayk: Olive has a feature where it uses this yaml file to automatically create model optimization workflows by mixing and matching compatible passes. 
The accelerator field refers to the compute hardware device. Olive uses a combination of ep,accelerator to categorize workflows. so the eps and accelerators here are used to create a matrix of supported combos.

I think you can remove this pass from here. Since the pass does not support data configs yet, the auto optimizer would not be able to use this pass as it won't have the dataloader func to provide to it. 
@trajepl can confirm?

riyadshairi979: I will remove for now. We can add them back later if needed.

olive/passes/onnx/nvmo_quantization.py
```diff
@@ -0,0 +1,129 @@
...
+        type_=int,
+        default_value=1,
+        description="Batch size for calibration.",
+    ),
+    "calib_size": PassConfigParam(
+        type_=int,
+        default_value=64,
+        description="Calibration data size.",
+    ),
+    "dataloader_func": PassConfigParam(
```
trajepl, line 34: nitpick: btw, it is fine not to add data_config here. I am planning to implent a standalone data container to unify dataloader_func and data_config.

examples/bert/bert_nvmo_ptq.json
```diff
@@ -0,0 +1,39 @@
...
+            "type": "OnnxConversion",
+            "config": {
+                "target_opset": 17
+            }
+        },
+        "transformers_optimization": {
+            "type": "OrtTransformersOptimization",
+            "disable_search": true,
+            "config":{
+                "model_type": "bert"
```
jambayk, line 26: Hi, I ran the example locally with the newly released linux package and it worked. 
image

Can you add `"opt_level":0` here? Otherwise, the transformers optimizer tries to use an inference session with the ep to do further session graph fusions. but that doesn't work if the ep is tensorrt as the session graph cannot be saved.

examples/bert/bert_nvmo_ptq.json
jambayk, line 1: Please also add a section for this workflow in the README.md. Thanks!

olive/passes/onnx/nvmo_quantization.py
```diff
@@ -0,0 +1,129 @@
...
+            logger.error("Only INT4 quantization with RTN and AWQ algorithm is supported.")
+            return False
+
+        return True
+
+    def _run_for_config(
+        self, model: OliveModelHandler, data_root: str, config: Dict[str, Any], output_model_path: str
+    ) -> OliveModelHandler:
+        try:
+            from modelopt.onnx.quantization.int4 import quantize_int4  # type: ignore[import]
```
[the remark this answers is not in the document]
riyadshairi979: Good suggestion, thanks. I will make the change in next release.

olive/passes/onnx/nvmo_quantization.py
jambayk: I think this was a follow up to another discussion about data configs. There is no change needed from your side at this point.

jambayk (dismissed):
Thanks for the contribution! Will merge once the CI passes.

examples/bert/README.md
```diff
@@ -99,6 +100,14 @@ This workflow performs BERT optimization on GPU with CUDA/TensorRT. It performs
 2. TensorRT: `TensorrtExecutionProvider`
     - *PyTorch Model -> Onnx Model -> ONNX Runtime performance tuning with trt_fp16_enable*
     Config file: [bert_trt_gpu.json](bert_trt_gpu.json)
+
+### BERT optimization with TensorRT-Model-Optimizer on CPU/GPU
```
riyadshairi979, line 104: @jambayk should add something about deployment? like below
```Users can deploy the quantized ONNX model using TensorRT 10.x but that is not supported through Olive right now, stay tuned!`

jambayk: The comment about the deployment would probably be about ORT and not Olive? Since olive only produces the model and it gets deployed using ort, tensorrt or some other engine. I will let @EmmaNingMS comment on this.

EmmaNingMS: Vote for this version if ORT support is in the plan. "Deployment support for TensorRT-Model-Optimizer quantized models is coming soon in ORT, in the meantime try TensorRT 10.x"

What's the timeline to add the support in ORT? Will Nvidia contribute to it?

riyadshairi979: @EmmaNingMS I will ask team about the roadmap regarding these support. Thanks.

olive/passes/onnx/nvmo_quantization.py
jambayk, line 108: Not a breaking issue since onnx opset 21 is not currently supported by ort. But I think the quantizer probably needs to update the model's opset to 21 if it is not already so? Or require the model's opset to be 21? (olive provides a pass for that too https://github.com/microsoft/Olive/blob/3e5ce3f6907eed5063d937af756e9a214a9342eb/olive/passes/onnx/conversion.py#L579)

Otherwise, the model graph is technically invalid for lower opset version since int4 in dq and q ops were only added in opset version 21. https://github.com/onnx/onnx/releases/tag/v1.16.0

riyadshairi979: Our tool does not care much about opset version either. Anything else before we can merge this PR?

jambayk: Sounds good to me. Everything looks good from my side. I will approve and merge once the CI passes. 

The only outstanding thing is the discussion about the deployment messaging but that can be updated in a follow up PR if necessary.

jambayk: The failing/cancelled test is not related to this PR. I will rerun it once the other tests complete.

jambayk (approved):
Thanks!

PR #1136, opened 2024-05-05
`additional_files` as file names and inheritance, transformers opt on host

jambayk:
## Describe your changes
**additional_files**
- `additional_files` has been standardized to file names that are relative to the `model_path` directory.
- This avoids issues and confusions when using full paths.
- Inheritance of `model_attributes` comes before copying of additional files. Otherwise, if the input_model has additional_files, the output model's `model_attributes` is created from a new dict with only "additional_files" as keys.

**ModelBuilder** 
- Updated to follow the file name standard. 
- New option `enable_GQA_on_CPU`

**ExportAdapters**
- Add adapter weight file name to `additional_files`

**OrtTransformersOptimization**
- Don't run this pass on target since inference session is only created if `opt_level` > 0. 
- This is not a common scenario and unnecessarily blocks other common use cases such as optimizing for DML EP using a Linux host


shaahji:
Relative paths won't work with passes like `OrtPerfTuning` because the pass itself does not generate any outputs. The pass spits the input model's path as output again of its own. Please verify the PR with examples/llama2/llama2_model_builder.py with both with and without the optional argument `--metadata-only` and ensure that the packaged zip file contains all the necessary metadata files.

jambayk:
Closing in favor of #1139

PR #1141, opened 2024-05-09
🐛 bugfix: model_builder `search` config might be None

trajepl:
## Describe your changes

model_build `search` config might be None

### DIFF
--- a/olive/passes/onnx/model_builder.py
+++ b/olive/passes/onnx/model_builder.py
@@ -216,7 +216,9 @@ class ModelBuilder(Pass):
         with open(genai_config_filepath) as istrm:
             genai_config = json.load(istrm)
 
-        genai_config["search"] = {**genai_config.get("search", {}), **config.get("search", {})}
+        new_search_options = config.get("search") or {}
+        new_search_options.update(genai_config.get("search") or {})
+        genai_config["search"] = new_search_options
 
         with open(genai_config_filepath, "w") as ostrm:
             json.dump(genai_config, ostrm, indent=4)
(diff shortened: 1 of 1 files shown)


olive/passes/onnx/model_builder.py
```diff
@@ -216,7 +216,9 @@ def _run_for_config(
         with open(genai_config_filepath) as istrm:
             genai_config = json.load(istrm)
 
-        genai_config["search"] = {**genai_config.get("search", {}), **config.get("search", {})}
+        new_search_options = config.get("search") or {}
+        new_search_options.update(genai_config.get("search") or {})
```
jambayk, line 220: I think the order the opposite. we want `genai_config.update(config.get("search) or {})`

jambayk: btw, python doesn't complain about common keys in the `{**a, **b}` solution 
image

`genai_config["search"] = {**genai_config.get("search", {}), **(config.get("search") or {})}` also works.

olive/passes/onnx/model_builder.py
```diff
@@ -216,7 +216,9 @@ def _run_for_config(
         with open(genai_config_filepath) as istrm:
             genai_config = json.load(istrm)
 
-        genai_config["search"] = {**genai_config.get("search", {}), **config.get("search", {})}
+        new_search_options = config.get("search") or {}
```
guotuofeng, line 219: don't change the original config

olive/passes/onnx/model_builder.py
trajepl: You are right. Updated like:
`genai_config["search"] = {**(genai_config.get("search") or {}), **(config.get("search") or {})}`

to ensure it is safe to merge genai_config, config's search config.

jambayk (approved):

PR #1147, opened 2024-05-10
⚱️ new interface for `set_generator_params` for genai > 0.1.0

trajepl:
## Describe your changes

1. genai 0.1.0, `set_generator_params` accept dict as argument. But after that, `set_generator_params` uncompresses the dict as kwargs. This PR is used to support all of the version of genai.
2. Typos fix

### DIFF
--- a/olive/engine/packaging/sample_code/GenAIOnnxModel/python/code_sample.py
+++ b/olive/engine/packaging/sample_code/GenAIOnnxModel/python/code_sample.py
@@ -11,6 +11,7 @@ import sys
 import time
 
 import onnxruntime_genai as og
+from packaging import version
 
 # ruff: noqa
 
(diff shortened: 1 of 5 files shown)


olive/engine/packaging/sample_code/GenAIOnnxModel/python/code_sample.py
```diff
@@ -108,7 +109,10 @@ def _main():
             if name in args and getattr(args, name)
         }
     )
-    params.set_search_options(search_options)
+    if version.parse(og.__version__) > version.parse("0.1.0"):
+        params.set_search_options(**search_options)
+    else:
```
guotuofeng, line 114: do we really need to support genai 0.1.0? I think it is ok to only support 0.2.0

trajepl: I think that too. But currently 0.1.0 is the official one. 
0.2.0 is in pre-release.

image

guotuofeng: we can stop support 0.1.0 since the example need use 0.2.0. otherwise, we need set trust_remote_code or something similar like what Emma run into.

trajepl: Okay, let me update it.

jambayk: agree. let's just support the latest version. their code went through a lot of changes since it was still in pre-release

trajepl: Done.

PR #1148, opened 2024-05-10
Fix exception handling in `load_hf_model_from_task`

jambayk:
## Describe your changes
`load_hf_model_from_task` catches `OSError,ValueError` to allow iterating over potentially multiple candidate classes. However, for most cases, there is only one model class to try and there are valid errors like device OOM, unsupported kwargs, that need to be rethrown. 
Currently, the model loading silently fails and the caller gets `None` as the model. This is unexpected behavior and makes it hard to debug without error logs or failure with traceback.
 
### DIFF
--- a/olive/model/utils/hf_utils.py
+++ b/olive/model/utils/hf_utils.py
@@ -51,7 +51,7 @@ def load_hf_model_from_task(task: str, name: str, **kwargs) -> "PreTrainedModel"
             # the ValueError need to be caught since there will be multiple model_class for single task.
             # if the model_class is not the one for the task, it will raise ValueError and
             # next model_class will be tried.
-            logger.debug(
+            logger.info(
                 "Failed to load model %s with name_or_path %s.\n kwargs: %s.\n Exception raised: %s",
                 model_class,
                 name,
@@ -59,12 +59,8 @@ def load_hf_model_from_task(task: str, name: str, **kwargs) -> "PreTrainedModel"
                 e,
             )
 
-    if model is None:
-        # don't expect for this condition to be true since class_tuple is never empty
-        # either model load worked or error raised
-        # just to be safe from_pretrained returned a model and to satisfy linter
-        raise ValueError(f"Failed to load model with name_or_path {name}. Class tuples were {class_tuple}.")
-
+    # this won't be None since class_tuple is never empty and we only reach here if model loaded successfully
+    # satisfies linter too
     return model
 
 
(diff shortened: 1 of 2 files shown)


olive/model/utils/hf_utils.py
```diff
@@ -41,22 +37,32 @@ def load_hf_model_from_task(task: str, name: str, **kwargs):
...
-    model = None
+    exceptions = []
     for model_class in class_tuple:
         try:
             model = model_class.from_pretrained(name, **kwargs)
             logger.debug("Loaded model %s with name_or_path %s", model_class, name)
             return model
-        except (OSError, ValueError):
+        except (OSError, ValueError) as e:
+            if len(class_tuple) == 1:
```
guotuofeng, line 47: what about we raise the exception if it is the last item in the tuple?
```python
for i, model_class in enumerate(class_tuple):
      except:
        if i == len(class_tuple) - 1:
           raise
```

olive/model/utils/hf_utils.py
```diff
@@ -37,26 +33,42 @@ def load_hf_model_from_task(task: str, name: str, **kwargs):
...
+                # error could be device OOM, device_map: "auto" not supported, etc
+
+                # len(class_tuple) > 1: not common - image-segmentation, conversational, etc
+                # there is no easy way to get tracebacks for earlier failures, so just raise from last
+                raise
             # the ValueError need to be caught since there will be multiple model_class for single task.
             # if the model_class is not the one for the task, it will raise ValueError and
             # next model_class will be tried.
-            continue
+            logger.debug(
```
guotuofeng, line 54: warning?

olive/model/utils/hf_utils.py
```diff
@@ -37,26 +33,42 @@ def load_hf_model_from_task(task: str, name: str, **kwargs):
...
+                "Failed to load model %s with name_or_path %s.\n kwargs: %s.\n Exception raised: %s",
+                model_class,
+                name,
+                kwargs,
+                e,
+            )
+
+    if model is None:
```
guotuofeng, line 62: we can safely remove the model None check and raise of ValueError?

jambayk: updated

guotuofeng (approved):

olive/model/utils/hf_utils.py
jambayk: changed to info

olive/model/utils/hf_utils.py
jambayk: done

PR #1149, opened 2024-05-13
Add OnnxIOFloat16ToFloat32 Pass

xiaoyu-work:
## Describe your changes

Add OnnxIOFloat16ToFloat32 Pass

### DIFF
--- a/docs/source/api/passes.rst
+++ b/docs/source/api/passes.rst
@@ -42,6 +42,12 @@ OnnxFloatToFloat16
 --------------------
 .. autoconfigclass:: olive.passes.OnnxFloatToFloat16
 
+.. _onnx_io_float16_to_float32:
+
+OnnxIOFloat16ToFloat32
+--------------------
+.. autoconfigclass:: olive.passes.OnnxIOFloat16ToFloat32
+
 .. _ort_mixed_precision:
 
 OrtMixedPrecision
--- a/docs/source/api/passes.rst
+++ b/docs/source/api/passes.rst
@@ -45,7 +45,7 @@ OnnxFloatToFloat16
 .. _onnx_io_float16_to_float32:
 
 OnnxIOFloat16ToFloat32
---------------------
+----------------------
 .. autoconfigclass:: olive.passes.OnnxIOFloat16ToFloat32
 
 .. _ort_mixed_precision:
--- a/examples/phi3/phi3.py
+++ b/examples/phi3/phi3.py
@@ -179,6 +179,7 @@ def genai_run(prompt, model_path, max_length):
     print("Generator created")
 
     first = True
+    first_token_timestamp = None
     new_tokens = []
 
     print("\n", prompt)
@@ -200,12 +201,15 @@ def genai_run(prompt, model_path, max_length):
     del generator
 
     run_time = time.time() - started_timestamp
-    print(
-        "\n\n"
-        f"Prompt tokens: {len(input_tokens)}, New tokens: {len(new_tokens)},"
-        f" Time to first: {(first_token_timestamp - started_timestamp):.2f}s,"
-        f" New tokens per second: {len(new_tokens)/run_time:.2f} tps"
-    )
+    if first_token_timestamp is None:
+        print("\n\nNo tokens generated")
+    else:
+        print(
+            "\n\n"
+            f"Prompt tokens: {len(input_tokens)}, New tokens: {len(new_tokens)},"
+            f" Time to first: {(first_token_timestamp - started_timestamp):.2f}s,"
+            f" New tokens per second: {len(new_tokens)/run_time:.2f} tps"
+        )
 
 
 if __name__ == "__main__":
--- a/examples/phi3/phi3.py
+++ b/examples/phi3/phi3.py
@@ -1,0 +1,212 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+
+import argparse
+import json
+import platform
+import time
+from pathlib import Path
+
+from onnxruntime import __version__ as OrtVersion
+import onnxruntime_genai as og
+from packaging import version
+
+from olive.workflows import run as olive_run
+
+
+TARGETS = ["cpu", "gpu", "mobile", "web"]
+
+TARGET_TO_EP = {
+    "cpu": "CPUExecutionProvider",
+    "mobile": "CPUExecutionProvider",
+    "gpu": "CUDAExecutionProvider",
+    "web": "JsExecutionProvider",
+}
+
+
+def get_args(raw_args):
+    parser = argparse.ArgumentParser(description="phi3 optimization")
+
+    parser.add_argument(
+        "--target",
+        type=str,
+        default=None,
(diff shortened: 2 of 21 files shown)


olive/passes/onnx/float32_conversion.py
```diff
@@ -0,0 +1,126 @@
...
+                default_value="logits",
+                description=(
+                    "Only convert inputs/outputs whose name matches this pattern. By defaultlooking for logits names"
+                ),
+            )
+        }
+        config.update(get_external_data_config())
+        return config
+
+    def create_io_mapping(self, graph, i_map, o_map):
```
jambayk, line 40: `OnnxDAG` doesn't cover this use case fully but it can be extended to allow us this type of graph manipulation if needed. The logic here is okay for now but we could update this in the future to use onnx dag. it has an easily accessible graph representation so that we don't have to loop through the graph to make changes.

jambayk: added this pass to workitem 31579 tracking onnx dag use cases.

olive/passes/onnx/float32_conversion.py
```diff
@@ -0,0 +1,126 @@
...
+            logger.debug("input %s from fp32", i.name)
+            for n in i_map[i.name]:
+                for j, o in enumerate(n.input):
+                    if o == i.name:
+                        n.input[j] = i.name + "_fp16"
+            cast = onnx.helper.make_node(
+                "Cast",
+                inputs=[i.name],
+                outputs=[i.name + "_fp16"],
+                to=onnx.TensorProto.FLOAT,
```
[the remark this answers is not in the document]
guschmue: to fp32 - javascript doesn't know what fp16 is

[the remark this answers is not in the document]
jambayk: Since this is a cast in the input from `{i.name}` to `{i.name}_fp16`, the cast to `FLOAT` did not make sense to me. We are essentially casting from `float` to `float` with this node since `{i.name}` is changed to float in line 72.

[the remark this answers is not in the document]
jambayk: My understanding is we want: `fp32 input` -> `cast to fp16` -> `Graph` -> `cast to fp32` ->`fp32 input`. In which case this cast `to` must be `FLOAT16`?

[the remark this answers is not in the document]
devang-ml: yes

docs/source/api/passes.rst
```diff
@@ -42,6 +42,12 @@ OnnxFloatToFloat16
 --------------------
 .. autoconfigclass:: olive.passes.OnnxFloatToFloat16
 
+.. _onnx_io_float16_to_float32:
+
+OnnxIOFloat16ToFloat32
+--------------------
```
jambayk, line 48: Doc build test is complaining about this line being shorter than the title.

olive/passes/onnx/float32_conversion.py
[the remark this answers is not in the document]
xiaoyu-work: In this case. I think line 54 should select FLOAT rather FLOAT16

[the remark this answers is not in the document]
jambayk: line 54 is fine.

docs/source/api/passes.rst
devang-ml: Fixed

examples/phi3/phi3.py
jambayk, line 1: This commit is probably still needed? https://github.com/microsoft/Olive/pull/1149/commits/158469316655148d3720d316bfff8f5c5af76eef

xiaoyu-work: yeah forgot about this...

examples/phi3/phi3.py
```diff
@@ -0,0 +1,216 @@
...
+from pathlib import Path
+
+import onnxruntime_genai as og
+
+from olive.workflows import run as olive_run
+
+# flake8: noqa: T201
+
+
+TARGETS = ["cpu", "gpu", "mobile", "web"]
```
jambayk, line 18: the targets don't match the actual options. needs to be "cuda" instead of "gpu"

examples/phi3/phi3.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def get_args(raw_args):
+    parser = argparse.ArgumentParser(description="phi3 optimization")
+
+    parser.add_argument(
+        "--target",
+        type=str,
+        default=None,
+        choices=["cpu", "cuda", "mobile", "web"],
```
jambayk, line 35: replace the choices with `TARGETS` or `TARGETS` can be removed.

examples/phi3/phi3.py
```diff
@@ -0,0 +1,216 @@
...
+}
+
+
+def get_args(raw_args):
+    parser = argparse.ArgumentParser(description="phi3 optimization")
+
+    parser.add_argument(
+        "--target",
+        type=str,
+        default=None,
```
jambayk, line 34: `required=True` is better than having default None and checking in line 70

examples/phi3/phi3_template.json
```diff
@@ -0,0 +1,36 @@
...
+                    }
+                ]
+            }
+        }
+    },
+    "passes": {
+
+    },
+    "engine": {
+        "cache_dir": "cache",
```
jambayk, line 33: please add `"target": "local_system"`, otherwise, the system is not used at all and it will just loop through the installed eps.

examples/phi3/phi3.py
```diff
@@ -0,0 +1,216 @@
...
+        choices=["cpu", "cuda", "mobile", "web"],
+        help="Choose from cpu, cuda, mobile or web",
+    )
+    parser.add_argument(
+        "--precision",
+        type=str,
+        default=None,
```
jambayk, line 41: the default can be set to `int4` here instead of line 73

examples/phi3/phi3.py
xiaoyu-work: Good call! i didn't get a chance to check phi3 example code. I've addressed all comments.

examples/phi3/phi3.py
```diff
@@ -0,0 +1,212 @@
...
+
+# flake8: noqa: T201
+
+
+TARGETS = ["cpu", "cuda", "mobile", "web"]
+
+TARGET_TO_EP = {
+    "cpu": "CPUExecutionProvider",
+    "mobile": "CPUExecutionProvider",
+    "gpu": "CUDAExecutionProvider",
```
jambayk, line 23: this still says "gpu"

examples/phi3/phi3.py
```diff
@@ -0,0 +1,212 @@
...
+        default=200,
+        help="Max length for generation. Not supported with Web target.",
+    )
+
+    return parser.parse_args(raw_args)
+
+
+def main(raw_args=None):
+    args = get_args(raw_args)
+    elif args.target in ("mobile", "web") and args.precision != "int4":
```
jambayk, line 71: if

examples/phi3/phi3.py
xiaoyu-work: Fixed.

jambayk (approved):

PR #1151, opened 2024-05-13
Remove HFConfig::dataset and update docs accordingly

shaahji:
## Remove HFConfig::dataset and update docs accordingly
All data configs should now be defined under "data_configs" tag in config and referenced by name for use.


olive/model/handler/mixin/dummy_inputs.py
```diff
@@ -33,19 +34,15 @@ def get_dummy_dataloader_from_io_config(self):
...
-                self.hf_config.model_name,
-                self.hf_config.task,
-                **self.hf_config.dataset,
-            ).to_data_container()
+            logger.debug("Using data_config to build dummy dataloader")
+            dataloader = data_config.to_data_container()
         return dataloader
 
-    def get_dummy_inputs(self, filter_hook=None, filter_hook_kwargs=None):
+    def get_dummy_inputs(self, data_config=None, filter_hook=None, filter_hook_kwargs=None):
```
jambayk, line 45: is there a use case for the `data_configs` option? I think it's safe to remove this code path entirely. We could just remove the `get_dummy_dataloader_from_hf_config` entirely.

shaahji: Done!

olive/model/handler/mixin/dummy_inputs.py
```diff
@@ -59,7 +47,7 @@ def get_dummy_inputs(self, filter_hook=None, filter_hook_kwargs=None):
             dummy_inputs = user_module_loader.call_object(self.dummy_inputs_func, self)
             # respect user's dummy_inputs_func, no hook
         else:
-            dataloader = self.get_dummy_dataloader_from_io_config() or self.get_dummy_dataloader_from_hf_config()
+            dataloader = self._get_dummy_dataloader_from_io_config()
             if dataloader:
                 dummy_inputs, _ = dataloader.get_first_batch()
             elif not self.hf_config.components:
```
jambayk, line 53: This is not caused by your PR but could you update this condition to `elif self.hf_config and not self.hf_config.components and self.hf_config.task`? This PR https://github.com/microsoft/Olive/pull/1121 had not added the full conditions for this case.

shaahji: Done!

jambayk (approved):

PR #1155, opened 2024-05-14
Fixing a few issues reported by newest version of pylint

shaahji:

### DIFF
--- a/olive/passes/onnx/vitis_ai_quantization.py
+++ b/olive/passes/onnx/vitis_ai_quantization.py
@@ -339,6 +339,7 @@ class VitisAIQuantization(Pass):
 
         # get the dataloader
         # TODO(XiaoSheng): only use data config
+        dataloader = None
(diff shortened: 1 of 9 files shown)


olive/passes/onnx/vitis_ai_quantization.py
```diff
@@ -339,6 +339,7 @@ def _run_for_config(
 
         # get the dataloader
         # TODO(XiaoSheng): only use data config
+        dataloader = None
```
jambayk, line 342: for this and openvino quantization, I think doing an else condition and raising `ValueError("dataloader_func or data_config is required") is better but we also don't do that in other places https://github.com/microsoft/Olive/blob/22f41383bbe35b246636c48cb59c85cbcec3a945/olive/passes/onnx/quantization.py#L233 

We could update this later so this is fine.

jambayk (approved):

PR #1159, opened 2024-05-15
fix dtype of config.head_dim to int for directml llm example

thevishalagarwal:
fix `config.head_dim` to `int` type. The changes in #1138 made the head_dim param to type `float`.

### DIFF
--- a/examples/directml/llm/llm.py
+++ b/examples/directml/llm/llm.py
@@ -52,7 +52,7 @@ def set_config_parameters(tokenizer: transformers.AutoTokenizer, repo_id: str, n
 
     config.hidden_size = llm_model.config.hidden_size
     config.num_heads = llm_model.config.num_attention_heads
-    config.head_dim = getattr(llm_model.config, "head_dim", config.hidden_size / config.num_heads)
+    config.head_dim = getattr(llm_model.config, "head_dim", config.hidden_size // config.num_heads)
     config.num_layers = num_layers or llm_model.config.num_hidden_layers
     config.vocab_size = llm_model.config.vocab_size
     config.model_type = main_model.config.model_type


thevishalagarwal:
@PatriceVignola Can you please review?

jambayk (approved):

examples/directml/llm/llm.py
```diff
@@ -52,7 +52,7 @@ def set_config_parameters(tokenizer: transformers.AutoTokenizer, repo_id: str, n
 
     config.hidden_size = llm_model.config.hidden_size
     config.num_heads = llm_model.config.num_attention_heads
-    config.head_dim = getattr(llm_model.config, "head_dim", config.hidden_size / config.num_heads)
+    config.head_dim = getattr(llm_model.config, "head_dim", config.hidden_size // config.num_heads)
```
jambayk, line 55: thanks for the change. did you hit this error when testing with a model?

thevishalagarwal: Yes, I'm getting the following error trace. This is because the config.head_dim is of type `float` and it is expected to be `int`

```sh
decoder_model.py 244 __init__
self.q_proj = torch.nn.Linear(config.hidden_size, config.num_heads * config.head_dim, bias=config.use_bias)

linear.py 98 __init__
self.weight = Parameter(torch.empty((out_features, in_features), **factory_kwargs))

TypeError:
empty() received an invalid combination of arguments - got (tuple, dtype=NoneType, device=NoneType), but expected one of:
 * (tuple of ints size, *, tuple of names names, torch.memory_format memory_format, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
 * (tuple of ints size, *, torch.memory_format memory_format, Tensor out, torch.dtype dtype, torch.layout layout, torch.device device, bool pin_memory, bool requires_grad)
 ```

jambayk: Thanks! that makes sense. for gemma model, it's model config has `head_dim` attribute but it's not there for phi3, etc.

jambayk (approved):

PatriceVignola:
Oh my bad, I missed this PR. Changes look good.

PR #1162, opened 2024-05-15
[DO NOT MERGE] enable fixed fp16 vae with dml

jambayk:


jambayk:
@thevishalagarwal I haven't tried it with dml since I did not have a suitable DML compatible machine back when I made the changes. I don't think @PatriceVignola got the chance to try it out either? 

However, I verified it works with CUDA ep so I expect it to probably work with DML ep too.

PR #1166, opened 2024-05-20
Add cloud model cache

xiaoyu-work:
## Describe your changes

Add cloud model cache

### DIFF
--- a/olive/common/utils.py
+++ b/olive/common/utils.py
@@ -417,3 +417,28 @@ def find_first_matched_value(original_dict: Dict, keys: Union[str, Tuple, List[s
     if raise_key_error:
         raise KeyError(f"Keys {keys} not found in {original_dict}")
     return None
+
+
+def get_credentials(default_auth_params: Dict = None):
+    """Get credentials for MLClient.
+
+    Order of credential providers:
+    1. Azure CLI
+    2. DefaultAzureCredential
+    3. InteractiveBrowserCredential
+    """
+    from azure.identity import DefaultAzureCredential, InteractiveBrowserCredential
+
+    logger.debug("Getting credentials for MLClient")
+    try:
+        default_auth_params = default_auth_params or {}
+        credential = DefaultAzureCredential(**default_auth_params)
+        # Check if given credential can get token successfully.
+        credential.get_token("https://management.azure.com/.default")
+        logger.debug("Using DefaultAzureCredential")
+    except Exception:
+        logger.warning("Using InteractiveBrowserCredential since of default credential errors", exc_info=True)
+        # Fall back to InteractiveBrowserCredential in case DefaultAzureCredential not work
+        credential = InteractiveBrowserCredential()
+
+    return credential
--- a/olive/engine/cloud_cache_helper.py
+++ b/olive/engine/cloud_cache_helper.py
@@ -2,12 +2,12 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
-from enum import Enum
 import io
 import json
 import logging
 import os
 from copy import deepcopy
+from enum import Enum
 from pathlib import Path
 from typing import Any, Dict
 
@@ -32,6 +32,7 @@ class CloudModelInvalidStatus(str, Enum):
     NO_MODEL_FILE = "NO_MODEL_FILE"
     FAILED_CONFIG = "FAILED_CONFIG"
 
+
 class CloudCacheHelper:
     def __init__(
         self,
@@ -43,10 +44,10 @@ class CloudCacheHelper:
     ):
         try:
             from azure.storage.blob import ContainerClient
-        except ImportError:
+        except ImportError as exc:
             raise ImportError(
                 "Please install azure-storage-blob and azure-identity to use the cloud model cache feature."
-            )
+            ) from exc
         credential = get_credentials()
         self.container_name = container_name
         self.container_client = ContainerClient(
@@ -63,14 +64,14 @@ class CloudCacheHelper:
         self.hf_cache_path = hf_cache_path
 
     def update_model_config(self, cloud_model_path, model_config, input_model_hash):
-        logger.info(f"Updating model config with cloud model path: {cloud_model_path}")
+        logger.info("Updating model config with cloud model path: %s", cloud_model_path)
         model_directory_prefix = f"{input_model_hash}/model"
         blob_list = self.container_client.list_blobs(name_starts_with=model_directory_prefix)
 
         for blob in blob_list:
             blob_client = self.container_client.get_blob_client(blob)
             local_file_path = self.output_model_path / blob.name[len(model_directory_prefix) + 1 :]
-            logger.info(f"Downloading {blob.name} to {local_file_path}")
+            logger.info("Downloading %s to %s", blob.name, local_file_path)
             local_file_path.parent.mkdir(parents=True, exist_ok=True)
             with open(local_file_path, "wb") as download_file:
                 download_stream = blob_client.download_blob()
@@ -81,7 +82,7 @@ class CloudCacheHelper:
         return model_config
 
     def download_model_cache_map(self):
-        logger.info(f"Downloading model cache map from {self.map_blob}")
+        logger.info("Downloading model cache map from %s", self.map_blob)
         blob_client = self.container_client.get_blob_client(self.map_blob)
 
         if not blob_client.exists():
@@ -95,9 +96,7 @@ class CloudCacheHelper:
             download_file.write(download_stream.readall())
 
         with open(self.local_map_path) as file:
-            map_dict = json.load(file)
-            
-        return map_dict
+            return json.load(file)
 
     def get_hash_key(self, model_config: ModelConfig, pass_search_point: Dict[str, Any], input_model_hash: str):
         hf_hub_model_commit_id = None
@@ -111,7 +110,7 @@ class CloudCacheHelper:
                 model_name = f"models--{model_config_copy.config['hf_config']['model_name'].replace('/', '--')}"
                 commit_path = model_cache_dir / model_name / "refs" / "main"
                 if not commit_path.exists():
-                    raise Exception(
+                    raise ValueError(
                         "Huggingface cache not found. Please specify 'hf_cache_path' in 'enable_cloud_cache' config."
                     )
                 with open(commit_path) as file:
@@ -143,7 +142,7 @@ class CloudCacheHelper:
         model_config_path = model_config_path / "model_config.json"
 
         blob_client = self.container_client.get_blob_client(model_config_blob)
-        logger.info(f"Downloading {model_config_blob} to {model_config_path}")
+        logger.info("Downloading %s to %s", model_config_blob, model_config_path)
 
         with open(model_config_path, "wb") as download_file:
             download_stream = blob_client.download_blob()
@@ -209,10 +208,10 @@ def check_model_cache(cloud_cache_helper: CloudCacheHelper, input_model_config, 
     output_model_config = None
     logger.info("Cloud model cache is enabled. Check cloud model cache ...")
 
-    output_model_hash = cloud_cache_helper.get_hash_key(
-        input_model_config, pass_search_point, input_model_hash
+    output_model_hash = cloud_cache_helper.get_hash_key(input_model_config, pass_search_point, input_model_hash)
+    cloud_model_path = cloud_cache_helper.exist_in_model_cache_map(
+        cloud_cache_helper.cloud_cache_map, output_model_hash
     )
-    cloud_model_path = cloud_cache_helper.exist_in_model_cache_map(cloud_cache_helper.cloud_cache_map, output_model_hash)
     if cloud_model_path is not None:
         if cloud_model_path == CloudModelInvalidStatus.FAILED_CONFIG:
             logger.info("Model is pruned in cloud cache.")
@@ -226,20 +225,20 @@ def check_model_cache(cloud_cache_helper: CloudCacheHelper, input_model_config, 
             pass_run_locally = False
     return pass_run_locally, output_model_config
 
+
 def update_input_model_config(cloud_cache_helper: CloudCacheHelper, input_model_config, input_model_hash):
     # download model files
     logger.info("Cloud model cache is enabled. Downloading input model files ...")
     cloud_model_path = cloud_cache_helper.exist_in_model_cache_map(cloud_cache_helper.cloud_cache_map, input_model_hash)
     if cloud_model_path is not None:
         cloud_cache_helper.update_model_config(cloud_model_path, input_model_config, input_model_hash)
-        
-def upload_model_to_cloud(cloud_cache_helper: CloudCacheHelper, input_model_config, pass_search_point, input_model_hash, output_model_config):
-    output_model_hash = cloud_cache_helper.get_hash_key(
-        input_model_config, pass_search_point, input_model_hash
-    )
+
+
+def upload_model_to_cloud(
+    cloud_cache_helper: CloudCacheHelper, input_model_config, pass_search_point, input_model_hash, output_model_config
+):
+    output_model_hash = cloud_cache_helper.get_hash_key(input_model_config, pass_search_point, input_model_hash)
     cloud_model_path = "FAILED_CONFIG"
     if output_model_config not in PRUNED_CONFIGS:
-        cloud_model_path = cloud_cache_helper.upload_model_to_cloud_cache(
-            output_model_hash, output_model_config
-        )
+        cloud_model_path = cloud_cache_helper.upload_model_to_cloud_cache(output_model_hash, output_model_config)
     cloud_cache_helper.update_model_cache_map(output_model_hash, cloud_model_path)
--- a/olive/engine/cloud_cache_helper.py
+++ b/olive/engine/cloud_cache_helper.py
@@ -152,7 +152,6 @@ class CloudCacheHelper:
 
 
 def check_model_cache(cloud_cache_helper: CloudCacheHelper, input_model_config, pass_search_point, input_model_hash):
-    pass_run_locally = True
     output_model_config = None
     logger.info("Cloud model cache is enabled. Check cloud model cache ...")
 
@@ -160,8 +159,7 @@ def check_model_cache(cloud_cache_helper: CloudCacheHelper, input_model_config, 
     if cloud_cache_helper.exist_in_cloud_cache(output_model_hash):
         logger.info("Model is found in cloud cache.")
         output_model_config = cloud_cache_helper.get_model_config_by_hash_key(output_model_hash)
-        pass_run_locally = False
-    return pass_run_locally, output_model_config
+    return output_model_config
 
 
 def update_input_model_config(cloud_cache_helper: CloudCacheHelper, input_model_config, input_model_hash):
--- a/olive/engine/cloud_cache_helper.py
+++ b/olive/engine/cloud_cache_helper.py
@@ -1,0 +1,245 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+from enum import Enum
+import io
+import json
+import logging
+import os
+from copy import deepcopy
+from pathlib import Path
+from typing import Any, Dict
+
+from olive.common.config_utils import ConfigBase
+from olive.common.utils import get_credentials, hash_dict
+from olive.engine.config import FAILED_CONFIG, PRUNED_CONFIGS
+from olive.model.config.model_config import ModelConfig
+
+logger = logging.getLogger(__name__)
+
+
+class CloudCacheConfig(ConfigBase):
+    enable_cloud_cache: bool = True
+    account_url: str = "https://olivepublicmodels.blob.core.windows.net"
+    contaier_name: str = "olivecachemodels"
+    upload_to_cloud: bool = True
+    input_model_config: ModelConfig = None
+    hf_cache_path: str = os.environ.get("HF_HOME", None) or str(Path("~/.cache/huggingface").expanduser())
+
+
+class CloudModelInvalidStatus(str, Enum):
+    NO_MODEL_FILE = "NO_MODEL_FILE"
+    FAILED_CONFIG = "FAILED_CONFIG"
+
+class CloudCacheHelper:
+    def __init__(
+        self,
+        cache_dir: str,
+        account_url: str,
+        container_name: str,
+        hf_cache_path: str,
+        input_model_config: ModelConfig = None,
+    ):
+        try:
+            from azure.storage.blob import ContainerClient
+        except ImportError:
+            raise ImportError(
+                "Please install azure-storage-blob and azure-identity to use the cloud model cache feature."
+            )
+        credential = get_credentials()
+        self.container_name = container_name
+        self.container_client = ContainerClient(
+            account_url=account_url, container_name=self.container_name, credential=credential
+        )
+
+        self.map_blob = "model_map.json"
+        self.local_map_path = Path(cache_dir) / "model_map.json"
+        self.local_map_path.parent.mkdir(parents=True, exist_ok=True)
+        self.cache_dir = cache_dir
+        self.output_model_path = Path(cache_dir) / "model"
+        self.cloud_cache_map = self.download_model_cache_map()
+        self.input_model_config = input_model_config
+        self.hf_cache_path = hf_cache_path
+
+    def update_model_config(self, cloud_model_path, model_config, input_model_hash):
+        logger.info(f"Updating model config with cloud model path: {cloud_model_path}")
+        model_directory_prefix = f"{input_model_hash}/model"
+        blob_list = self.container_client.list_blobs(name_starts_with=model_directory_prefix)
+
+        for blob in blob_list:
+            blob_client = self.container_client.get_blob_client(blob)
+            local_file_path = self.output_model_path / blob.name[len(model_directory_prefix) + 1 :]
+            logger.info(f"Downloading {blob.name} to {local_file_path}")
+            local_file_path.parent.mkdir(parents=True, exist_ok=True)
+            with open(local_file_path, "wb") as download_file:
+                download_stream = blob_client.download_blob()
+                download_file.write(download_stream.readall())
+
+        model_config.config["model_path"] = str(self.output_model_path / cloud_model_path)
+
+        return model_config
+
+    def download_model_cache_map(self):
+        logger.info(f"Downloading model cache map from {self.map_blob}")
+        blob_client = self.container_client.get_blob_client(self.map_blob)
+
+        if not blob_client.exists():
+            return {}
+
+        self.local_map_path.unlink(missing_ok=True)
+        self.local_map_path.touch()
+
+        with open(self.local_map_path, "wb") as download_file:
+            download_stream = blob_client.download_blob()
+            download_file.write(download_stream.readall())
+
+        with open(self.local_map_path) as file:
+            map_dict = json.load(file)
+            
+        return map_dict
+
+    def get_hash_key(self, model_config: ModelConfig, pass_search_point: Dict[str, Any], input_model_hash: str):
+        hf_hub_model_commit_id = None
+        if input_model_hash is None:
+            model_config_copy = self.input_model_config
+            model_path = model_config_copy.config.get("model_path")
+            if model_path is None and model_config_copy.config.get("hf_config") is not None:
+                # hf model, get model commit from hf cache
+                cache_dir = Path(self.hf_cache_path)
+                model_cache_dir = Path(cache_dir) / "hub"
+                model_name = f"models--{model_config_copy.config['hf_config']['model_name'].replace('/', '--')}"
+                commit_path = model_cache_dir / model_name / "refs" / "main"
+                if not commit_path.exists():
+                    raise Exception(
+                        "Huggingface cache not found. Please specify 'hf_cache_path' in 'enable_cloud_cache' config."
+                    )
+                with open(commit_path) as file:
+                    hf_hub_model_commit_id = file.read().strip()
+        else:
+            model_config_copy = deepcopy(model_config)
+            if model_config_copy.config.get("model_path") is not None:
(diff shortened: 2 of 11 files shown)


guotuofeng:
do we need unify the local cache and cloud cache by an abstracted cache interface and both local cache and cloud cache implement the interface?

xiaoyu-work:
> do we need unify the local cache and cloud cache by an abstracted cache interface and both local cache and cloud cache implement the interface?

Good call. Though I don't see too many overlaps between local cache and cloud cache, I will look into this. It would be better to unify them.

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,244 @@
...
+from olive.common.utils import get_credentials, hash_dict
+from olive.engine.config import FAILED_CONFIG, PRUNED_CONFIGS
+from olive.model.config.model_config import ModelConfig
+
+logger = logging.getLogger(__name__)
+
+
+class CloudCacheConfig(ConfigBase):
+    enable_cloud_cache: bool = True
+    account_url: str = "https://olivepublicmodels.blob.core.windows.net"
```
trajepl, line 24: Should we remove this and keep it as required by default?

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,244 @@
...
+    contaier_name: str = "olivecachemodels"
```
trajepl, line 25: ditto.

olive/engine/cloud_cache_helper.py
xiaoyu-work: No, the common use case is all users will share a same model storage account. I make this configurable just in case there is any custom use case.

xiaoyu-work:
> do we need unify the local cache and cloud cache by an abstracted cache interface and both local cache and cloud cache implement the interface?

Cloud model cache does not share too many common logics as local cache. I will just keep this as it is and revisit this later.

olive/common/utils.py
```diff
@@ -415,3 +415,28 @@ def find_first_matched_value(original_dict: Dict, keys: Union[str, Tuple, List[s
...
+    try:
+        default_auth_params = default_auth_params or {}
+        credential = DefaultAzureCredential(**default_auth_params)
+        # Check if given credential can get token successfully.
+        credential.get_token("https://management.azure.com/.default")
+        logger.debug("Using DefaultAzureCredential")
+    except Exception:
+        logger.warning("Using InteractiveBrowserCredential since of default credential errors", exc_info=True)
+        # Fall back to InteractiveBrowserCredential in case DefaultAzureCredential not work
+        credential = InteractiveBrowserCredential()
```
shaahji, line 440: `InteractiveBrowserCredential` is only valid if launched in interactive mode. On CI this will result in infinite wait. This should be wrapped in a condition of whether the launch is interactive or not. In other cases, this should just fail.

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,244 @@
...
+        blob_client = self.container_client.get_blob_client(self.map_blob)
+
+        if not blob_client.exists():
+            return {}
+
+        self.local_map_path.unlink(missing_ok=True)
+        self.local_map_path.touch()
+
+        with open(self.local_map_path, "wb") as download_file:
+            download_stream = blob_client.download_blob()
```
shaahji, line 95: Should `download_stream` use `with` to support auto close when done? It might probably close automatically when reaching EOF but being explicit might help.

shaahji (changes requested):
Versioning - Config files are being uploaded to cache without versioning. So, what happens if a new *required* key is added to the config? Would it fail with unexpected exceptions which the user can't act on or would the logic recover gracefully?

Telemetry - How often is any particular entry in cache accessed. Not that it should be part of this PR but something to consider. How do we differentiate between merely checking the status of something in cache vs. actually being used by someone during their pipeline run. For the latter, it would require additional calls from Olive to record such use.

olive/common/utils.py
shaahji: For automation purposes, we could consider using some environment variable that is available only on CI. This would be default and can be overriden by a command line argument.

xiaoyu-work: How does current pipeline work with `InteractiveBrowserCredential()`? I assume it will timeout and fall back to next credential.

olive/engine/cloud_cache_helper.py
xiaoyu-work: I was following the pattern from documentation here: https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blob-download-python#download-to-a-file-path

xiaoyu-work:
> Versioning - Config files are being uploaded to cache without versioning. So, what happens if a new _required_ key is added to the config? Would it fail with unexpected exceptions which the user can't act on or would the logic recover gracefully?
> 
> Telemetry - How often is any particular entry in cache accessed. Not that it should be part of this PR but something to consider. How do we differentiate between merely checking the status of something in cache vs. actually being used by someone during their pipeline run. For the latter, it would require additional calls from Olive to record such use.

Good call! How often do you think we will update a required attribute in model config? Any case in your mind now? If it happens rarely, can we manually clean all the cached models?

olive/engine/cloud_cache_helper.py
shaahji: Hmm, I guess `download_blob` doesn't actually return a stream - the returned object actually wraps multiple 'get data in range' calls. You could, however, replace the two calls with one [readinto](https://learn.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.storagestreamdownloader?view=azure-python#azure-storage-blob-storagestreamdownloader-readinto) call

olive/common/utils.py
shaahji: Apparently, as per the documentation [here](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.interactivebrowsercredential?view=azure-dotnet) there is no timeout or anyway to cancel it unless the request is launched as an async job. I guess, it's expected as the name says _interactive_. If there are other places using the same/similar calls, they are all wrong. Those calls will wait indefinitely for user action.

shaahji:
> Good call! How often do you think we will update a required attribute in model config? Any case in your mind now? If it happens rarely, can we manually clean all the cached models?

Your guess is just as good as mine. I am introducing the `generative` flag on `input_model` as part of my current task.

Even if it happens extremely rare, invalidating existing cache might not be a very user-friendly solution, especially, given the expense involved in training/optimizing a model. I don't have a solution and maybe it; shouldn't even be part of this PR but it surely needs attention/discussion. If we fill up the cache with data tomorrow only to throw it away the next day, we are not realizing the true potential of the cache. I think at the very least a follow up PR should add a version to the config to quickly check if the version in cache and the running versions are the same.

On the telemetry front, again, needs further design and discussion. This can wait until we solidify how the cache is/will be used. The data from Azure telemetry alone might be enough or we might need to add explicit calls in Olive to record the telemetry.

xiaoyu-work:
> > Good call! How often do you think we will update a required attribute in model config? Any case in your mind now? If it happens rarely, can we manually clean all the cached models?
> 
> Your guess is just as good as mine. I am introducing the `generative` flag on `input_model` as part of my current task.
> 
> Even if it happens extremely rare, invalidating existing cache might not be a very user-friendly solution, especially, given the expense involved in training/optimizing a model. I don't have a solution and maybe it; shouldn't even be part of this PR but it surely needs attention/discussion. If we fill up the cache with data tomorrow only to throw it away the next day, we are not realizing the true potential of the cache. I think at the very least a follow up PR should add a version to the config to quickly check if the version in cache and the running versions are the same.
> 
> On the telemetry front, again, needs further design and discussion. This can wait until we solidify how the cache is/will be used. The data from Azure telemetry alone might be enough or we might need to add explicit calls in Olive to record the telemetry.

I just realized that the model config is only used as a hash key here. So if the model config changed, it will get a new hash key which leads to a new cache. Does this sound good to you?

olive/common/utils.py
shaahji: Please add a task in our backlog to revisit the use of `InteractiveBrowserCredentials`. I don't think its use in the codebase is correct. Resolve the conversation with link to the task.

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,244 @@
...
+        hf_hub_model_commit_id = None
+        if input_model_hash is None:
+            model_config_copy = self.input_model_config
+            model_path = model_config_copy.config.get("model_path")
+            if model_path is None and model_config_copy.config.get("hf_config") is not None:
+                # hf model, get model commit from hf cache
+                cache_dir = Path(self.hf_cache_path)
+                model_cache_dir = Path(cache_dir) / "hub"
+                model_name = f"models--{model_config_copy.config['hf_config']['model_name'].replace('/', '--')}"
+                commit_path = model_cache_dir / model_name / "refs" / "main"
```
shaahji, line 111: Seems like this logic basing decisions on undocumented internals of how hfhub works. Is there no documented API to get some of this information? This can likely change tomorrow behind our back and break entire caching logic. On the other end, is this path even required to be part of the hash? There should be an explicit _commit\_id_ or something that hfhub uses.

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,244 @@
...
+
+        # model config
+        with open(model_config_path) as file:
+            model_config_dict = json.load(file)
+            return ModelConfig.from_json(model_config_dict)
+
+    def update_model_cache_map(self, output_model_hash, cloud_model_path):
+        logger.info("Updating model cache map.")
+        if not self.local_map_path.exists():
+            self.local_map_path.touch()
```
shaahji, line 159: This is not required. Very next statement is opening the same file for write.

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,244 @@
...
+            with open(self.local_map_path, "w") as file:
+                file.write(b"{}")
+        with open(self.local_map_path, "r+b") as file:
+            self.cloud_cache_map = json.load(file)
+        self.cloud_cache_map[output_model_hash] = cloud_model_path
```
shaahji, line 164: The strategy here is seriously flawed. Let's discuss this offline.

shaahji (changes requested):

olive/common/utils.py
xiaoyu-work: Done https://aiinfra.visualstudio.com/Model%20optimization%20Toolkit/_workitems/edit/38586

olive/engine/cloud_cache_helper.py
xiaoyu-work: There is no documentation about this. and yes this might be fragile but I don't think this is a normal use case and they won't have an API for this. In case it breaks the logic, I can add a fall-back logic here to skip caching if it breaks.

You are right this path is the commit-id they use. This `main` file just includes a commit id.

olive/engine/cloud_cache_helper.py
xiaoyu-work: Good call. Will update this.

olive/engine/cloud_cache_helper.py
shaahji: As I understand it, there is a single file to track the status of the cache i.e. a map of what’s in the cloud cache. This has a major flaw.
Consider two jobs – A & B, and t0 < t1 < t2 < t3
Job A starts at time t0 – downloads the current status of the cache
Job B start at time t1 - downloads the current status of the cache
Job B finishes at time t2 – updates the map, and uploads the new status of the cache
Job A finishes at time t3 – updates the map, and uploads the new status of the cache
NOTE: at t3, the record for Job B’s update is lost.

This is a simple illustration of the problem but when considering we could have large number of jobs running parallel, the scenario could be worse. A long running job will end up overwriting entries for all smaller jobs.
This can easily be fixed by eliminating the need for the map that tracks the cache status. Instead, each job creates its own record in the cloud using the hash which is already being computed. So, we treat the cloud cache as a directory rather than a file. There might be a few limitations to this strategy –
* Azure might have limitation on file names
  * Hash must be alphanumeric only
  * Length of the hash as string

To check whether a cache entry exists, we just need to compute the expected hash and look for the file itself. The file will contain the json object that was used to generate the hash.

olive/engine/cloud_cache_helper.py
shaahji: https://github.com/huggingface/huggingface_hub/issues/2045#issuecomment-1963787595
```
>>> from huggingface_hub import model_info

>>> model_info("google/gemma-7b-it").sha
'bee54ce1bd2a5b36d68a06857fc0871a3389ca0b'

>>> model_info("google/gemma-7b-it", revision="flax").sha
'90fe4a773beffece715c57300f9f2c31b230724b'

>>> model_info("google/gemma-7b-it", revision="float16").sha
'cda5a5189d4606b969ca2b316298debf5fc2953d'
```

shaahji: Or even better, https://github.com/huggingface/huggingface_hub/issues/2045#issuecomment-1964137527

xiaoyu-work: Good to know this! Not sure if the local huggingface api will always use the latest commit from hub. Will it check the commit id difference between local and hub. But I assume this should be always true if the we don't specify the commit info or version info.

xiaoyu-work: Seems this is not true. I have tested to load `Intel/bert-base-uncased-mrpc` model (from my local cache), however, the commit id is not updated to the latest one: https://huggingface.co/Intel/bert-base-uncased-mrpc/commit/844e804944328e45c3f3bf69b35467e0ce0b26af. Can you double check this?

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,191 @@
...
+        hf_hub_model_commit_id = None
+        model_config_copy = deepcopy(model_config)
+        if input_model_hash is None:
+            from huggingface_hub import repo_info
+
+            if model_config.has_hf_config():
+                hf_hub_model_commit_id = repo_info(model_config.get_hf_model_name()).sha
+        else:
+            if model_config_copy.config.get("model_path") is not None:
+                model_config_copy.config["model_path"] = None
```
shaahji, line 82: The assignment of `model_path` should be unconditional otherwise you will end up with two different hashes for same model - one with `model_path: None` and another with no `model_path` which but both really mean the same for this practical use case.

Ideally, I would just replace that whole else part with `model_config_copy.config.pop("model_path", None)` so the `model_path` is never taken into account for computing the hash.

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,191 @@
...
+        with io.BytesIO(model_config_bytes) as data:
+            self.container_client.upload_blob(f"{output_model_hash}/model_config.json", data=data, overwrite=False)
+
+        # upload model file
+        model_blob = str(Path(output_model_hash) / "model")
+
+        with tempfile.TemporaryDirectory() as temp_dir:
+            model_path_file = Path(temp_dir) / "model_path.json"
+            with open(model_path_file, "w") as file:
+                file.write({"model_path": model_path.name})
```
shaahji, line 137: Isn't `model_path` local file/dir path here? Why record a local file/dir path? If this is just the name of the file/folder, why not directly in the path computed on line 139 below? Why bother making a separate file merely to store the model name?

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,191 @@
...
+            self._upload_dir_to_blob(model_path, f"{model_blob}")
+
+    def _upload_dir_to_blob(self, dir_path, blob_folder_name):
+        for item in dir_path.iterdir():
+            if item.is_dir():
+                self._upload_dir_to_blob(item, f"{blob_folder_name}/{item.name}")
+            else:
+                blob_name = f"{blob_folder_name}/{item.name}"
+                with open(item, "rb") as data:
+                    self.container_client.upload_blob(name=blob_name, data=data, overwrite=False)
```
shaahji, line 154: nit: Good opportunity to parallelize this with multiprocessing. Some of these files can be very large and uploading them to remote can be time consuming. Uploading them in parallel will be a low hanging fruit to optimize it.

olive/engine/engine.py
```diff
@@ -224,6 +235,7 @@ def run(
         output_dir: str = None,
         output_name: str = None,
         evaluate_input_model: bool = True,
+        enable_cloud_cache: "CloudCacheConfig" = None,
```
shaahji, line 238: Is this variable named correctly? Should this be `cloud_cache_config`? `enable_cloud_cache` is member of the config. Is the type specifier wrong? Should that be bool?

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,191 @@
...
+    pass_run_locally = True
+    output_model_config = None
+    logger.info("Cloud model cache is enabled. Check cloud model cache ...")
+
+    output_model_hash = cloud_cache_helper.get_hash_key(input_model_config, pass_search_point, input_model_hash)
+    if cloud_cache_helper.exist_in_cloud_cache(output_model_hash):
+        logger.info("Model is found in cloud cache.")
+        output_model_config = cloud_cache_helper.get_model_config_by_hash_key(output_model_hash)
+        pass_run_locally = False
+    return pass_run_locally, output_model_config
```
shaahji, line 167: nit: Unnecessary return value `pass_run_locally`. Just return `None` for `output_model_config`. Two variables create ambiguity where what if `output_model_config` is not None but `pass_run_locally` is still true.

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,191 @@
...
+
+from olive.common.config_utils import ConfigBase
+from olive.common.utils import get_credentials, hash_dict
+from olive.model.config.model_config import ModelConfig
+
+logger = logging.getLogger(__name__)
+
+
+class CloudCacheConfig(ConfigBase):
+    enable_cloud_cache: bool = True
```
shaahji, line 21: I would recommend splitting this flag into two - `enable_read` and `enable_write`. This gives better control to user when they want to use the cache but not contribute to it and vice-versa.

This can especially be useful for developers who are doing trial and error runs where they know the output might be not useful.

shaahji (changes requested):

olive/engine/cloud_cache_helper.py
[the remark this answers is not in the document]
xiaoyu-work: I think this is not true. The initial idea about this "cloud cache" is meant to be using Azure. So `cloud` also means sharable between different teams (or users). If not using Azure, that might be a new type of cache and can be called "remote".

olive/engine/cloud_cache_helper.py
xiaoyu-work: This sounds good to me. Will update this.

olive/engine/cloud_cache_helper.py
xiaoyu-work: Sounds good to me. I can improve this later in a different PR.

olive/engine/cloud_cache_helper.py
xiaoyu-work: No matter which format that the model file is, a file or folder, it can have different names, model file name or folder name. Olive doesn't have a standard to unify all output model names. So this file is only used to track model name (file name or folder name). When downloading the model from cloud cache, this name will be updated in the model config.

olive/engine/engine.py
xiaoyu-work: I intentionally to make this config can be both `bool` and `CloudCacheConfig`. If it is a `bool`, it will use all default configurations (one azure storage for all users). If they want to use their own storage, they can use `CloudCacheConfig` to configure the details. Otherwise those who want to use default cloud cache will have to write unnecessary configs.

olive/engine/cloud_cache_helper.py
xiaoyu-work: Sounds good to me. Will update this.

olive/engine/cloud_cache_helper.py
xiaoyu-work: Do you mean `upload_to_cloud` flag?

olive/engine/cloud_cache_helper.py
xiaoyu-work: https://aiinfra.visualstudio.com/Model%20optimization%20Toolkit/_workitems/edit/39356

olive/engine/cloud_cache_helper.py
shaahji: Yes, something along the lines of `enable_upload_to_cloud` and `enable_download_from_cloud`
or `enable_cloud_cache_read` and `enable_cloud_cache_write`

olive/engine/cloud_cache_helper.py
shaahji: That's engine's logic to decide. I don't see a reason for this function to return the same result in two different logical ways. If engine wants some way to override the result from here it still can do it.

```python
pass_run_locally = not check_model_cache( .... )
# Later down in the logic
if some_other_condition:
  pass_run_locally = False
```

Basically, the caller already has all the information with just returning the `output_model_config`.

```python
output_model_config == (pass_run_locally == False)
(not output_model_config) == (pass_run_locally == True)
```

olive/engine/cloud_cache_helper.py
shaahji: As I understand it, the cloud is already mimicking the local file/folder path and names. So, I am not seeing why we need another map to track the same information. Can't we just ask the cloud itself to iterate a path and get the required information? 

In fact, the map isn't even complete because you are storing only the root folder path in the map and not all the files within the folder. In cloud_cache_helper.py:151 you call `upload_blob` but that file is not registered in the model path.

olive/engine/cloud_cache_helper.py
xiaoyu-work: `enable_download_from_cloud` would be always true when you enabling cloud cache. And `enable_upload_to_cloud ` is same as `upload_to_cloud ` flag we already have.

olive/engine/cloud_cache_helper.py
xiaoyu-work: As for a model path field in Model Config, if it is a folder, it is a folder name. If it is a file, it is a file name. So I record its folder name or file name here. If using blob API to determine its path, it needs some additional steps like checking if it is a file or folder, what is its name something like this.

olive/engine/cloud_cache_helper.py
xiaoyu-work: Not sure how Huggingface cache works, but if you just directly load model from pretrained api in Python (without using Olive), seems it will not fetch the latest commit. But somehow if i run Olive, it can fetch the latest commit from hf hub.

olive/engine/cloud_cache_helper.py
xiaoyu-work: Sounds good to me. Will update this.

olive/engine/cloud_cache_helper.py
shaahji: Use cases -
* If I iterating on a workflow, I could use `enable_download_from_cloud` but disable `enable_upload_to_cloud` to avoid polluting the cache with bad results.
*  If I have already had a successful workflow post iterations with `enable_upload_to_cloud` disabled, I could run the workflow again with `enable_upload_to_cloud` enabled to push the results to cache so that recurring runs can use the cache. The workflow wouldn't do anything because it will use results from previous run but the results should now be uploaded and cached.

olive/engine/engine.py
shaahji: Should that decision be made here? Shouldn't that be part of the config validation so everything downstream from there will get the same input. Making the decision here between `bool` vs. `CloudCacheConfig` means every other system/code will have to also implement similar behavior which is unnecessary. We should have that decision be made once and, in my opinion, the config validation is the right place for it (similar logic exists for `DataConfig` also where the input could be either a `str` or a `DataConfig` object).

olive/engine/cloud_cache_helper.py
shaahji: If it is a folder, is there a need to record each file within the folder? If yes, then that's not happening.

In my experience using remote caches, having these secondary records (ones that try to mimic cache content) always cause more trouble than being useful. Especially, given the cost of asking the cache itself what it has is very small. These records tend to go out of sync and issues from it are very hard to debug. The model file might say one thing but the cache something entirely different. Consider the fact that remote caches are primarily asynchronous in nature and don't implement an exclusive lock. - any number of paralleling running pipelines could potentially be updating the records here. There is absolutely no guarantee that two workflows might not try to update the same cache at the same time. You might end up with a scenario where half the files are from one workflow and other half from another.

In short, I would recommend not depending on the model path and using the remote API. 

Another improvement to consider - Upload the output of the workflow (all results including the model_path file into a temporary remote folder and then renaming it. If you use a unique identifier in the temporary path that should avoid the potential collision (and overwrites) from other running parallel workflows.

olive/engine/cloud_cache_helper.py
shaahji: Yes, look like the behavior is inconsistent between APIs. Olive uses `AutoModel.from_pretrained` which passes `revision=None` as input to other APIs but `PretrainedModel.from_pretrained` passes `revision=main`. This difference is the source of the discrepancy. I don't think it is still wrong as long as the usage works consistently within Olive. If the commit hash that the API returns is the same as the model that is loaded, we are good. Whether the _newest_ model is downloaded or not is secondary problem outside the scope of this PR.

olive/engine/cloud_cache_helper.py
xiaoyu-work: If it is a folder, the model path is just the folder name, it doesn't need to record each file. Even we check the cache itself rather than recording the model path, it is still hard to address the paralleling running concern right? 

blob is a flat structure. There is no direct api to determine if it is a folder or file. There is also another case where we have an onnx model with external data, the model path should be the onnx model file path, while the model files could include onnx model file and external data file which is a folder. The iteration way to determine model path doesn't work here.

The model cache is based on Pass level but not workflow level. Do you mean output of pass?

olive/engine/engine.py
xiaoyu-work: It is validated by config. Please check line 89 in run/config.py file.

olive/engine/cloud_cache_helper.py
xiaoyu-work: For your use cases:

- User can enable cloud model cache with true flag in Engine, but set `upload_to_cloud` to false.
- This use case is not being considered when I implement this. Remote cache is mainly focused on downlaoding but not uploading, which means the cache workflow is like: run a pass -> check local cache -> if yes, continue workflow, if no, check cloud cache -> if neither, run pass locally. So we don't have a use that it exist in local cache but we want to update the cloud cache with local cache.

olive/engine/cloud_cache_helper.py
xiaoyu-work: Yeah the concern here is user loaded model locally commit id is inconsistent with remote API. That's the reason I would prefer to check local cache. Remote API can be inconsistent but local cache is always right.

shaahji: That shouldn't be inconsistent because both models are being loaded from Olive. The inconsistency is only if one end loads in Olive and the other outside Olive.

I still think using the available API is better than hardcoding some undocumented paths. I would fallback to this solution only if we have a demonstrable scenario where both ends mismatch. Alternatively, we could report this as a bug on HuggingFace repo.

olive/engine/cloud_cache_helper.py
shaahji: Let's run with the solution for now. We will revisit if we have specific scenarios where the model_path.json goes out of sync with cache.

olive/engine/cloud_cache_helper.py
shaahji: Let's reconsider this if/when we have a use case.

olive/engine/engine.py
shaahji: > If it is a bool, it will use all default configurations (one azure storage for all users).
> It is validated by config. Please check line 89 in run/config.py file.

I am confused now. In your previous statement you said it could be either/or. But if the `validate_config` is always returns an instance of `CloudCacheConfig` then it cannot be a `bool` as parameter to `Engine`. It will always be an instance of `CloudCacheConfig` or None. - in which case, the variable should be named `cloud_cache_config`.

The config accepting either `bool` or an instance of `CloudCacheConfig` is good. But, the engine should only care for it as an instance of `CloudCacheConfig` (or None) and never as `bool`.

xiaoyu-work: Yeah you are right i misunderstood you. For engine it is always `CloudCacheConfig`. I can rename this parameter.

olive/engine/cloud_cache_helper.py
xiaoyu-work: Yeah i have already made a change here to use HF api. The inconsistence comes from, when Olive loads model from local cache with old commit id, but we always call the remote API to fetch the latest commit id for hashing, turns out the hash key is wrong.

xiaoyu-work: Since as for my testing, seems Olive somehow will download and use the latest commit from HF, I'm good for this API now. But I still have concerns this could be potential issue and this would be hard to debug in the future if there was any issue.

olive/engine/cloud_cache_helper.py
```diff
@@ -0,0 +1,186 @@
...
+            return
+
+        model_path = Path(model_path)
+        model_config_copy = deepcopy(output_model_config)
+        model_config_copy.config["model_path"] = None
+
+        # upload model config file
+        model_config_bytes = json.dumps(model_config_copy.to_json()).encode()
+        with io.BytesIO(model_config_bytes) as data:
+            self.container_client.upload_blob(f"{output_model_hash}/model_config.json", data=data, overwrite=False)
```
shaahji, line 126: You might want to move this after all the files/folders are uploaded to avoid scenarios where the `model_config.json` is available in the cache but the necessary files are still uploading.

shaahji (approved):

olive/engine/cloud_cache_helper.py
xiaoyu-work: Yeah we probably should upload large files first. Updated.

shaahji (approved):

PR #1168, opened 2024-05-21
🇵🇭 phi3 finetune+genai

trajepl:
## Describe your changes

Existing genai example only cover the case of export model from official huggingface hub. This PR shows how to use model builder to optimize the finetune lora weights.

**Note that**: genai currently only works for LORA finetuned model. Qlora will use nf4 or fp4 which leads several issues in genai:
1. nf4/fp4 will force the model on cuda when loading. This lead runtime error when convert weights from tensor to numpy (need fix from genai).
2. weights shape dismatch error. (investigating...)

### DIFF
--- a/olive/model/handler/mixin/hf_config.py
+++ b/olive/model/handler/mixin/hf_config.py
@@ -16,7 +16,6 @@ from olive.model.utils.hf_utils import (
     load_hf_model_from_model_class,
     load_hf_model_from_task,
     save_hf_model_config,
-    save_hf_model_generation_config,
     save_hf_model_tokenizer,
 )
 
@@ -50,7 +49,7 @@ class HfConfigMixin:
 
         return get_hf_model_config(self.get_model_path_or_name(), **self.hf_config.get_loading_args_from_pretrained())
 
-    def _get_hf_model_generation_config(self):
+    def get_hf_model_generation_config(self):
         if self.hf_config is None:
             raise ValueError("HF model_config is not available")
 
@@ -58,7 +57,7 @@ class HfConfigMixin:
             self.get_model_path_or_name(), **self.hf_config.get_loading_args_from_pretrained()
         )
 
-    def _get_hf_model_tokenizer(self, **kwargs):
+    def get_hf_model_tokenizer(self, **kwargs):
         if self.hf_config is None:
             raise ValueError("HF model_config is not available")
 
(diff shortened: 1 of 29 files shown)


olive/passes/onnx/model_builder.py
```diff
@@ -150,6 +168,29 @@ def _run_for_config(
             if not isinstance(model, PyTorchModelHandler):
                 raise ValueError("metadata_only has to be true with ONNXModel as input.")
 
+        merged_model_path = None
```
jambayk, line 171: Is this feature requested by some customer? If this is needed, I think it is better if we provide a separate pass to merge the adapter weights. Otherwise, we are adding similar merge options and functionalities to multiple passes like onnxconversion and modelbuilder when these are technically not related to the pass. 
If this is done, we could also remove the merge weights options from onnx conversion pass.

trajepl: Yes, there is a offline session will finetune phi3 and run genai.
But not urgent, let me update it with seprated pass.

trajepl: Updated.
But for conversion, since the quantized_modules need to be unload before run `merge_and_unload`. I did not remove the logics directly. How about update it with separated PR.

olive/passes/pytorch/merge_lora_weights.py
```diff
@@ -0,0 +1,51 @@
...
+from olive.model import PyTorchModelHandler
+from olive.passes import Pass
+from olive.passes.pass_config import PassConfigParam
+
+logger = logging.getLogger(__name__)
+
+
+class MergeLoraWeights(Pass):
+    """Merge Lora weights into a complete model."""
+
```
devang-ml, line 20: - Please use "adapter" instead of "lora" in naming. Say MergeAdapterWeights. LoRA is too specific.
- Does this merge the adapter weights into master weights OR add rank adaption (code + weight) into the model?

devang-ml (commented):
Please update Phi3 example readme.

olive/passes/pytorch/merge_lora_weights.py
trajepl: Updated. 
Also for ` adapter weights into master weights` and `rank adaption (code + weight)`, could you help explain the differences? 
From my understanding, current pass will merge the adapter weights into base model, at the time it also unload the rank adaption code from base model. So I guess the answer is `adapter weights into master weights`.
https://huggingface.co/docs/peft/v0.6.2/en/package_reference/tuners#peft.LoraModel.merge_and_unload

devang-ml: I see. Thanks!

olive/model/handler/mixin/hf_config.py
```diff
@@ -66,28 +66,31 @@ def _get_hf_model_tokenizer(self, **kwargs):
         # TODO(anyone): only provide relevant kwargs, no use case for now to provide kwargs
         return get_hf_model_tokenizer(self.get_model_path_or_name(), **kwargs)
 
-    def save_metadata_for_token_generation(self, output_dir: str, **kwargs) -> List[str]:
+    def save_metadata_for_token_generation(
```
guotuofeng, line 69: do we need change the signature? I think kwargs should contains these values if user specify them?

olive/passes/pytorch/merge_adapter_weights.py
```diff
@@ -0,0 +1,51 @@
...
+            raise RuntimeError(
+                "No adapter path found in the model. Please check your input "
+                "model type or remove `MergeAdapterWeights` from passes configs"
+            )
+        pytorch_model = model.load_model()
+        merged_model = pytorch_model.merge_and_unload()
+
+        merged_model.save_pretrained(output_model_path)
+        model.save_metadata_for_token_generation(
+            output_model_path,
```
guotuofeng, line 39: do we need add the files to additional_files like https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/conversion.py#L156?

olive/passes/pytorch/merge_adapter_weights.py
```diff
@@ -0,0 +1,51 @@
...
+        if not model.adapter_path:
+            raise RuntimeError(
+                "No adapter path found in the model. Please check your input "
+                "model type or remove `MergeAdapterWeights` from passes configs"
+            )
+        pytorch_model = model.load_model()
+        merged_model = pytorch_model.merge_and_unload()
+
+        merged_model.save_pretrained(output_model_path)
+        model.save_metadata_for_token_generation(
```
guotuofeng, line 38: it seems here only save the tokenizer config?

trajepl: Updated as offline discussion.

guotuofeng (approved):

olive/passes/pytorch/merge_adapter_weights.py
trajepl: We can ignore this configs for merge_adapter_weights. It is used in packaging module to save model's dependencies.

PR #1173, opened 2024-05-23
🔗 linter fix

trajepl:

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -376,7 +376,7 @@ class Engine:
         # clean the passes
         self.passes.clear()
         for name, config in self.pass_config.items():
-            pass_cls: Type["Pass"] = config["type"]
+            pass_cls: Type[Pass] = config["type"]
             pass_cfg = config["config"]
             pass_cfg = pass_cls.generate_search_space(accelerator_spec, pass_cfg, config["disable_search"])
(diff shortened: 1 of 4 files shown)


olive/engine/engine.py
```diff
@@ -376,7 +376,7 @@ def setup_passes(self, accelerator_spec: "AcceleratorSpec"):
         # clean the passes
         self.passes.clear()
         for name, config in self.pass_config.items():
-            pass_cls: Type["Pass"] = config["type"]
```
guotuofeng, line 379: will fail

guotuofeng: https://stackoverflow.com/questions/77661139/how-do-i-stop-the-ruff-linter-from-moving-imports-into-a-type-checking-if-block

guotuofeng (changes requested):
the typecheck  import need use double quotes

trajepl:
> the typecheck import need use double quotes

In python, the annotions in function argument will be evaluated as a forward reference. Type annotations can be quoted to avoid forward references.

But when it comes to function variables, python always evaluate type annotations in a deferred manner, making the quotes unnecessary. https://docs.astral.sh/ruff/rules/quoted-annotation/

guotuofeng (approved):

PR #1177, opened 2024-05-28
🎃 Support ORT 1.18 QNN Preprocess

trajepl:
## Describe your changes

Support ORT 1.18 QNN Preprocess by adding configs:
- init_overrides
- add_qtype_converts
- activation_symmetric
- weight_symmetric

### DIFF
--- a/olive/passes/onnx/quantization.py
+++ b/olive/passes/onnx/quantization.py
@@ -267,9 +267,6 @@ _static_optional_config = {
                 Ex: if you override the output of an Add to 16-bit, this option ensures that the activation inputs
                 of the Add are also up-converted to 16-bit and that data types for surrounding ops are converted
                 appropriately. Refer to the documentation in mixed_precision_overrides_utils.py for additional details.
-            - `activation_symmetric: bool = False`: Symmetrize calibration data for activations.
-            - `weight_symmetric: bool = None`: True if weights should be quantized symmetrically (i.e., rmax == -rmin)
-                Defaults to None. If set to None, weight_symmetric is assumed true if the weight_type is a signed int.
             To be noted that the options might be updated in the further version of onnxruntime.
         """,
     ),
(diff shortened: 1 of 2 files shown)


devang-ml (approved):

olive/passes/onnx/quantization.py
```diff
@@ -227,6 +227,52 @@
...
+                                            tensor that should be converted to another quantization type.
+                    'convert["recv_nodes"] = Set : Set of node names that consume the converted activation,
+                                                other nodes get the original type. If not specified,
+                                                assume all consumer nodes get the converted type.
+            - `add_qtype_converts: bool = True`: True if this function should automatically add "convert" entries to
+                the provided `init_overrides` to ensure that operators use valid input/output types (activations only).
+                Ex: if you override the output of an Add to 16-bit, this option ensures that the activation inputs
+                of the Add are also up-converted to 16-bit and that data types for surrounding ops are converted
+                appropriately. Refer to the documentation in mixed_precision_overrides_utils.py for additional details.
+            - `activation_symmetric: bool = False`: Symmetrize calibration data for activations.
```
jambayk, line 270: we also have these values exposed https://github.com/microsoft/Olive/blob/ee10bd7adbab91215da157737b1862af52c2844c/olive/passes/onnx/quantization.py#L99

jambayk: I was thinking about something related recently. We need to be clear about which config options are used or ignored for qnn ep, especially the `exposed_extra_options`.

trajepl: True. But I just noticed that the default value for  `weight_symmetric` is different between qnn-config and normal quantization config.
1. qnn-config:
```python
    if weight_symmetric is None:
        weight_symmetric = weight_type in {QuantType.QInt8, QuantType.QInt16}
```
2.  onnxurntime base quantization
https://github.com/microsoft/onnxruntime/blob/d44be41e1cf79fd0d5361854480c92ea3daac3f1/onnxruntime/python/tools/quantization/base_quantizer.py#L91

The difference is only from `QuantType.QFLOAT8E4M3FN`.

trajepl: Maybe currently, we can keep this PR as is? Since qnn ep team may have some customized changes later. That might bring conflicts between qnn_config and base quantization's config.

jambayk: looking at their code, looks like their documentation for the default value of the `WeightSymmetric` outdated?https://github.com/microsoft/onnxruntime/blob/d44be41e1cf79fd0d5361854480c92ea3daac3f1/onnxruntime/python/tools/quantization/quantize.py#L113 
From the base quantizer code, it is None and the boolean value depends on the weight_type. Not sure if we should update our default value since it changes default behavior. 

In order to not have overlap between the params, I think we could update the default value for `WeightSymmetric` to `None` when EP is QNN EP link in https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/quantization.py#L570? We could then pop the value from extra_options and provide it to the qnn config generator.

jambayk: or if you want to keep them separate, please add a note in the param descriptions about which one is used/ignored based on EP.

trajepl: Yes, the docs is outdated. The default value for WeightSymmetric is None actually if user did not set it.

trajepl: Updated as offline discussion.

jambayk (approved):
Thanks!

PR #1181, opened 2024-05-29
UT: Pin `peft` to `0.10.0`

jambayk:
## Describe your changes
- The CI test on Windows fails occasionally during import of peft. At https://github.com/huggingface/peft/blob/e7b75070c72a88f0f7926cc6872858a2c5f0090d/src/peft/tuners/boft/layer.py#L30, the error is `OSError: [WinError 122] The data area passed to a system call is too small`. This code was added in https://github.com/huggingface/peft/commit/811169939ff466199cdd48767a10ca48497e03f7 which is part of peft version 0.11.0+.
- Revert #1180 since more tests are affected by the same issue. Will just pin peft version to the previous stable version 0.10.0 until the issue is fully investigated and resolved.

### DIFF
--- a/test/requirements-test.txt
+++ b/test/requirements-test.txt
@@ -22,7 +22,9 @@ onnxruntime-extensions
 openvino==2023.2.0
 optimum>=1.17.0
 pandas
-peft
+# TODO(anyone): Unpin peft once the issue is resolved
+# occasional import error on Windows with peft 0.11.1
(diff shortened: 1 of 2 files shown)


guotuofeng:
could we add the commit that makes the perf failure with new version?

guotuofeng (approved):

PR #1187, opened 2024-06-05
Cleanup DataConfig implementation

shaahji:
## Cleanup DataConfig implementation
* Removed DataConfig::params_config
* Removed DataConfig::components/component_args

All components specific parameters are now grouped in four separate objects:
 + load_dataset_config
 + pre_process_data_config
 + post_process_data_config
 + dataloader_config


examples/bert/bert_ptq_cpu.json
```diff
@@ -31,18 +31,28 @@
...
-                "batch_size": 1
+            "load_dataset_config": {
+                "params": {
+                    "data_name": "glue",
+                    "subset": "mrpc",
+                    "split": "validation"
+                }
+            },
+            "pre_process_data_config": {
+                "params": {
```
guotuofeng, line 42: do we really need the "params" key name? it would be better if like
```
            "load_dataset_config": {
                    "data_name": "glue",
                    "subset": "mrpc",
                    "split": "validation"
            },
```

guotuofeng: the `type` info is needless since the "load_dataset_config" already contains the type info

shaahji: > do we really need the `params` key name? it would be better if like
> 
> ```
>             "load_dataset_config": {
>                     "data_name": "glue",
>                     "subset": "mrpc",
>                     "split": "validation"
>             },
> ```

I agree that params seems redundant. I wasn't sure, what would the config class look like since `type` is required parameter and other are all optional. There is no strict list of what the params can be (think of custom data set/ data container implementation which can take any user arguments).

shaahji: > the `type` info is needless since the `load_dataset_config` already contains the type info

Hmm, I am unsure what you mean config already contains the type info. Please elaborate.

docs/architecture.md
```diff
@@ -284,16 +284,14 @@ Then the design for `DataContainer` interface will be like:
...
-            "load_dataset": {
-                "type": "test_dataset",
-                "params": {"test_value": "test_value"},
-            },
-            "dataloader": {
-                "type": "_test_dataloader",  # This is the key to get dataloader
-                "params": {"test_value": "test_value"},
-            },
-        }
+        "load_dataset_config": {
```
jambayk, line 287: since this is in the python api, this needs to be `load_dataset_config = ..., ....`

examples/bert/bert_ptq_cpu.json
jambayk: I also agree `params` nesting is not needed. But like Hitesh said, `type` is still needed if the user wants to override the default component type.

docs/architecture.md
shaahji: Fixed.

examples/bert/bert_ptq_cpu.json
shaahji: I am open to suggestions on how to remove the `params` node in the hierarchy. Is there a way to define a sub class of `ConfigBase` that can take any key-value pair as argument without the need for a dict member i.e. without being explicitly defined as a member of the class? I guess that would be counter-intuitive to the use of `ConfigBase` itself.

jambayk: Oh yeah, I forgot about the logistics of doing that. 

There is a `ConfigBase` subclass `ConfigWithExtraArgs` https://github.com/microsoft/Olive/blob/3d3277b646c2c5a0fa96bcf629a117aa78423176/olive/common/config_utils.py#L145 where any undefined field gets aggregated into the `extra_args` field. If we use this, then instead of accessing `params` in the backend, we could access `extra_args`.

shaahji: If the changes in this PR look good, I would like to merge this change before making the decorative `params` change. It's modifying too many files, and I don't want to sit on them for too long. Merging will be very hard.

olive/data/config.py
```diff
@@ -77,86 +58,74 @@ def __init__(self, **kwargs):
...
 
-    def _update_default_component(self):
-        """Resolve the default component type."""
-        for k, v in self.default_components_type.items():
-            self.default_components[k] = DataComponentConfig(type=v, params={})
+        # 2. get default_components from default_components_type
+        return {k: DataComponentConfig(type=v, params={}) for k, v in default_components_type.items()}
 
-    def fill_in_params(self):
+    def _fill_in_params(self):
```
jambayk, line 103: I don't think this is needed anymore? If the function as default values, then we don't need to add it to `params` which is passed as `kwargs` anyways.

olive/data/container/raw_data_container.py
```diff
@@ -18,15 +18,19 @@ class RawDataContainer(DataContainer):
...
-            params_config={
-                "data_dir": data_dir,
-                "input_names": input_names,
-                "input_shapes": input_shapes,
-                "input_types": input_types, # optional
-                "input_dirs": input_dirs, # optional
-                "input_suffix": input_suffix, # optional
-                "input_order_file": input_order_file, # optional
-                "annotations_file": annotations_file, # optional
+            components={
```
jambayk, line 21: `components` doesn't exist anymore.

olive/data/config.py
shaahji: That's change is going to require a full suite of tests again. Can I remove the call in a follow up PR?

jambayk: sure! Doesn't affect the behavior so it's not high priority. just think it's redundant.

olive/workflows/run/config.py
```diff
@@ -99,25 +99,30 @@ def validate_data_configs(cls, v, values):
...
             # auto insert trust_remote_code from input model hf config
             # won't override if value was set to False explicitly
-            for key in ["trust_remote_code"]:
-                if (
-                    hf_config.get("from_pretrained_args", {}).get(key, None)
-                    and v["params_config"].get(key, None) is None
-                ):
-                    v["params_config"][key] = hf_config["from_pretrained_args"][key]
+            if hf_config.get("from_pretrained_args", {}).get("trust_remote_code"):
+                v["load_dataset_config"] = load_dataset_config = v.get("load_dataset_config") or {}
```
jambayk, line 120: `trust_remote_code` is used in `preprocess_data` and not `load_dataset`

examples/bert/bert_ptq_cpu.json
jambayk: sounds good to me! as you said, it's a cosmetic change that only looks different on the user side. In the backend, we need a way to hold the params anyways.

olive/workflows/run/config.py
shaahji: Fixed.

olive/data/container/raw_data_container.py
shaahji: Fixed.

olive/data/config.py
shaahji: Done.

shaahji: Apparently, this is still required. The call appends keys to the `params` that are initialized as `None` and will potentially be filled in later. Few tests failed if I removed this call -

https://aiinfra.visualstudio.com/PublicPackages/_build/results?buildId=480801&view=logs&j=dbbdd4d8-ffd3-515a-31fd-659375dbdff8&t=0a2248d7-6ef4-5599-1bc5-d55a003ee0aa

Change reverted.

examples/bert/bert_ptq_cpu.json
guotuofeng: Is the `type` required? I don't see it in the example.
what I mean is not change the `DataComponentConfig` and just change the plain json dict to `DataComponentConfig` in validator.

jambayk: example of type override image

shaahji: Even without that params is still required to collect only the parameters that are passed down to the creator function. I noticed that DataSet/DataContainer constructors fail if they are passed unknown arguments (including in kwargs). So, it might not actually be possible to remove the params entirely. If we did, we will still be doing custom include/exclude list to collect only the arguments that are passed down to the constructor (but then what are the other parameters for? discard?).

jambayk (approved):

PR #1198, opened 2024-06-15
Llama2: Multi-lora generate using genai, Update default kv cache io names

jambayk:
## Describe your changes
- The default input-output names are changed to `past_key_values.<id>.key/value` and `present.<id>.key/value` since this is the most commonly used template (by both genai and optimum)
- Ignore `onnxruntime-genai` package when checking for `onnxruntime` dependency
- `llama2_multilora` notebook now has example generation for different lora adapters using genai.

### DIFF
--- a/docs/source/overview/options.md
+++ b/docs/source/overview/options.md
@@ -119,13 +119,13 @@ find more details in [Olive Models](https://microsoft.github.io/Olive/api/models
           - If it is `False`, Olive will not use key value cache.
           - If it is `True`, Olive will infer the cache configuration from the input_names/input_shapes and input model based on default `kv_cache`.
           - If it is a dictionary, it should contains the key value cache configuration. Here is an default configuration example:
-            - `ort_past_key_name`: "past_key_<id>"
+            - `ort_past_key_name`: "past_key_values.<id>.key"
                 Template for the past key name. The `<id>` will be replaced by the id of the past key.
-            - `ort_past_value_name`: "past_value_<id>"
+            - `ort_past_value_name`: "past_key_values.<id>.value"
                 Template for the past value name. The `<id>` will be replaced by the id of the past value.
-            - `ort_present_key_name`: "present_key_<id>"
+            - `ort_present_key_name`: "present.<id>.key"
                 Template for the present key name. The `<id>` will be replaced by the id of the present key.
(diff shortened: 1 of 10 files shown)


docs/source/overview/options.md
```diff
@@ -119,13 +119,13 @@ find more details in [Olive Models](https://microsoft.github.io/Olive/api/models
           - If it is `False`, Olive will not use key value cache.
           - If it is `True`, Olive will infer the cache configuration from the input_names/input_shapes and input model based on default `kv_cache`.
           - If it is a dictionary, it should contains the key value cache configuration. Here is an default configuration example:
-            - `ort_past_key_name`: "past_key_<id>"
+            - `ort_past_key_name`: "past_key_values.<id>.key"
```
trajepl, line 122: If the purpose to update io_config names is to provide a simpliest example for llama2, then I am ok with this changes.

But in general, the original value looks clearer and simpler.

jambayk: The reason is not for a simpler example. But like I mentioned in the PR description, this template is the more common template. It is used by both genai and optimum.

I have only seen the previous template in the ort phi2 example.

jambayk: This makes the model directly compatible with both generate apis. Otherwise, we have to edit the genai config for genai api. It probably doesn't work for optimum.

docs/source/overview/options.md
```diff
@@ -119,13 +119,13 @@ find more details in [Olive Models](https://microsoft.github.io/Olive/api/models
...
                 Template for the past key name. The `<id>` will be replaced by the id of the past key.
-            - `ort_past_value_name`: "past_value_<id>"
+            - `ort_past_value_name`: "past_key_values.<id>.value"
                 Template for the past value name. The `<id>` will be replaced by the id of the past value.
-            - `ort_present_key_name`: "present_key_<id>"
+            - `ort_present_key_name`: "present.<id>.key"
```
guotuofeng, line 126: just two cents about the past/present key names. it seems the key name is different for different model. need we make these keys configurable or specified by user?

docs/source/overview/options.md
trajepl: Ic. Sorry I missed the PR description.
The changes look good to me.

trajepl (approved):

PR #1203, opened 2024-06-19
Add workflow host

xiaoyu-work:
## Describe your changes

Add workflow host to run Olive workflow on a AzureML system. 

- User can submit workflow run to AzureML system.
- The local machine can be safely turned off while the workflow can still be running on the AML workspace compute.

## TODO:

- User can retrieve the logs and artifacts from the remote VM by `--retrieve` flag.


shaahji:
Is there really a need/reason to setup the cloud system in every workflow file? Shouldn't something be setup once per machine and then any workflow launched from that machine use it - much like how azure login works. Basically, a separate cloud config file that will only be used if the workflow calls out to use cloud deployment.

Solves multiple problems -
* The cloud configuration requires paths to local files (like the ssh private key). This makes the workflow files not share-able across team and can't be committed to a shared repository. If the cloud configuration was pulled out, the workflow can be shared and each user can create their own cloud configuration with local context.
* If I have multiple workflows, I will have to go thru' each file and modify something - not the best user experience. I should be able to modify it in one place and have it apply to all my workflows.
* The path to the cloud configuration can be a command line argument allowing user to manage multiple different configurations easily.

xiaoyu-work:
> Is there really a need/reason to setup the cloud system in every workflow file? Shouldn't something be setup once per machine and then any workflow launched from that machine use it - much like how azure login works. Basically, a separate cloud config file that will only be used if the workflow calls out to use cloud deployment.
> 
> Solves multiple problems -
> 
> * The cloud configuration requires paths to local files (like the ssh private key). This makes the workflow files not share-able across team and can't be committed to a shared repository. If the cloud configuration was pulled out, the workflow can be shared and each user can create their own cloud configuration with local context.
> * If I have multiple workflows, I will have to go thru' each file and modify something - not the best user experience. I should be able to modify it in one place and have it apply to all my workflows.
> * The path to the cloud configuration can be a command line argument allowing user to manage multiple different configurations easily.

Yeah this is a good idea. I can abstract the cloud configuration out to a separate cloud config file.

guotuofeng:
image
typo

docs/source/features/run_workflow_remotely.md
```diff
@@ -0,0 +1,99 @@
...
+    "hostname": "8.8.8.8",
+    "os": "Linux",
+    "username": "ucm",
+    "key_filename": "/home/ucm/.ssh/key.pem"
+}
+```
+
+## Run Olive Workflow
+
+To run the Olive workflow, add a dispatcher configuation at the top level of the Olive config file:
```
jambayk, line 65: It might be simpler if we don't make the dispatcher a part of the `olive run` command and add a new command like `olive dispatch --config workflow.json --target remote_config.json`? `--retrieve` could be an option for this command.

xiaoyu-work: I've thought about this. The reason I didn't do this is because I would like to simplify the command to run Olive. But I'm open to this.

olive/common/utils.py
```diff
@@ -417,3 +417,13 @@ def find_first_matched_value(original_dict: Dict, keys: Union[str, Tuple, List[s
     if raise_key_error:
         raise KeyError(f"Keys {keys} not found in {original_dict}")
     return None
+
+
+def get_path_by_os(path: Union[Path, str], target_system: OS):
```
jambayk, line 422: doesn't `str(path)` also do the same thing? 
image

olive/systems/azureml/aml_workflow_runner.py
```diff
@@ -0,0 +1,51 @@
...
+    olive_run(olive_config, setup=True)
+    olive_run(olive_config)
+
+    workflow_artifacts_path = Path(olive_config_args.workflow_artifacts)
+
+    remote_cache_dir = workflow_artifacts_path / "cache"
+    remote_output_dir = workflow_artifacts_path / "output"
+
+    # Clean up the remote cache and output directories if they exist
+    if remote_cache_dir.exists():
```
jambayk, line 41: should we provide an option to reuse cache? For example, the workflow might fail at a later step but we want to reuse the previous steps?

olive/systems/local.py
```diff
@@ -19,6 +19,9 @@
 class LocalSystem(OliveSystem):
     system_type = SystemType.Local
 
+    def __init__(self, accelerators: List[AcceleratorConfig] = None, hf_token: bool = None):
```
jambayk, line 22: Is there a reason this is added? Since it's just calling the super()'s init, this is redundant.

olive/common/utils.py
xiaoyu-work: This was used to get path from a different OS. Forgot to remove this since we go for aml system approach. Will remove this.

olive/systems/azureml/aml_workflow_runner.py
xiaoyu-work: Most AML workspace compute should be provisioned as a new one for next job. I got some errors when a node in a cluster was not released and immediately used for next job so I added this. I don't think we would have any use case that reuse cache. What do you mean by reuse the previous steps?

Regarding to your question, this `cache` folder is not the Olive workflow `cache`. This is a remote path with same name.

olive/systems/local.py
xiaoyu-work: Good catch. Forgot to remove this, this was added for remote dispatcher.

olive/systems/azureml/aml_system.py
```diff
@@ -125,14 +126,70 @@ def __init__(
...
+        with tempfile.TemporaryDirectory() as tempdir:
+            workflow_pipeline = self._create_pipeline_for_workflow(workflow_id, tempdir)
+            self._run_workflow_job(
+                ml_client,
+                workflow_pipeline,
+                "olive-workflow",
+                tags={"Workflow": workflow_id},
+            )
+
+    def _create_pipeline_for_workflow(self, workflow_id: str, tmp_dir):
```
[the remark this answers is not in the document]
xiaoyu-work: I was thinking whether we should add this support as the local model and data could be large and uploading them would take a long time. @devang-ml what do you think?

olive/systems/azureml/aml_workflow_runner.py
jambayk: I see that the `workflow_artifacts_path` is mounted as a datastore folder. So, I thought this was used as the workflow cache. But I now see that the cache is copied after the workflow is run.

jambayk: this brings up another question. if a user runs two workflows concurrently or sequentially without changing the workflow id, wouldn't this cause the results of the run that completes first to disappear? or in an edge case, case a race condition on the directory save? 

Since the workflow cache is not persistent across run, the `workflow_id` doesn't seem useful if the user needs to specify it themselves. It made sense when the engine host was a remote vm. 
Maybe the windows_artifacts_path should be created using a unique_id for the specific config being run? Maybe a hash for the workflow config? Instead of `f"azureml://datastores/{self.config.datastores}/paths/{workflow_id}"`

olive/systems/azureml/aml_system.py
```diff
@@ -125,14 +126,70 @@ def __init__(
         self.env_vars = self._get_hf_token_env(self.azureml_client_config.keyvault_name) if self.hf_token else None
         self.temp_dirs = []
 
+    def submit_workflow(self, workflow_id: str):
```
jambayk, line 129: how does someone retrieve the output of the workflow? Do they have to go to the datastore and download the files manually?

xiaoyu-work: Yes, I got some errors when downloading from datastore by azure file system api. So I didn't add retrieve feature for now.

olive/systems/azureml/aml_workflow_runner.py
xiaoyu-work: Yeah I was thinking cache is not useful for this remote run as we cannot use this cache at all. The reason I save cache folder as well to datastore is because: 1. Output models is in cache folder. Unless we copy output models out. 2. I (may) will add retrieve feature later. Then the cache can be downloaded locally for local use.

olive/systems/azureml/aml_system.py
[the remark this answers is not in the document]
devang-ml: We should not limit the use case.

But, we should be smart because the data is usually coming from HF or other places anyway. In those cases, there is no need to download the data locally and upload it to AML system.

olive/systems/azureml/aml_workflow_runner.py
devang-ml: What would be the local use of downloaded cache?

There is no use of local (i.e. on the remote workflow host) cache. Ideally, it would use cloud cache here from PR 1166; 
 which can be done as a follow up PR.

xiaoyu-work: For the same workflow-id, I was thinking a use case that a user runs workflow remotely, retrieve the cache and outputs to local. When next time he was trying to run the same workflow, Olive can directly go for local cache in this case. This is a different use case from could model cache.

olive/systems/azureml/aml_system.py
```diff
@@ -115,14 +117,104 @@ def __init__(
...
+                [(data_config.name, data_config.to_json(check_object=True)) for data_config in run_config.data_configs]
+            )
+
+        resources_check_list = [run_config.passes, run_config.evaluators, run_config.systems]
+        for resources in resources_check_list:
+            if resources:
+                config_list = self._get_dict_resources_config(resources, config_list)
+
+        for name, config_json in config_list:
+            name_inputs, name_args = self.create_inputs_and_args(
```
jambayk, line 168: `create_inputs_and_args` has been updated to take all configs at once to avoid duplicate mounts. 

It should be possible to just do 
`inputs, args = self.create_inputs_and_args({"run_config" : run_config.to_json()}, "tmp_dir", ignore_keys=["engine"m "model_attributes"]` 

Similarly, in the aml workflow runner it can just parse the run_config itself at once.

jambayk: although I am not sure about ignoring the engine entirely since it must have been valided to replace evaluators with the full configuration.

xiaoyu-work: Will update this.

olive/systems/azureml/aml_system.py
```diff
@@ -115,14 +117,78 @@ def __init__(
...
+
+        run_config.workflow_host = None
+
+        config_root = tmp_dir / "config"
+        cur_dir = Path(__file__).resolve().parent
+        code_files = [cur_dir / script_name]
+        self.copy_files(code_files, config_root)
+
+        inputs, args = self.create_inputs_and_args(
+            {WORKFLOW_CONFIG: run_config.to_json()},
```
jambayk, line 149: nit: since the run_config is already is validated, should we pop `systems`, `evaluators`, `data_configs`,  from the run_config json first? Those are not used at all and complicate the config as well as resources. For example, I see that the aml system related resources are uploaded in your test job.

xiaoyu-work: What do you mean by they are not used? The remote workflow can also run evaluation and submit a job to aml system (though not verified yet, will verify this later).

jambayk: They are not used since their values are already inserted directly into engine and passes during the runconfig validation. So, there is nothing in the workflow referring to these anymore.

xiaoyu-work: Updated. I will merge after your AML client insertion PR.

docs/source/features/run_workflow_remotely.md
jambayk, line 1: This file needs to be added to the index https://github.com/microsoft/Olive/blob/aab163c561f151b572b4df116f9d3429f7888793/docs/source/index.rst?plain=1#L40

olive/systems/azureml/aml_system.py
```diff
@@ -405,6 +476,27 @@ def pass_runner_pipeline():
...
+        tags = {**self.tags, **(tags or {})}
+        job = retry_func(
+            ml_client.jobs.create_or_update,
+            [pipeline_job],
+            {"experiment_name": experiment_name, "tags": tags},
+            max_tries=self.azureml_client_config.max_operation_retries,
+            delay=self.azureml_client_config.operation_retry_interval,
+            exceptions=(HttpResponseError, JobException),
+        )
+        ml_client.jobs.stream(job.name)
```
jambayk, line 496: if we stream, the process keeps waiting for it to finish. I think we don't want that?

olive/systems/azureml/aml_system.py
```diff
@@ -405,6 +476,27 @@ def pass_runner_pipeline():
 
         return pass_runner_pipeline()
 
+    def _run_workflow_job(
```
jambayk, line 479: nit: maybe we can generalize the `_run_job` method to make the job streaming and output download optional? The job submission logic is the same.

olive/systems/azureml/aml_system.py
xiaoyu-work: After the job was submitted, it's okay to kill this streaming process, the job will be still running on the AML.

olive/systems/azureml/aml_system.py
xiaoyu-work: Sounds good. Updated.

olive/systems/azureml/aml_system.py
jambayk: Just wanted to make sure if the streaming and process kill is intended behavior. Or we just want to submit the job and return from the process without streaming.

olive/systems/azureml/aml_system.py
```diff
@@ -505,6 +486,7 @@ def _run_job(
         tmp_dir: Union[str, Path],
```
jambayk, line 486: this needs to be a kwarg too? The workflow run doesn't provide this

olive/systems/azureml/aml_system.py
xiaoyu-work: Keep this steaming is an easier way to track the workflow run status without having to check job status online, whether it succeed or fail. But user still can terminate this anytime, so I keep this streaming here.

jambayk: That makes sense. Maybe the log can tell the user it's safe to exit the program? So that they don't assume it needs to be kept alive

xiaoyu-work: Yeah i can add a log for this. I added such log before when running the workflow on a remote VM but not for AML computes. Azure cli already has such log if you terminate this: 
```
azure.ai.ml.exceptions.JobException: The output streaming for the run interrupted.
But the run is still executing on the compute target.
Details for canceling the run can be found here: https://aka.ms/aml-docs-cancel-run
```

jambayk: Since this log is printed only when terminated, I think an olive log saying it's safe to exit would be helpful.

jambayk (approved):

PR #1209, opened 2024-06-21
Use DummyDataLoader instead of declaring new types for tests

shaahji:
## Use DummyDataLoader instead of declaring new types for tests

* Deprecate skip_dataloader because it ignores forced treating Dataset as Dataloader and we have no more use case for it.
* Updated DummyDataset
  * to generate random data instead of constant ones 
  * to accept a seed for randomness and a size
  * to avoid generating the data until explicitly requested
* Removed clones, yet slightly different, implementations of DummyDataset. Used primarily to tests.
* Renamed global create_dataloader to create_dummy_dataloader to be explicit. Used only in tests.

### DIFF
--- a/olive/data/component/dataset.py
+++ b/olive/data/component/dataset.py
@@ -94,41 +94,48 @@ class BaseDataset(TorchDataset):
 
 
 class DummyDataset(BaseDataset):
-    def __init__(self, input_shapes, input_names: Optional[List] = None, input_types: Optional[List] = None):
-        """Initialize the dummy dataset.
+    def __init__(
+        self,
+        input_shapes,
+        input_names: Optional[List] = None,
+        input_types: Optional[List] = None,
+        max_samples: Optional[int] = 32,
+    ):
+        """Initialize the dataset with dummy data.
 
-        if input_names is None, the dummy dataset will return a tuple of tensors
-        else the dummy dataset will return a dict of tensors
+        if input_names is not provided, the dataset will return a tuple of tensors
+        else the dataset will return a dict of tensors
         """
         # pylint: disable=super-init-not-called
-        self.input_shapes = input_shapes
-        self.input_names = input_names
-        self.input_types = input_types or ["float32"] * len(input_shapes)
+        if not input_types:
+            input_types = ["float32"] * len(input_shapes)
+        input_types = [resolve_torch_dtype(dtype_str) for dtype_str in input_types]
+
+        if input_names:
+            dummy_data = {}
+            for input_name, input_shape, input_type in zip(input_names, input_shapes, input_types):
+                dummy_data.update({input_name: torch.ones(input_shape, dtype=input_type)})
+            dummy_data = dummy_data if len(dummy_data) > 1 else dummy_data[input_names[0]]
+        else:
+            dummy_data = []
+            for shape, dtype in zip(input_shapes, input_types):
+                dummy_data.append(torch.ones(shape, dtype=dtype))
+            dummy_data = tuple(dummy_data) if len(dummy_data) > 1 else dummy_data[0]
+
+        self.max_samples = max_samples
+        self.dummy_data = dummy_data, torch.tensor([0])
 
     def __len__(self):
-        return 256
+        return self.max_samples
 
     def __getitem__(self, index):
         # From https://docs.python.org/3/reference/datamodel.html#object.__getitem__,
         # __getitem__ should raise IndexError when index is out of range
         # Otherwise, the enumerate function will enter infinite loop
-        if index < 0 or index >= len(self):
+        if index < 0 or index >= self.max_samples:
             raise IndexError("Index out of range")
 
-        input_types = [resolve_torch_dtype(dtype_str) for dtype_str in self.input_types]
-
-        if not self.input_names:
-            dummy_inputs = []
-            for shape, dtype in zip(self.input_shapes, input_types):
-                dummy_inputs.append(torch.ones(shape, dtype=dtype))
-            dummy_inputs = tuple(dummy_inputs) if len(dummy_inputs) > 1 else dummy_inputs[0]
-        else:
-            dummy_inputs = {}
-            for input_name, input_shape, input_type in zip(self.input_names, self.input_shapes, input_types):
-                dummy_inputs.update({input_name: torch.ones(input_shape, dtype=input_type)})
-            dummy_inputs = dummy_inputs if len(dummy_inputs) > 1 else dummy_inputs[self.input_names[0]]
-        label = 0
-        return dummy_inputs, label
+        return self.dummy_data
 
 
 class RawDataset(BaseDataset):
--- a/olive/data/container/dummy_data_container.py
+++ b/olive/data/container/dummy_data_container.py
@@ -18,17 +18,17 @@ class DummyDataContainer(DataContainer):
         dummy_data_config = DataConfig(
             name="dummy",
             type="DummyDataContainer",
-            "load_dataset_config"={
-                "params": {
+            load_dataset_config=DataComponentConfig(
+                params={
                     "input_names": metric.user_config.input_names,
                     "input_shapes": metric.user_config.input_shapes,
                     "input_types": metric.user_config.input_types,
                 }
-            }
+            ),
         )
     """
 
     default_components_type: ClassVar[dict] = {
         DataComponentType.LOAD_DATASET.value: "dummy_dataset",
-        DataComponentType.DATALOADER.value: "skip_dataloader",
+        DataComponentType.DATALOADER.value: "no_auto_batch_dataloader",
     }
--- a/test/unit_test/passes/onnx/test_perf_tuning.py
+++ b/test/unit_test/passes/onnx/test_perf_tuning.py
@@ -5,7 +5,7 @@
 import logging
 import math
 import re
-from test.unit_test.utils import create_dataloader, get_onnx_model
+from test.unit_test.utils import create_dummy_dataloader, get_onnx_model
 from unittest.mock import MagicMock, PropertyMock, patch
 
 import psutil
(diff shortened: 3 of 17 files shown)


test/unit_test/passes/onnx/test_perf_tuning.py
```diff
@@ -5,26 +5,33 @@
...
 from olive.evaluator.metric_result import flatten_metric_result
 from olive.evaluator.olive_evaluator import OliveEvaluator, OnnxEvaluator
 from olive.hardware.accelerator import DEFAULT_CPU_ACCELERATOR, DEFAULT_GPU_CUDA_ACCELERATOR, AcceleratorSpec, Device
 from olive.passes.olive_pass import create_pass_from_dict
 from olive.passes.onnx.perf_tuning import PERFTUNING_BASELINE, OrtPerfTuning, PerfTuningRunner, generate_test_name
 
 
+# TODO(shaahji): Remove this once perf_tuning pass supports DataConfig
+def _create_dummy_dataloader(data_dir, batch_size=1, **kwargs):
+    return DataLoader(DummyDataset([[1, 1]]), batch_size=batch_size)
```
jambayk, line 25: doesn't the shape `[1,1]` already include the batch dimension. Is the data from this dataloader going to have shape `[1,1,1]` then?

test/unit_test/workflows/test_workfow_run.py
jambayk, line 1: not related to the PR but just realized there is a typo in the name of this file

test/unit_test/passes/onnx/test_perf_tuning.py
shaahji: There is some ambiguity between the way tests are implementing and what perf tuning/metrics expect as the data.  The implementation of `DummyDataset` hasn't changed in what it used to return (except for using `rand()` instead of `ones()`). I have a follow up PR to make both tests and perf_tuning work consistently with batching input.

test/unit_test/utils.py
```diff
@@ -193,8 +189,9 @@ def get_accuracy_metric(
         name="accuracy",
         type=MetricType.ACCURACY,
         sub_types=sub_types,
-        user_config=user_config or accuracy_metric_config,
+        user_config=user_config,
```
jambayk, line 192: looks like `accuracy_score_metric_config` is getting ignored?

jambayk: nevermind. I mixed the two configs.

test/unit_test/utils.py
```diff
@@ -31,29 +33,9 @@ def forward(self, x):
...
-        self.data = torch.tensor(self.rng.random((size, 1)))
-        self.labels = torch.tensor(self.rng.random(1))
-
-    def __getitem__(self, idx):
-        return self.data[idx], self.labels[idx]
-
-    def __len__(self):
-        return self.size
+# TODO(shaahji): Remove this once perf_tuning pass supports DataConfig
+def create_dummy_dataloader(data_dir, batch_size=1, **kwargs):
```
jambayk, line 37: can you make the size of this dataset be `1`? Otherwise, the quantization_aware_training is taking a long time to run.

jambayk: the old dataloader also only had size 1

test/unit_test/passes/onnx/test_perf_tuning.py
jambayk: as long as the test passes, sounds good to me. Wasn't sure if it would pass since the shape `[1,1,1]` doesn't match the input shape of the model.

test/unit_test/utils.py
shaahji: Done!

test/unit_test/workflows/test_workfow_run.py
shaahji: Renamed!

test/unit_test/passes/onnx/test_perf_tuning.py
shaahji: All tests are passing. But, as you noticed, there are inconsistencies in inputs across tests and passes. Each test seems to be authored to work with specific data input, some accounting for batching and others don't.

jambayk (dismissed):

olive/data/component/dataset.py
```diff
@@ -115,20 +129,32 @@ def __getitem__(self, index):
...
-                dummy_inputs.append(torch.ones(shape, dtype=dtype))
-            dummy_inputs = tuple(dummy_inputs) if len(dummy_inputs) > 1 else dummy_inputs[0]
-        else:
+        # NOTE: torch.rand doesn't support all types, especially not int64 and bool
+        # So, always generate the data with dtype=float32 and convert it to expected input_type.
+        if self.input_names:
             dummy_inputs = {}
             for input_name, input_shape, input_type in zip(self.input_names, self.input_shapes, input_types):
-                dummy_inputs.update({input_name: torch.ones(input_shape, dtype=input_type)})
+                data = torch.rand(input_shape, generator=generator, dtype=torch.float32)
```
jambayk, line 143: I think we still need an option or dataset that gives us constant ones for data. The dummy data for pytorch model also uses the dummy data container https://github.com/microsoft/Olive/blob/722749ecaba2117f36474b65b6efb38813ac6b3b/olive/model/handler/mixin/dummy_inputs.py#L28. This might cause issues with the `input_ids` and `attention_masks` being invalid or lead to unexpected behavior.

jambayk: `torch.rand` and then cast to `int` leads to all zeros since the sampling is in `[0,1)`.

trajepl: Also, dummy inputs may be used for model latency evaluation. Should we keep the inputs be consistent among model evaluations like torch, onnx, dlc and etc.?

olive/data/container/dummy_data_container.py
```diff
@@ -18,17 +18,21 @@ class DummyDataContainer(DataContainer):
...
+                # TODO(shaahji): Remove this override once DataConfig transition is complete
+                params={ "batch_size": None }
+            )
         )
     """
 
     default_components_type: ClassVar[dict] = {
         DataComponentType.LOAD_DATASET.value: "dummy_dataset",
-        DataComponentType.DATALOADER.value: "skip_dataloader",
+        DataComponentType.DATALOADER.value: "default_dataloader",
```
trajepl, line 37: `no_auto_batch_dataloader` takes the same effect with `default_dataloader` and `params={ "batch_size": None }`
image

olive/data/component/dataset.py
shaahji: Reverted to use `torch.ones` instead of `torch.rand`. A truly random dataset needs discussion and design. Fallback to what worked in the past for now.

olive/data/container/dummy_data_container.py
shaahji: Done!

test/unit_test/test_data_root.py
```diff
@@ -53,6 +32,17 @@ def get_dataloader_config():
...
         },
+        "data_configs": [
+            {
+                "name": "test_data_config",
+                "type": "DummyDataContainer",
+                "load_dataset_config": {
+                    "type": "dummy_dataset",
+                    "params": {"data_dir": "data", "input_shapes": [1], "size": 1},
+                },
+                "dataloader_config": {"params": {"batch_size": 16}},
```
jambayk, line 43: the dataloader component is `no_auto_batch_dataloader` so batch size is ignored

test/unit_test/passes/onnx/test_perf_tuning.py
```diff
@@ -5,26 +5,33 @@
...
+
+
 @pytest.mark.parametrize(
     "config",
     [
         {"input_names": ["input"], "input_shapes": [[1, 1]]},
         {},
-        {"dataloader_func": create_dataloader},
-        {"dataloader_func": create_dataloader, "dataloader_func_kwargs": {"dummy_kwarg": 1}},
+        {"dataloader_func": _create_dummy_dataloader},
```
jambayk, line 33: can this use the `create_dummy_dataloader` from `test.unit_tests.utils`?

test/unit_test/passes/pytorch/test_quantization_aware_training.py
```diff
@@ -2,18 +2,26 @@
...
+from torch.utils.data import DataLoader
+
+from olive.data.component.dataset import DummyDataset
 from olive.hardware.accelerator import AcceleratorSpec
 from olive.passes.olive_pass import FullPassConfig, create_pass_from_dict
 from olive.passes.pytorch.quantization_aware_training import QuantizationAwareTraining
 
 
+# TODO(shaahji): Remove this once QuantizationAwareTraining pass supports DataConfig
+def _create_dummy_dataloader(data_dir, batch_size=1, size=1, **kwargs):
```
jambayk, line 16: same here

test/unit_test/utils.py
```diff
@@ -162,23 +144,31 @@ def get_mock_openvino_model():
...
+        type="DummyDataContainer",
+        load_dataset_config=DataComponentConfig(
+            params={
+                "input_shapes": input_shapes,
+                "input_names": input_names,
+                "input_types": input_types,
+                "size": size,
+            }
+        ),
+        dataloader_config=DataComponentConfig(params={"batch_size": batch_size}),
```
jambayk, line 159: batch size is ignored here too. will probably have to update the callers to provide the correct shape?

shaahji: Fixed

test/unit_test/passes/pytorch/test_quantization_aware_training.py
shaahji: This was done in response to your comment about time it takes for training. Required to force a size of 1 for dataset.

test/unit_test/passes/onnx/test_perf_tuning.py
shaahji: Done!

test/unit_test/test_data_root.py
shaahji: Fixed.

jambayk (approved):

PR #1214, opened 2024-06-27
🗝️ Provide built in config for Random data with kv cache

trajepl:
## Describe your changes

Provide built in config for Random data with kv cache


examples/llama2/llama2_generate.json
```diff
@@ -60,6 +60,32 @@
                     "joiner": "\n\n"
                 }
             }
+        },
+        {
+            "name": "prompt_transformer_dummy_data",
+            "type": "TransformersDummyDataContainer",
+            "load_dataset_config": {
+                "params": {
+                    "batch_size": 2,
```
devang-ml, line 69: - Please make these params (batch_size, max_seq_len etc., ) for TransformerDummyDataContainer have proper default values and remove them from .json configs in examples.
- Please have two separate builtin containers (one for prompt and another for token) with appropriate default values for seq lens

trajepl: Updated.
Keep `generative` as false by default since it will skip `position_id` and `past_key_value`. Also, the `generative=true` is used only for llama2_generate.json.

Please let me know if you want to set `generative` as true by default.

devang-ml: Why is llama2_generate.json is special regarding this? 
@shaahji

shaahji: The dataset was evaluation is different because the expected input shapes are different for running as a generative model vs. non-generative model. Ideally, the difference should be driven by change in data config only but the setup was using dataloader_func so just carry forward.

This will be swapped out by a near future PR to support only data configs for metric configuration.

olive/data/container/dummy_data_container.py
```diff
@@ -32,3 +38,96 @@ class DummyDataContainer(DataContainer):
...
+class TransformersDummyDataContainer(DummyDataContainer):
+    """Dummy data container for transformer model.
+
+    The way to create a dummy data container for transformer model:
+        dummy_data_config = DataConfig(
+            name="dummy",
+            type="TransformersDummyDataContainer",
+            "load_dataset_config"={
+                "params": {
+                    "batch_size": 1,
```
jambayk, line 53: Do you think it might be more consistent to have batch_size as parameter for `data_loader_config`? I tested it locally and torch dataloader is able to handle batching of the list of key-values correctly (although it turns the tuple into list but I think that's fine?). 

Otherwise, it feels weird to have `batch_size` as a parameter of load dataset.

trajepl: True. I agree with you. Will update it.

examples/llama2/llama2_generate.json
devang-ml: llama2 is a generative model so I'm not sure what is the difference between running as a generative model vs. non-generative model.

olive/data/container/dummy_data_container.py
trajepl: Updated.

examples/llama2/llama2_generate.json
devang-ml: BTW, this is a just a nit-pick that should not block this PR.

docs/source/tutorials/configure_data.rst
```diff
@@ -185,6 +185,62 @@ Convert the raw data config to the data container.
...
+
+        .. code-block:: json
+
+            {
+                "name": "transformers_dummy_data_config",
+                "type": "TransformersDummyDataContainer",
+                "load_dataset_config": {
+                    "params": {
+                        "batch_size": 1,
+                        "seq_len": 128,
```
devang-ml, line 202: Please avoid listing parameters with defaults values here.

olive/data/template.py
```diff
@@ -123,3 +123,141 @@ def raw_data_config_template(
             }
         ),
     )
+
+
+def transformers_dummy_data_config_template(
+    model_name,
```
devang-ml, line 129: What's the use of these templates?

devang-ml (commented):
Please update opt, phi, mistral etc., examples also.

olive/data/template.py
trajepl: API purpose. In olive, we can directly call this function to create a dc without json config.

docs/source/tutorials/configure_data.rst
trajepl: Updated.

trajepl:
> Please update opt, phi, mistral etc., examples also.

Updated opt/phi/mistral.

But for open_llama, there is the extra input `use_cache` / `use_cache_branch`. I need to kick off another round updates in this PR.

jambayk:
> > Please update opt, phi, mistral etc., examples also.
> 
> Updated opt/phi/mistral.
> 
> But for open_llama, there is the extra input `use_cache` / `use_cache_branch`. I need to kick off another round updates in this PR.

For the openllama example, maybe we can move away from the legacy merged model to the single model we use for the llama2 example? the merged model is an old solution and not used much anymore.

olive/data/template.py
devang-ml: I don't see any use of this function currently.

docs/source/tutorials/configure_data.rst
```diff
@@ -185,6 +185,42 @@ Convert the raw data config to the data container.
...
+    .. tab:: Config JSON
+
+        .. code-block:: json
+
+            {
+                "name": "transformers_dummy_data_config",
+                "type": "TransformersDummyDataContainer",
+                "load_dataset_config": {
+                    "params": {
+                        "model_name": "meta-llama/Llama-2-7b-hf"
```
devang-ml, line 201: Do you need model_name ? "input_model" has that information so there is no need to duplicate it in the .json config.

trajepl: If it is the workflow-level json config, user do not need to provide, since befor start `olive engine`, olive will try to validate the input json config and try to fill-in the corresponding value.
https://github.com/microsoft/Olive/blob/d930b2c001961e7aac7e15c37502e147ee99d7b7/olive/workflows/run/config.py#L210

But if only data_config or user use the python api to create data_config, they have to provide `model_name`.
I added the comments here.
image

I was thinking the example in `configure_data.rst` should be executable solely. Then I added the model_name.
Please let me know if you want to remove.

trajepl:
> > > Please update opt, phi, mistral etc., examples also.
> > 
> > 
> > Updated opt/phi/mistral.
> > But for open_llama, there is the extra input `use_cache` / `use_cache_branch`. I need to kick off another round updates in this PR.
> 
> For the openllama example, maybe we can move away from the legacy merged model to the single model we use for the llama2 example? the merged model is an old solution and not used much anymore.

Updated.

olive/data/template.py
trajepl: Just removed. Anyone can add if they need it like we did for `dummy data config, huggingface data config`.

docs/source/tutorials/configure_data.rst
devang-ml: Please remove. Thanks!

examples/llama2/llama2_generate.json
```diff
@@ -60,6 +60,24 @@
...
+            "name": "transformer_prompt_dummy_data",
+            "type": "TransformersPromptDummyDataContainer",
+            "load_dataset_config": {
+                "params": {
+                    "generative": true
+                }
+            }
+        },
+        {
+            "name": "transformer_token_dummy_data",
```
jambayk, line 74: looks like this is not used?

examples/llama2/llama2_generate.json
```diff
@@ -89,44 +107,8 @@
                             "priority": 1
                         }
                     ],
+                    "data_config": "transformer_prompt_dummy_data",
```
jambayk, line 110: maybe rename this metric to `generation_latency`?

trajepl: Updated

examples/llama2/llama2_generate.json
trajepl: Removed.

olive/data/component/dataloader.py
```diff
@@ -3,14 +3,25 @@
...
 
 from olive.data.registry import Registry
 
 
 @Registry.register_default_dataloader()
 def default_dataloader(dataset, batch_size=1, **kwargs):
-    return DataLoader(dataset, batch_size=batch_size, **kwargs)
+    def ignore_batch_collate_fn(batch):
+        # ignore to batch the data fields for given inputs
+        to_ignore_fields = kwargs.get("ignore_fields") or ["step"]
```
jambayk, line 15: hard-coding `["step"]` might not be a good idea here. especially since this is the `default_dataloader` component.

jambayk (approved):

PR #1220, opened 2024-06-30
Update OrtPerfTuning pass to accept data only using data_config

shaahji:

### DIFF
--- a/examples/directml/squeezenet/squeezenet_config.json
+++ b/examples/directml/squeezenet/squeezenet_config.json
@@ -56,10 +56,7 @@
         "perf_tuning": {
             "type": "OrtPerfTuning",
             "config": {
-                "user_script": "user_script.py",
-                "dataloader_func": "create_dataloader",
                 "device": "gpu",
-                "batch_size": 1,
                 "execution_mode_list": [ "ORT_SEQUENTIAL" ],
                 "providers_list": [ "DmlExecutionProvider" ]
             }
--- a/examples/phi2/phi2_optimize_template.json
+++ b/examples/phi2/phi2_optimize_template.json
@@ -89,7 +89,7 @@
                     ],
                     "user_config": {
                         "user_script": "user_script.py",
-                        "dataloader_func": "create_dataloader",
+                        "dataloader_func": "phi2_dataloader",
                         "batch_size": 2,
                         "inference_settings": {
                             "onnx": {
@@ -200,9 +200,7 @@
         "perf_tuning": {
             "type": "OrtPerfTuning",
             "config": {
-                "user_script": "user_script.py",
-                "dataloader_func": "create_dataloader",
-                "batch_size": 2,
+                "data_config": "perf_tuning_data_config",
                 "enable_profiling": false
             }
         }
@@ -213,7 +211,7 @@
             "type": "HuggingfaceContainer",
             "user_script": "user_script.py",
             "load_dataset_config": {
-                "type": "load_tiny_code_dataset",
+                "type": "tiny_code_dataset",
                 "params": {
                     "data_name": "nampdn-ai/tiny-codes",
                     "split": "train",
@@ -258,6 +256,20 @@
             "pre_process_data_config": {
                 "params": {
                     "source_max_len": 2048
+                }
+            }
+        },
+        {
+            "name": "perf_tuning_data_config",
+            "type": "DummyDataContainer",
+            "user_script": "user_script.py",
+            "load_dataset_config": {
+                "type": "local_dataset"
+            },
+            "dataloader_config": {
+                "type": "phi2_dataloader",
+                "params": {
+                    "batch_size": 2
                 }
             }
         }
--- a/examples/resnet/resnet_dynamic_ptq_cpu.json
+++ b/examples/resnet/resnet_dynamic_ptq_cpu.json
@@ -28,6 +28,28 @@
             }
         }
     },
+    "data_configs": [
+        {
+            "name": "cifar10_data_config",
+            "type": "HuggingfaceContainer",
+            "user_script": "user_script.py",
+            "load_dataset_config": {
+                "type": "cifar10_dataset",
+                "params": {
+                    "data_dir": "data"
+                }
+            },
+            "pre_process_data_config": {
+                "type": "skip_pre_process"
+            },
+            "dataloader_config": {
+                "params": {
+                    "batch_size": 16,
+                    "drop_last": true
+                }
+            }
+        }
+    ],
     "evaluators": {
         "common_evaluator":{
             "metrics":[
@@ -84,10 +106,7 @@
         "perf_tuning": {
             "type": "OrtPerfTuning",
             "config": {
-                "user_script": "user_script.py",
-                "dataloader_func": "create_dataloader",
-                "batch_size": 16,
-                "data_dir": "data"
+                "data_config": "cifar10_data_config"
             }
         }
     },
--- a/olive/passes/onnx/perf_tuning.py
+++ b/olive/passes/onnx/perf_tuning.py
@@ -9,19 +9,20 @@ import logging
 import tempfile
 import time
 from pathlib import Path
-from typing import Any, Callable, Dict, Union
+from typing import Any, Dict, Union
 
+from olive.common.config_utils import validate_config
 from olive.common.ort_inference import check_and_normalize_provider_args
 from olive.data.config import DataConfig
+from olive.data.template import dummy_data_config_template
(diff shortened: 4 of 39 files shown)


examples/directml/squeezenet/squeezenet_config.json
```diff
@@ -23,6 +23,12 @@
             }
         }
     },
+    "data_configs": [
+        {
+            "name": "perf_tuning_data_config",
```
jambayk, line 28: the type says dummydatacontainer but there is no configuration for any of the components. is this a valid data config?

examples/phi2/phi2_optimize_template.json
```diff
@@ -260,6 +258,17 @@
                     "source_max_len": 2048
                 }
             }
+        },
+        {
+            "name": "perf_tuning_data_config",
+            "type": "DummyDataContainer",
```
jambayk, line 264: this says dummydatacontainer but looks like this data config is supposed to only have a `dataloader_config`? What happens with the load_data component? Will it work even when the input shapes are missing?

examples/resnet/resnet_dynamic_ptq_cpu.json
```diff
@@ -28,6 +28,28 @@
             }
         }
     },
+    "data_configs": [
+        {
+            "name": "cifar10_data_config",
+            "type": "HuggingfaceContainer",
```
jambayk, line 34: I think it might be better to introduce an empty container type? This config is huggingface container in name only and doesn't behave like a regular hf container since almost all component types are overridden. 

Same applies to the dummydatacontainer config in the phi2 example.

olive/passes/onnx/perf_tuning.py
```diff
@@ -313,24 +290,37 @@ def __init__(self, accelerator_spec: AcceleratorSpec, config: Dict[str, Any], da
...
-        if config_dict.get("dataloader_func_kwargs"):
-            latency_user_config["func_kwargs"] = {"dataloader_func": config_dict.get("dataloader_func_kwargs")}
+        if not self.config.data_config:
+            if model.io_config:
+                if not is_io_config_static(model.io_config):
+                    # since Olive will not save the pytorch model's io_config to olive onnx model
+                    # we cannot generate dummy data for the onnx model if this model has dynamic input shapes
+                    # TODO(trajep): try to get static input shapes from onnx model.
+                    # If so, we can move the dataloader for latency measurement.
+                    logger.debug(
```
jambayk, line 300: should we make this a failure since this case is bound to fail?

examples/directml/squeezenet/squeezenet_config.json
shaahji: Yes, this would use `dummy_dataset` and `no_auto_batch_dataloader` because those are setup as default values in the definition of `DummyDataContainer`.

jambayk: doesn't the dummydatacontainer require `input_shapes` in load data component?

examples/phi2/phi2_optimize_template.json
shaahji: Not true. The `dataloader_config` is defaulted to `no_auto_batch_dataloader`.

examples/resnet/resnet_dynamic_ptq_cpu.json
shaahji: I had the same thought but an empty data config would still need to return something from `load_dataset_config` to `dataloader_config`. I don't know if an empty `DataConfig` would be a good idea.

olive/passes/onnx/perf_tuning.py
shaahji: Keeping the same behavior as earlier. Handling of `io_config` should be its own PR.

examples/phi2/phi2_optimize_template.json
jambayk: what I meant is this config doesn't appear to be using the dummy data container's load dataset component. it only needs the dataloader since it just ignores the dataset. 

Similar to https://github.com/microsoft/Olive/pull/1220#discussion_r1669373510, I don't know if the data config is even valid since it's missing input shapes. 

That's why I suggested https://github.com/microsoft/Olive/pull/1220#discussion_r1669361944

examples/resnet/resnet_dynamic_ptq_cpu.json
jambayk: for example the updated phi2 example is a case where the dataset passed to dataloader is completely ignored.

jambayk: perhaps in the case of the "empty" ("custom") data container, it should be the responsibility of the user to provide components that are compatible with one another and where they want to use it? Although this might just encourage users to wrap the pre-existing "dataloader_func" with this container like how the phi2 example in this PR is doing

examples/directml/squeezenet/squeezenet_config.json
shaahji: Depends on the use case. When used in the context of perf tuning, this information comes from `input_model/io_config`.
Here in this case, the data config is used only for perf tuning.

examples/phi2/phi2_optimize_template.json
shaahji: Same here. Since the data config is used only for the purpose of perf tuning, the `input_shapes` come from `input_model/io_config`. However, the dataset that get created will be ignored since `create_phi2_dataloader` in phi2/user_script.py` ignores it and creates its own dataset/dataloder.

This is confusing and a roundabout way of implementing it, because phi2 doesn't actually implement an explicit dataset - it only has a dataloader. We could define an `EmpyDataContainer` but, in my opinion, that would be wrong path forward. Users should implement a dataset and not a dataloader. DataLoader is generic collecting logic and should be _smart_ - query the dataset batch size times and return the data - and so there shouldn't be a need to implement explicit DataLoaders.

jambayk: maybe in this example, it should explicitly bypass the load data component. or else, I think this would also fail without the input shapes https://github.com/microsoft/Olive/blob/7e6f5129a0d486c7a00cd27c62f57cd33dfe7922/olive/data/component/load_dataset.py#L38

jambayk: but like in the other discussion below, my concern is - if we are overriding all components, then using the type as dummydatacontainer or huggingfacecontainer is a bit confusing and/or misleading.

examples/directml/squeezenet/squeezenet_config.json
shaahji: I updated the logic in `OliveEvaluator` to handle this scenario. In my opinion, providing a built-in definition for `EmptyContainer` is not a good idea. Such implementation will push users to do the minimum (i.e. implement the `DataContainer`) and break the required modularity. Such implementation will also prevent us from implementing any possible optimizations, in future, that would bank heavily on these components being modular.

examples/phi2/phi2_optimize_template.json
shaahji: I can see that as an argument but I don't think `EmptyContainer` is a solution. Needs further discussion and so will push it for a follow up PR.

examples/resnet/resnet_dynamic_ptq_cpu.json
shaahji: Commented above.

olive/evaluator/olive_evaluator.py
```diff
@@ -236,12 +234,28 @@ def generate_metric_user_config_with_model_io(metric: Metric, model: "OliveModel
                 "Model input shapes are not static. Cannot use inferred input shapes for creating dummy data. This will"
                 " cause an error when creating dummy data for tuning."
             )
-        if io_config and not metric.user_config.input_names and not metric.user_config.input_shapes:
+
+        if metric.data_config:
```
jambayk, line 238: I don't think this is a good idea. If a user chose the type as `dummydatacontainer` and overrode the load_dataset parameter (like how the hf container is overridden in the resnet examples) these params are not relevant. It also doesn't sound like a good idea for the evaluator to update a pre-existing container. 

Since the only use case right now is the squeezenet example. Why not just remove the `data_config` value passed in the perf tuning config. the perf tuning pass is able to create it's own data config just like this if the data config is not provided.

shaahji: Updated.

examples/directml/squeezenet/squeezenet_config.json
shaahji: Removed.

jambayk (approved):

PR #1224, opened 2024-07-08
Report duplicate DataComponentType registration as fatal

shaahji:
## Report duplicate DataComponentType registration as fatal

Duplicate registration causes undesirable results and hard to debug errors.

### DIFF
--- a/olive/data/registry.py
+++ b/olive/data/registry.py
@@ -2,6 +2,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+import inspect
 import logging
 from typing import ClassVar, Dict, Union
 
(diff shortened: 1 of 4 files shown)


olive/data/registry.py
```diff
@@ -36,12 +37,27 @@ def register(cls, sub_type: Union[DataComponentType, DataContainerType], name: s
 
         def decorator(component):
             component_name = name if name is not None else component.__name__
-            if component_name in cls._REGISTRY[sub_type.value]:
-                # don't want to warn here since user script is loaded everytime data config is initialized
-                # there is nothing user can do to fix this warning
-                logger.debug(
-                    "Component %s already registered in %s, will override the old one.", component_name, sub_type.value
-                )
+            if (component_name in cls._REGISTRY[sub_type.value]) and (type(component).__name__ != "MagicMock"):
```
jambayk, line 40: do we know where the "MagicMock" case is coming from? Would it be possible to fix from there instead of adding a case for a unit test scenario into source code?

shaahji: It's coming from `test_data_root::test_data_root_for_dataset`. I tried a few different ways to implement that test but none worked. Ideas welcome!

jambayk: This commit works by patching `inspect` https://github.com/microsoft/Olive/commit/be76b69b937c36a9dab80f319c9a05553d747be2

shaahji: Updated.

jambayk (dismissed):

jambayk (approved):

PR #1228, opened 2024-07-11
Use DataConfig based metric configuration

shaahji:
## Use DataConfig based metric configuration

* Deprecated dataloader_func, func_args, data_dir, batch_size, input_names, input_shapes, and input_types for metric configuration
* Default data component is still a data component and should be registered as such. Also, certain data component can be registered with multiple different names.
* Fixed a bug in docker system where the config wasn't being serialized correctly.
* Added support for DataConfig/user_script and DataConfig/script_dir to both AMLSystem and DockerSystem.
* Updated all examples and tests to use the DataConfig based metric configuration.

### DIFF
--- a/examples/bert/nv_user_script.py
+++ b/examples/bert/nv_user_script.py
@@ -3,27 +3,28 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 import torch
-from datasets import load_dataset  # type: ignore[import]
 from datasets.utils import logging as datasets_logging  # type: ignore[import]
 from transformers import AutoTokenizer, BertModel  # type: ignore[import]
+
+from olive.data.registry import Registry
 
 datasets_logging.disable_progress_bar()
 datasets_logging.set_verbosity_error()
 
 
-def load_pytorch_origin_model():
+def load_pytorch_origin_model(model_path):
     model = BertModel.from_pretrained("bert-base-uncased")
     model.eval()
     return model
 
 
-def create_calibration_dataloader(batch_size, calib_size, *args, **kwargs):
+@Registry.register_dataloader("nvmo_calibration_dataloader")
+def create_calibration_dataloader(dataset, batch_size, calib_size=64, **kwargs):
     tokenizer = AutoTokenizer.from_pretrained("bert-base-uncased")
 
     def tokenization(example):
         return tokenizer(example["text"], padding="max_length", max_length=128, truncation=True)
 
-    dataset = load_dataset("rotten_tomatoes", split="validation[:10%]")
     dataset = dataset.map(tokenization, batched=True)
     dataset.set_format(type="torch", columns=["input_ids", "token_type_ids", "attention_mask"])
 
--- a/examples/resnet/user_script.py
+++ b/examples/resnet/user_script.py
@@ -72,7 +72,8 @@ class PytorchResNetDataset(Dataset):
 # -------------------------------------------------------------------------
 
 
-def post_process(output):
+@Registry.register_post_process()
+def cifar10_post_process(output):
     # max_elements, max_indices = torch.max(input_tensor, dim)
     # This is a two classes classification task, result is a 2D array [[ 1.1541, -0.6622],[[-0.2137,  0.0360]]]
     # the index of this array among dimension 1 will be [1, 0], which are labels of this task
@@ -81,18 +82,13 @@ def post_process(output):
 
 
 # -------------------------------------------------------------------------
-# Dataloader for Evaluation and Performance Tuning
+# DataSet for Evaluation and Performance Tuning
 # -------------------------------------------------------------------------
 
 
 @Registry.register_dataset()
 def cifar10_dataset(data_dir):
     return PytorchResNetDataset(CIFAR10DataSet(data_dir).val_dataset)
-
-
-# TODO(shaahji): Remove this once metrics transitions to using DataConfig only!
-def create_dataloader(data_dir, batch_size, *args, **kwargs):
-    return DataLoader(cifar10_dataset(data_dir), batch_size=batch_size, drop_last=True)
 
 
 # -------------------------------------------------------------------------
@@ -130,7 +126,7 @@ def resnet_calibration_reader(data_dir, batch_size, *args, **kwargs):
 # keep this to demo/test custom evaluation function
 def eval_accuracy(model: OliveModelHandler, data_dir, batch_size, device, execution_providers):
     sess = model.prepare_session(inference_settings=None, device=device, execution_providers=execution_providers)
-    dataloader = create_dataloader(data_dir, batch_size)
+    dataloader = DataLoader(cifar10_dataset(data_dir), batch_size=batch_size, drop_last=True)
 
     preds = []
     target = []
@@ -148,14 +144,14 @@ def eval_accuracy(model: OliveModelHandler, data_dir, batch_size, device, execut
                 result = torch.Tensor(res[0])
             else:
                 result = torch.Tensor(res)
-            outputs = post_process(result)
+            outputs = cifar10_post_process(result)
             preds.extend(outputs.tolist())
             target.extend(labels.data.tolist())
 
     elif model.framework == Framework.PYTORCH:
         for input_data, labels in dataloader:
             result = model.run_session(sess, input_data)
-            outputs = post_process(result)
+            outputs = cifar10_post_process(result)
             preds.extend(outputs.tolist())
             target.extend(labels.data.tolist())
 
--- a/olive/data/component/post_process_data.py
+++ b/olive/data/component/post_process_data.py
@@ -8,8 +8,20 @@ import transformers
 from olive.data.registry import Registry
 
 
+@Registry.register_post_process()
 @Registry.register_default_post_process()
+@Registry.register_post_process("skip_post_process")
 def post_process(output_data, **kwargs):
+    """Post-process data.
+
+    Args:
+        output_data (object): Model output to be post-processed.
+        **kwargs: Additional named arguments.
+
+    Returns:
+        object: Post-processed data.
+
+    """
     return output_data
 
 
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -14,14 +14,15 @@ from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, NamedTuple, Tuple, 
 import numpy as np
 import torch
 
-import olive.data.template as data_config_template
-from olive.common.config_utils import ConfigBase
+from olive.common.config_utils import ConfigBase, validate_config
 from olive.common.ort_inference import OrtInferenceSession, prepare_io_bindings
 from olive.common.pydantic_v1 import validator
 from olive.common.user_module_loader import UserModuleLoader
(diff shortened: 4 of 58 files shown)


test/integ_test/evaluator/azureml_eval/utils.py
```diff
@@ -10,89 +10,99 @@
...
 from olive.systems.azureml import AzureMLDockerConfig, AzureMLSystem
 
 # pylint: disable=redefined-outer-name
 
+_current_dir, _models_dir, _data_dir, _user_script = None, None, None, None
+
 
 def get_directories():
-    current_dir = Path(__file__).resolve().parent
+    global _current_dir, _models_dir, _data_dir, _user_script
```
jambayk, line 24: Since pylint is complaining about the global variables, I think you could implement this get_directories as a pytest fixture like in https://github.com/microsoft/Olive/blob/fc4fc903f09da82d3205a79a011a7bf89089f955/test/unit_test/workflows/test_whisper_workflow.py#L23 

If you scope it as "module", it would only run through the contents once if you don't want to be executed multiple times.

jambayk: looks like it's already updated to not use globals

shaahji: Implemented it as recommended by PyLint docs.

olive/evaluator/olive_evaluator.py
```diff
@@ -216,15 +215,18 @@ def evaluate(
 
     @staticmethod
     def generate_metric_user_config_with_model_io(metric: Metric, model: "OliveModelHandler"):
```
jambayk, line 217: if we want the previous behavior, I think the logic is: 
- if metric.data_config, return
- if not io_config, return
- if not io_config.input_shapes, return
- if not is_io_config_static, return 
- create new dummy config from io_config

If the metric comes with its own data config, then we should not do anything with it even though the type is dummy_dataset.

jambayk: in the perf tuning pass, it can just pass data_config is None if not provided.

shaahji: > If the metric comes with its own data config, then we should not do anything with it even though the type is dummy_dataset.

That would break auto perf tuning. This pass is automatically added to the workflow and so don't come with a data config. In some other cases, the metric defines a dummy dataset but doesn't provide the required input_xxxx fields. I think there a lot of assumptions about how/when/where we get this input if not provided. 

Your proposed logic will break a lot of examples and tests. I don't like the conditional branching either but this is the safe solution to move forward for this PR. I can add another task on my plate to revisit this and remove the branching.

shaahji: > create new dummy config from io_config

This assumes that the logic only runs for the latency metric, which in itself is a bad assumption at this level. At the perf tuning level, it is acceptable but not here.

jambayk: I think my proposed logic is what is currently happening. The priority was dataloader_func > data_config > metric io_config > model_io_config. 

In my proposed logic, (dataloader_func and data_config) is 1, metric io_config is also step 1 with user provided data config, rest of the logic is for model io_config

jambayk: if the perf tuning pass just provides none as data config if not provided, the evaluator would automatically fall back on the model's io_config.

shaahji: > if the perf tuning pass just provides none as data config if not provided, the evaluator would automatically fall back on the model's io_config.

Evaluator never created the `DataConfig` itself. It only populated the config if the expected keys were not found. You are proposing creating `DataConfig` in evaluator with default type being 'DummyDataContainer`.  That is not the same.

Currently, PerfTuning pass sends in a data config with the `dummy_dataset` and the evaluator is only populating the fields. Instead of the PerfTuning it could be any other pass or coming directly from user config as well.

However, in your logic, if the user provide a dummy data config without providing the input_xxx the workflow will fail. Your logic forces the user to repeat what's already in the input model.

jambayk: Evaluator created the data config here https://github.com/microsoft/Olive/blob/fc4fc903f09da82d3205a79a011a7bf89089f955/olive/evaluator/olive_evaluator.py#L309. This case is only true if the conditions I listed above are met (except for the static io condition which is needed but was overlooked)

if the user wants the models io config to be used, they can just choose to not provide a `data_config` to the metric config. That's the behavior currently. they don't provide dataloader_func or data config. 
the perf tuning can do the same

jambayk: https://github.com/microsoft/Olive/blob/fc4fc903f09da82d3205a79a011a7bf89089f955/olive/evaluator/olive_evaluator.py#L194

shaahji: Updated.

olive/data/component/post_process_data.py
```diff
@@ -8,8 +8,20 @@
 from olive.data.registry import Registry
 
 
+@Registry.register_post_process()
```
jambayk, line 11: is there a reason it is registered using this also?

olive/evaluator/olive_evaluator.py
```diff
@@ -141,10 +142,12 @@ def _evaluate_custom(
     ) -> MetricResult:
         raw_res = None
         if metric.user_config.evaluate_func:
+            data_dir = metric.data_config.load_dataset_params.get("data_dir") if metric.data_config else None
+            batch_size = metric.data_config.dataloader_params.get("batch_size", 1) if metric.data_config else 1
             raw_res = eval_func(
```
jambayk, line 147: since, I already made a PR which is able to automatically detect data resources, I think it might be simpler here to pass any values the evaluate func needs as `evaluate_kwargs`. This way we don't need a data config for a custom evaluator func.

examples/resnet/user_script.py
```diff
@@ -130,7 +126,7 @@ def resnet_calibration_reader(data_dir, batch_size, *args, **kwargs):
 # keep this to demo/test custom evaluation function
 def eval_accuracy(model: OliveModelHandler, data_dir, batch_size, device, execution_providers):
```
jambayk, line 127: per https://github.com/microsoft/Olive/pull/1228/files#r1681720056, I think we could make `data_dir` and `batch_size` kwargs here.

olive/evaluator/olive_evaluator.py
```diff
@@ -226,87 +230,46 @@ def generate_metric_user_config_with_model_io(metric: Metric, model: "OliveModel
                 "Model input shapes are not static. Cannot use inferred input shapes for creating dummy data. This will"
```
jambayk, line 230: this can be a direct failure since we know static shapes are needed to create dummy data for latency metric.

test/integ_test/evaluator/azureml_eval/utils.py
```diff
@@ -10,89 +10,99 @@
...
 
 def get_onnx_model():
-    return str(models_dir / "model.onnx")
+    return str(_models_dir / "model.onnx")
 
 
 def delete_directories():
-    shutil.rmtree(data_dir)
-    shutil.rmtree(models_dir)
+    global _current_dir, _models_dir, _data_dir, _user_script
```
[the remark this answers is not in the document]
jambayk: resolved

examples/bert/nv_user_script.py
```diff
@@ -3,27 +3,28 @@
...
-def load_pytorch_origin_model():
+def load_pytorch_origin_model(model_path):
     model = BertModel.from_pretrained("bert-base-uncased")
     model.eval()
     return model
 
 
-def create_calibration_dataloader(batch_size, calib_size, *args, **kwargs):
+@Registry.register_dataloader("nvmo_calibration_dataloader")
+def create_calibration_dataloader(dataset, batch_size, calib_size=64, *args, **kwargs):
```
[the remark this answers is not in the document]
jambayk: resolved

test/integ_test/evaluator/docker_eval/utils.py
```diff
@@ -10,114 +10,125 @@
...
 from olive.systems.docker import DockerSystem, LocalDockerConfig
 
 # pylint: disable=redefined-outer-name
 
+_current_dir, _models_dir, _data_dir, _user_script = None, None, None, None
+
 
 def get_directories():
-    current_dir = Path(__file__).resolve().parent
+    global _current_dir, _models_dir, _data_dir, _user_script
```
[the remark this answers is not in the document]
jambayk: resolved

test/integ_test/evaluator/docker_eval/utils.py
```diff
@@ -10,114 +10,125 @@
...
 
 def get_openvino_model():
-    return {"model_path": str(models_dir / "openvino")}
+    return {"model_path": str(_models_dir / "openvino")}
 
 
 def delete_directories():
-    shutil.rmtree(data_dir)
-    shutil.rmtree(models_dir)
+    global _current_dir, _models_dir, _data_dir, _user_script
```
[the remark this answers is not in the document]
jambayk: resolved

test/integ_test/evaluator/local_eval/utils.py
```diff
@@ -7,163 +7,118 @@
...
-    import transformers
-
-    if isinstance(res, transformers.modeling_outputs.SequenceClassifierOutput):
-        _, preds = torch.max(res.logits, dim=1)
-    else:
-        _, preds = torch.max(res, dim=1)
-    return preds
-
+def get_directories():
+    global _current_dir, _models_dir, _data_dir, _user_script
```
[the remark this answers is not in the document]
jambayk: resolved

test/integ_test/evaluator/local_eval/utils.py
```diff
@@ -7,163 +7,118 @@
...
-        zip_ref.extractall(models_dir)
-    return {"model_path": str(models_dir / "openvino")}
+        zip_ref.extractall(_models_dir)
+    return {"model_path": str(_models_dir / "openvino")}
 
 
 def delete_directories():
-    shutil.rmtree(data_dir)
-    shutil.rmtree(models_dir)
+    global _current_dir, _models_dir, _data_dir, _user_script
```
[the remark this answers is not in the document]
jambayk: resolved

test/multiple_ep/utils.py
```diff
@@ -17,59 +18,67 @@
 
 # pylint: disable=redefined-outer-name
 
+_current_dir, _models_dir, _data_dir, _user_script = None, None, None, None
+
 
 def get_directories():
-    current_dir = Path(__file__).resolve().parent
+    global _current_dir, _models_dir, _data_dir, _user_script
```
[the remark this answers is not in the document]
jambayk: resolved

olive/data/component/post_process_data.py
shaahji: With default registration alone, it can't explicitly be used in json configs. A default component type is still a legal component type to be used in config and this registration allows that.

olive/evaluator/olive_evaluator.py
shaahji: This is to keep the old behavior intact. This is not a failure in current implementation.

jambayk (approved):

PR #1230, opened 2024-07-12
AzureMLSystem: Generalize resource mount

jambayk:
## Describe your changes
- Implemented a `resource_path` utils called `find_all_resources` which traverses a config dictionary and returns all resources (local files/folders, aml resources)
- Instead of hard coding expected resource params or inferring them from `ParamCategory`, the AMLSystem now uses automatically finds the resources and mounts them. It also keeps tracks of the dict keys to update them in the runners.
- Removed the ort version check in aml pass runner since 1.16 is too old so we don't need to officially support it.

In follow up PRs:
- the DockerSystem will also be updated to use the same logic. 
- Migrate the `_prepare_non_local_model` logic from engine into the DockerSystem and LocalSystem so that the engine is agnostic to resources completely.

### DIFF
--- a/examples/test/test_resnet_ptq_cpu.py
+++ b/examples/test/test_resnet_ptq_cpu.py
@@ -31,7 +31,7 @@ def setup():
 
 @pytest.mark.parametrize("search_algorithm", ["random"])
 @pytest.mark.parametrize("execution_order", ["pass-by-pass"])
-@pytest.mark.parametrize("system", ["aml_system"])
+@pytest.mark.parametrize("system", ["local_system", "aml_system"])
 @pytest.mark.parametrize(
     "olive_json",
     [
--- a/examples/test/test_resnet_ptq_cpu.py
+++ b/examples/test/test_resnet_ptq_cpu.py
@@ -8,7 +8,7 @@ from pathlib import Path
 import pytest
 from onnxruntime import __version__ as OrtVersion
 from packaging import version
-from utils import check_output, patch_config, set_azure_identity_logging
+from utils import check_output, patch_config
 
 from olive.common.utils import retry_func, run_subprocess
 
@@ -24,14 +24,14 @@ def setup():
     # retry since it fails randomly
     retry_func(run_subprocess, kwargs={"cmd": "python prepare_model_data.py", "check": True})
 
-    set_azure_identity_logging()
+    # set_azure_identity_logging()
     yield
     os.chdir(cur_dir)
 
 
 @pytest.mark.parametrize("search_algorithm", ["random"])
 @pytest.mark.parametrize("execution_order", ["pass-by-pass"])
-@pytest.mark.parametrize("system", ["local_system", "aml_system"])
+@pytest.mark.parametrize("system", ["aml_system"])
 @pytest.mark.parametrize(
     "olive_json",
     [
--- a/examples/test/utils.py
+++ b/examples/test/utils.py
@@ -79,8 +79,12 @@ def update_azureml_config(olive_config):
     if workspace_name is None:
         raise Exception("Please set the environment variable WORKSPACE_NAME")
 
+    exclude_managed_identity_credential = (
+        {"exclude_managed_identity_credential": True} if "EXCLUDE_MANAGED_IDENTITY_CREDENTIAL" in os.environ else {}
+    )
+
     client_id = os.environ.get("MANAGED_IDENTITY_CLIENT_ID")
-    if client_id is None:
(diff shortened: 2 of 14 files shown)


examples/test/test_resnet_ptq_cpu.py
```diff
@@ -24,7 +24,7 @@ def setup():
     # retry since it fails randomly
     retry_func(run_subprocess, kwargs={"cmd": "python prepare_model_data.py", "check": True})
 
-    set_azure_identity_logging()
+    # set_azure_identity_logging()
```
trajepl, line 27: Any reason to disable this logging?

examples/test/utils.py
```diff
@@ -79,16 +79,20 @@ def update_azureml_config(olive_config):
     if workspace_name is None:
         raise Exception("Please set the environment variable WORKSPACE_NAME")
 
+    exclude_managed_identity_credential = (
+        {"exclude_managed_identity_credential": True} if "EXCLUDE_MANAGED_IDENTITY_CREDENTIAL" in os.environ else {}
```
trajepl, line 83: Is this `EXCLUDE_MANAGED_IDENTITY_CREDENTIAL` set in our pipeline variables?

olive/common/utils.py
```diff
@@ -151,6 +151,26 @@ def flatten_dict(dictionary, stop_condition=None):  # pragma: no cover
     return result
 
 
+def get_dict_value(dictionary: dict, key: Union[str, Tuple, List[str]]):
+    """Get value from a nested dictionary."""
```
trajepl, line 155: Maybe update the function name to indicate that it is used to get value from nested dictionary.

olive/common/utils.py
```diff
@@ -151,6 +151,26 @@ def flatten_dict(dictionary, stop_condition=None):  # pragma: no cover
...
+    if isinstance(key, str):
+        key = [key]
+
+    for k in key:
+        dictionary = dictionary[k]
+    return dictionary
+
+
+def set_dict_value(dictionary: dict, key: Union[str, Tuple, List[str]], new_value):
```
trajepl, line 164: Ditto.

trajepl (approved):

examples/test/test_resnet_ptq_cpu.py
jambayk: the azure identity logs are too verbose and makes it hard to see the relevant information. This was added by mike when we had azure credential issues due to permissions but it's not needed anymore.

trajepl: Ic, maybe we can try to remove it later. Or you want to keep it to remind the follwoing developer to use it for debugging purpose?

jambayk: just kept it for now if we need to debug similar issues but I can remove it in the follow up.

examples/test/utils.py
jambayk: This is for local testing purposes. Azure VMs have system managed identity so it messes with the azure login. We can set this env so that it uses our azure cli login instead.

olive/common/utils.py
jambayk: Sure. I can update it in the follow up. Maybe to something like `get_nested_dict_value`?

examples/test/utils.py
xiaoyu-work: I always add this config by myself. Probably we can add this as a note to our doc.

olive/resource_path.py
```diff
@@ -107,32 +107,32 @@ def create_resource_path(self) -> ResourcePath:
...
+
 def create_resource_path(
-    resource_path: Optional[Union[str, Path, Dict[str, Any], ResourcePathConfig, ResourcePath]]
+    resource_path: OLIVE_RESOURCE_ANNOTATIONS,
 ) -> Optional[Union[ResourcePath, List[ResourcePath]]]:
     """Create a resource path from a string or a dict.
 
-    If a string is provided, it is inferred to be a file, folder, or string name.
-    If a Path is provided, it is inferred to be a file or folder.
+    If a string or Path is provided, it is inferred to be a file, folder, or string name.
```
xiaoyu-work, line 119: This change might be more confusing.

jambayk: The logic did not change. the doc was not update when the behavior was changed by https://github.com/microsoft/Olive/pull/352 so I updated it to fix the doc. 

Looks like sometimes, this is being called with `Path` object even though it is a string name. Might be because of some validator somewhere but it is not related to this PR.

PR #1232, opened 2024-07-12
CI: Separate Bert CUDA local and aml test

jambayk:
## Describe your changes
- Separate the bert cuda example test into local and aml tests
- AML test:
  - doesn't require to be run on a GPU pipeline agent. 
  - Reduced the cases since we only want to test CUDA EP on aml. No need to test the pass_flows or search strategies. Saves test time since the aml gpu cluster is small and has longer queue times.
  - Updated base image with cuda 11.8. Both torch and ort don't support 11.6 anymore
  - Updated the conda yamls

### DIFF
--- a/examples/bert/bert_cuda_gpu.json
+++ b/examples/bert/bert_cuda_gpu.json
@@ -118,7 +118,6 @@
         "perf_tuning": {
             "type": "OrtPerfTuning",
             "config": {
-                "enable_cuda_graph": true,
                 "data_config": "glue_mrpc"
             }
         }
--- a/examples/test/test_bert_cuda_gpu_aml.py
+++ b/examples/test/test_bert_cuda_gpu_aml.py
@@ -19,16 +19,17 @@ def setup():
     os.chdir(cur_dir)
 
 
+# NOTE: this doesn't need to run on a host with a GPU, as it's testing the CUDA EP on AML
 @pytest.mark.parametrize("search_algorithm", ["tpe"])
-@pytest.mark.parametrize("execution_order", ["joint", "pass-by-pass"])
-@pytest.mark.parametrize("system", ["local_system"])
+@pytest.mark.parametrize("execution_order", ["joint"])
+@pytest.mark.parametrize("system", ["aml_system"])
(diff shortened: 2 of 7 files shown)


examples/test/test_bert_cuda_gpu_aml.py
```diff
@@ -0,0 +1,35 @@
+# -------------------------------------------------------------------------
```
xiaoyu-work, line 1: Can you add this test to https://github.com/microsoft/Olive/blob/main/examples/test/test_bert_ptq_cpu_aml.py? We have too many bert tests I'm already confused now lol.

examples/bert/bert_cuda_gpu.json
```diff
@@ -118,7 +118,6 @@
         "perf_tuning": {
             "type": "OrtPerfTuning",
             "config": {
-                "enable_cuda_graph": true,
```
xiaoyu-work, line 121: Will this speed up the test?

examples/test/test_bert_cuda_gpu_aml.py
jambayk: I separated them since the aml tests take time so it's better to run them in parallel.

jambayk: the bert cpu aml already itself takes over an hour. cuda aml also takes time since it needs to queue for the gpu cluster.

examples/bert/bert_cuda_gpu.json
jambayk: This is not for test speed up. turns out the onnx model here is not compatible with cuda graph so perf tuning only returned the input model as best model. so I disabled it so that the perf tuning pass here actually finds something useful.

examples/test/test_bert_cuda_gpu_aml.py
jambayk: I think we need to spend some time in the future to re-evaluate our example tests and probably break them out into their own pipeline.

xiaoyu-work (approved):

PR #1235, opened 2024-07-15
Remove `data_root`, `prepare_resources_for_local` for all configs

jambayk:
## Describe your changes
- Remove the concept of `data_root` entirely. It is not used at all in any of our examples and we only have one customer who use a pinned version of Olive. `data_root` is too complicated to maintain as it is tied into all parts of Olive and most passes, etc ignore it. The cost of maintaining the feature is not worth it.
  - If needed, we can revisit this requirement and maybe create a new type of resource path that gets resolved based on an environment variable (like `DATA_ROOT`) or the workflow/engine resolves at the beginning. There is no need to propagate the `data_root` throught olive.
- All cache related functionalities are now abtracted under a new `olive.engine.cache.OliveCache` class.
- `prepare_non_local_model` is now generalized as `OliveCache.prepare_resources_for_local`. This method sets up config for local execution.


olive/cache.py
```diff
@@ -9,82 +9,103 @@
...
-    """Return the subdirectories of the cache directory.
+def set_cache_dir(cache_dir: Union[str, Path]):
+    """Set the cache directory."""
+    os.environ["OLIVE_CACHE_DIR"] = str(cache_dir)
 
-    There are three subdirectories: models, runs, and evaluations.
+
+def unset_cache_dir():
+    """Unset the cache directory."""
+    if "OLIVE_CACHE_DIR" in os.environ:
```
trajepl, line 26: nit: -> `os.environ.pop('OLIVE_CACHE_DIR', None)`

olive/cache.py
```diff
@@ -9,82 +9,103 @@
...
-    return cache_dir / "models", cache_dir / "runs", cache_dir / "evaluations", cache_dir / "non_local_resources"
+    return Path(os.environ.get("OLIVE_CACHE_DIR", ".olive-cache")).resolve()
 
 
-def clean_cache(cache_dir: Union[str, Path] = ".olive-cache"):
+def get_cache_sub_dirs() -> Dict[str, Path]:
+    """Return the subdirectories of the cache directory."""
+    cache_dir = get_cache_dir()
+    return {
+        "models": cache_dir / "models",
```
trajepl, line 43: Maybe you missed `workflow_id` which is added by Xiaoyu.
image

olive/cache.py
```diff
@@ -9,82 +9,103 @@
...
+from olive.common.config_utils import ConfigBase, convert_configs_to_dicts, serialize_to_json, validate_config
+from olive.common.utils import hash_dict, set_nested_dict_value
+from olive.resource_path import ResourcePath, create_resource_path, find_all_resources
 
 logger = logging.getLogger(__name__)
 
 
-def get_cache_sub_dirs(cache_dir: Union[str, Path] = ".olive-cache"):
-    """Return the subdirectories of the cache directory.
+def set_cache_dir(cache_dir: Union[str, Path]):
```
trajepl, line 19: open to discuss: it is a good idea to set the cache folder with os.env. Also, think further, a cli interface to clean cache folder can be useful, just like huggingface did.

olive/cache.py
```diff
@@ -9,82 +9,103 @@
...
-    cache_dir = Path(cache_dir)
-    return cache_dir / "models", cache_dir / "runs", cache_dir / "evaluations", cache_dir / "non_local_resources"
+    return Path(os.environ.get("OLIVE_CACHE_DIR", ".olive-cache")).resolve()
 
 
-def clean_cache(cache_dir: Union[str, Path] = ".olive-cache"):
+def get_cache_sub_dirs() -> Dict[str, Path]:
+    """Return the subdirectories of the cache directory."""
+    cache_dir = get_cache_dir()
+    return {
```
trajepl, line 42: Do think if it is needed to return a class but not a dict? Then any down-stream users can directly use `dot` to get the models/evaluation/etc..

olive/cache.py
jambayk: This is set during engine init as cache_dir attribute https://github.com/microsoft/Olive/blob/97df601aae7648f8482e8f8c819ac765350508ef/olive/engine/engine.py#L91

olive/cache.py
jambayk: We could add that interface. But I think that could be a future work item?

olive/cache.py
jambayk: this is a good idea. I can create a cache class with these subdirectories. And also put the cache utils as methods for it.

jambayk: for the time being I will just make the cache dirs as a class with dot notation. We can refactor all of the cache methods into a single class as a follow up since that requires more work that is out of scope for this PR.

olive/cache.py
trajepl: Yeah, just an idea.

olive/cache.py
trajepl: But when you see `get_cache_dir`, it still use the default value of `.olive-cache`.
Actually, in olive when you call `get_cache_dir`, the `OLIVE_CACHE_DIR` must be set already, right?

trajepl: For the case where user set the cache dir by `OLIVE_CACHE_DIR`, then `get_cache_dir` will return the path without `workflow_id`.

jambayk: The environment variable is currently not expected to be set by the user but by the engine during initialization. The environment variable is only expected to be used by other parts of olive code to know where the cache dir is. 

However, this feature is not currently needed so I am removing it in my next commit. will re-introduce it when I need it for my hf model changes.

olive/cache.py
jambayk: Moved all cache utils into an `OliveCache` class

guotuofeng:
the data_root is designed to concatenate the data path in AzureML..

trajepl (approved):

PR #1239, opened 2024-07-17
`HfModelHandler` separated from `PyTorchModelHandler`

jambayk:
## Describe your changes
Huggingface models are now separated from regular PyTorch models. Previously, we used `hf_config` to define hf models but this has multiple drawbacks:
  - confusion between `model_path` and `hf_config.model_name`
  - complicated model loading, io config and dummy input logic
  - unnecessary nesting of hf_config
 
The changes made are:
- `HfModelHandler`: that only deals with hf models. 
  - Based on current use cases, there is no user script associated with it.
  - `feature` is removed. Automatically inferred from `task`.
  - default `task` is `text-generation-with-past` which is equivalent to a causallm model with kv cache.
  - Can automatically infer MLFlow models and prepare them as needed. 
- `HfComponent` concept is removed. It was only used for whisper example which is easily implemented using `CompositeModelHandler`.
- `PyTorchModelHandler`: `hf_config` and related logic are removed entirely
- `DistributedPyTorchModelHandler` replaced with `DistributedHfModelHandler`. 
- `CompositePyTorchModelHandler` removed. There is no use case for it and as shown in whisper example, can be implemented using `CompositeModelHandler`.
- `KVCache`: order of past and present names is corrected as `[key0, value0, key1, value1, ...]` from `[key0, key1, ..., value0, value1, ...]`
- Relevant passes are updated to handler either or both of `PyTorchModelHandler` and `HfModelHandler`

  - New `HfModelHandler` introduced and all huggingface related features from `PyTorchModelHandler` removed.


olive/passes/pytorch/slicegpt.py
```diff
@@ -76,10 +76,10 @@
 
     @torch.no_grad()
     def _run_for_config(
-        self, model: PyTorchModelHandler, config: Dict[str, Any], output_model_path: str
+        self, model: HfModelHandler, config: Dict[str, Any], output_model_path: str
     ) -> PyTorchModelHandler:
-        if sys.version_info < (3, 10):
-            raise ValueError("SliceGPT requires python3.10 or higher")
+        # if sys.version_info < (3, 10):
+        #     raise ValueError("SliceGPT requires python3.10 or higher")
```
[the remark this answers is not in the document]
jambayk: Resolved

olive/model/config/kv_cache_config.py
```diff
@@ -70,17 +71,20 @@ def _get_v_names(self, direction="inputs"):
...
+        return list(chain.from_iterable(zip(self._get_k_names(direction), self._get_v_names(direction))))
+
+    def get_ort_past_kv_names(self):
+        return self._get_kv_names("inputs")
+
     def get_ort_past_key_names(self):
         return self._get_k_names("inputs")
 
     def get_ort_past_value_names(self):
         return self._get_v_names("inputs")
```
trajepl, line 84: Better to remove these two functions and update the corresponding calling part with `get_ort_past_kv_names`.

olive/model/handler/hf.py
```diff
@@ -0,0 +1,213 @@
...
+            model_path=model_path,
+            model_attributes=model_attributes,
+            io_config=io_config,
+            generative=generative,
+        )
+        self.add_resources(locals())
+        self.task = task
+        self.load_kwargs = validate_config(load_kwargs, HfLoadKwargs, warn_unused_keys=False) if load_kwargs else None
+
+        self.mlflow_model_path = None
```
trajepl, line 53: nit: mlflow is not strongly bound with transformers. Maybe update the corresponding name to `mlflow_transformers`? Then we can distinguish the normal mlflow model and mlflow-transoformers models.

olive/model/handler/mixin/mlflow.py
```diff
@@ -6,41 +6,62 @@
...
+
+        # some mlflow models have config and tokenizer directories but model directory also
+        # contains the same files
+        model_dir_contents = set(model_dir.iterdir())
+        if set(config_dir.iterdir()) <= model_dir_contents and set(tokenizer_dir.iterdir()) <= model_dir_contents:
+            self.mlflow_model_path = str(model_dir)
+            return
+
+        # have to gather all contents into a single directory
+        cache = OliveCache.from_cache_env()
```
trajepl, line 60: Seems you removed `mlflow_transformer_model_cache_dir` and set the default mlflow model folder under cache directory to bypass the issue of `read-only` in default local downloading folder we met before, right?

jambayk: Yes. This HfModelHandler automatically prepares mlflow models so the user can just use it like it's a model id or local path. We already have a cache folder and the mlflow model is already downloaded there (when running locally) so it's easier to just copy or hard link there.

olive/model/config/kv_cache_config.py
jambayk: done

olive/model/handler/hf.py
jambayk: done

olive/model/handler/mixin/mlflow.py
jambayk: Just pushed a commit which executes this logic only if we ask for `model_name_or_path`. for most of the common uses cases like loading model, config and tokenizer, we can just call them on the correct subdirectory. 

Otherwise, creating a copy of the model files during init is inefficient if we don't really need it.

test/unit_test/common/test_hf.py
```diff
@@ -0,0 +1,74 @@
+# -------------------------------------------------------------------------
```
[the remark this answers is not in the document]
jambayk: resolved

trajepl (approved):

PR #1243, opened 2024-07-18
RunConfig: Generalize `azureml_client` auto-insert

jambayk:
## Describe your changes
- Instead of looking for specific config parameters, use a `root_validator` to automatically find relevant configs and insert `azureml_client` if needed.
- Also abstracted duplicate logic in `evaluator` and `data_config` validation.


olive/workflows/run/config.py
```diff
@@ -240,12 +234,6 @@ def validate_data_configs_with_hf_model(cls, v, values):
     def validate_evaluators(cls, v, values):
         for idx, metric in enumerate(v.get("metrics", [])):
             v["metrics"][idx] = _resolve_data_config(metric, values, "data_config")
-
-            data_dir_config = v["metrics"][idx].get("user_config", {}).get("data_dir", None)
-            if isinstance(data_dir_config, dict):
-                if _have_aml_client(data_dir_config, values):
-                    data_dir_config["config"]["azureml_client"] = values["azureml_client"]
-                v["metrics"][idx]["user_config"]["data_dir"] = data_dir_config
```
shaahji, line 248: This might still be required if `eval_func_args` are to contain the `data_dir` & `batch_size` as discussed in the earlier PR.

jambayk: The root validator works directly on the whole workflow config. It applies the insertion recursively so that case is automatically covered.

olive/workflows/run/config.py
```diff
@@ -309,6 +291,48 @@ def validate_pass_search(cls, v, values):
...
+
+    Return the path to insert azureml_client and the type of the config.
+    """
+    if not isinstance(config, dict):
+        return None, None
+
+    config_type = config.get("type")
+
+    if config_type in AZUREML_RESOURCE_TYPES and not config.get("config", {}).get("azureml_client"):
+        return ("config", "azureml_client"), "AzureML Resource Path"
```
xiaoyu-work, line 328: - Does this only cover one level down insertion? Like `config -> azureml_client`.
- Will this be conflicting with the future change if we want to flatten the `config` (for `config.get("config", {})`)?

olive/workflows/run/config.py
```diff
@@ -354,36 +378,14 @@ def _resolve_config_str(v, values, alias, component_name):
...
-
-    # resolve sub component name to component config
-    if sub_component not in component_map:
-        raise ValueError(f"{data_config_alias} {sub_component} not found in {components}")
-
-    v[data_config_alias] = component_map[sub_component]
-    return v
+    # it is okay to insert a new member into values
+    # doesn't affect the original values
+    values["data_configs_dict"] = {dc.name: dc for dc in values.get("data_configs") or []}
```
xiaoyu-work, line 387: Though this doesn't affect the functionality, this value will be added to the parsed run config obj.

jambayk: When I checked last time, I didn't see this behavior but you are right, it shows up in the object. I will pass this directly as {"data_configs_dict": ...} into the next call

olive/workflows/run/config.py
jambayk: Here we know that we are dealing with azureml resource path (same for aml system in the next condition) so we only look at `config -> azureml_client` because that's where it is expected to be in. Key point note here is that we are only checking to see if this is an azureml related config. the recursion logic is handled by the caller.

Good point about the nesting change (I am starting work on it today). In that case, we can just check directly `config.get("azureml_client")`.

olive/workflows/run/config.py
jambayk: updated

olive/workflows/run/config.py
xiaoyu-work: Ah yeah I missed the recursion part in `_insert_azureml_client`. Looks good to me.

xiaoyu-work (approved):

PR #1245, opened 2024-07-19
`NestedConfigs`: config fields don't need nesting

jambayk:
## Describe your changes
- Introduce a new configuration class called `NestedConfig` that automatically gathers fields into a `_nested_field_name` dictionary. This is used as the base class for the config classes like path, model, pass, etc. 
- After validation, the objects are the same as before so there is no change necessary on the backend. 
- Examples updated to remove the unnecessary nestings. 
- Fixed a bug in the dependency setup caused by using `ClassVar` to annotate module dependencies.

  - It is not necessary to nest the configuration parameters inside a `config`, `params`, `engine` field. These values can be provided directly to the parent dictionary.


olive/common/config_utils.py
```diff
@@ -306,8 +308,12 @@
     if isinstance(config, dict):
         user_keys = set(config.keys())
         config = instance_class(**config)
-        config_keys = set(config.dict().keys())
+        # check for unused keys
+        config_dict = config.dict()
+        config_keys = set(config_dict.keys())
         unused_keys = user_keys - config_keys
+        if isinstance(config, NestedConfig):
+            unused_keys -= set((config_dict.get(config._nested_field_name) or {}).keys())
```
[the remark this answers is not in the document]
jambayk: Resolved

shaahji:
Can you split this PR to remove the decorative changes only? It will make reviewing this a lot easier.

jambayk:
> Can you split this PR to remove the decorative changes only? It will make reviewing this a lot easier.

@shaahji I have removed the json formatter related changes. I originally added the linter here since I found it easier to format automatically than manually fix each file. But separated them for the sake of easier review. Now the main changes in the jsons are just the removal of the nestings.

test/unit_test/workflows/mock_data/ner_task_dataset.json
jambayk, line 1: These files were removed since we don't use them in our tests

test/unit_test/workflows/mock_data/default_engine.json
jambayk, line 1: updated the mock_data jsons to only keep minimal information needed to test the relevant features.

shaahji (approved):

PR #1249, opened 2024-07-22
Custom `FORMAT-JSON` linter

jambayk:
## Describe your changes
- Created a custom json linter that formats all json files based on `max-line-length` and `indent` settings.
- Unlike `json.dumps`, this format also adds a space before or after brackets and braces for better readability.


.azure_pipelines/performance_check/configs/bert.json
```diff
@@ -1,41 +1,17 @@
 {
     "input_model": {
         "type": "PyTorchModel",
-        "config": {
-            "hf_config": {
-                "model_name": "Intel/bert-base-uncased-mrpc",
-                "task": "text-classification"
-            }
-        }
+        "config": { "hf_config": { "model_name": "Intel/bert-base-uncased-mrpc", "task": "text-classification" } }
```
xiaoyu-work, line 4: My personal json format preference is how it currently looks like....This looks more crowded.

jambayk: You can ignore the files under the perf check since they haven't been updated and have lots of unnecessary nesting. They also had to be updated since the linter runs on all files. 
For the rest of the codebase, most of the nesting have been removed so it doesn't look as cluttered. 

Although, I agree that for deep nested dicts, this formatting doesn't look as nice but the default json.dump formatting with indent makes the jsons, especially the io configs, appear large than they are. @shaahji can confirm this is the reason he formatted the files earlier after talking with Devang. 
We need a standard way of formatting the jsons to keep them all consistent and not require guess work or manual indentations. 

The linter could be improved to make it aware of the the max level of nesting allowed on a line but at this point, I don't think it's worth the extra complexity.

shaahji: Unfortunately, for humans, perception is everything. If the files look smaller, even though the content is the same, user perceives them as easier compared to the longer more readable version.

shaahji (approved):

PR #1268, opened 2024-07-29
🤖  Quarot

trajepl:
## Describe your changes

1. Integration QuaRot into olive, which is a new Quantization scheme based on Rotations, which is able to quantize LLMs end-to-end. 
2. Add phi3-quarot example

Release Note: New pass QuaRot to quantize transformer to improve performance and reduce memory footprint.

Example:
`w_gptq` needs flash-atten which only supports Ampere GPUs or newer.

Test on A100:
image


olive/passes/pytorch/quarot.py
```diff
@@ -0,0 +1,299 @@
...
+
+    @staticmethod
+    def _default_config(accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        return {
+            "input_model_dtype": PassConfigParam(
+                type_=QuaRot.ModelDtype,
+                default_value=QuaRot.ModelDtype.FP16,
+                description="Input model's data type.",
+            ),
+            "calibration_data_config": PassConfigParam(
```
jambayk, line 61: Hitesh and I had a related discussion before. Would it be possible to add the functionality for sampling directly to our dataloader component so that this and slicegpt pass don't have to expose `calibration_nsamples` and `calibration_batch_size`? 
wrt the `"labels"` as inputs, we could just wrap the original dataloader and override its iteration logic to return a single dict.

trajepl: Sounds good. How about implement it in follow-up PR?

jambayk: Sure! Follow up is good too.

jambayk (approved):

PR #1270, opened 2024-07-29
Add adapter path support for cloud model cache

xiaoyu-work:
## Describe your changes

Add adapter path support for cloud model cache

### DIFF
--- a/olive/engine/cloud_cache_helper.py
+++ b/olive/engine/cloud_cache_helper.py
@@ -255,5 +255,6 @@ def is_hf_repo_exist(repo_name: str):
             "huggingface_hub is not installed. "
             "Please install huggingface_hub to use the cloud model cache feature for Huggingface model."
         )
+        raise
 
     return HfApi().repo_exists(repo_name)
--- a/olive/engine/cloud_cache_helper.py
+++ b/olive/engine/cloud_cache_helper.py
@@ -48,43 +48,50 @@ class CloudCacheHelper:
         self.model_path_map = "model_path.json"
         self.input_model_config = input_model_config
 
-    def update_model_config(self, cloud_model_path, adapter_path, model_config, input_model_hash, output_model_path):
+    def update_model_config(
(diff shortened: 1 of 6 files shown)


xiaoyu-work:
Will update this RP in this week.

olive/engine/cloud_cache_helper.py
```diff
@@ -46,40 +45,61 @@ def __init__(
...
+            model_config.config["adapter_path"] = str(output_model_path / adapter_path)
+
+        return model_config
+
+    def download_blob_list(self, blob_list, directory_prefix, output_model_path, prefix: str = None):
         for blob in blob_list:
             blob_client = self.container_client.get_blob_client(blob)
-            local_file_path = self.output_model_path / blob.name[len(model_directory_prefix) + 1 :]
+            local_file_path = (
+                output_model_path / prefix / blob.name[len(directory_prefix) + 1 :]
```
jambayk, line 75: instead of doing string slice, maybe you could do something like `Path(blob.name).name`?

xiaoyu-work: `Path(blob.name).name` will lose the subfolder structure of this blob. For example if I have `prefix/subfolder/file.name`, `Path(blob.name).name` will only get `file.name`.

jambayk: Oh I see.

olive/engine/cloud_cache_helper.py
```diff
@@ -29,7 +29,6 @@ class CloudCacheConfig(ConfigBase):
 class CloudCacheHelper:
```
jambayk, line 29: While you are updating this could you also add some docstrings for the methods and comments about the cloud cache helpers? Otherwise, it's hard to follow the logic if this needs to be updated as part of other changes.

xiaoyu-work: Added

PR #1272, opened 2024-07-31
Separate aml example test from pipeline

xiaoyu-work:
## Describe your changes

- Separate aml example test from pipeline
- New aml example test pipeline had been setup.
- AML example pipeline will run once a week.
- AML example pipeline will be triggered when code merged, and pr trigger for azureml client and aml system.

### DIFF
--- a/.azure_pipelines/job_templates/olive-example-template.yaml
+++ b/.azure_pipelines/job_templates/olive-example-template.yaml
@@ -5,7 +5,7 @@ parameters:
   pool: ''
   python_version: '3.8'
   onnxruntime: 'onnxruntime'
-  subfolder: ''
+  subfolder: 'local'
   torch: 'torch'
 
 jobs:
--- a/.azure_pipelines/job_templates/olive-example-template.yaml
+++ b/.azure_pipelines/job_templates/olive-example-template.yaml
@@ -5,13 +5,14 @@ parameters:
   pool: ''
   python_version: '3.8'
   onnxruntime: 'onnxruntime'
+  subfolder: ''
   torch: 'torch'
 
 jobs:
-  - job: ${{parameters.name}}_Test_Examples
+  - job: ${{ parameters.name }}_Test_Examples
     timeoutInMinutes: 300
     pool:
-      name: ${{ parameters.pool}}
+      name: ${{ parameters.pool }}
     strategy:
       matrix:
         ${{ insert }}: ${{ parameters.examples }}
@@ -38,7 +39,7 @@ jobs:
           python -m pip install pytest
           python -m pip install azure-identity azure-storage-blob tabulate
           python -m pip install -r $(Build.SourcesDirectory)/examples/$(exampleFolder)/$(exampleRequirements)
-          python -m pytest -v -s --log-cli-level=WARNING --junitxml=$(Build.SourcesDirectory)/logs/test_examples-TestOlive.xml $(Build.SourcesDirectory)/examples/test/test_$(exampleName).py --basetemp $(PYTEST_BASETEMP)
+          python -m pytest -v -s --log-cli-level=WARNING --junitxml=$(Build.SourcesDirectory)/logs/test_examples-TestOlive.xml $(Build.SourcesDirectory)/examples/test/${{ parameters.subfolder }}/test_$(exampleName).py --basetemp $(PYTEST_BASETEMP)
       displayName: Test Examples
       env:
         WORKSPACE_SUBSCRIPTION_ID: $(workspace-subscription-id)
--- a/.azure_pipelines/olive-aml-ci.yaml
+++ b/.azure_pipelines/olive-aml-ci.yaml
@@ -1,4 +1,5 @@
 trigger:
+  batch: true
   branches:
     include:
     - main
@@ -17,14 +18,6 @@ pr:
     include:
     - olive/azureml/*
     - olive/systems/azureml/*
-
-schedules:
-- cron: 0 6 * * 5
-  displayName: Scheduled Build
-  branches:
-    include:
-    - main
-  always: true
 
 jobs:
 # Linux examples test
--- a/.azure_pipelines/olive-aml-ci.yaml
+++ b/.azure_pipelines/olive-aml-ci.yaml
@@ -1,5 +1,22 @@
-trigger: none
-pr: none
+trigger:
+  branches:
+    include:
+    - main
+  paths:
+    exclude:
+    - docs/*
+    - examples/*
+    - README.md
+    - CONTRIBUTING.md
+    - LICENSE
+pr:
+  branches:
+    include:
+    - main
+  paths:
+    include:
+    - olive/azureml/*
+    - olive/systems/azureml/*
 
 schedules:
 - cron: 0 6 * * 5
--- a/.azure_pipelines/olive-aml-ci.yaml
+++ b/.azure_pipelines/olive-aml-ci.yaml
@@ -1,0 +1,47 @@
+trigger: none
+pr: none
+
+schedules:
+- cron: 0 6 * * 5
+  displayName: Scheduled Build
+  branches:
+    include:
+    - main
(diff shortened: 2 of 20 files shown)


examples/test/__init__.py
jambayk, line 1: Do you know if the package structure with these `__init__.py` files let us run individual tests from any directory? Or they need to be run from the root of the repository like in the ci?

examples/test/local/test_bert_cuda_gpu.py
```diff
@@ -6,13 +6,14 @@
...
-from utils import check_output, patch_config
+
+from ..utils import check_output, patch_config
 
 
 @pytest.fixture(scope="module", autouse=True)
 def setup():
     """Setups any state specific to the execution of the given module."""
-    cur_dir = Path(__file__).resolve().parent.parent
+    cur_dir = Path(__file__).resolve().parent.parent.parent
```
jambayk, line 16: not directly related to this PR but perhaps we can make this a util like `get_cur_dir`

.azure_pipelines/olive-aml-ci.yaml
jambayk, line 1: where do you think this pipeline should sit in? Our internal ADO or the publicpackages ADO?

.azure_pipelines/job_templates/olive-example-template.yaml
```diff
@@ -5,13 +5,14 @@ parameters:
   pool: ''
   python_version: '3.8'
   onnxruntime: 'onnxruntime'
+  subfolder: ''
```
jambayk, line 8: the default could be `local`. Or else, this PR also needs to update the stable and nightly pipelines also.

examples/test/__init__.py
xiaoyu-work: Yeah we can run this in any directory.

examples/test/local/test_bert_cuda_gpu.py
xiaoyu-work: The test still needs `cur_dir` 3 lines after (or maybe not. I can verify this)

.azure_pipelines/olive-aml-ci.yaml
xiaoyu-work: I added this pipeline to our internal ADO.

.azure_pipelines/job_templates/olive-example-template.yaml
xiaoyu-work: Sounds good. Updated.

examples/test/local/test_bert_cuda_gpu.py
jambayk: oh I meant, we can make this line 16 use a utils function since almost all examples do the same thing and we are hardcoding the directory structure in each one. I just called it `get_examples_dir` since it is actually returning the `olive/examples` directory path and not the current directory (cur_dir is an old name from before example test refactor)

.azure_pipelines/olive-ci.yaml
```diff
@@ -85,20 +85,14 @@ jobs:
     name: Linux_CI
     pool: $(OLIVE_POOL_UBUNTU2004)
     onnxruntime: onnxruntime
+    subfolder: local
```
jambayk, line 88: nit: subfolder is not required anymore. not an issue, just think this keeps it consistent with the stable and nightly yamls

examples/test/azureml/test_bert_ptq_cpu_aml.py
```diff
@@ -24,8 +25,6 @@ def setup():
     [
         # aml system test
         ("bert_ptq_cpu.json", "tpe", "joint", "aml_system"),
-        # aml model test in local system
-        ("bert_ptq_cpu_aml.json", False, None, "local_system"),
```
jambayk, line 28: I think this test case is still useful. Helps us make sure aml model paths can also be used for local runs.

xiaoyu-work: Oh yeah that's true. I thought this will be tested in aml system as well, while aml system doesn't download aml models to local. I'll add this back.

examples/test/local/test_bert_cuda_gpu.py
xiaoyu-work: I just refactored all of this.

.azure_pipelines/olive-ci.yaml
xiaoyu-work: Removed.

examples/test/azureml/test_bert_ptq_cpu_aml.py
xiaoyu-work: Added back

.azure_pipelines/olive-aml-ci.yaml
```diff
@@ -0,0 +1,64 @@
+trigger:
```
jambayk, line 1: This trigger should probably be set to none? Otherwise, it will trigger the pipeline everytime a pr merges

xiaoyu-work: I intentionally added this trigger. While this is not blocking the PR, it might be okay to run this when code got merged. But this just reminds me that I probably should remove scheduled trigger in this case.

jambayk: yes, if we want this trigger then the schedule trigger is unnecessary. Since these tests don't take too long, it sounds okay to run them post merge. 
could you add `batch: true` like in https://github.com/microsoft/Olive/blob/ff929df6a3bd5b3082641f5ce34b728ba2361e7a/.azure_pipelines/build-doc.yaml#L8? Just in case multiple PR merge together, there is no need to run once for each

xiaoyu-work: Sounds good. Updated.

.azure_pipelines/olive-aml-ci.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - CONTRIBUTING.md
+    - LICENSE
+pr:
+  branches:
+    include:
+    - main
+  paths:
+    include:
+    - olive/azureml/*
+    - olive/systems/azureml/*
```
jambayk, line 20: maybe also include `examples/test/azureml`?

.azure_pipelines/olive-aml-ci.yaml
```diff
@@ -0,0 +1,57 @@
+trigger:
+  batch: true
+  branches:
+    include:
+    - main
+  paths:
+    exclude:
+    - docs/*
+    - examples/*
```
jambayk, line 9: we might still want to run after examples changes since the aml test configs might have been updated

jambayk: image
this looks like a good enough filter

jambayk: Can you also confirm the pipeline is set up in the publicpackages ado and runs as expected?

.azure_pipelines/olive-aml-ci.yaml
```diff
@@ -0,0 +1,57 @@
...
+  paths:
+    include:
+    - olive/azureml/*
+    - olive/systems/azureml/*
+
+jobs:
+# Linux examples test
+- template: job_templates/olive-example-template.yaml
+  parameters:
+    name: Linux_CI
```
jambayk, line 26: nit: name include aml so that we know which pipeline it belongs to in the github status

jambayk: Nevermind, github displays the pipeline name before the job name

.azure_pipelines/olive-aml-ci.yaml
xiaoyu-work: I've run the pipeline.

jambayk (dismissed):
Thanks!

jambayk (approved):

PR #1275, opened 2024-08-02
Support for custom evaluators

shaahji:
## Support for custom evaluators

* Move evaluation registration out of OliveEvaluator into its own class
* Evaluators are registered using both the Framework and the class name
* Add "type", "type_args", "user_script" and "script_dir" to OliveEvaluationConfig to support user specific custom implementation
* Remove some dead code

### DIFF
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -9,14 +9,16 @@ from abc import ABC, abstractmethod
 from copy import deepcopy
 from functools import partial
 from numbers import Number
-from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, NamedTuple, Tuple, Type, Union
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Dict, List, NamedTuple, Tuple, Union
 
 import numpy as np
 import torch
 
-from olive.common.config_utils import ConfigBase, validate_config
+from olive.common.config_utils import NestedConfig, validate_config
+from olive.common.import_lib import import_user_module
 from olive.common.ort_inference import OrtInferenceSession, prepare_io_bindings
-from olive.common.pydantic_v1 import validator
+from olive.common.pydantic_v1 import Field, root_validator, validator
 from olive.common.user_module_loader import UserModuleLoader
 from olive.common.utils import tensor_data_to_device
 from olive.constants import Framework
@@ -26,6 +28,7 @@ from olive.data.template import dummy_data_config_template
 from olive.evaluator.metric import LatencySubType, Metric, MetricType, ThroughputSubType, get_latency_config_from_metric
 from olive.evaluator.metric_backend import MetricBackend
 from olive.evaluator.metric_result import MetricResult, SubMetricResult, flatten_metric_result, joint_metric_key
+from olive.evaluator.registry import Registry
 from olive.hardware import Device
 from olive.model import DistributedOnnxModelHandler, ONNXModelHandler
 from olive.model.config.io_config import is_io_config_static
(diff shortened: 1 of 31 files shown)


olive/evaluator/olive_evaluator.py
```diff
@@ -1106,6 +1141,12 @@ def validate_metrics(cls, v):
...
-            raise ValueError(f"Priorities must be unique and in the range 1 to {metric_len}")
+            raise ValueError(
+                f"Priorities must be unique and in the range 1 to {metric_len}\n"
+                f"\tActual: {rank_set}\n\tExpected: {expected_rank_set}"
+            )
 
         return v
+
+    def create_evaluator(self, model: "OliveModelHandler" = None) -> OliveEvaluator:
+        return Registry.get(self.type or str(model.framework))(**(self.type_args or {}))
```
xiaoyu-work, line 1152: qq: In which case we will use type to get an evaluator?

shaahji: I am introducing custom evaluators to support internal i.e. MS specific evaluation techniques. Since these are internal, they cannot be part of public Olive repo.

xiaoyu-work:
Overall this looks good to me. Can you please fix the tests and Github warnings? Can you also add a documentation about how to use custom evaluators for advanced users?

olive/evaluator/olive_evaluator.py
```diff
@@ -1076,6 +1089,16 @@
...
 
+    # @validator("type", pre=True)
+    # def validate_type(cls, v, values):
+    #     if "user_script" in values and values["user_script"]:
+    #         import_user_module(values["user_script"], values.get("script_dir"))
+    #
+    #     if v is not None and Registry.get(v) is None:
+    #         raise ValueError(f"Invalid evaluator type: {v}")
+    #
+    #     return v
```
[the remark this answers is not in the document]
shaahji: Removed.

olive/workflows/run/run.py
```diff
@@ -143,7 +141,12 @@
     set_ort_logger_severity(run_config.engine.ort_py_log_severity_level)
 
     # ort_log_severity_level: C++ logging levels
-    ort.set_default_logger_severity(run_config.engine.ort_log_severity_level)
+    try:
+        import onnxruntime as ort
+
+        ort.set_default_logger_severity(run_config.engine.ort_log_severity_level)
```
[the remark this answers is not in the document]
shaahji: Added a log comment to inform the user.

olive/evaluator/registry.py
```diff
@@ -0,0 +1,68 @@
...
+        Args:
+            name (str): the name of the evaluator, if name is None, uses the class name
+
+        Returns:
+            Callable: the decorator function
+
+        """
+
+        def decorator(component):
+            component_name = name if name is not None else component.__name__
```
jambayk, line 30: nit: not important since users register their own evaluators but I recently made a PR to make all types case-insensitive by storing all keys in lowercase. https://github.com/microsoft/Olive/blob/64b79dcddd029ae2927637bb03074b46958763e2/olive/data/registry.py#L43

olive/evaluator/registry.py
```diff
@@ -0,0 +1,68 @@
...
+            if component_name in cls._REGISTRY:
+                component_1 = cls._REGISTRY[component_name]
+                component_2 = component
+
+                component_file_1 = inspect.getfile(component_1)
```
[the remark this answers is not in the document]
shaahji: I think we need a generic implementation of the registry itself. We have at least three different implementations (DataConfig, Evaluator, and Search). Likely more!

I will move the code into a shared util in a follow up PR. I have another one coming to deprecate custom metric and updating the docs.

olive/evaluator/registry.py
shaahji: Will make the change in the follow up PR.

olive/evaluator/olive_evaluator.py
```diff
@@ -1058,16 +1075,24 @@ def _prepare_dataloader(
...
+
+    type: str = None
+    type_args: Dict = Field(default_factory=dict)
 
+    # user script to define and register the evaluator
+    user_script: Union[Path, str] = None
+    script_dir: Union[Path, str] = None
 
-class OliveEvaluatorConfig(ConfigBase):
     metrics: List[Metric] = []  # noqa: RUF012
```
jambayk, line 1088: I am not really sure how the custom evaluator would work since it still needs metric. Is the idea that the metric for the custom evaluator (unless they are latency or accuracy) would be defined as `custom` metric type? 

In that case, there is overlap between the custom metric type's options https://github.com/microsoft/Olive/blob/64b79dcddd029ae2927637bb03074b46958763e2/olive/evaluator/metric_config.py#L29 and https://github.com/microsoft/Olive/blob/64b79dcddd029ae2927637bb03074b46958763e2/olive/evaluator/metric_config.py#L18 
The custom metric is also able to take an `eval_func` which can be run directly on the model https://github.com/microsoft/Olive/blob/64b79dcddd029ae2927637bb03074b46958763e2/olive/evaluator/olive_evaluator.py#L145

can you share an example or test for how such an evaluator config would be defined and used?

jambayk: just saw your comment about custom metric deprecation. If that addresses this question, then the PR looks good to me.

jambayk (approved):

PR #1277, opened 2024-08-02
CLI: `finetune` command

jambayk:
## Describe your changes
- Add `finetune` cli command to run lora/qlora finetuning workflow
  - currently it does finetuning -> export -> transformers opt -> extract adapters
  - other options such as using model builder or only running finetuning will be added later as we have more clarity on the requirements for this command. 

Some code improvements found while implementing the command:
- Clean up on the training arguments which are the same as the hf training argument defaults
- `add_special_tokens` default is changed back to `True` #1259
- `corpus_strategy` -> `strategy` rename which was missed in #1273
- removed unused data configs from examples
- changed some lora related logs to debug since they are not needed for basic users
 

devang-ml (commented):
Please document the new command 
features/cli.html

jambayk:
> Please document the new command features/cli.html

@devang-ml  the commands are automatically documented https://github.com/microsoft/Olive/blob/main/docs/source/features/cli.rst?plain=1

olive/cli/finetune.py
```diff
@@ -0,0 +1,267 @@
...
+        from olive.workflows import run as olive_run
+
+        set_tempdir(self.args.tempdir)
+
+        run_config = self.get_run_config()
+        with tempfile.TemporaryDirectory() as tempdir:
+            run_config["output_dir"] = tempdir
+            olive_run(run_config)
+
+            # need to improve the output structure of olive run
```
trajepl, line 142: Agree with this.
Olive workflow will copy same model multi-times. Is it possible only run fine-tune passes out of engine?  I met the similar concern in perf-tuning cli.

olive/cli/finetune.py
```diff
@@ -0,0 +1,267 @@
...
+            eval_data_config = deepcopy(config["data_configs"][0])
+            eval_data_config["name"] = "eval_data"
+            eval_data_config["load_dataset_config"]["split"] = self.args.eval_split
+            config["data_configs"].append(eval_data_config)
+            config["passes"]["f"]["eval_data_config"] = "eval_data"
+
+        return config
+
+
+TEMPLATE = {
```
trajepl, line 210: How about move this kind of template under "auto_optimzier" ? It will be helpful for coming auto optimizer with well-know workflow templates.
image

jambayk: Can I do it as a follow up PR? There is no use case for the auto optimizer currently so I don't have anything to test this for?

olive/cli/finetune.py
jambayk: We definitely need to reconsider the output structure. Before when the engine was run in no search mode, its output was a dict with model config but now it gives us a dictionary of accelerator:footprint which is not easy to parse 

We have similar issues in the other example scripts too. Where we have to hardcode the expected output structure.

trajepl: Let us add it into todo items.

olive/cli/finetune.py
trajepl: Sure.

trajepl (approved):

PR #1281, opened 2024-08-06
🧗 perf tuning cli

trajepl:
## Describe your changes

Perf tuning cli.

E.g. for phi2-int4-cpu got from model builder, user can directly tuning the session parameters in cpu and cuda ep with:
`olive tune-session-params --model model.onnx --device gpu --providers_list CUDA CPU --hf_model_name microsoft/phi-2`
Also, user can provide their customized data_config:
`olive tune-session-params --model  model.onnx --device gpu --providers_list CUDA CPU  --data_config_path data_config.json`
where the data config json looks like
```json
{
    "name": "gpu_cpu",
    "type": "TransformersTokenDummyDataContainer",
    "load_dataset_config": {
        "model_name": "microsoft/phi-2"
    }
}
```

### DIFF
--- a/olive/auto_optimizer/workflow_template/ort_perf_tuning.yaml
+++ b/olive/auto_optimizer/workflow_template/ort_perf_tuning.yaml
@@ -1,0 +1,17 @@
+input_model:
+  type: "ONNXModel"
+  model_path: <<model_path>>
+passes:
+  perf_tuning:
+    type: "OrtPerfTuning"
+    data_config:
+      name: "perf_tuning_data"
+      type: "TransformersTokenDummyDataContainer"
+systems:
+  local_system:
+    type: "LocalSystem"
+    accelerators:
+      - device: <<device>>
+engine:
+  target: "local_system"
+  host: "local_system"
--- a/olive/auto_optimizer/workflow_template/ort_perf_tuning.yaml
+++ b/olive/auto_optimizer/workflow_template/ort_perf_tuning.yaml
@@ -1,12 +1,14 @@
 input_model:
   type: "ONNXModel"
   model_path: <<model_path>>
+data_configs:
+  - name: "perf_tuning_data"
+    type: "TransformersTokenDummyDataContainer"
 passes:
   perf_tuning:
     type: "OrtPerfTuning"
-    data_config:
-      name: "perf_tuning_data"
-      type: "TransformersTokenDummyDataContainer"
+    # the name will be updated if user provides customized data_config
+    data_config: "perf_tuning_data"
 systems:
   local_system:
     type: "LocalSystem"
--- a/olive/cli/perf_tuning.py
+++ b/olive/cli/perf_tuning.py
@@ -15,6 +15,7 @@ import yaml
 from olive.auto_optimizer.template_mapping import PERF_TUNING_TEMPLATE
 from olive.cli.base import BaseOliveCLICommand
 from olive.common.utils import set_nested_dict_value, set_tempdir
+from olive.data.config import DataConfig
 from olive.workflows import run as olive_run
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


olive/cli/perf_tuning.py
```diff
@@ -0,0 +1,317 @@
...
+            "--past_seq_len",
+            type=int,
+            help="Past sequence length to use for the input data.",
+        )
+        dataset_group.add_argument(
+            "--max_seq_len",
+            type=int,
+            help="Max sequence length to use for the input data.",
+        )
+        dataset_group.add_argument(
```
jambayk, line 79: this might not be required? since our evaluator can cast input data to the expected dtype?

olive/cli/perf_tuning.py
```diff
@@ -0,0 +1,317 @@
...
+            action="store_true",
+            help="Whether to use key-value cache for perf_tuning",
+        )
+        dataset_group.add_argument(
+            "--hf_model_name",
+            required=True,
+            help="Huggingface model name used to load model configs from huggingface.",
+        )
+        dataset_group.add_argument(
+            "--batch_size",
```
jambayk, line 60: are these options always required to be passed? If so, can we set some default values instead?

olive/cli/perf_tuning.py
```diff
@@ -0,0 +1,317 @@
...
+
+        # dataset options
+        dataset_group = sub_parser.add_argument_group("dataset options")
+        dataset_group.add_argument(
+            "--data_config_path",
+            type=str,
+            help="Path to the data config file. It allows to customize the data config for the model.",
+        )
+        dataset_group.add_argument(
+            "--predict_with_kv_cache",
```
jambayk, line 50: maybe we could be smart about this and infer this and past-value names directly from the model inputs itself? We could still provide the user options to pass these if they use non-standard input names.

devang-ml (commented):
We need to be clear that this command tunes ORT session parameters. May be 'tune-session-params' as the sub command would help?

olive/cli/perf_tuning.py
```diff
@@ -0,0 +1,317 @@
...
+
+class PerfTuningCommand(BaseOliveCLICommand):
+
+    @staticmethod
+    def register_subcommand(parser: ArgumentParser):
+        sub_parser = parser.add_parser(
+            "perftuning",
+            help=(
+                "Automatically tune the inference performance(latency) for a given onnx model. "
+                "Currently, we only support tuning the model converted from huggingface model."
```
devang-ml, line 31: Automatically tune the session parameters for inference performance ...

xiaoyu-work:
Can you add a documentation [here](https://microsoft.github.io/Olive/features/cli.html)?

olive/cli/perf_tuning.py
trajepl: It is not required. I think only the fileds with `required=True` are required as input.
For those value which are not specified, if their `action` is `store_true|store_false`, the default value is false or true. If their type is `int|str|..`, the default value is none which will be ignored later.

trajepl: alt text

olive/cli/perf_tuning.py
trajepl: It is not required. I think only the fileds with `required=True` are required as input.
For those value which are not specified, if their action is `store_true|store_false`, the default value is false or true. If their type is `int|str|..`, the default value is `none` which will be ignored later.

olive/cli/perf_tuning.py
trajepl: It is used to distinguish prompt latency or token latency and required the onnx model with kv pairs in input_names.

Maybe we can change a name? I did not use prompt or token in name, because somehow the `token` happens on other classification task as well.

olive/cli/perf_tuning.py
trajepl: Removed.

trajepl:
> Can you add a documentation [here](https://microsoft.github.io/Olive/features/cli.html)?

The cli document will be auto-generated by sphinx doc framework.

olive/auto_optimizer/workflow_template/ort_perf_tuning.yaml
```diff
@@ -0,0 +1,17 @@
+input_model:
+  type: "ONNXModel"
+  model_path: <<model_path>>
+passes:
+  perf_tuning:
+    type: "OrtPerfTuning"
+    data_config:
```
jambayk, line 7: @shaahji brought up that we don't want to support having data config dicts passed directly to metrics or passes anymore in a workflow config. 
In order to avoid conflicts in the future, can you define this in data_configs and refer to the name like in our examples?

olive/cli/perf_tuning.py
```diff
@@ -0,0 +1,312 @@
...
+            "--data_config_path",
+            type=str,
+            help="Path to the data config file. It allows to customize the data config for the model.",
+        )
+        dataset_group.add_argument(
+            "--predict_with_kv_cache",
+            action="store_true",
+            help="Whether to use key-value cache for perf_tuning",
+        )
+        dataset_group.add_argument(
```
jambayk, line 56: maybe the hf dataset params should be in a separate group than the `dataset options` group? If data config path is provided, all of the other options are not relevant anymore. Don't know if it's possible to create two mutually exclusive groups but atleast grouping them separately would be less confusing.

olive/cli/perf_tuning.py
```diff
@@ -0,0 +1,312 @@
...
+        args_dict = vars(self.args)
+        pass_config.update({k: args_dict[k] for k in pass_config_keys if args_dict[k] is not None})
+        return pass_config
+
+    def _get_data_config(self, template_config=None) -> Dict:
+        if not self.args.data_config_path:
+            if template_config is None:
+                raise ValueError("Template config is required when data config is not provided.")
+            return self._update_default_data_config_params(template_config["passes"]["perf_tuning"]["data_config"])
+        data_config_path = Path(self.args.data_config_path)
```
jambayk, line 255: we also have `DataConfig.parse_file_or_obj` that handles this logic automatically.

olive/auto_optimizer/workflow_template/ort_perf_tuning.yaml
trajepl: Updated.

olive/cli/perf_tuning.py
trajepl: Updated.

jambayk (dismissed):

olive/cli/perf_tuning.py
```diff
@@ -0,0 +1,314 @@
...
+        args_dict = vars(self.args)
+        pass_config.update({k: args_dict[k] for k in pass_config_keys if args_dict[k] is not None})
+        return pass_config
+
+    def _get_data_config(self, template_config=None) -> Dict:
+        if not self.args.data_config_path:
+            if template_config is None:
+                raise ValueError("Template config is required when data config is not provided.")
+            return self._update_default_data_config_params(template_config["data_configs"][0])
+        return DataConfig.parse_file_or_obj(self.args.data_config_path)
```
jambayk, line 263: Btw, might have to update the name of the data config here or the pass's param to match the user given name?

jambayk (approved):

PR #1284, opened 2024-08-07
Add cli to remove model from cloud cache

xiaoyu-work:
## Describe your changes

Add cli to remove model from cloud cache

New cli command to delete models from cloud cache. Please check cli documentation for more details.

### DIFF
--- a/olive/cli/cloud_cache.py
+++ b/olive/cli/cloud_cache.py
@@ -1,0 +1,64 @@
+import logging
+
+from olive.cli.base import BaseOliveCLICommand
+from olive.common.utils import get_credentials
+
+logger = logging.getLogger(__name__)
+
+
+class CloudCacheCommand(BaseOliveCLICommand):
+    @staticmethod
+    def register_subcommand(parser):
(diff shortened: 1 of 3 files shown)


olive/cli/cloud_cache.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        sub_parser.add_argument(
+            "--container",
+            type=str,
+            required=True,
+            help="The container name for the cloud cache.",
+        )
+        sub_parser.add_argument(
+            "--model_hash",
+            type=str,
```
devang-ml, line 32: How would user know which model_hash to remove?

xiaoyu-work: User has to check the Storage blob or workflow log by himself to get the model hash. There is no way to directly get the model hash. Another option in my head is to have a config file as input, a specific command to run this workflow, but instead of "running the workflow in order to get the output model", this run will delete all model cache about this config in the cloud cache.

jambayk (dismissed):

jambayk (approved):

PR #1293, opened 2024-08-09
Update remote workflow and huggingface doc

xiaoyu-work:
## Describe your changes

Update remote workflow and huggingface doc

### DIFF
--- a/docs/source/features/run_workflow_remotely.md
+++ b/docs/source/features/run_workflow_remotely.md
@@ -45,4 +45,4 @@ The artifacts of the workflow, including cache and outputs, will be automaticall
 }
 ```
 
-The cache and outputs will be exported to `<datastore>/workflow_artifacts/<workflow_id>/cache` and `<datastore>/workflow_artifacts/<workflow_id>/output`.
+The cache and outputs will be exported to `<datastore>/workflow_artifacts/<workflow_id>/cache` and `<datastore>/workflow_artifacts/<workflow_id>/output`. You can check logs in your job to find the detailed output path.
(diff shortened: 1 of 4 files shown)


docs/source/features/run_workflow_remotely.md
```diff
@@ -35,3 +31,18 @@ python -m olive run --config <config_file.json>
...
+```json
+"aml_system": {
+    "type": "AzureML",
+    "config": {
+        "datastore": "<datastore-name>"
+    }
+}
+```
+
+The cache and outputs will be exported to `<datastore>/workflow_artifacts/<workflow_id>/cache` and `<datastore>/workflow_artifacts/<workflow_id>/output`.
```
jambayk, line 48: can you update aml system to also log the resolved path. The user has no information about the default values of `datastore` and `workflow_id`

xiaoyu-work: Updated

olive/systems/azureml/aml_workflow_runner.py
```diff
@@ -43,14 +43,17 @@ def main(raw_args=None):
     remote_cache_dir = workflow_artifacts_path / "cache"
     remote_output_dir = workflow_artifacts_path / "output"
 
+    print("The cache of this workflow is stored at: ", str(remote_cache_dir))
```
jambayk, line 46: I don't think it's useful here. this will just print the path where the datastore is mounted as. something like `/tmp/...`. I was suggesting adding the log to the aml system before it submits the job. like here https://github.com/microsoft/Olive/blob/213ac7ae5361351d6d7a7a5f5f26039355f16818/olive/systems/azureml/aml_system.py#L130

xiaoyu-work: If this is a relative path then it would be fine. Yeah but you are right I can add the log to aml system.

jambayk (dismissed):

jambayk (approved):

PR #1294, opened 2024-08-09
LoRA: Remove `ORTTrainer` support, embedding resize, `eval_dataset_size`, duplicate examples

jambayk:
## Describe your changes
**Data:**
- Rename `label_cols` to `label_col` and make its type string. We only use `label_cols[0]` which makes the name and list typing confusing

**LoRA/QLoRA:**
- Remove `smart_tokenizer_and_embedding_resize` which was used to add a pad_token if not present. This was copied from the original qlora implementation, but it is unnecessary and not efficient. Most training scripts use the eos_token as pad_token. This is fine as long as the attention mask and labels are correctly masked.
- Remove `eval_dataset_size` option which was used to split train data into train and eval. this is a specific use case which is supported using data configs that use slices of the train data. We want to remove as much of the data processing logic from the pass to make it generic
- Data collator is kept for now since the trainer has its own batch size (which depends on batch size params as well as number of gpus). It is more generic now.
- Remove support for `ORTTrainer`. This feature is not used and hasn't been validated for newer models or transformers version.
- Remove `phi` and `open_llama` qlora examples since these are copies of the examples for `llama`, `phi2` and `phi3`. Keep multiple copies of essentially the same workflow is hard to maintain when we make changes. 
- Other lora examples updated to remove redundant options.


jambayk:
Will split into multiple PRs

PR #1295, opened 2024-08-09
Handle metric priority correctly so validation can pass

shaahji:
## Handle metric priority correctly so validation can pass

priority works across metrics, however, aml evaluation works only a single metric at a time. So, metric validation was failing.

### DIFF
--- a/olive/systems/azureml/aml_system.py
+++ b/olive/systems/azureml/aml_system.py
@@ -612,10 +612,15 @@ class AzureMLSystem(OliveSystem):
             outputs = {}
             for metric in evaluator_config.metrics:
                 metric_tmp_dir = tmp_dir / metric.name
+                metric_clone = deepcopy(metric)
+                if metric_clone.sub_types:
+                    metric_clone.sub_types = sorted(metric_clone.sub_types, key=lambda st: st.priority, reverse=True)
+                    for priority, sub_type in enumerate(metric_clone.sub_types):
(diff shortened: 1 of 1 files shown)


olive/systems/azureml/aml_system.py
```diff
@@ -612,10 +612,15 @@ def evaluate_pipeline():
             outputs = {}
             for metric in evaluator_config.metrics:
                 metric_tmp_dir = tmp_dir / metric.name
+                metric_clone = deepcopy(metric)
+                if metric_clone.sub_types:
+                    metric_clone.sub_types = sorted(metric_clone.sub_types, key=lambda st: st.priority, reverse=True)
+                    for priority, sub_type in enumerate(metric_clone.sub_types):
+                        sub_type.priority = priority + 1
```
jambayk, line 619: this works too or we can just assign all priorities to `-1` since the priority is irrelevant for the evaluation.

shaahji: I tried that and it doesn't work. The logic triggers this check -

https://github.com/microsoft/Olive/blob/b1e051e06122d914ca7aeaeabcc78486f476bb20/olive/evaluator/olive_evaluator.py#L1140

For it to work, I will have to remove the `name` field of the subtype.

jambayk: Oh that makes sense. Thanks! Approved

jambayk (approved):

PR #1297, opened 2024-08-10
Add cloud cache support for aml models

xiaoyu-work:
## Describe your changes

Add cloud cache support for aml models

### DIFF
--- a/olive/engine/cloud_cache_helper.py
+++ b/olive/engine/cloud_cache_helper.py
@@ -13,7 +13,7 @@ from typing import Any, Dict, Optional
 from olive.common.config_utils import ConfigBase
 from olive.common.utils import get_credentials, hash_dict
 from olive.model.config.model_config import ModelConfig
-from olive.resource_path import ResourceType, create_resource_path
+from olive.resource_path import create_resource_path
 
 logger = logging.getLogger(__name__)
 
@@ -337,9 +337,6 @@ def is_valid_cloud_cache_model(model_config: ModelConfig):
     """
     if model_config.type == "hfmodel":
         resource_path = create_resource_path(model_config.config.get("model_path"))
-        if (
-            resource_path.is_string_name()
-            or resource_path.is_azureml_models()
-        ):
+        if resource_path.is_string_name() or resource_path.is_azureml_models():
             return True
     return False
--- a/olive/engine/cloud_cache_helper.py
+++ b/olive/engine/cloud_cache_helper.py
@@ -13,7 +13,7 @@ from typing import Any, Dict, Optional
 from olive.common.config_utils import ConfigBase
 from olive.common.utils import get_credentials, hash_dict
 from olive.model.config.model_config import ModelConfig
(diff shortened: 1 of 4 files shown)


olive/engine/cloud_cache_helper.py
jambayk, line 1: this PR supports aml models when running from local workflow host. However, it still doesn't support aml models when running on an aml workflow host. To the aml host, the model appears as a local model so some more bookkeeping would be required to inform it that it is an aml model. 

Will this be implemented soon? As a follow-up PR of part of this?

olive/engine/engine.py
```diff
@@ -803,21 +802,15 @@ def _run_passes(
...
-            if not (
-                model_config.type.lower() == "hfmodel"
-                and create_resource_path(model_config.config.get("model_path")).is_string_name()
-            ):
+            if not is_valid_cloud_cache_model(model_config):
                 logger.warning(
-                    "Only HfModel with huggingface id as model_path is supported by cloud cache. Setting"
-                    " enable_cloud_cache=False."
+                    "Only HfModel with huggingface id as model_path "
+                    "or HFModel from Azure ML registry model is supported by cloud cache. "
```
devang-ml, line 808: - HfModel vs HFModel
- HFModel is Olive concept that user may need to be familiar with.

olive/engine/cloud_cache_helper.py
xiaoyu-work: Yeah i'm still looking into aml models with aml workflow host.

olive/engine/cloud_cache_helper.py
```diff
@@ -92,22 +99,20 @@ def get_hash_key(
...
         hf_hub_model_commit_id = None
         model_config_copy = deepcopy(model_config)
-        if (
-            input_model_hash is None
-            and model_config.type.lower() == "hfmodel"
-            and create_resource_path(model_config.config["model_path"]).is_string_name()
-        ):
-            hf_hub_model_commit_id = repo_info(model_config.config["model_path"]).sha
+        model_path_resource_path = create_resource_path(model_config.config.get("model_path"))
+        if input_model_hash is None:
```
jambayk, line 105: following the logic in engine `input_model_hash` is always called with `none` so the else part is never run. Is there a reason for this?

xiaoyu-work: Only the first input model hash for first Pass is None. The following pass's input model hash will be the output model hash from previous pass run returns in the for loop.

olive/engine/engine.py
jambayk: the inconsistency between `HfModel` and `HFModel` in this docstring still exists. My intention was to use `HfModel` as the casing but I missed updating here https://github.com/microsoft/Olive/blob/9e2c817b126f38d462680d72b20def711b26c668/olive/model/handler/hf.py#L27

xiaoyu-work: Will update to HfModel.

jambayk (approved):

PR #1311, opened 2024-08-14
Add aml model support for remote workflow

xiaoyu-work:
## Describe your changes

Add aml model support for remote workflow

### DIFF
--- a/olive/systems/azureml/aml_system.py
+++ b/olive/systems/azureml/aml_system.py
@@ -156,6 +156,15 @@ class AzureMLSystem(OliveSystem):
         cur_dir = Path(__file__).resolve().parent
         code_files = [cur_dir / script_name]
         self.copy_files(code_files, config_root)
+
+        if run_config.input_model.config.get("model_path"):
+            input_model_resource_path = create_resource_path(run_config.input_model.config["model_path"])
+            if (
+                input_model_resource_path.is_azureml_models()
+                and run_config.engine.cloud_cache_config.enable_cloud_cache
+            ):
+                run_config.engine.cloud_cache_config.input_model_identifier = input_model_resource_path.get_path()
+
         workflow_config = run_config.to_json(make_absolute=False)
         [
             workflow_config.pop(component, None)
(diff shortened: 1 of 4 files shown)


olive/systems/azureml/aml_system.py
```diff
@@ -301,6 +304,9 @@ def create_inputs_and_args(
                         self._create_arg_and_input_from_resource_path(resource_path)
                     )
                     all_resources[resource_str] = resource_input_name
+                    # For cloud cache, we need to pass the azreml model path to the job
+                    if name == WORKFLOW_CONFIG and resource_path.is_azureml_models():
```
jambayk, line 308: there is no guarantee that this resource is the `model_path`. It could be a data file, user_script, etc.

jambayk: Instead of doing it hackily like this, I think it's easier for the workflow pipeline created to check the input model's model path and directly update the cloud cache config with this info. that way, we don't reserve an arbitrary key and there's no need to update the arg parser. 

somewhere around https://github.com/microsoft/Olive/blob/35f29f12db322eefd461773fd5a759ca928b6107/olive/systems/azureml/aml_system.py#L164
just need to also ignore `cloud_cache_config` here https://github.com/microsoft/Olive/blob/35f29f12db322eefd461773fd5a759ca928b6107/olive/systems/azureml/aml_system.py#L168 so that it doesn't get replaced.

xiaoyu-work: what do you mean by there is no guarantee it is not the model_path?

jambayk: create_input_and_args iterates through all of the resources in a config. So this condition will be true even when it meets a resource that is not the input model_path.

xiaoyu-work: To my understanding, even aml model path is used in other places, as for this config, the input model must be this same aml model, is this correct? Do we use aml model path somewhere else but having different input model?

jambayk: I can see a scenario where we might use aml model as part of a data config but not in the input model. So I thought it's better to be explicit about it and just update the cloud cache config directly in the aml system without going through a complicated path

xiaoyu-work: i see. I can update this.

jambayk (approved):

olive/systems/azureml/aml_system.py
jambayk: Thanks!

PR #1314, opened 2024-08-14
👓  phi3-vision tutorial

trajepl:
## Describe your changes

 phi3-vision tutorial

image

### DIFF
--- a/examples/phi3/README_VISION.md
+++ b/examples/phi3/README_VISION.md
@@ -1,0 +1,50 @@
+# Phi3 optimization with Olive
+This folder contains an example of optimizing the microsoft/Phi-3-vision-128k-instruct model from [Hugging Face](https://huggingface.co/microsoft/Phi-3-vision-128k-instruct) with Olive.
+
+## Prerequisites
+Please refer to the [pre-requisites of Phi3-vision optimization with GenAI](https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/phi-3-vision.md#0-pre-requisites) for more details.
+Basically, you need to install the following packages:
+- `huggingface_hub[cli]`
+- `numpy`
+- `onnx`
+- `ort-nightly>=1.19.0.dev20240601002` or `ort-nightly-gpu>=1.19.0.dev20240601002`
+    - [ORT nightly package](https://onnxruntime.ai/docs/install/#inference-install-table-for-all-languages) is needed until the latest changes are in the newest ORT stable package
+    - For CPU:
+    ```bash
+    pip install ort-nightly --index-url=https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/
+    ```
+    - For CUDA 11.X:
+    ```bash
+    pip install ort-nightly-gpu --index-url=https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ORT-Nightly/pypi/simple/
+    ```
+    - For CUDA 12.X:
+    ```bash
+    pip install ort-nightly-gpu --index-url=https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ort-cuda-12-nightly/pypi/simple/
+    ```
+- `pillow`
+- `requests`
+- `torch`
+- `torchvision`
+- `transformers`
+
+
+### For optimizing model from Hugging Face
+if you have not logged in Hugging Face account,
+- Install Hugging Face CLI and login your Hugging Face account for model access
+```
+huggingface-cli login
+```
+
+## Usage
+we will use the `phi3_vision.py` script to optimize model for a chosen hardware target by running the following commands.
+
+```
+python phi3.py [--target HARDWARE_TARGET(cpu|cuda)] [--precision DATA_TYPE] [--output_dir OUTPUT_DIR] [--cache_dir CACHE_DIR] [--inference] [--optimized_model_path OPT_MODEL_PATH]
+
+# Examples
+# optimize model and start a inference process
+python phi3_vision.py --cache_dir cache/ --output_dir output --inference
+
+# inference with optimized model
+python phi3_vision.py --optimized_model_path --inference
+```
--- a/examples/phi3/README_VISION.md
+++ b/examples/phi3/README_VISION.md
@@ -2,11 +2,11 @@
 This folder contains an example of optimizing the microsoft/Phi-3-vision-128k-instruct model from [Hugging Face](https://huggingface.co/microsoft/Phi-3-vision-128k-instruct) with Olive.
 
 ## Prerequisites
-Please refer to the [pre-requisites of Phi3-vision optimization with GenAI](https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/phi-3-vision.md#0-pre-requisites) for more details.
-Basically, you need to install the following packages:
-- `huggingface_hub[cli]`
-- `numpy`
-- `onnx`
+Install python packages
+```
+pip install -r requirements-vision.txt
+```
+Also, this example requires org-nightly package for ORT. Please install the following packages as you need:
 - `ort-nightly>=1.19.0.dev20240601002` or `ort-nightly-gpu>=1.19.0.dev20240601002`
     - [ORT nightly package](https://onnxruntime.ai/docs/install/#inference-install-table-for-all-languages) is needed until the latest changes are in the newest ORT stable package
     - For CPU:
@@ -21,12 +21,6 @@ Basically, you need to install the following packages:
     ```bash
     pip install ort-nightly-gpu --index-url=https://aiinfra.pkgs.visualstudio.com/PublicPackages/_packaging/ort-cuda-12-nightly/pypi/simple/
     ```
-- `pillow`
-- `requests`
-- `torch`
-- `torchvision`
-- `transformers`
-
 
 ### For optimizing model from Hugging Face
 if you have not logged in Hugging Face account,
--- a/examples/phi3/README_VISION.md
+++ b/examples/phi3/README_VISION.md
@@ -43,8 +43,8 @@ python phi3.py [--target HARDWARE_TARGET(cpu|cuda)] [--precision DATA_TYPE] [--o
 
 # Examples
 # optimize model and start a inference process
-python phi3_vision.py --cache_dir cache/ --output_dir output --inference
+python phi3_vision.py --cache_dir cache --output_dir output --inference
 
 # inference with optimized model
-python phi3_vision.py --optimized_model_path --inference
+python phi3_vision.py --optimized_model_path optimized_model_path --inference
 ```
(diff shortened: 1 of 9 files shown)


examples/phi3/README_VISION.md
```diff
@@ -0,0 +1,50 @@
+# Phi3 optimization with Olive
+This folder contains an example of optimizing the microsoft/Phi-3-vision-128k-instruct model from [Hugging Face](https://huggingface.co/microsoft/Phi-3-vision-128k-instruct) with Olive.
+
+## Prerequisites
+Please refer to the [pre-requisites of Phi3-vision optimization with GenAI](https://github.com/microsoft/onnxruntime-genai/blob/main/examples/python/phi-3-vision.md#0-pre-requisites) for more details.
+Basically, you need to install the following packages:
+- `huggingface_hub[cli]`
```
jambayk, line 7: maybe the requirements other than ort-nightly can be moved into a requirement.txt file? Otherwise, the user needs to copy paste each one

examples/phi3/phi3_vision.py
```diff
@@ -0,0 +1,362 @@
...
+        default="cpu",
+        required=False,
+        choices=TARGETS,
+        help=f"Choose from {TARGETS}",
+    )
+    parser.add_argument(
+        "--precision",
+        type=str,
+        default="int4",
+        choices=["int4"],
```
jambayk, line 38: float models are not supported?

examples/phi3/vision/config_templates/vision_config.json
```diff
@@ -0,0 +1,46 @@
...
+            "input_types": [ "float32", "int64" ],
+            "input_shapes": [ [ 1, 17, 3, 336, 336 ], [ 1008, 1344 ] ],
+            "output_names": [ "visual_features" ],
+            "dynamic_axes": {
+                "pixel_values": { "0": "num_images", "1": "max_num_crops", "3": "height", "4": "width" },
+                "image_sizes": { "0": "num_images" },
+                "visual_features": { "0": "batch_size", "1": "num_img_tokens" }
+            }
+        },
+        "dummy_inputs_func": "get_dummy_inputs"
```
jambayk, line 18: Is the dummy input funce needed since we have all of the input shapes and types already in the ioconfig?

examples/phi3/phi3_vision.py
```diff
@@ -0,0 +1,362 @@
...
+    import numpy as np
+    import torch
+    from onnx import TensorProto, external_data_helper, helper, numpy_helper, save_model
+    from transformers import AutoConfig, AutoModelForCausalLM
+
+    model = AutoModelForCausalLM.from_pretrained(input_model_path, trust_remote_code=True)
+    config = AutoConfig.from_pretrained(input_model_path, trust_remote_code=True)
+
+    # User inputs
+    io_dtype = TensorProto.FLOAT16 if args.precision == torch.float16 else TensorProto.FLOAT
```
jambayk, line 97: args.precision appears to be a string with only "int4" as option. when would this case be true? I think we expect to use fp16 if the target is cuda and fp32 when it is cpu?

examples/phi3/vision/config_templates/vision_config.json
```diff
@@ -0,0 +1,46 @@
+{
+    "input_model": {
+        "type": "PyTorchModel",
+        "model_path": "<<str: model_path>>",
+        "model_loader": "vision_embed_tokens_loader",
+        "model_script": "scripts/user_script.py",
```
jambayk, line 6: this path is relative to the "vision" folder but we are running from "phi3"

examples/phi3/phi3_vision.py
```diff
@@ -0,0 +1,362 @@
...
+    else:
+        config["passes"]["convert"]["torch_dtype"] = "float16"
+        config["passes"]["convert"]["device"] = "cuda"
+        config["systems"]["local_system"]["accelerators"] = [
+            {"device": "GPU", "execution_providers": ["CUDAExecutionProvider"]}
+        ]
+
+    config["engine"] = {
+        "cache_dir": Path(args.cache_dir).resolve() / "vision",
+        "output_dir": Path(args.output_dir).resolve() / "vision",
```
jambayk, line 283: the output of the config run still stays in the output directory. It should probably be cleaned up after the resave

examples/phi3/phi3_vision.py
```diff
@@ -0,0 +1,362 @@
...
+        processor = AutoProcessor.from_pretrained(input_model_path, trust_remote_code=True)
+        model = AutoModelForCausalLM.from_pretrained(input_model_path, trust_remote_code=True)
+        del model, processor, config
+    except Exception as e:
+        print(f"Error loading model: {e}")
+        return False
+    return True
+
+
+def build_text_embedding(args, input_model_path):
```
jambayk, line 84: Is this manual building of the model necessary? I am able to generate the same graph using onnx conversion of the embed_tokens submodule https://github.com/microsoft/Olive/commit/777d82250e8a858b1fe584aa90e6b5b56164b478

examples/phi3/README_VISION.md
trajepl: Updated.

examples/phi3/phi3_vision.py
trajepl: Add fp16 support. The fp32 of onnx model run failed, 

[E:onnxruntime:onnxruntime-genai, sequential_executor.cc:516 ExecuteKernel] Non-zero status code returned while running Expand node. Name:'/model/attn_mask_reformat/input_ids_subgraph/Expand' Status Message: /model/attn_mask_reformat/input_ids_subgraph/Expand: left operand cannot broadcast on dim 3 LeftShape: {1,1,2500,2500}, RightShape: {1,1,2500,5572}

examples/phi3/phi3_vision.py
trajepl: Updated.

examples/phi3/vision/config_templates/vision_config.json
trajepl: The value will be updated in phi3-vision.py. To avoid confusion, updated.

examples/phi3/phi3_vision.py
trajepl: The output_dir can be changed to another one. Since the model name must follow the given format to aligh the that in generation_config.json.

examples/phi3/phi3_vision.py
trajepl: https://huggingface.co/microsoft/Phi-3-vision-128k-instruct-onnx-cpu/blob/main/onnx/builder.py
It is borrowed from huggingface builder.py which is mentined in genai's tutorial as well.
I am not sure if it is totally same with build_text_embedding.

jambayk: Since the graph looks the same, could you try the model generated from the config. 

It feels unnecessary and non standard to manually generate the graph is a simple exported model works.

examples/phi3/phi3_vision.py
```diff
@@ -0,0 +1,378 @@
...
+        save_as_external_data=pass_config.get("save_as_external_data", False),
+        all_tensors_to_one_file=pass_config.get("all_tensors_to_one_file", True),
+        external_data_name=pass_config.get("external_data_name", None),
+        size_threshold=pass_config.get("size_threshold", 1024),
+        convert_attribute=pass_config.get("convert_attribute", False),
+    )
+
+
+def main(raw_args=None):
+    args = get_args(raw_args)
```
jambayk, line 201: maybe you can check that fp16 is not allowed for cpu target?

examples/phi3/phi3_vision.py
```diff
@@ -0,0 +1,378 @@
...
+        pass_config = output_node.pass_run_config
+        resave_onnx_model(model_path, output_dir / "phi-3-v-128k-instruct-text.onnx", pass_config)
+    except Exception:
+        config_link = "https://huggingface.co/microsoft/Phi-3-vision-128k-instruct-onnx-cpu/tree/main/cpu-int4-rtn-block-32-acc-level-4."
+        print(
+            "Ignore the error during Olive run.\nThis script will copy the "
+            "genai_config.json and processor_config.json from ",
+            config_link,
+        )
+        # even if the Olive run fails, partially generated model
```
jambayk, line 260: Do you know what the failure could be? Does it fail during quantization?

examples/phi3/phi3_vision.py
```diff
@@ -0,0 +1,378 @@
...
+    config["input_model"]["model_path"] = input_model_path
+    config["input_model"]["model_script"] = config_path.parent / "scripts" / "user_script.py"
+
+    if args.precision == "fp16" or (args.precision == "int4" and args.target == "cuda"):
+        config["passes"]["convert"]["torch_dtype"] = "float16"
+    else:
+        config["passes"]["convert"]["torch_dtype"] = "float32"
+
+    if args.target == "cpu":
+        config["passes"]["convert"]["torch_dtype"] = "float32"
```
jambayk, line 285: nit: this looks redundant? Since the float16 case should only happen for cpu if we enforce the check earlier.

examples/phi3/phi3_vision.py
jambayk: Since the workflow output_dir (not the output dir of this script) contents are already resaved into the final directory, I think this still needs to be cleaned up regardless of where it is. 

Otherwise, we just have copies of the same model lying around.

examples/phi3/phi3_vision.py
trajepl: Failure in model builder. They had a fix in main branch already. After the new version of genai is released, this failure will be resolved.

trajepl: https://github.com/microsoft/onnxruntime-genai/blob/cb313b49adec51f8679701df7d2451c9d21374a2/src/python/py/models/builder.py#L267C1-L269C129

examples/phi3/phi3_vision.py
trajepl: image
Conversion cannot not work directly. Seems the generator requires the inputs name of `inputs_embeds`.

trajepl: It worked. Just updated.

examples/phi3/phi3_vision.py
trajepl: Removed.

examples/phi3/phi3_vision.py
```diff
@@ -0,0 +1,315 @@
...
+        cache_dir = text_config.engine.cache_dir
+        pass_config = text_config.passes["builder"].config
+        # find the model file in the cache directory
+        model_path = list(cache_dir.rglob("*.onnx"))[-1]
+        resave_onnx_model(model_path, output_dir / "phi-3-v-128k-instruct-text.onnx", pass_config)
+    print("Model generation completed, output saved to ", output_dir)
+    # clean up the output directory in olive.workflows
+    to_remove_folders = [
+        Path(args.output_dir).resolve() / "vision",
+        Path(args.output_dir).resolve() / "text",
```
jambayk, line 182: text_embedding is missing here.

Otherwise, everything else looks good to me!

trajepl: Updated.

jambayk (approved):
Thanks!

PR #1322, opened 2024-08-16
🏪 Support opversion conversion for models with external data 

trajepl:


olive/passes/onnx/conversion.py
```diff
@@ -518,12 +518,27 @@ def _run_for_config(
         self, model: ONNXModelHandler, config: Dict[str, Any], output_model_path: str
     ) -> ONNXModelHandler:
         # get current models's opset version
-        model_proto = model.load_model()
+        if config["save_as_external_data"]:
```
jambayk, line 521: I don't know if this is a good indicator of the model having external data. The user might not have provided this as true even though the input model has external data since olive will save as external data if the model is too big anyways. 

Since there is a way to load the external weights into the model proto at the end, why not just always load without external data? 
If this is prohibitive for some reason, a better indicator could be whether the `"model_path"` resource is a directory or a file.

trajepl: > Since there is a way to load the external weights into the model proto at the end, why not just always load without external data?

`onnx.version_converter.convert_version` will fail with the error protobuf size > max_size if the input model goes big than the limited size.

jambayk (dismissed):

jambayk (approved):

jambayk (approved):

PR #1324, opened 2024-08-19
Shorten cache folder path

xiaoyu-work:
## Describe your changes

Shorten cache folder path

### DIFF
--- a/olive/cache.py
+++ b/olive/cache.py
@@ -12,7 +12,7 @@ from typing import TYPE_CHECKING, Dict, Optional, Union
 
 from olive.common.config_utils import ConfigBase, convert_configs_to_dicts, serialize_to_json, validate_config
 from olive.common.constants import DEFAULT_CACHE_DIR, DEFAULT_WORKFLOW_ID
-from olive.common.utils import hash_dict, set_nested_dict_value
+from olive.common.utils import hash_dict, hash_string, set_nested_dict_value
 from olive.resource_path import ResourcePath, create_resource_path, find_all_resources
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


olive/cache.py
```diff
@@ -105,12 +105,10 @@ def get_run_json_path(
         """Get the path to the run json."""
         pass_config_hash = hash_dict(pass_config)[:8]
         if not accelerator_spec:
-            run_json_path = self.dirs.runs / f"{pass_name}-{input_model_number}-{pass_config_hash}.json"
+            run_json_path = hash_string(f"{pass_name}-{input_model_number}-{pass_config_hash}")[:8]
```
trajepl, line 108: Will this bring inconvininence to debug/look up specific pass results in the cache folder?

shaahji: Rather than `[:8]` pass a parameter to the `hash_string` to generate a fixed length hash string.

shaahji (changes requested):
Update both `hash_string` and `hash_dict` to take the length of the output as input and update the code accordingly.

olive/cache.py
jambayk: I agree with Jiapeng's comment. Also, we currently have this feature where we can clean up the results of specific pass (that we might be testing) and its children output models https://github.com/microsoft/Olive/blob/7ab3f92a7b3151a25042067106644ebf2b4c9448/olive/cache.py#L115 

This feature depends on the name being part of the cache name. 

Is there really a need to shorten the cache path length. I think using a shorter hash than `8` might be sufficient if it's really needed.

xiaoyu-work: The basic idea is no user would care the intermediate models, as long as the output model path is correct. @jambayk is that `clean_pass_run_cache` used in our core logic?

jambayk: yes https://github.com/microsoft/Olive/blob/0a57b9d0095059a8ca8775e119ca540579405c82/olive/engine/engine.py#L120

olive/engine/engine.py
```diff
@@ -918,16 +918,18 @@ def _run_pass(
         input_model_number = input_model_id.split("_")[0]
         # Note: the final output model id need contains the accelerator information
         # if the output model is accelerator dependent.
+
+        model_number = self.cache.get_new_model_number()
```
jambayk, line 922: if we don't have any models that are cached with models, then this model_number is useless. It would always just return 0 (or maybe 1). 

It was originally added to get an alias for the model without having to refer to the whole model's id in cache

olive/engine/engine.py
```diff
@@ -918,16 +918,18 @@ def _run_pass(
         input_model_number = input_model_id.split("_")[0]
```
jambayk, line 918: this also doesn't work since there no nothing to split

PR #1326, opened 2024-08-21
ExtractAdapters: `dynamic_lora_r` and `optional_inputs`

jambayk:
## Describe your changes
ONNXModelHandler:
- New `change_model_path_to_dir` method. Currently, there are two different places where this logic is implemented manually. In onnx conversion, it doesn't update the model_file_name which leads to warning when resolving the onnx model file path 

ExtractAdapter:
-  `dynamic_lora_r`: Option to make the dimension for LoRA r static or dynamic.
- `optional_inputs`: Onnx and ort allow the same name to be used for graph initializers and inputs. When this happens, the input behaves like an optional input where the initializer is used if the input is not provided. https://github.com/onnx/onnx/blob/main/docs/IR.md#nodes 
  Using this behavior, we can set the default initializer for the lora weights to be empty tensors with shapes `[in_dim, 0]` and `[0, out_dim]`. When the lora weights are not provided during session run, these initializers would be used which is equivalent to running the base model with no adapters. 

### DIFF
--- a/olive/passes/onnx/extract_adapters.py
+++ b/olive/passes/onnx/extract_adapters.py
@@ -193,42 +193,47 @@ class ExtractAdapters(Pass):
         for node_name in nodes_to_remove:
             dag.remove_node(node_name, check_no_consumers=True)
 
-        if config["make_inputs"] and not config["pack_inputs"]:
-            # create inputs for the weights
-            for weight_name in weights:
-                dag.convert_initializer_to_input(weight_name)
-                self._make_dynamic_optional(dag, weights, weight_name, config)
+        if config["make_inputs"]:
+            if quant_modules and (config["dynamic_lora_r"] or config["optional_inputs"]):
+                # dynamic shape and optional inputs not supported for quantized modules yet
+                logger.warning("Quantized modules are not supported with dynamic_lora_r or optional_inputs. Ignoring.")
+                logger.debug("Quantized modules: %s", quant_modules)
 
-        elif config["make_inputs"] and config["pack_inputs"]:
-            # what weights are packed together
-            packed_weights, packings = self.pack_weights(weights, lora_modules, float_modules, quant_modules)
+            if not config["pack_inputs"]:
+                # create inputs for the weights
+                for weight_name in weights:
+                    dag.convert_initializer_to_input(weight_name)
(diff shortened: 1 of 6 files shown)


olive/passes/onnx/extract_adapters.py
```diff
@@ -350,3 +373,23 @@ def _create_empty_initializer(cls, dag: OnnxDAG, weights: Dict[str, "NDArray"],
         new_initializer = cls._copy_initializer(old_proto, new_name)
         # add the new initializer to the graph
         dag.add_initializer(new_initializer, dag.get_io(old_name).graph_idx)
+
+    @classmethod
+    def _make_dynamic_optional(cls, dag: OnnxDAG, weights: Dict[str, "NDArray"], name: str, config: Dict[str, Any]):
+        """Make the input dynamic and optional."""
+        if "quant" in name:
+            # dynamic shape and optional inputs not supported for quantized modules yet
```
devang-ml, line 381: nit pick: add warning.

olive/passes/onnx/extract_adapters.py
```diff
@@ -350,3 +373,23 @@ def _create_empty_initializer(cls, dag: OnnxDAG, weights: Dict[str, "NDArray"],
...
+            return
+
+        dim_idx = 1 if "lora_A" in name else 0
```
devang-ml, line 384: Do we have to hard code "lora_A"?

olive/passes/onnx/extract_adapters.py
jambayk: Will do the config check first and log a warning here for quant models.

olive/passes/onnx/extract_adapters.py
jambayk: the lora r dimension index depends on whether it's the A matrix or B matrix, so we need this information. The names here are after they have already been standardized https://github.com/microsoft/Olive/blob/056741839f1c0f8b7f9b7c1785b928c8781f37d7/olive/passes/onnx/extract_adapters.py#L319 so lora_A or lora_B is already present.

olive/passes/onnx/extract_adapters.py
jambayk: Done!

devang-ml (approved):

PR #1342, opened 2024-09-03
update mlflow logic

xiaoyu-work:
## Describe your changes

1. Update and simplify mlflow logic. 
New mlflow folder structure:
```
mlflow_folder/
├── model/
│   ├── model_files
│   └── config.json
└── components/
    └── tokenizer/
        └── tokenizer_files
```
2. Update downloading non-local resource logic. Add support to download azureml curated model (not sure which PR changes this logic before).

### DIFF
--- a/olive/common/hf/mlflow.py
+++ b/olive/common/hf/mlflow.py
@@ -49,7 +49,7 @@ def get_pretrained_name_or_path(model_name_or_path: str, name: str) -> str:
     mlflow_data = _load_mlflow_model_data(model_name_or_path)
     flavors = mlflow_data.get("flavors", {}).keys()
 
-    if flavors in ("hftransformers", "hftransformersv2"):
+    if {"hftransformers", "hftransformersv2"} & flavors.keys():
         path = parent_dir / "data" / name
         if path.exists():
             return str(path)
--- a/olive/common/hf/mlflow.py
+++ b/olive/common/hf/mlflow.py
@@ -47,7 +47,7 @@ def get_pretrained_name_or_path(model_name_or_path: str, name: str) -> str:
         path = parent_dir / "data" / name
         if path.exists():
             return str(path)
-    
+
     elif "transformers" in flavors:
         if name in ("model", "config"):
             path = parent_dir / "model"
--- a/olive/common/hf/mlflow.py
+++ b/olive/common/hf/mlflow.py
@@ -7,23 +7,26 @@ from pathlib import Path
 import yaml
 
 
-def load_mlflow_model_data(model_name_or_path: str) -> dict:
+def _load_mlflow_model_data(model_name_or_path: str) -> dict:
     yaml_path = Path(model_name_or_path) / "MLmodel"
 
     if not yaml_path.exists():
-        raise FileNotFoundError(f"No MLmodel file found at path: {yaml_path}")
+        return None
 
     with open(yaml_path) as fp:
         return yaml.safe_load(fp)
 
 
 def is_mlflow_transformers(model_name_or_path: str) -> bool:
-    mlflow_data = load_mlflow_model_data(model_name_or_path)
+    mlflow_data = _load_mlflow_model_data(model_name_or_path)
+
+    if not mlflow_data:
+        return False
 
     # TODO(trajep): let user specify flavor name if needed
     # to support other flavors in mlflow not only hftransformers
     flavors = mlflow_data.get("flavors", {})
-    if not flavors or not ({"hftransformersv2", "transformers"} & flavors.keys()):
+    if not flavors or not ({"hftransformers", "hftransformersv2", "transformers"} & flavors.keys()):
         raise ValueError(
             "Invalid MLFlow model format. Please make sure the input model"
             " format is the same as the result of mlflow.transformers.save_model,"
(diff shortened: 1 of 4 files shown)


jambayk (changes requested):
Please don't change the pre-existing logic. The logic was added to support different use cases for the mlflow models (hf model handler, data configs, etc). Some parts of the olive codebase expect the model, tokenizer and other metadata to be in the same format as hf checkpoints. But we also don't want to do this if not needed. 

since the only thing that changed is the mlflow version and the folder structure, we can just check the version in the yaml and do conditional logic for finding the tokenizer and config.

xiaoyu-work:
> Please don't change the pre-existing logic. The logic was added to support different use cases for the mlflow models (hf model handler, data configs, etc). Some parts of the olive codebase expect the model, tokenizer and other metadata to be in the same format as hf checkpoints. But we also don't want to do this if not needed.
> 
> since the only thing that changed is the mlflow version and the folder structure, we can just check the version in the yaml and do conditional logic for finding the tokenizer and config.

Given that the new MLflow structure has been released for almost a year, as for me I don't think we need to support the old one anymore. However, regarding your concerns, could you please point out where the old MLflow logic is still being used? I would like to double-check how those logics work.

olive/common/hf/mlflow.py
jambayk, line 1: Commenting here so that we can have a discussion thread instead of quote replies. 

Wrt to the ml flow version, it would have been fine if we only cared about custom aml models where the user uploads mlflow models themselves. However, aml registry models (curated models) are still on older versions of mlflow model structures
https://ml.azure.com/models/Phi-3-mini-4k-instruct/version/10/catalog/registry/azureml
https://ml.azure.com/models/microsoft-deberta-base-mnli/version/5/catalog/registry/HuggingFace

jambayk: all of the code that uses the config, tokenizer, model load from this utils are expected to work with the mlflow models https://github.com/microsoft/Olive/blob/main/olive/common/hf/utils.py 

like https://github.com/microsoft/Olive/blob/625161942dedf2660b6b53714149a4019fd06ea2/olive/data/component/pre_process_data.py#L10

jambayk: My suggestion would be to either read the mlflow version used to save the model from the yaml and check for the corresponding directory or just keep a list of possible directory structures and loop through them for the first match

xiaoyu-work: That looks weird. Those models have latest mlflow version but also using old structure. Thanks for this example. Let me look into this.

jambayk: Looks like the model catalog models were saved using azureml-evaluate-mlflow instead of mlflow directly. Like in https://github.com/microsoft/Olive/blob/625161942dedf2660b6b53714149a4019fd06ea2/test/unit_test/model/test_hf_model.py#L34

jambayk: looking at the files for the aml mlflow and regular mlflow models, I think we don't even need to look at the mlflow version. Only need to check for the flavor ("hftransformers", "hftransformersv2", "transformers")

xiaoyu-work: I added `transformers` while keeping the original logic. Thanks for the heads-up about the HF checkpoint folder structure.

olive/common/hf/mlflow.py
```diff
@@ -7,25 +7,28 @@
...
-            raise ValueError(
-                "Invalid MLFlow model format. Please make sure the input model"
-                " format is same with the result of mlflow.transformers.save_model,"
-                " or aml_mlflow.hftransformers.save_model from azureml.evaluate.mlflow"
-            )
+        return yaml.safe_load(fp)
+
+
+def is_mlflow_transformers(model_name_or_path: str) -> bool:
+    mlflow_data = load_mlflow_model_data(model_name_or_path)
```
jambayk, line 21: this would raise error for non mlflow models. line 14 should still return false

olive/common/hf/mlflow.py
```diff
@@ -37,14 +40,23 @@ def get_pretrained_name_or_path(model_name_or_path: str, name: str) -> str:
...
-    if pretrained_path.exists():
-        return str(pretrained_path)
+    mlflow_data = load_mlflow_model_data(model_name_or_path)
+    flavors = mlflow_data.get("flavors", {}).keys()
+
+    if "hftransformersv2" in flavors:
+        path = parent_dir / "data" / name
+        if path.exists():
+            return str(path)
+
```
jambayk, line 50: the logic from line 45-48 is still needed here. There are some models where the data subdirectories are not present

xiaoyu-work: Line 46 is `parent_dir / "data" / "model"`, do you mean `parent_dir / "model"`?

jambayk: `parent_dir / "data" / "model"` An example model https://ml.azure.com/models/Llama-2-7b-chat/version/24/catalog/registry/azureml-meta

I don't remember the details, but I think it depends on whether some experimental features are turned on when the model was saved.

jambayk: btw, could you add these models as comments for each case so we have context in the future? Thanks!

xiaoyu-work: Done

olive/common/hf/mlflow.py
jambayk, line 1: not as high priority since there are no current use cases for mlflow saved models but maybe we could add a test case similar to https://github.com/microsoft/Olive/blob/aa50a710f9f109e661cc6a347095992ab690e60e/test/unit_test/model/test_hf_model.py#L21

xiaoyu-work: Updated

jambayk: Thanks!

jambayk (approved):

PR #1343, opened 2024-09-03
🚘 Auto opt cli

trajepl:
## Describe your changes

Auto opt cli.

e.g. for bert model, we can optimize the model from huggingface and conduct onnx model with:
```sh

olive auto-opt --model Intel/bert-base-uncased-mrpc --data_config_path data_config.json --task text-classification

olive auto-opt --model Intel/bert-base-uncased-mrpc --data_config_path data_config.json --task text-classification --precision int4 --providers CPU

olive auto-opt --model Intel/bert-base-uncased-mrpc --data_config_path data_config.json --task text-classification --precision fp16 --providers CUDA

# use model builder
olive auto-opt --model microsoft/Phi-3-mini-4k-instruct --precision fp16 --providers CUDA --use_model_builder

```


olive/cli/auto_opt.py
```diff
@@ -0,0 +1,230 @@
...
+            help="Onnx input model path.",
+        )
+        model_group.add_argument(
+            "--task",
+            type=str,
+            default=None,
+            help="The task of the input model. Need to be specified if the input model is downloaded from huggingface.",
+        )
+        model_group.add_argument(
+            "--model_type",
```
devang-ml, line 88: Let's avoid overloading "model_type". May be "model_framework"? or "model_kind"? or something else?

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,230 @@
...
+            choices=["HfModel", "ONNXModel"],
+            help="model type, choose from HfModel and ONNXModel",
+        )
+
+        system_group = sub_parser.add_argument_group("system options")
+        system_group.add_argument(
+            "--device",
+            type=str,
+            default=None,
+            choices=["cpu", "gpu"],
```
devang-ml, line 100: Add "npu" also

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,230 @@
...
+                " will deduce from the value of execution providers, CPU/VistisAi for cpu and"
+                " CUDA/Tensorrt/Dml for gpu. If both device and execution providers are not specified,"
+                " default to cpu device with CPUExecutionProvider."
+            ),
+        )
+        system_group.add_argument(
+            "--providers",
+            type=str,
+            nargs="*",
+            choices=["CPU", "CUDA", "Tensorrt", "Dml", "VitisAI"],
```
devang-ml, line 113: QnnEP also

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,230 @@
...
+            ),
+        )
+
+        search_strategy_group = sub_parser.add_argument_group("search strategy options")
+        search_strategy_group.add_argument(
+            "--num-samples", type=int, default=5, help="Number of samples for search algorithm"
+        )
+        search_strategy_group.add_argument("--seed", type=int, default=0, help="Random seed for search algorithm")
+        search_strategy_group.add_argument(
+            "--execution_order",
```
devang-ml, line 146: "search_order" is clearer term here.

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,230 @@
...
+            type=str,
+            default="joint",
+            choices=["joint", "pass-by-pass"],
+            help="Execution order for search strategy",
+        )
+        search_strategy_group.add_argument(
+            "--search_algorithm",
```
devang-ml, line 153: We want to disable search, evaluation by default. Which means. dataset options should be optional.

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,230 @@
...
+            choices=["fp16", "fp32", "int4", "int8"],
+            help=(
+                "The output precision of the optimized model. If not specified, "
+                "the default precision is fp32 for cpu and fp16 for gpu"
+            ),
+        )
+
+        search_strategy_group = sub_parser.add_argument_group("search strategy options")
+        search_strategy_group.add_argument(
+            "--num-samples", type=int, default=5, help="Number of samples for search algorithm"
```
devang-ml, line 142: Do we really need to expose this in the CLI?

devang-ml (commented):
Let's add "auto-opt" needed packages list under extra dependencies so that user can do `pip install olive-ai[auto-opt]`

olive/cli/auto_opt.py
trajepl: adopted `model_framework`

olive/cli/auto_opt.py
trajepl: Updated.

olive/cli/auto_opt.py
trajepl: Updated.

olive/cli/auto_opt.py
trajepl: Updated.

olive/cli/auto_opt.py
trajepl: Updated.

olive/cli/auto_opt.py
trajepl: I was thinking if the search takes long time, user can somehow reduce the numbers of num-samples to stop the search in time.

trajepl:
> Let's add "auto-opt" needed packages list under extra dependencies so that user can do `pip install olive-ai[auto-opt]`

I think it might be hard to define a unified extra dependencies lists for auto-opt. Since auto-opt may be used in different device which cause the conflicts on onnxruntime version.

I think we have a feature to dynamicly `get_required_packages` based on the given accelerators which can be used in auto-opt.

devang-ml:
We can exclude OnnxRuntime and IHV toolkits. But we should be able to install other required packages using `pip install olive-ai[auto-opt]`.

trajepl:
> We can exclude OnnxRuntime and IHV toolkits. But we should be able to install other required packages using `pip install olive-ai[auto-opt]`.

Updated. Basically, `pip install olive-ai` is adequate to run auto-opt cli. Furthermore, to simplify the conversion, since olive currently use `optimum` to get onnx model, I added `optimum` to olive-ai[auto-opt]. 

Also I tested that, after the installation olive-ai[auto-opt], user only needs to install corresponding onnxruntime or onnxruntime-genai(_only enable --use_model_builder_) based on the device(cpu/gpu etc.), olive auto-opt cli can conduct reasonable results.

samuel100:
What models will be supported by `auto-opt`? When I have tried the auto optimizer using:

```json
{
    "input_model":{
        "type": "HfModel",
        "model_path": "microsoft/phi-3.5-mini-instruct",
        "task": "text-generation"
    },
    "systems": {
        "local_system": {
            "type": "LocalSystem",
            "accelerators": [
                {
                    "device": "cpu",
                    "execution_providers": [
                        "CPUExecutionProvider"
                    ]
                }
            ]
        }
    },
    "auto_optimizer_config": {
        "opt_level": 0,
        "disable_auto_optimizer": false,
        "precision": "int4"
    },
    "host": "local_system",
    "target": "local_system",
    "cache_dir": "cache",
    "output_dir" : "models"
}
```
I get an error message on the OrtTransformerOptimizer pass:

```
ValueError: Unsupported model type: phi3, please select one from [bart, bert, bert_tf, bert_keras, clip, gpt2, gpt2_tf, gpt_neox, swin, tnlr, t5, unet, vae, vit, conformer, phi] which need to be set under OrtTransformersOptimization.config
```

If these are the only models supported then it is a little underwhelming because they are pretty outdated. It is also a bit odd because the optimizer works for optimum models when I run `olive finetune`. 

As a general rule, a user should be able to plug in:

- A wide range of model architectures (for example, if Llama4 is released and it has the same architecture as Llama3, then I'd expect the tool to work).
- Device target
- precision

The user should not need to worry about adding data.

devang-ml:
The list and the error message is from OnnxRuntime's transformer optimizer.

devang-ml:
Model supported by the transformer optimizer are listed at https://github.com/microsoft/onnxruntime/blob/f4d62eeb2e058e2c7b5de0eaa9599368d32b23d5/onnxruntime/python/tools/transformers/optimizer.py#L50

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,255 @@
...
+
+    @staticmethod
+    def register_subcommand(parser: ArgumentParser):
+        sub_parser = parser.add_parser(
+            "auto-opt",
+            help=("Automatically performance optimize input model"),
+        )
+
+        # model options
+        model_group = sub_parser.add_argument_group("model options")
```
xiaoyu-work, line 71: Can you use [this](https://github.com/microsoft/Olive/blob/main/olive/cli/base.py#L105) and [this](https://github.com/microsoft/Olive/blob/main/olive/cli/base.py#L42) method here?

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,255 @@
...
+            "--search_algorithm",
+            type=str,
+            default="tpe",
+            choices=["exhaustive", "tpe", "random"],
+            help="Search algorithm for search strategy",
+        )
+
+        # output options
+        output_group = sub_parser.add_argument_group("output options")
+        output_group.add_argument(
```
xiaoyu-work, line 174: Not related to this PR but we could gather this to base.py as well.

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,255 @@
...
+
+        set_tempdir(self.args.output_path)
+        with tempfile.TemporaryDirectory() as tempdir:
+            run_config = self.get_run_config(tempdir)
+            olive_run(run_config)
+            if is_remote_run(self.args):
+                # TODO(jambayk): point user to datastore with outputs or download outputs
+                # both are not implemented yet
+                return
+            output_path = Path(self.args.output_path)
```
xiaoyu-work, line 199: I updated the cli output logic for other clis. Can you take a look?

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,255 @@
...
+            config["input_model"]["task"] = self.args.task
+        config["cache_dir"] = Path(tempdir) / "cache"
+        config["output_dir"] = self.args.output_path
+
+        device = self.args.device
+        if not device:
+            device = (
+                "gpu"
+                if self.args.providers
+                and any(p[: -(len("ExecutionProvider"))] in ["CUDA", "Tensorrt", "Dml"] for p in self.args.providers)
```
xiaoyu-work, line 224:
```
        system_group.add_argument(
            "--providers",
            type=str,
            nargs="*",
            choices=["CPU", "CUDA", "Tensorrt", "Dml", "VitisAI", "Qnn"],
            help="List of execution providers to use for optimization",
        )
```
without `ExecutionProvider`?

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,255 @@
...
+                else "cpu"
+            )
+        providers = self.args.providers or ["CPUExecutionProvider"] if device == "cpu" else ["CUDAExecutionProvider"]
```
xiaoyu-work, line 227: Same as above

olive/cli/auto_opt.py
xiaoyu-work: Should device be set to cpu if `VitisAI` and `Qnn` provided here?

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,255 @@
...
+        }
+
+        config["search_strategy"] = {
+            # TODO(anyone): rename execution_order to search_order in search_strategy
+            "execution_order": self.args.search_order,
+            "search_algorithm": self.args.search_algorithm,
+            "num_samples": self.args.num_samples,
+            "seed": self.args.seed,
+        }
+        # TODO(anyone): add remote options to auto opt passes
```
xiaoyu-work, line 246: Is remote option not working here?

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,255 @@
...
+    "search_strategy": {"execution_order": "joint", "search_algorithm": "tpe", "num_samples": 5, "seed": 0},
+    "systems": {
+        "local_system": {
+            "type": "LocalSystem",
+            "accelerators": [{"device": "cpu", "execution_providers": ["CPUExecutionProvider"]}],
+        }
+    },
+    "host": "local_system",
+    "evaluators": EVALUATE_TEMPLATE,
+    "evaluator": "common_evaluator",
```
xiaoyu-work, line 55: What does this `evaluator` do?

olive/cli/auto_opt.py
trajepl: Done.

olive/cli/auto_opt.py
trajepl: Done.

olive/cli/auto_opt.py
trajepl: Good call. But I saw there are different arguments attributes. Some of them are required, some of them are not. And they have different default values. 

Can we update it in follow-up PR?

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,265 @@
...
+            "--excluded_passes",
+            type=str,
+            nargs="*",
+            help=(
+                "List of passes to disable for optimization, if not specified, "
+                "auto-opt will disable ModelBuilder/OrtPerfTuning by default."
+            ),
+        )
+        auto_opt_config_group.add_argument(
+            "--use_model_builder",
```
samuel100, line 144: Can we abstract MB decision point for user? As a first-time user, this will add cognitive load ("what is model builder; should I use this"). Instead, could we test whether MB supports the model and if it does just use MB?

image

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,265 @@
...
+            choices=["CPU", "CUDA", "Tensorrt", "Dml", "VitisAI", "Qnn"],
+            help="List of execution providers to use for optimization",
+        )
+
+        # dataset options
+        dataset_group = sub_parser.add_argument_group(
+            "dataset options, required for some optimization passes like quantization, and evaluation components"
+        )
+        dataset_group.add_argument(
+            "--data_config_path",
```
samuel100, line 118: Could we make this a path to HF/Local to the actual data without requiring additional configuration? What does the data need to look like, how would the user generate it?

olive/cli/auto_opt.py
```diff
@@ -0,0 +1,265 @@
...
+}
+
+
+class AutoOptCommand(BaseOliveCLICommand):
+    allow_unknown_args: ClassVar[bool] = True
+
+    @staticmethod
+    def register_subcommand(parser: ArgumentParser):
+        sub_parser = parser.add_parser(
+            "auto-opt",
```
samuel100, line 74: Should we simplify the name of the command to `optimize`?

olive/cli/auto_opt.py
devang-ml: We need an API from MB to check whether the model is supported or not. We can do this as a follow up PR.

olive/cli/auto_opt.py
devang-ml: We need to do the same as 'finetune` command and avoid having user provide json/yaml config file.

jambayk (approved):

PR #1344, opened 2024-09-03
Add cli output validation & improve cli logging

xiaoyu-work:
## Describe your changes

- Add huggingface model url support for cli model name: `https://huggingface.co/microsoft/Phi-3.5-MoE-instruct`
- Add log level configuration. The default value is 3 to avoid verbose logging. 
- Add export model number check for `capture-onnx`, `finetune` and `perf-tuning` cli. If there is no output model, will recommend user to change log level to get detailed logs.
- Add extras for `capture-onnx`, `finetune` and `perf-tuning` cli.
- Remove `validate_evaluate_input_model` as this is duplicated as [this check](https://github.com/microsoft/Olive/blob/main/olive/engine/engine.py#L328-L332). We want all logs controlled by logging.py

### DIFF
--- a/olive/cli/base.py
+++ b/olive/cli/base.py
@@ -43,15 +43,31 @@ def get_model_name_or_path(model_name_or_path) -> Union[str, Dict[str, str]]:
     pattern = r"^(?P<registry_name>[^:]+):(?P<model_name>[^:]+):(?P<version>[^:]+)$"
     match = re.match(pattern, model_name_or_path)
 
-    if not match:
-        return model_name_or_path
+    if match:
+        return {
+            "type": "azureml_registry_model",
+            "registry_name": match.group("registry_name"),
+            "name": match.group("model_name"),
+            "version": match.group("version"),
+        }
 
-    return {
-        "type": "azureml_registry_model",
-        "registry_name": match.group("registry_name"),
-        "name": match.group("model_name"),
-        "version": match.group("version"),
-    }
+    pattern = r"https://huggingface\.co/([^/]+/[^/]+)(?:/.*)?"
+    match = re.search(pattern, model_name_or_path)
+
+    if match:
+        return match.group(1)
+
+    return model_name_or_path
+
+
+def add_logging_options(sub_parser):
+    log_group = sub_parser.add_argument_group("logging options")
+    log_group.add_argument(
+        "--log_level",
+        type=int,
+        default=3,
+        help="Logging level. Default is 3. level 0: DEBUG, 1: INFO, 2: WARNING, 3: ERROR, 4: CRITICAL",
+    )
 
 
 def add_remote_options(sub_parser):
@@ -146,3 +162,8 @@ def update_remote_option(config, args, cli_action, tempdir):
             "hf_token": bool(args.keyvault_name),
         }
         config["workflow_host"] = "aml_system"
+
+
+# TODO(team): Remove this function once the output structure is refactored
+def get_output_model_number(outputs: Dict) -> int:
+    return sum(len(f.nodes) for f in outputs.values())
--- a/olive/cli/capture_onnx.py
+++ b/olive/cli/capture_onnx.py
@@ -12,8 +12,10 @@ from typing import ClassVar, Dict
 from olive.cli.base import (
     BaseOliveCLICommand,
     add_hf_model_options,
+    add_logging_options,
     add_remote_options,
     get_model_name_or_path,
+    get_output_model_number,
     is_remote_run,
     update_remote_option,
 )
@@ -38,6 +40,8 @@ class CaptureOnnxGraphCommand(BaseOliveCLICommand):
             "capture-onnx-graph",
             help=("Capture ONNX graph using PyTorch Exporter or Model Builder from the Huggingface model."),
         )
+
+        add_logging_options(sub_parser)
 
         # model options
         add_hf_model_options(sub_parser)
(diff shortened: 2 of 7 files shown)


olive/cli/base.py
```diff
@@ -43,15 +43,31 @@ def get_model_name_or_path(model_name_or_path) -> Union[str, Dict[str, str]]:
...
-        return model_name_or_path
+    if match:
+        return {
+            "type": "azureml_registry_model",
+            "registry_name": match.group("registry_name"),
+            "name": match.group("model_name"),
+            "version": match.group("version"),
+        }
+
+    pattern = r"https://huggingface\.co/([^/]+/[^/]+)(?:/.*)?"
```
jambayk, line 54: is this one of the customer scenarios? I have usually only seen people provide the id directly. just wondering

olive/cli/capture_onnx.py
```diff
@@ -207,7 +218,7 @@ def get_run_config(self, tempdir: str) -> Dict:
 
 
 TEMPLATE = {
-    "input_model": {"type": "HfModel"},
+    "input_model": {"type": "HfModel", "load_kwargs": {"trust_remote_code": True}},
```
jambayk, line 221: I don't think we want to force this to always be true. There are various undesirable scenarios with this:
- user should opt in since the remote code might be malicious
- for models like phi3, it forces the model to load using local code even though the model type is already supported directly by transformers 

trust_remote_code is already an option for hf_model group https://github.com/microsoft/Olive/blob/625161942dedf2660b6b53714149a4019fd06ea2/olive/cli/base.py#L101 so it can be added to the model config like in https://github.com/microsoft/Olive/blob/625161942dedf2660b6b53714149a4019fd06ea2/olive/cli/finetune.py#L200

olive/olive_config.json
```diff
@@ -118,6 +118,10 @@
         "bnb": [ "bitsandbytes" ],
         "ort": [ "onnxruntime", "onnxruntime-directml", "onnxruntime-gpu", "onnxruntime-openvino", "numpy<2.0" ],
         "nvmo": [ "nvidia-modelopt~=0.11.0", "onnx-graphsurgeon" ],
-        "ort-genai": [ "onnxruntime-genai" ]
+        "ort-genai": [ "onnxruntime-genai" ],
+        "capture-onnx-graph": [ "onnxruntime-genai" ],
+        "cloud-cache": [ "azure-storage-blob" ],
+        "finetune": [ "onnxruntime-genai" ],
```
jambayk, line 124: can you add the other dependencies that come from `optimum` and `lora` also?

olive/cli/base.py
xiaoyu-work: Yes this is a case we need to support.

olive/cli/capture_onnx.py
xiaoyu-work: Oh i didn't realize it is one of the model options. I'll update this.

olive/olive_config.json
xiaoyu-work: Done

olive/cli/capture_onnx.py
```diff
@@ -179,13 +179,15 @@ def run(self):
...
             config["input_model"]["task"] = self.args.task
 
-        config["log_severity_level"] = self.args.log_level
-
-        to_replace = [("output_dir", tempdir)]
+        to_replace = [
+            ("output_dir", tempdir),
+            ("log_severity_level", self.args.log_level),
+            (("input_model", "model_path"), get_model_name_or_path(self.args.model_name_or_path)),
+            (("input_model", "load_kwargs", "trust_remote_code"), self.args.trust_remote_code),
```
jambayk, line 189: this needs an empty `load_kwargs` dict in the template or we could set `load_kwargs` as a dict here.

olive/cli/base.py
```diff
@@ -115,10 +131,10 @@ def update_remote_option(config, args, cli_action, tempdir):
 
         try:
             subscription_id = json.loads(subprocess.check_output("az account show", shell=True).decode("utf-8"))["id"]
-            logger.info("Using Azure subscription ID: %s", subscription_id)
+            print("Using Azure subscription ID: %s", subscription_id)
```
jambayk, line 134: thanks! this is a good idea to print instead of log.

olive/workflows/run/config.py
```diff
@@ -246,13 +243,6 @@ def validate_engine(cls, v, values):
         v = _resolve_system(v, values, "target")
         return _resolve_evaluator(v, values)
 
-    @validator("engine")
```
jambayk, line 249: is this removed since the engine does the check anyways and the info log is too verbose?

xiaoyu-work: Yeah 2 reasons:
- Engine does the check.
- This log level is not controlled by logging.py as this runs before we are initializing logging.py

olive/cli/capture_onnx.py
```diff
@@ -207,7 +226,7 @@ def get_run_config(self, tempdir: str) -> Dict:
 
 
 TEMPLATE = {
```
jambayk, line 228: not directly related to this PR since the template was already present. But wouldn't precision "fp16" with cpu ep fail for the model builder?

xiaoyu-work: Yeah fp16 doesn't work with cpu. There will be an error message saying this. But I updated the cli to support model builder gpu as well.

jambayk: the `device` argument is now being used for two different things. for conversion, it is the device where the model is placed on while exporting. for model builder, it is the target device to use the model on.

jambayk: Could also confirm, if just providing the device is fine in the accelerators? I think it might loop over both cpuep and cudaep for gpu device.

xiaoyu-work: 1. Yeah this argument can be different meanings for different pass, as long as we don't run these 2 passes together. 
2. Yes it will for loop all available eps, which confuses me. I thought if this is a no search and no evaluator run, only one available ep should be picked up to run the workflow, cause Olive doesn't need metrics to compare anything. Is this expected?

jambayk: Yeah, Mike added the logic to automatically infer the ep from device. That was when we wanted to do multi-ep workflows: user provided target device, and we produce all valid ep models.

If we just provide a single ep, the device can be auto inferred.

jambayk (approved):

PR #1345, opened 2024-09-05
ModelBuilder: Use `TRANSFORMERS_CACHE` as `cache_dir`

jambayk:
## Describe your changes
We currently try to infer the transformers cache from environment variables but those are not set most of the time. `transformers.utils` has a variable which handles this logic internally so we can just use that instead. 

Also added some exception handling to remove the intermediate files that the model builder adds to the cache dir. Ideally, the transformers cache and intermediate cache should be dealt separately in order to not pollute the transformers cache.

### DIFF
--- a/olive/passes/onnx/model_builder.py
+++ b/olive/passes/onnx/model_builder.py
@@ -7,10 +7,10 @@
 import copy
 import json
 import logging
-import os
-import tempfile
 from pathlib import Path
 from typing import Any, Dict, Union
+
+import transformers
 
 from olive.common.utils import IntEnumBase, StrEnumBase
 from olive.hardware.accelerator import AcceleratorSpec, Device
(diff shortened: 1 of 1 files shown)


olive/passes/onnx/model_builder.py
```diff
@@ -184,15 +174,26 @@ def _run_for_config(
...
+                cache_dir=transformers.utils.TRANSFORMERS_CACHE,
+                **extra_args,
+            )
+        except Exception:
+            # if model building fails, clean up the intermediate files in the cache_dir
+            cache_dir = Path(transformers.utils.TRANSFORMERS_CACHE)
+            if cache_dir.is_dir():
+                for file in cache_dir.iterdir():
+                    if file.suffix == ".bin":
+                        file.unlink()
```
shaahji, line 195: Do we really want to do this? Transformer cache may contain data from other models as well. Blindly deleting everything doesn't seem like a good idea.

shaahji (changes requested):

olive/passes/onnx/model_builder.py
jambayk: Not really needed but it cleans up the cache directory if the model builder fails for some reason. I checked the cache structure and hf cache creates subdirectories inside it for models, datasets, etc. We are only iterating over the top level where the .bin files are generated by model builder. It is the same logic the model builder does when it succeeds https://github.com/microsoft/onnxruntime-genai/blob/c7eba3c63a454edd6662eb007ff397d1146cc081/src/python/py/models/builder.py#L378

jambayk: for reference
image

shaahji (approved):

PR #1356, opened 2024-09-13
Quantize: CLI command to quantize input model

shaahji:
## Quantize: CLI command to quantize input model

Usage:
```bash
  olive quantize                      \
    --model_name_or_path <model-name> \
    --trust_remote_code               \
    --device <cpu|gpu|npu>            \
    --algorithms <awq,gptq>           \
    --data_name <data-name>           \
    --subset <subset-name>            \
    --split <split-name>              \
    --batch_size <batch-size>         \
    --output_path <output-dir>
```

### DIFF
--- a/olive/cli/base.py
+++ b/olive/cli/base.py
@@ -14,7 +14,7 @@ import yaml
 
 from olive.cli.constants import CONDA_CONFIG
 from olive.common.user_module_loader import UserModuleLoader
-from olive.common.utils import hardlink_copy_dir, hash_dict, set_nested_dict_value
+from olive.common.utils import hardlink_copy_dir, hash_dict, set_nested_dict_value, unescaped_str
 from olive.resource_path import OLIVE_RESOURCE_ANNOTATIONS, find_all_resources
 
 
@@ -229,6 +229,10 @@ def get_input_model_config(args: Namespace) -> Dict:
     return _get_hf_input_model(args, model_name_or_path)
 
 
+def update_input_model_options(args, config):
+    config["input_model"] = get_input_model_config(args)
+
+
 def add_logging_options(sub_parser: ArgumentParser):
     """Add logging options to the sub_parser."""
     log_group = sub_parser.add_argument_group("logging options")
@@ -238,6 +242,7 @@ def add_logging_options(sub_parser: ArgumentParser):
         default=3,
         help="Logging level. Default is 3. level 0: DEBUG, 1: INFO, 2: WARNING, 3: ERROR, 4: CRITICAL",
     )
+    return log_group
 
 
 def add_remote_options(sub_parser: ArgumentParser):
@@ -272,8 +277,10 @@ def add_remote_options(sub_parser: ArgumentParser):
         help="The compute name to run the workflow on.",
     )
 
+    return remote_group
 
-def add_model_options(
+
+def add_input_model_options(
     sub_parser: ArgumentParser,
     enable_hf: bool = False,
     enable_hf_adapter: bool = False,
@@ -318,6 +325,7 @@ def add_model_options(
             "--trust_remote_code", action="store_true", help="Trust remote code when loading a model."
         )
         model_group.add_argument("-t", "--task", type=str, help="Task for which the model is used.")
+
     if enable_hf_adapter:
         assert enable_hf, "enable_hf must be True when enable_hf_adapter is True."
         model_group.add_argument(
@@ -345,6 +353,7 @@ def add_model_options(
         default=default_output_path,
         help="Path to save the command output.",
     )
+    return model_group
 
 
 def output_path_type(path: str) -> str:
@@ -363,7 +372,7 @@ def is_remote_run(args: Namespace) -> bool:
     return all([args.resource_group, args.workspace_name, args.aml_compute])
 
 
-def update_remote_option(config: Dict, args: Namespace, cli_action: str, tempdir: Union[str, Path]):
+def update_remote_options(config: Dict, args: Namespace, cli_action: str, tempdir: Union[str, Path]):
     """Update the config for remote run."""
     if args.resource_group or args.workspace_name or args.aml_compute:
         if not is_remote_run(args):
@@ -373,7 +382,7 @@ def update_remote_option(config: Dict, args: Namespace, cli_action: str, tempdir
 
         try:
             subscription_id = json.loads(subprocess.check_output("az account show", shell=True).decode("utf-8"))["id"]
-            print("Using Azure subscription ID: %s", subscription_id)
+            print(f"Using Azure subscription ID: {subscription_id}")
 
         except subprocess.CalledProcessError:
             print(
@@ -444,3 +453,220 @@ def save_output_model(config: Dict, output_model_dir: Union[str, Path]):
             json.dump(model_config, f, indent=4)
 
     print(f"Command succeeded. Output model saved to {output_model_dir}")
+
+
+def add_dataset_options(sub_parser, required=True, include_train=True, include_eval=True):
+    dataset_group = sub_parser.add_argument_group("dataset options")
+    dataset_group.add_argument(
+        "-d",
+        "--data_name",
+        type=str,
+        required=required,
+        help="The dataset name.",
+    )
+
+    if include_train:
+        dataset_group.add_argument("--train_subset", type=str, help="The subset to use for training.")
+        dataset_group.add_argument("--train_split", type=str, default="train", help="The split to use for training.")
+
+    if include_eval:
+        dataset_group.add_argument("--eval_subset", type=str, help="The subset to use for evaluation.")
+        dataset_group.add_argument("--eval_split", default="", help="The dataset split to evaluate on.")
+
+    if not (include_train and include_eval):
+        dataset_group.add_argument("--subset", type=str, help="The subset of the dataset to use.")
+        dataset_group.add_argument("--split", type=str, help="The dataset split to use.")
+
+    # TODO(jambayk): currently only supports single file or list of files, support mapping
+    dataset_group.add_argument(
+        "--data_files", type=str, help="The dataset files. If multiple files, separate by comma."
+    )
+
+    text_group = dataset_group.add_mutually_exclusive_group(required=False)
+    text_group.add_argument(
+        "--text_field",
+        type=str,
+        help="The text field to use for fine-tuning.",
+    )
+    text_group.add_argument(
+        "--text_template",
+        # using special string type to allow for escaped characters like \n
+        type=unescaped_str,
+        help=r"Template to generate text field from. E.g. '### Question: {prompt} \n### Answer: {response}'",
+    )
+    dataset_group.add_argument(
+        "--max_seq_len",
+        type=int,
+        default=1024,
+        help="Maximum sequence length for the data.",
+    )
+    dataset_group.add_argument(
+        "--add_special_tokens",
+        type=bool,
+        default=False,
+        help="Whether to add special tokens during preprocessing.",
+    )
+    dataset_group.add_argument(
+        "--max_samples",
+        type=int,
+        default=256,
+        help="Maximum samples to select from the dataset.",
+    )
+    dataset_group.add_argument(
+        "--batch_size",
+        type=int,
+        default=1,
+        help="Batch size.",
+    )
+
+    return dataset_group, text_group
+
+
+def update_dataset_options(args, config):
+    load_key = ("data_configs", 0, "load_dataset_config")
+    preprocess_key = ("data_configs", 0, "pre_process_data_config")
+    dataloader_key = ("data_configs", 0, "dataloader_config")
+    split = args.train_split if hasattr(args, "train_split") else args.split
+    subset = args.train_subset if hasattr(args, "train_subset") else args.subset
+
+    to_replace = [
+        ((*load_key, "data_name"), args.data_name),
+        ((*load_key, "split"), split),
+        ((*load_key, "subset"), subset),
+        (
+            (*load_key, "data_files"),
+            args.data_files.split(",") if args.data_files else None,
+        ),
+        ((*preprocess_key, "text_cols"), args.text_field),
+        ((*preprocess_key, "text_template"), args.text_template),
+        ((*preprocess_key, "max_seq_len"), args.max_seq_len),
+        ((*preprocess_key, "add_special_tokens"), args.add_special_tokens),
+        ((*preprocess_key, "max_samples"), args.max_samples),
+        ((*dataloader_key, "batch_size"), args.batch_size),
+    ]
+    for keys, value in to_replace:
+        if value is not None:
+            set_nested_dict_value(config, keys, value)
+
+
+def add_hf_dataset_options(sub_parser):
+    hf_dataset_group = sub_parser.add_argument_group(
+        "huggingface dataset options, if dataset options are not provided, "
+        "user should provide the following options to modify the default data config. "
+        "Please refer to olive.data.container.TransformersTokenDummyDataContainer for more details."
+    )
+    hf_dataset_group.add_argument(
+        "--hf_model_name",
+        help="Huggingface model name used to load model configs from huggingface.",
+    )
+    hf_dataset_group.add_argument(
+        "--batch_size",
+        type=int,
+        help="Batch size of the input data.",
+    )
+    hf_dataset_group.add_argument(
+        "--seq_len",
+        type=int,
+        help="Sequence length to use for the input data.",
+    )
+    hf_dataset_group.add_argument(
+        "--past_seq_len",
+        type=int,
+        help="Past sequence length to use for the input data.",
+    )
+    hf_dataset_group.add_argument(
+        "--max_seq_len",
+        type=int,
+        help="Max sequence length to use for the input data.",
+    )
+    hf_dataset_group.add_argument(
+        "--shared_kv",
+        action="store_true",
+        help="Whether to enable share kv cache in the input data.",
+    )
+    hf_dataset_group.add_argument(
+        "--generative",
+        action="store_true",
+        help="Whether to enable generative mode in the input data.",
+    )
+    hf_dataset_group.add_argument(
+        "--ort_past_key_name",
+        type=str,
+        help="Past key name for the input data.",
+    )
+    hf_dataset_group.add_argument(
+        "--ort_past_value_name",
+        type=str,
+        help="Past value name for the input data.",
+    )
+    # TODO(all): Argument conflicting with use of the same name in model options
+    # hf_dataset_group.add_argument(
+    #     "--trust_remote_code",
+    #     action="store_true",
+    #     help="Whether to trust remote code in the input data.",
+    # )
+    hf_dataset_group.add_argument(
+        "--max_samples",
+        type=int,
+        help="Max samples to use for the input data.",
+    )
+    hf_dataset_group.add_argument(
+        "--fields_no_batch",
+        nargs="*",
+        help="List of fields that should not be batched.",
+    )
+
+    return hf_dataset_group
+
+
+def add_accelerator_options(sub_parser):
+    accelerator_group = sub_parser.add_argument_group("accelerator group")
+
+    accelerator_group.add_argument(
+        "--device",
+        type=str,
+        default="cpu",
+        choices=["gpu", "cpu", "npu"],
+        help="Device to use for the model.",
+    )
+
+    accelerator_group.add_argument(
+        "--providers_list",
+        type=str,
+        nargs="*",
+        choices=[
+            "CUDAExecutionProvider",
+            "CPUExecutionProvider",
+            "DmlExecutionProvider",
+            "JsExecutionProvider",
+            "MIGraphXExecutionProvider",
+            "OpenVINOExecutionProvider",
+            "OpenVINOExecutionProvider",
+            "QNNExecutionProviderROCMExecutionProvider",
+            "TensorrtExecutionProvider",
+        ],
+        help=(
+            "List of execution providers to use for ONNX model. They are case sensitive. "
+            "If not provided, all available providers will be used."
+        ),
+    )
+
+    return accelerator_group
+
+
+def update_accelerator_options(args, config):
+    to_replace = [
+        (("systems", "local_system", "accelerators", 0, "device"), args.device),
+    ]
+
+    args.providers_list = args.providers_list or []
+    for idx, provider in enumerate(args.providers_list):
+        if not provider.endswith("ExecutionProvider"):
+            args.providers_list[idx] = f"{provider}ExecutionProvider"
+
+    if args.providers_list:
+        to_replace.append((("systems", "local_system", "accelerators", 0, "execution_providers"), args.providers_list))
+
+    for k, v in to_replace:
+        if v is not None:
+            set_nested_dict_value(config, k, v)
--- a/olive/cli/capture_onnx.py
+++ b/olive/cli/capture_onnx.py
@@ -9,13 +9,13 @@ from typing import ClassVar, Dict
 
 from olive.cli.base import (
     BaseOliveCLICommand,
+    add_input_model_options,
     add_logging_options,
-    add_model_options,
     add_remote_options,
     get_input_model_config,
     is_remote_run,
     save_output_model,
-    update_remote_option,
+    update_remote_options,
 )
 from olive.common.utils import IntEnumBase, set_nested_dict_value
 
@@ -40,7 +40,7 @@ class CaptureOnnxGraphCommand(BaseOliveCLICommand):
         add_logging_options(sub_parser)
 
         # model options
-        add_model_options(
+        add_input_model_options(
             sub_parser, enable_hf=True, enable_hf_adapter=True, enable_pt=True, default_output_path="onnx-model"
         )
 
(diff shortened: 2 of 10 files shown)


olive/cli/base.py
```diff
@@ -38,132 +38,241 @@ def register_subcommand(parser: ArgumentParser):
     def run(self):
         raise NotImplementedError
 
+    @staticmethod
+    def _get_model_name_or_path(model_name_or_path) -> Union[str, Dict[str, str]]:
```
xiaoyu-work, line 42: This will be called in the child class, so it's not truly private to this class. Can we remove the underscore? Same for below.

xiaoyu-work:
Can you also add a unit test for this?

olive/cli/quantize.py
```diff
@@ -0,0 +1,132 @@
...
+            }
+        },
+        "data_configs": [],
+        "passes": {
+            "awq": {"type": "AutoAWQQuantizer"},
+            "gptq": {
+                # Ref: https://github.com/AutoGPTQ/AutoGPTQ/pull/651/files
+                "type": "GptqQuantizer",
+                "data_config": None,
+                "outside_layer_modules": ["model.embed_tokens", "embed_dropout", "model.norm"],
```
jambayk, line 35: Maybe we should not hardcode these values which are for a specific hf model type? We could instead use the mappings in https://github.com/microsoft/Olive/blob/main/olive/common/hf/mappings.py to add support for phi3 in the pass.

olive/cli/base.py
```diff
@@ -38,132 +38,241 @@ def register_subcommand(parser: ArgumentParser):
...
-        config["workflow_host"] = "aml_system"
-
-
-# TODO(team): Remove this function once the output structure is refactored
-def get_output_model_number(outputs: Dict) -> int:
-    return sum(len(f.nodes) for f in outputs.values())
+    @staticmethod
+    def _add_dataconfig_options(sub_parser):
+        dataset_group = sub_parser.add_argument_group(
+            "data config options, which mutually exclusive with huggingface dataset options"
```
devang-ml, line 125: User should not need to provide a config json/yaml on the CLI. See how `finetune` CLI handles data

olive/cli/base.py
```diff
@@ -38,132 +38,241 @@ def register_subcommand(parser: ArgumentParser):
...
+    @staticmethod
+    def _add_hf_dataset_options(sub_parser):
+        hf_dataset_group = sub_parser.add_argument_group(
+            "huggingface dataset options, if dataset options are not provided, "
+            "user should provide the following options to modify the default data config. "
+            "Please refer to olive.data.container.TransformersTokenDummyDataContainer for more details."
+        )
+        hf_dataset_group.add_argument(
+            "--predict_with_kv_cache",
+            action="store_true",
```
devang-ml, line 142: If this option is only for `tune-session=params` CLI then we don't need it in base.py

olive/cli/quantize.py
```diff
@@ -0,0 +1,132 @@
...
+        config = deepcopy(QuantizeCommand._CONFIG_TEMPLATE)
+        for k, v in to_replace:
+            if v is not None:
+                set_nested_dict_value(config, k, v)
+
+        return config
+
+    def run(self):
+        run_config = self._get_run_config()
+        olive_run(run_config)
```
devang-ml, line 132: Please handle the output like other CLIs and report appropriate msg to the user.

olive/cli/base.py
shaahji: As per Python PEP [here](https://peps.python.org/pep-0008/#method-names-and-instance-variables)

_Use one leading underscore only for non-public methods and instance variables._

These functions are _protected_ i.e. can be used in sub-classes/derived classes but not be invoked outside.
Unfortunately, there is no recommendation to differentiate between protected and private member functions. However, some recommend using single underscore and double underscores.

samuel100:
Some feedback....


- whack-a-mole package install. I had to install auto-awq, which was not easy as the package name is different to the module (i.e. `awq module not found` error leads a user to try `pip install awq` which is not correct.)
- It is not really clear what is expected in the data_config YAML/JSON. In general data_configs are hard - what data does a user need to use, how would different datasets impact the results? what do they need to put into data_config? If the dataset impacts results we really need amazing documentation and guidance on what a user needs to provide (including how to generate [synthetic] data). If the data does not really impact the results can we get rid of the option and provide either dummy data or a static option (e.g. wikitext)?
- The help file should give some more information on the different algorithms. For example, it would be good to know that AWQ will output a 4bit model.
- The `--providers_list` assumes the user knows about ORT e.g. `CPUExecutionProvider`. The help information should enumerate the different options for the user. 
- How would a user evaluate the results for speed up/memory utilization/quality. Taking a step back, the motivation for quantization is to lower footprint and speed up execution without sacrificing efficacy for the task. The CLI command allows a user to try different algorithms (good) but it needs some evaluation information so that the user can make a decision on the "best" method.

A E2E here would be to run Quantization -> Capture the ONNX Graph for using in ORT. I therefore tried to take the output of quantization and run through `capture-onnx-graph`. Below is what happened:

Firstly, I tried to use Dynamo exporter option:

```bash
olive capture-onnx-graph \
    --model_name_or_path models/qwen-awq/awq/cpu-cpu_model/model \
    --use_dynamo_exporter True \
    --use_ort_genai True \
    --output_path models/qwen-awq/captured \
    --device cpu \
    --log_level 1
```

This hit the following error:

```
[2024-09-17 09:35:36,294] [INFO] [engine.py:874:_run_pass] Running pass c:OnnxConversion
You have loaded an AWQ model on CPU and have a CUDA device available, make sure to set your model on a GPU device in order to run your model.
`low_cpu_mem_usage` was None, now set to True since model is quantized.
We detected that you are passing `past_key_values` as a tuple and this is deprecated and will be removed in v4.43. Please use an appropriate `Cache` class (https://huggingface.co/docs/transformers/v4.41.3/en/internal/generation_utils#transformers.Cache)
[2024-09-17 09:35:40,494] [ERROR] [engine.py:972:_run_pass] Pass run failed.
Traceback (most recent call last):
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/olive/engine/engine.py", line 960, in _run_pass
    output_model_config = host.run_pass(p, input_model_config, output_model_path, pass_search_point)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/olive/systems/local.py", line 30, in run_pass
    output_model = the_pass.run(model, output_model_path, point)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/olive/passes/olive_pass.py", line 206, in run
    output_model = self._run_for_config(model, config, output_model_path)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/olive/passes/onnx/conversion.py", line 116, in _run_for_config
    output_model = self._run_for_config_internal(model, config, output_model_path)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/olive/passes/onnx/conversion.py", line 149, in _run_for_config_internal
    return self._convert_model_on_device(model, config, output_model_path, device, torch_dtype)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/olive/passes/onnx/conversion.py", line 367, in _convert_model_on_device
    converted_onnx_model = OnnxConversion._export_pytorch_model(
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/olive/passes/onnx/conversion.py", line 205, in _export_pytorch_model
    pytorch_model(*dummy_inputs)
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/transformers/models/qwen2/modeling_qwen2.py", line 1104, in forward
    outputs = self.model(
              ^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/transformers/models/qwen2/modeling_qwen2.py", line 915, in forward
    layer_outputs = decoder_layer(
                    ^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/transformers/models/qwen2/modeling_qwen2.py", line 655, in forward
    hidden_states, self_attn_weights, present_key_value = self.self_attn(
                                                          ^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/transformers/models/qwen2/modeling_qwen2.py", line 542, in forward
    query_states = self.q_proj(hidden_states)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1532, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/nn/modules/module.py", line 1541, in _call_impl
    return forward_call(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/awq/modules/linear/gemm.py", line 243, in forward
    out = WQLinearMMFunction.apply(
          ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/torch/autograd/function.py", line 598, in apply
    return super().apply(*args, **kwargs)  # type: ignore[misc]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda/envs/quant-cli/lib/python3.11/site-packages/awq/modules/linear/gemm.py", line 47, in forward
    out = awq_ext.gemm_forward_cuda(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: t == DeviceType::CUDA INTERNAL ASSERT FAILED at "/usr/share/miniconda3/envs/build/lib/python3.11/site-packages/torch/include/c10/cuda/impl/CUDAGuardImpl.h":28, please report a bug to PyTorch. 
```

So, next I tried Model Builder option. This worked... BUT it is not clear what really happened - I had to set the `--precision int4` option... does that re-quantize the model using RTN? The model output from MB was 1.3GB, which compared to 0.6MB for the safetensor file after the AWQ quantization.

shaahji:
> whack-a-mole package install. I had to install auto-awq, which was not easy as the package name is different to the module (i.e. awq module not found error leads a user to try pip install awq which is not correct.)

Ironically, I ran into the same problem initially. I had some thoughts about addressing this at a larger level. Olive already knows the dependencies for each Pass. They are defined in olive_config.json. We *could* potentially iterate and verify that those dependencies are present even before we start running the passes. That avoids long wait before a workflow fails after running a few expensive passes. Thoughts?

> It is not really clear what is expected in the data_config YAML/JSON.

The data_config requirement is removed in a follow up commit. The arguments to the command line are inline with the finetune commend i.e. data_name, train_subset, eval_subset, etc.

> The help file should give some more information on the different algorithms. For example, it would be good to know that AWQ will output a 4bit model.

There can never be enough information in help. :) One thing could be argued to be more important the other. I propose to provide a link to specific algorithm's documentation. 

> The --providers_list assumes the user knows about ORT e.g. CPUExecutionProvider. The help information should enumerate the different options for the user.

I will add the available options in the choices list.

> How would a user evaluate the results for speed up/memory utilization/quality. Taking a step back, the motivation for quantization is to lower footprint and speed up execution without sacrificing efficacy for the task. The CLI command allows a user to try different algorithms (good) but it needs some evaluation information so that the user can make a decision on the "best" method.

As I understand the intent, CLI commands are meant to be "do one job only". For evaluation, we might introduce a separate cli command that user can chain along with this.

test/unit_test/cli/test_base.py
```diff
@@ -269,7 +279,9 @@
     expected_config = {"type": "PyTorchModel", "model_path": "model_path"}
 
     # execute
-    config = get_input_model_config(args)
+    config = {}
+    MockCommand(None, args)._update_input_model_options(config)
```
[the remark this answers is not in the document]
shaahji: Disabled!

test/unit_test/cli/test_base.py
```diff
@@ -253,7 +263,7 @@
 
     # execute and assert
     with pytest.raises(ValueError, match="invalid-name is not a valid Huggingface model name."):
-        get_input_model_config(args)
+        MockCommand(None, args)._update_input_model_options({})
```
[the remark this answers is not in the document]
shaahji: Disabled!

test/unit_test/cli/test_base.py
```diff
@@ -238,7 +248,7 @@
 
     # execute and assert
     with pytest.raises(ValueError, match="Either _model_loader or model_name_or_path is required for PyTorch model."):
-        get_input_model_config(args)
+        MockCommand(None, args)._update_input_model_options({})
```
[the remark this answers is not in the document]
shaahji: Disabled!

test/unit_test/cli/test_base.py
```diff
@@ -218,10 +227,11 @@
         mock_instance.has_function.side_effect = has_function_side_effect
 
         # execute
-        config = get_input_model_config(args)
+        MockCommand(None, args)._update_input_model_options(config)
```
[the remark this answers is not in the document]
shaahji: Disabled!

shaahji:
All comments/inputs addressed.

olive/cli/base.py
devang-ml: We don't need to make them @staticmethod.

olive/cli/capture_onnx.py
```diff
@@ -43,10 +33,10 @@ def register_subcommand(parser: ArgumentParser):
             help=("Capture ONNX graph using PyTorch Exporter or Model Builder from the Huggingface model."),
         )
 
-        add_logging_options(sub_parser)
+        CaptureOnnxGraphCommand._add_logging_options(sub_parser)
```
devang-ml, line 37: See above comment.

olive/cli/base.py
shaahji: So, make them member functions or global? `self.xxxxx()` vs. `xxxx()`

Making them members doesn't make sense because there's no use of `self` in the implementation.
The issue with making them global is each of them have to be imported for use.

shaahji: Class methods just adds another unused argument, just like member functions (with self) do. I don't see any reason to change them to member functions or class functions, except for aesthetic reasons.

Logically, them being static is correct because they don't have any use of cls or self. Global just makes the importa look complicated.

Either way, @devang-ml recommended reverting all the _beautification_/_code organization_ changes and focusing the PR only on the quantize command. So, I will do that and report the PR.

jambayk: yes, I think we can revisit this later if needed. My suggestion was to make the register_subcommand into a class method so that calls like `CaptureOnnxGraphCommand._add_logging_options` could be replaced with `cls._add_logging_options`. for the rest staticmethod is cleaner. 

But it's mostly aesthetic as you said since we don't enforce a particular style for these cases.

olive/cli/launcher.py
```diff
@@ -37,6 +38,7 @@ def get_cli_parser(called_as_console_script: bool = True) -> ArgumentParser:
     PerfTuningCommand.register_subcommand(commands_parser)
     ConfigureQualcommSDKCommand.register_subcommand(commands_parser)
     ManageAMLComputeCommand.register_subcommand(commands_parser)
+    QuantizeCommand.register_subcommand(commands_parser)
```
jambayk, line 41: maybe you could move this somewhere higher (35?) manage aml and cloud cache don't run workflows so this order feels a bit weird in the documentation.

olive/cli/quantize.py
```diff
@@ -0,0 +1,148 @@
...
+
+        has_failures = False
+        with tempfile.TemporaryDirectory(prefix="olive-cli-tmp-", dir=self.args.output_path) as tempdir:
+            run_config = self._get_run_config(tempdir)
+            olive_run(run_config)
+
+            if is_remote_run(self.args):
+                return
+
+            output_path = Path(self.args.output_path)
```
jambayk, line 134: https://github.com/microsoft/Olive/blob/180dffb8af7643d4dfcffcb044a14ba171ab2647/olive/cli/base.py#L410 handles this automatically. If there is only one pass_flow, the `algorithm_name` nesting doesn't exist.

olive/cli/quantize.py
```diff
@@ -0,0 +1,148 @@
...
+            "--algorithms",
+            type=str,
+            nargs="*",
+            required=True,
+            choices=sorted(QuantizeCommand._CONFIG_TEMPLATE["passes"].keys()),
+            help="List of quantization algorithms to run.",
+        )
+
+        # dataset options
+        add_dataset_options(sub_parser)
```
jambayk, line 96: the train/eval names in this doesn't make sense in this scenario. Maybe the option to just have just one `split`,`subset` or one each for train/eval can be a parameter for this function like in `add_input_model_options`

test/unit_test/cli/test_cli.py
```diff
@@ -251,5 +251,49 @@ def test_cloud_cache_command(_, mock_container_client, test_set):
...
+    tmpdir.mkdir()
+
+    output_dir = tmp_path / "output_dir"
+
+    # setup
+    mock_repo_exists.return_value = True
+    mock_tempdir.return_value = tmpdir.resolve()
+    mock_run.return_value = {}
+
+    for algo_name in algorithm_names:
```
jambayk, line 270: related to the above comment. this nesting doesn't exist if there is only one pass_flow https://github.com/microsoft/Olive/blob/180dffb8af7643d4dfcffcb044a14ba171ab2647/olive/engine/engine.py#L250

shaahji: The generated config never has more than one pass_flow and the output directories are created as expected. It is verified as part of the unit test as well.

olive/cli/quantize.py
shaahji: The pass_flows never have more than one entry. The nesting does exist in either case. It is verified as part of unit test as well - parameterized with one pass and two pass_flows.

test/unit_test/cli/test_cli.py
shaahji: Nevermind, I think you mean in the output folder of the cli, not in the output folder of the olive_run.

olive/cli/quantize.py
shaahji: Nevermind, I think you mean in the output folder of the cli, not in the output folder of the olive_run.

olive/cli/quantize.py
```diff
@@ -0,0 +1,133 @@
...
+            required=True,
+            choices=sorted(QuantizeCommand._CONFIG_TEMPLATE["passes"].keys()),
+            help="List of quantization algorithms to run.",
+        )
+
+        # dataset options
+        add_dataset_options(sub_parser, include_train=False, include_eval=False)
+
+        # accelerator options
+        add_accelerator_options(sub_parser)
```
jambayk, line 99: since only pytorch models are currently supported, I don't think this is needed? The ep+device doesn't matter for pytorch only passes. they are just placeholders.

jambayk: even device is not necessary. if the pass needs to run on gpu, it will run on gpu. 

might not be an issue for cpu device but if you choose gpu, with ep not provided, the behavior is that it will loop over all gpu related eps. not necessary here

olive/cli/quantize.py
```diff
@@ -0,0 +1,133 @@
...
+                "name": "default_data_config",
+                "type": "HuggingfaceContainer",
+                "load_dataset_config": {},
+                "pre_process_data_config": {},
+                "dataloader_config": {},
+                "post_process_data_config": {},
+            }
+        ],
+        "passes": {
+            "awq": {"type": "AutoAWQQuantizer"},
```
jambayk, line 50: I tried running with only awq which should not need a data config. but I get this error 
image

shaahji: So, one algorithm needs dataset and the other doesn't. I don't think there's an easy way to configure `ArgumentParser` which this kind of requirement. This might need some `ArgumentParser` trickery to make it work. Somehow, the `required` option of dataset will need to be toggled based on the user input.

olive/cli/quantize.py
jambayk: oh, the looping doesn't happen since the template has one ep already in the system. but regardless, I think the accelerator option is not needed for now. it is confusing when they don't do anything.

olive/cli/quantize.py
jambayk: thats not possible since the arg parse is defined before the user input. you could just make it optional but print a failure while creating the run config. That's what i did for some options that were optional based on user input in the generate-adapter command https://github.com/microsoft/Olive/blob/d3070aa5da2bf77e30bee3df0ce6659f79c192e3/olive/cli/generate_adapter.py#L96

olive/cli/quantize.py
shaahji: Wouldn't that equate to removing _systems_, _host_ and _target_ settings from config?? If I include those options then I have to choose either cpu or gpu explicitly, which might be right for one algorithm but not the other.

olive/cli/quantize.py
shaahji: Isn't that misleading. Documentation says one thing but when used as per the documentation, it reports a failure.

jambayk: you could provide your own help text if you want. I don't think there is an issue making some options conditionally required.

olive/cli/quantize.py
jambayk: since the only tools that it currently supports are pytorch quantization tools, I think it's okay to just keep the current default system as is in the config. they are just there as placeholders. 
- the target accelerator spec doesn't matter for pytorch passes but we need some placeholder to ensure the engine doesn't do an unnecessary loop of all possible accelerators.
- accelerator spec for host device also doesn't really make sense. If it needs to run on gpu, it should run on gpu. I think Mike added it, but its usage is arbitrary so far.
Most of these are artifacts of our originally ort focused design decisions.

jambayk: you can keep the options if you want. just think it's unnecessary and it currently works fine because the config already has a default ep that prevents the looping behavior.

jambayk (approved):

PR #1358, opened 2024-09-16
Leverage the new `dynamo=True` API for PyTorch model export

justinchuby:
## Describe your changes

This PR adds the `dynamo=True` usage to leverage the improved dynamo exporter available since PyTorch 2.5. Since PyTorch 2.5 is not yet released, it also supports development/nightly builds of PyTorch.

### DIFF
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -201,30 +201,31 @@ class OnnxConversion(Pass):
                     f"torch.onnx.dynamo_export is not available for torch version {torch_version}. "
                     "Please upgrade your torch version to 2.5.0 or above."
                 )
-            elif legacy_dynamo_supported_version <= torch_version < dynamo_supported_version:
-                from torch._dynamo import config as dynamo_config
+            from torch._dynamo import config as dynamo_config
 
-                dynamo_config.capture_scalar_outputs = True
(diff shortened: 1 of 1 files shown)


olive/passes/onnx/conversion.py
```diff
@@ -190,33 +190,50 @@ def _export_pytorch_model(
...
+            if torch_version < legacy_dynamo_supported_version:
                 raise ImportError(
                     f"torch.onnx.dynamo_export is not available for torch version {torch_version}. "
-                    "Please upgrade your torch version to 2.2.0 or above."
+                    "Please upgrade your torch version to 2.5.0 or above."
                 )
-            from torch._dynamo import config as dynamo_config
-            from torch.onnx import dynamo_export
+            elif legacy_dynamo_supported_version <= torch_version < dynamo_supported_version:
+                from torch._dynamo import config as dynamo_config
```
jambayk, line 205: since the if condition raises an error, I think we can make this an if condition and move the common lines 205-208 from the elif/else blocks.
Also line 218 seems unnecessary since that is the condition for the else block?

justinchuby: Done. PTAL

jambayk (approved):
Thanks!

PR #1361, opened 2024-09-18
Engine: Improve output structure, CLI: Configurable model options, Separate `finetune`+`generate-adapter`

jambayk:
## Describe your changes
**Engine**:
- Improve the output folder structure of a workflow run. The current structure was meant for multiple-ep, multiple-passflow worfklows but that is not the common usage for olive.
- Unnecessary nesting for accelerator spec and pass flows is removed for single ep, single passflow scenario.
- `output_name` is removed from both pass config and engine config.
  - The behavior of `output_name` is arbitrary. User can get the output in a specific folder by directly providing the `output_dir` like `parent-dir/specific-dir`.
  - `output_name` was allowed for pass config to save intermediate models. But this can be achieved by providing multiple pass flows like `[[A, B], [A, B, C]]`. This is cleaner than the former.
- Refer to `Engine.run` for more details on the new output structure.

**CLI**:
- `add_model_options` is made configurable so that only the desired model type related options are added.
- `save_output_model` uses the new engine output directory structure to copy the output model into the final output directory.
- `finetune` command separated into `finetune` and `generate-adapter` commands. These commands can be chained as shown in the llama2 multilora notebook.

### DIFF
--- a/olive/cli/base.py
+++ b/olive/cli/base.py
@@ -66,10 +66,28 @@ def _get_onnx_input_model(model_path: str) -> Dict:
     Only supports local ONNX model file path.
     """
     print(f"Loading ONNX model from {model_path}")
-    return {
+    model_config = {
         "type": "OnnxModel",
         "model_path": model_path,
     }
+
+    # additional processing for the model folder
+    model_path = Path(model_path).resolve()
+    if model_path.is_dir():
+        onnx_files = list(model_path.glob("*.onnx"))
+        if len(onnx_files) > 1:
+            raise ValueError("Found multiple .onnx model files in the model folder. Please specify one.")
+        onnx_file_path = onnx_files[0]
+        model_config["onnx_file_name"] = onnx_file_path.name
+
+        # all files other than the .onnx file and .onnx.data file considered as additional files
+        additional_files = sorted(
+            set({str(fp) for fp in model_path.iterdir()} - {str(onnx_file_path), str(onnx_file_path) + ".data"})
+        )
+        if additional_files:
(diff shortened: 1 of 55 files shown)


olive/cli/generate_adapter.py
```diff
@@ -0,0 +1,141 @@
...
+        sub_parser.add_argument(
+            "--precision",
+            type=str,
+            default="float16",
+            choices=["float16", "float32"],
+            help="The precision of the optimized model and adapters.",
+        )
+
+        # Model options
+        add_model_options(sub_parser, enable_hf=True, enable_hf_adapter=True)
```
devang-ml, line 43: This CLI should also accept ONNX model as an input. In that case, skip onnx graph capturing & optimizations and just run extract adapter pass. MB metadata should be generated based on use-ort-genai flag, as it is right now.

olive/cli/base.py
```diff
@@ -231,37 +255,76 @@ def add_remote_options(sub_parser):
     )
 
 
-def add_model_options(sub_parser):
+def add_model_options(
+    sub_parser: ArgumentParser,
```
devang-ml, line 259: Should we add output options also here?

olive/cli/generate_adapter.py
jambayk: I will add this part. Just saw in the design doc that this is supported scenario

olive/cli/base.py
jambayk: sure. I can move it here too.

devang-ml (approved):

devang-ml (approved):

PR #1363, opened 2024-09-18
Make certain gptq options customizable via model specific mapping

shaahji:
## Make certain gptq options customizable via model specific mapping

This is to avoid hardcoding these paramters in config files for models that aren't (like phi3) yet officially supported by auto-gptq.

- [ x Lint and apply fixes to your code by running `lintrunner -a`

### DIFF
--- a/olive/passes/pytorch/gptq.py
+++ b/olive/passes/pytorch/gptq.py
@@ -11,6 +11,7 @@ from typing import Any, Dict, List, Union
 import torch
 
 from olive.common.config_utils import validate_config
+from olive.common.hf.mappings import MODEL_INSIDE_LAYER_MODULES, MODEL_LAYERS_BLOCK_NAME, MODEL_OUTSIDE_LAYER_MODULES
 from olive.data.config import DataConfig
 from olive.hardware.accelerator import AcceleratorSpec, Device
 from olive.model import HfModelHandler, PyTorchModelHandler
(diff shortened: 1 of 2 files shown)


olive/passes/pytorch/gptq.py
```diff
@@ -166,20 +167,35 @@ def _run_for_config(
...
             quantized_model.inside_layer_modules = config["inside_layer_modules"]
+        elif model_type not in GPTQ_CAUSAL_LM_MODEL_MAP:
+            if model_type in MODEL_INSIDE_LAYER_MODULES:
+                quantized_model.inside_layer_modules = MODEL_INSIDE_LAYER_MODULES[model_type]
+            else:
+                raise ValueError(
+                    "Can't get inside_layer_modules to quantize automatically, please provide it in config."
+                )
+
+        if model_type not in GPTQ_CAUSAL_LM_MODEL_MAP:
```
jambayk, line 194: should we also do the same for `layer_block_name` and remove the default value for this? If you look at the autogptq base, there is nothing special about layer_block_name compared to the other two https://github.com/AutoGPTQ/AutoGPTQ/blob/097dd04e2fd02bc6664908cd91b7ef8c6302a332/auto_gptq/modeling/_base.py#L88

jambayk: also since the logic is the same, could you abstract the logic and parameterize the three?

jambayk (approved):
Thanks!

olive/passes/pytorch/gptq.py
```diff
@@ -41,7 +42,7 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
             ),
             "layers_block_name": PassConfigParam(
                 type_=str,
-                default_value="model.layers",
+                default_value=None,
                 description=(
                     "Block name to quantize. Default value is model.layers. "
```
jambayk, line 47: oh, btw this still refers to a default value.

shaahji: If I leave that as default value then line 181 will always succeed making the else part non-reachable code.

jambayk: sorry, I meant the description still refers to a default value when there is no default value anymore.

jambayk (approved):

PR #1366, opened 2024-09-20
Fix SDXL optimization for DirectML

thevishalagarwal:
Issue: SD XL optimization for DirectML doesn't work due to incorrect config


jambayk (approved):
Thanks for the fix! We tested the stable diffusion example in the CI but not this one since the model to too large.

Kotomi-Du:
Hi, @thevishalagarwal 
May I ask what kind of optimization it does for SDXL? I have tried to convert a new optimized model based on latest commit, but didn't see the performance improve for unet on NV4060ti and Intel discrete graphics.

thevishalagarwal:
@Kotomi-Du For SDXL 1.0 base, I can see the number of nodes reduced from 15345 to 3322 and a magnitude of reduced inference time.

PR #1374, opened 2024-09-25
Restrict initialization of passes to what's called out in pass_flows

shaahji:
## Restrict initialization of passes to what's called out in pass_flows

Config can list any number of passes, but the ones that are actually run are the ones listed in pass_flows. This avoids the case where a pass may not be able to successfully be initialized which shouldn't matter because its not going to be run unless it is explicitly called out in the pass_flows.

### DIFF
--- a/olive/workflows/run/run.py
+++ b/olive/workflows/run/run.py
@@ -231,18 +231,19 @@ def run_engine(package_config: OlivePackageConfig, run_config: RunConfig):
     # to remove engine level loop and pass the accelerator_specs to the engine directly.
     for accelerator_spec, (passes, pass_flows) in zip(acc_list, pass_list):
         engine.reset_passes()
-        if passes:
-            # First pass registers the necessary module implementation
-            for pass_config in passes.values():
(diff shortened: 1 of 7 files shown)


olive/workflows/run/run.py
```diff
@@ -231,28 +231,32 @@ def run_engine(package_config: OlivePackageConfig, run_config: RunConfig):
     # to remove engine level loop and pass the accelerator_specs to the engine directly.
     for accelerator_spec, (passes, pass_flows) in zip(acc_list, pass_list):
         engine.reset_passes()
-        if passes:
+        pass_flows_to_run = pass_flows or [[*passes.keys()]]
```
jambayk, line 234: can you combine this into a single set? There might be cases like `[["a", "b"], ["a", "b","c"]]` where we would be registering the same pass multiple times.

xiaoyu-work: I suppose this dedupe logic should be added to register method.

jambayk (dismissed):

jambayk (approved):

PR #1376, opened 2024-09-26
Deprecate support for inline data configs

shaahji:

### DIFF
--- a/olive/workflows/run/config.py
+++ b/olive/workflows/run/config.py
@@ -402,6 +402,11 @@ def _resolve_system(v, values, system_alias):
 
 
 def _resolve_data_config(v, values, data_config_alias):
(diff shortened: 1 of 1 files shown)


olive/workflows/run/config.py
```diff
@@ -402,6 +402,9 @@ def _resolve_system(v, values, system_alias):
 
 
 def _resolve_data_config(v, values, data_config_alias):
+    if isinstance(data_config_alias, dict):
```
jambayk, line 405: Can you also check for DataConfig object to cover all possible cases?

jambayk: Please also make the error more informative. Something like `"Inline data configs are not supported! Use the name of a data config defined under 'data_configs'"`

jambayk (approved):

PR #1399, opened 2024-10-03
LoRA: Expose `deepspeed` option with default value if True

jambayk:
## Describe your changes
Make `deepspeed` an exposed `training_args` option. Along with the dict and string types supported by default, we also allow bool. If set to `True`, it will use a default deepspeed config which enables ZeRO3 with cpu offloading.


shaahji (commented):
Should olive_config.json be updated to include deepspeed as a dependency?

shaahji (approved):

PR #1400, opened 2024-10-03
Refactor cache

xiaoyu-work:
## Describe your changes

Refactor cache system.

- Rename cloud cache to shared cache.
- Unify local cache structure and shared cache structure.
- Update cache structure and config.


olive/cache.py
```diff
@@ -201,53 +321,27 @@
...
-        if isinstance(config, ConfigBase):
-            config_class = config.__class__
+            return local_resource_path
 
-        # find and download all non-local resources
-        config = convert_configs_to_dicts(config)
-        all_resources = find_all_resources(config)
-        for resource_key, resource_path in all_resources.items():
-            set_nested_dict_value(config, resource_key, self.get_local_path(resource_path))
+        return local_resource_path
```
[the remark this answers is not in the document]
shaahji: Resolved.

olive/cache.py
```diff
@@ -32,127 +37,243 @@
...
+            values["container_name"] = values.get("container_name") or "olivecachemodels"
+        else:
+            values["enable_shared_cache"] = False
+        return values
+
+    @staticmethod
+    def _get_shared_cache_match(cache_dir):
+        cache_dirs = [cache_dir] if not isinstance(cache_dir, list) else cache_dir
+        pattern = r"https://([^.]+)\.blob\.core\.windows\.net/([^/]+)"
+        for cache_dir in cache_dirs:
```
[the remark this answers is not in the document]
shaahji: Resolved.

olive/cache.py
```diff
@@ -32,127 +37,243 @@ class CacheSubDirs:
...
+
+    @staticmethod
+    def _get_shared_cache_match(cache_dir):
+        cache_dirs = [cache_dir] if not isinstance(cache_dir, list) else cache_dir
+        pattern = r"https://([^.]+)\.blob\.core\.windows\.net/([^/]+)"
+        for cache in cache_dirs:
+            match = re.match(pattern, str(cache))
+
+            if match:
+                return match
```
shaahji, line 90: You could just split the string with '/' as separator to make this simpler.

And, what if there's are more than one cache_dir i.e. the input is a list and there are multiple valid urls?

olive/cache.py
```diff
@@ -32,127 +37,243 @@ class CacheSubDirs:
...
+    account_name: str = None
+    container_name: str = None
+    enable_shared_cache: bool = False
+    update_shared_cache: bool = True
+
+    @validator("account_name", pre=True)
+    def validate_account_name(cls, v, values):
+        if v:
+            return v
+        match = cls._get_shared_cache_match(values.get("cache_dir"))
```
shaahji, line 61: `cache_dir` can be a list of strings also and you need to handle that.

olive/cache.py
```diff
@@ -32,127 +37,243 @@ class CacheSubDirs:
...
+        return match.group(1) if match else None
+
+    @validator("container_name", pre=True)
+    def validate_container_name(cls, v, values):
+        if v:
+            return v
+        match = cls._get_shared_cache_match(values.get("cache_dir"))
```
shaahji, line 68: Same as above. `cache_dir` can be a list of strings also and you need to handle that.

olive/cache.py
```diff
@@ -32,127 +37,243 @@ class CacheSubDirs:
...
-        cache_dir: Union[str, Path],
-        clean_cache: bool = False,
-        clean_evaluation_cache: bool = False,
-    ):
-        self.cache_dir = Path(cache_dir).resolve()
-        logger.info("Using cache directory: %s", self.cache_dir)
-        self.dirs = CacheSubDirs.from_cache_dir(self.cache_dir)
+    def __init__(self, cache_config: Union[CacheConfig, Dict]):
+        cache_config = validate_config(cache_config, CacheConfig)
+        cache_dir = Path(cache_config.cache_dir).resolve()
```
shaahji, line 101: Assuming `cache_dir` is a str or path. What about List[str]?

olive/cache.py
```diff
@@ -32,127 +37,243 @@ class CacheSubDirs:
...
 
-        if clean_evaluation_cache and self.dirs.evaluations.exists():
+        if cache_config.clean_evaluation_cache and self.dirs.evaluations.exists():
             shutil.rmtree(self.dirs.evaluations, ignore_errors=True)
 
+        cache_dir.mkdir(parents=True, exist_ok=True)
         for sub_dir in asdict(self.dirs).values():
-            if clean_cache and sub_dir.exists():
-                shutil.rmtree(sub_dir)
             sub_dir.mkdir(parents=True, exist_ok=True)
```
shaahji, line 113: Create the directories only if the `self.enable_shared_cache` is True.

olive/cache.py
```diff
@@ -32,127 +37,243 @@ class CacheSubDirs:
...
-        return self.dirs.models / f"{model_id}.json"
+        except Exception:
+            logger.exception("Failed to cache model to local cache.")
 
-    def get_model_output_path(self, model_id: str) -> Path:
-        """Get the path to the model output directory."""
-        output_path = self.dirs.models / model_id / "output_model"
-        output_path.mkdir(parents=True, exist_ok=True)
-        return output_path
+        if self.enable_shared_cache and self.update_shared_cache:
```
shaahji, line 149: Should we check for `self.enable_shared_cache`??
Without it you could support "upload only" scenario i.e. "no downloads but still update the remote cache".

olive/cache.py
```diff
@@ -32,127 +37,243 @@ class CacheSubDirs:
...
+        run_json["output_model_id"] = output_model_id
+        run_json_path = self.get_run_json_path(output_model_id)
+        try:
+            with run_json_path.open("w") as f:
+                json.dump(run_json, f, indent=4)
+            logger.debug("Cached run %s to %s", output_model_id, run_json_path)
+        except Exception:
+            logger.exception("Failed to cache run to local cache.")
+
+        if self.enable_shared_cache and self.update_shared_cache:
```
shaahji, line 190: May be remove `self.enable_shared_cache`

shaahji (changes requested):

olive/cache.py
xiaoyu-work: This case is handled in `_get_shared_cache_match`

olive/cache.py
xiaoyu-work: Same as above

olive/cache.py
xiaoyu-work: it would be hard to handle invalid case if splitting by `/`. And it will just return the first valid match.

olive/cache.py
shaahji: It's not because that returns only a single match. What if the input was actually a list of paths and there were more than one match?

olive/cache.py
xiaoyu-work: This is for local cache. What do you mean by if `self.enable_shared_cache` is true?

olive/cache.py
xiaoyu-work: The default value of `update_shared_cache` could be a problem. Right now `enable_shared_cache` is not needed for users if they don't want to enalbe shared cache, if `update_shared_cache` always be True, `enable_shared_cache` will always be needed for users.

olive/cache.py
xiaoyu-work: Will fix this.

olive/cache.py
shaahji: Move the mkdir calls after line 116 so user isn't left with empty directories when cache is disabled.

xiaoyu-work: Even if shared cache is enabled, local cache is still valid. Files will be cached to both local and shared cache. As for now, local cache is always enabled.

olive/cache.py
xiaoyu-work: done

olive/cache.py
xiaoyu-work: fixed

shaahji (approved):

PR #1403, opened 2024-10-03
MatMulNBitsToQDQ: Pass to convert MatMulNBits contrib op to standard QDQ

jambayk:
- Onnxruntime originally only support sub-4bit quantization using the `MatMulNBits` contrib operator. But int4/uint4 are now officially supported by onnx using the standard QDQ operators and is used by EPs such as DirectML and QNN.
- This pass converts a MatMulNBits node into a sequence of `DequantizeLinear`->`Transpose`(optional)->`MatMul` nodes. This way, existing quantization tools don't need to be modified to export in QDQ format.
- Updated the `quantize` CLI to use QDQ encoding by default. There is an option to use `MatMulNBits` if they pefer.

   New `MatMulNBitsToQDQ` pass to convert `MatMulNBits` to `DQ->MatMul`.

### DIFF
--- a/olive/passes/onnx/mnb_to_qdq.py
+++ b/olive/passes/onnx/mnb_to_qdq.py
@@ -1,0 +1,307 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+import logging
+import math
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, Dict
+
+import numpy as np
+import onnx
+
+from olive.hardware.accelerator import AcceleratorSpec
+from olive.model import ONNXModelHandler
+from olive.model.utils import resolve_onnx_path
+from olive.passes import Pass
+from olive.passes.onnx.common import get_external_data_config, model_proto_to_olive_model
+from olive.passes.onnx.onnx_dag import OnnxDAG
+from olive.passes.pass_config import PassConfigParam
+
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+
+logger = logging.getLogger(__name__)
+
+
+class MatMulNBitsToQDQ(Pass):
+    @classmethod
+    def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        return {
+            "use_transpose_op": PassConfigParam(
+                type_=bool,
+                # TODO(jambayk): decide whether to enable this by default or not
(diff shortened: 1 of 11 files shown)


devang-ml (commented):
Please surface this in quantize CLI while quantizing ONNX models. By default QDQ encoding is used.

olive/passes/onnx/mnb_to_qdq.py
```diff
@@ -0,0 +1,307 @@
...
+    def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        return {
+            "use_transpose_op": PassConfigParam(
+                type_=bool,
+                # TODO(jambayk): decide whether to enable this by default or not
+                # False gives same output on arm Mac/Windows, but not on x64 Linux/Windows
+                default_value=True,
+                description=(
+                    "Whether to use a Transpose operator after the DequantizeLinear operator. If False, the weight"
+                    " initializer will be transposed instead."
```
devang-ml, line 38: Add a note in the description regarding in which scenarios user may want to set this to false.

olive/passes/onnx/mnb_to_qdq.py
```diff
@@ -0,0 +1,307 @@
...
+                node_output_proto = dag.get_io(node_output).proto[-1]
+            node_attributes = dag.get_node_attributes(node_name)
+            K = node_attributes["K"]  # noqa: N806
+            N = node_attributes["N"]  # noqa: N806
+            block_size = node_attributes["block_size"]
+            num_k_blocks = math.ceil(K / block_size)
+
+            # only deal with 4 bits for now
+            if node_attributes["bits"] != 4:
+                continue
```
devang-ml, line 81: Add a info message indicating that this is only dealing with 4 bits.

devang-ml: INT4 only or FP4/NF4 are fine?

jambayk: int4 only. Matmulnbits only does int4 4bit quantization.

olive/passes/onnx/mnb_to_qdq.py
jambayk: I updated the default value to be True since it is more standard to just have DQ->MatMul. Added a note that using transpose op might be more efficient for eps such as DirectML. For other EPs, I am not sure which one is better since the int4 support is new and might not be optimized either way.

olive/passes/onnx/mnb_to_qdq.py
jambayk: done. Added log and updated comment.

devang-ml (approved):

PR #1407, opened 2024-10-07
Adding new SearchParameter type and simplifying the implementation

shaahji:
## Adding new SearchParameter type and simplifying the implementation

* New type "Proxy" which uses another parameter's suggestions as its own
* Simplify the implementation to avoid having to check subclass type for special handling.
* Add unit tests


olive/passes/olive_pass.py
```diff
@@ -338,7 +339,15 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
     def _resolve_defaults(cls, config: Dict[str, Any], default_config: Dict[str, PassConfigParam]) -> Dict[str, Any]:
         """Resolve default values."""
         for key, value in config.items():
-            if value == PassParamDefault.DEFAULT_VALUE:
+            if isinstance(default_config[key].searchable_values, CategoricalValue):
+                # Replace CategoricalValue with Categorical parameter
+                v = value
+                if v is None or isinstance(v, CategoricalValue):
```
devang-ml, line 345: Is this check flipped?
When do you want to override the value with default_value?

shaahji: Override the value with default if the user hasn't provided an explicit value for the parameter as part of the config.

This is correct. In the above case, the value is of type `CategoricalValue`.

olive/strategy/search_parameter.py
```diff
@@ -66,6 +69,33 @@ def to_json(self):
         return {"olive_parameter_type": "SearchParameter", "type": "Categorical", "support": self.support}
 
 
+class CategoricalValue(Categorical):
```
jambayk, line 72: why is this required? It looks to be the same as a categorical.

olive/strategy/utils.py
```diff
@@ -65,8 +65,8 @@ def _search_space_graph(search_space: Dict[str, SearchParameter]) -> DirectedGra
     """Create a directed graph from the search space."""
     graph = DirectedGraph(list(search_space.keys()))
     for name, param in search_space.items():
-        if isinstance(param, Conditional):
-            for parent in param.parents:
+        if isinstance(param, SearchParameter):
```
jambayk, line 68: this condition is unnecessary since all params in the search space must be search parameters

olive/strategy/search_parameter.py
```diff
@@ -130,21 +160,19 @@ def __init__(
...
 
-    def get_support_with_args(
-        self, parent_values: Dict[str, Any]
-    ) -> Union[List[str], List[int], List[float], List[bool]]:
+    def get_support(self, parent_values: Dict[str, Any] = None) -> Union[List[str], List[int], List[float], List[bool]]:
         """Get the support for the search parameter for a given parent value."""
-        # pylint: disable=arguments-differ
-        assert parent_values.keys() == set(self.parents), "parent values keys do not match the parents"
+        assert parent_values is not None
+        assert set(self.parents).issubset(parent_values.keys()), "Not all parents included in parent values keys"
```
jambayk, line 169: the error message doesn't sound correct. the condition is more like "no parents included in parent values keys"

jambayk: nevermind this is fine

olive/strategy/search_parameter.py
```diff
@@ -130,21 +160,19 @@ def __init__(
...
         parent_values = tuple(parent_values[parent] for parent in self.parents)
         return self.support.get(parent_values, self.default).get_support()
 
     def condition(self, parent_values: Dict[str, Any]) -> SearchParameter:
         """Fix the parent value and return a new search parameter."""
-        assert set(parent_values.keys()).issubset(set(self.parents)), "parent values keys not a subset of the parents"
+        assert set(self.parents).issubset(parent_values.keys()), "Not all parents includes in parent values keys"
```
jambayk, line 175: This check makes `condition` the same as `get_support`. In condition, only a subset of the original parents can be provided.

olive/passes/olive_pass.py
```diff
@@ -338,7 +339,15 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
     def _resolve_defaults(cls, config: Dict[str, Any], default_config: Dict[str, PassConfigParam]) -> Dict[str, Any]:
         """Resolve default values."""
         for key, value in config.items():
-            if value == PassParamDefault.DEFAULT_VALUE:
+            if isinstance(default_config[key].searchable_values, CategoricalValue):
```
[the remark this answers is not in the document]
shaahji: PR #1401 is the example of using this search parameter.

In simple words, the requirement is that "suggestions for a `Categorical` parameter are user input driven".

In more words, consider execution providers as search parameter. User can cherry-pick, say two, of the many supported execution providers. Search needs to be run on only the chosen ones i.e. two search points.

Another example, consider a search parameter like QLoRA rank that has to search on a few different rank values. At the moment, this is hardcoded to [16, 32, 64] in the pass implementation. There is no way for the user to run this pass for [128, 256, 512]. `CategoricalValue` supports this use case. 

Longer explanation with example -
`OrtSessionParamsTuning` takes a list of execution providers as input. For the search to work, each of those entry in that user provided list is a search point. `CategoricalValue` is sort of a placeholder which is replaced by 'Categorical` in olive_pass.py::342. There are other such config parameters.

olive/strategy/search_parameter.py
shaahji: Answered [here](https://github.com/microsoft/Olive/pull/1407#discussion_r1793011403)

olive/strategy/search_parameter.py
shaahji: > In condition, only a subset of the original parents can be provided.

Not sure what you mean by "only a subset of the original parents". What's _original parents_? I don't see `self.parents` changing in this context. 

Just so I understand your proposal, are you proposing the following -
```
parent_values = {k: v for k, v in parent_values.items() if k in self.parents}
assert set(parent_values.keys()).issubset(set(self.parents))
```
This assertion would never ever be false and will never fire. Limiting `parent_values` to keys in `parents` and then asserting that `parent_values` has only such keys?

`parent_values` is always a superset because it can contain values for potentially all search parameters in the search space. `parents` is, however, is a subset of that same set. That's exactly what the code is asserting.

jambayk: True, the assert is not needed. The original assert was to ensure at least one of the parents is present in `parent_values` (which previously did not include non parents). But I missed the fact that an empty set is also a subset.
An equivalent now would be replacing the old assert with
```
parent_values = {k: v for k,v in parent_values.items() if k in self.parents}
if not parent_values:
    return self
```

The `set(self.parents).issubset(parent_values.keys())` is not what we want. It asserts that all parents are present in `parent_values` but condition doesn't need that.

olive/passes/olive_pass.py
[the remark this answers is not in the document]
jambayk: so, what this does is tell the pass config that whatever the user provides, make it a categorical? This feels hacky and it also ignores the `disable_search` flag since it always becomes a search parameter. Is this expected? A no-search workflow will be invalid if perf-tuning pass is present? 

I think it would be better if the categoricalvalue change is part of the perf tuning PR so that the logic can be reviewed with it's usage. The way it is now, I cannot follow the logic and review/approve the changes related to it.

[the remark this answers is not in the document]
jambayk: if my understanding is correct, I think the `categoricalvalue` behavior might need to a separate flag in the passparamconfig. searchable_value is only used to hold the default search param when search is enabled or user asks for it using `SEARCHABLE_VALUES`. What you want looks more like a config validation to convert values to search parameter.

olive/strategy/search_parameter.py
shaahji: > which previously did not include non parents.

It still doesn't include non-parents. The `parent_values` is built using whatever the `get_parents` returns which is not any different from previous implementation.

> It asserts that all parents are present in parent_values but condition doesn't need that.

Hmm is that true? Wouldn't all parents be present in `parent_values`, even if one's value is default?

Regardless, this is not the intended change for the PR itself so I will revert it back to whatever it was.

olive/passes/olive_pass.py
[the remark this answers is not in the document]
jambayk: cmiiw, what we want is an easy (maybe automatic?) way for a user to provide search values without having to configure it using something like `{"olive_parameter_type": "SearchParameter", "type": "Categorical", "support": [1, 2, 3]}`. The param can be configured directly using `[1,2,3]`. 

I think one way to achieve this would be to add a validator to the pass config class if search is enabled. this validator could convert all (or only some flagged paramters if we add such a flag) non-search parameters into categoricals. this is fine since `Categorical([1])` is the same a providing `1`. 
However, a potential issue for this is how do we recognize if it's a fixed value or searchable list if the type of the param is also a list/sequence? Maybe a flag would help with that so that the flag is only allowed on params whose type is not a sequence.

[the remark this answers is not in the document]
shaahji: >  The param can be configured directly using [1,2,3]

Yes, that's correct and that's what current implementation supports.

> ... non-search parameters into categoricals. this is fine since Categorical([1]) is the same as providing 1

I don't think that works. We need to convert searchable parameters to `Categorical`. How does changing non-searchable parameters help (yes `Categoriclas([1])` is the same as `1` but that's a moot point). The non-searchable `Categorical`s will get converted to constants again down the pipeline giving us the same config as before the change. It has absolutely no affect on the searchable parameters.

The problem is under certain conditions we want to treat user provided input as suggestions for `Categorical`, overriding the default, if present. Yes, we can do that swap/replacement as part of validation or somewhere else.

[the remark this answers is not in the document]
jambayk: by "non-search parameters`, I meant values that are provided like `1`, `[1,2]` instead of using `SearchParameter` object or json. There is nothing stopping a parameter from search and making it "non-searchable". Any parameter that has a search parameter as its value is "searchable". having default_value or default_searchable_value is not meant to be a marker for whether a param is searchable or not. they are just defaults that get used when search is enabled/disabled. 

what i was suggesting is that, if we want to convert user inputs to search params so that they don't need to fully specify a search parameter, we can just convert the values (non searchparameter types) into categorical. this works even if the user provided a single fixed value because of the moot point.
This way, the peft tuning pass is not a special case. we still have to tackle the list type issue regardless.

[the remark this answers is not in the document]
shaahji: Wouldn't such a solution be pass specific unless we add a new flag to the `PassConfigParam` and handle it generically in the validation logic.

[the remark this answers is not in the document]
jambayk: Yes, I think this is better done in the validation logic. We probably need a flag to enable such behavior per param using a flag because of the following issue otherwise:
> However, a potential issue for this is how do we recognize if it's a fixed value or searchable list if the type of the param is also a list/sequence? Maybe a flag would help with that so that the flag is only allowed on params whose type is not a sequence.

Not directly related, even this method is better handled in the validation logic https://github.com/microsoft/Olive/blob/c473a3b73458d55ff5d0da5fe0464338f258e824/olive/passes/olive_pass.py#L338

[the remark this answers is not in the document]
devang-ml: Each parameter has three values.
A) User provided input value.
B) Default value selected by the pass implementor.
C) Searchable values selected by the pass implementer.

Today,
- If search is not enabled then A or B is used.
- If search is enabled and A is not provided, then C is used to generate search points. If A is provided than this parameter is not searched.

The purpose of this work is to provide the pass implementer new (if we don't already support it today) ability to support the following scenario where input type is a list.
- If search is not enabled then A or B is used.
- If search is enabled and A is not provided, then C is used to generate search points. If A is provided than A should be used to generate search points.

The pass implementor should be able to select how they want to treat the parameter. Make A searchable or not.

shaahji:
Abandoned in favor of #1428

PR #1410, opened 2024-10-07
remove unnecessary top level torch imports

xiaoyu-work:
## Describe your changes

remove unnecessary top level torch imports

### DIFF
--- a/olive/data/component/post_process_data.py
+++ b/olive/data/component/post_process_data.py
@@ -2,7 +2,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
-import torch
 import transformers
 
 from olive.data.registry import Registry
@@ -37,6 +36,8 @@ def text_classification_post_process(output_data, **kwargs):
         object: Post-processed data.
 
     """
+    import torch
+
     if isinstance(output_data, transformers.modeling_outputs.SequenceClassifierOutput):
         _, preds = torch.max(output_data.logits, dim=1)
     else:
@@ -57,6 +58,8 @@ def ner_post_process(output_data, **kwargs):
         object: Post-processed data.
 
     """
+    import torch
+
     if isinstance(output_data, transformers.modeling_outputs.TokenClassifierOutput):
         logits = output_data.logits
     else:
--- a/olive/data/component/dataloader.py
+++ b/olive/data/component/dataloader.py
@@ -5,14 +5,14 @@
 
 from typing import List, Union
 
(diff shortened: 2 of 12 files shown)


olive/data/component/post_process_data.py
jambayk, line 1: there are imports like this too https://github.com/microsoft/Olive/blob/add3f639a60b9797b40d6efec893594ce75adc06/olive/data/component/dataloader.py#L8

jambayk: also what about torchmetrics import. importing torchmetrics import torch also.

xiaoyu-work: Good call. I will update those imports also.

xiaoyu-work: It is difficult to lazily import torchmetrics, so I am leaving it as is. Not all of the torch import can be simply updated to lazy import. Though we can do that I doubt if it worths. So I just updated those straightforward imports.

jambayk: if torchmetrics is not lazily imported, then there is nothing to gain from lazy importing torch. torchmetrics imports torch.

xiaoyu-work: This change is only for optimizing warm import torch. As I tried myself, torch cold import on my cloud vm takes 5.8s, then warm import takes 0.8s. At my laptop, code import takes 8~9s, then warm import takes 4.1s. And it is not worthing make all torch imports be lazy imports, so this PR is only for warm import.

jambayk: sorry, I don't get how only making some torch imports helps here. when there are other torch imports that are as the same levels as the ones you made lazy in this PR (like https://github.com/microsoft/Olive/blob/c473a3b73458d55ff5d0da5fe0464338f258e824/olive/data/component/dataset.py#L11)

Does this help the time it takes it init and run an olive with a config where torch is not needed? if so, unless we can run a workflow with no torch requirements + no torch installed without a torch import error, the lazy and non-lazy torch imports feels arbitrary.

xiaoyu-work: If I can't update all torch imports in a specific file to lazy imports, I won't make any changes to that file. This avoids having a mix of lazy and non-lazy imports within the same file. However, if it's feasible to update all torch imports to lazy in a particular file, I'll do with the change. We can't make all torch imports lazy because some classes are inherited from torch classes. For example Dataset class (while it's possible to refactor the code to avoid this issue like defining the class inside a function, but this would overly complicate the code). I agree this will make our code arbitrary, but this may reduce the workflow running time if we avoid warm import torch as many as possible.

jambayk (approved):

PR #1418, opened 2024-10-11
Add example pipeline

xiaoyu-work:
## Describe your changes

Add example pipeline


.azure_pipelines/olive-examples.yaml
```diff
@@ -0,0 +1,67 @@
+trigger: none
+
+pr: none
+
+schedules:
+- cron: '0 0 * * 1'
```
jambayk, line 6: do you think weekly will be enough? Since we are actively developing the code, the examples might break frequently.

.azure_pipelines/olive-examples.yaml
jambayk, line 1: should we move some of the tests from the ci to here? maybe the inc, vitisai, bert docker tests?

.azure_pipelines/olive-examples.yaml
jambayk: also we might need another weekly run to test the examples with the nightly ort package also. unless we pin the ort requirements for the tests

xiaoyu-work: I think once a week is enough, even daily we don't check it daily. Once a week we can have a weekly schedule to check the pipeline. As for nightly version, I checked with Devang and he said we don't need it for now. We could add it later when we need it.

.azure_pipelines/olive-examples.yaml
xiaoyu-work: Sounds good to me! I can move them here.

.azure_pipelines/olive-examples.yaml
jambayk: sounds good. thanks!

.azure_pipelines/olive-examples.yaml
jambayk: thanks. and also resnet_qat.

olive/resource_path.py
```diff
@@ -40,7 +40,6 @@ class ResourceType(CaseInsensitiveEnum):
 LOCAL_RESOURCE_TYPES = (ResourceType.LocalFile, ResourceType.LocalFolder)
 AZUREML_RESOURCE_TYPES = (
     ResourceType.AzureMLModel,
-    ResourceType.AzureMLRegistryModel,
```
jambayk, line 43: is there a reason for this change?

jambayk: since registrymodel does not require workspace information?

xiaoyu-work: Yes, a couple days ago I added this for shared cache: https://github.com/microsoft/Olive/pull/1417. Then I realize it is used for https://github.com/microsoft/Olive/blob/main/olive/workflows/run/config.py#L314 as well. So instead of updating config file, i reverted my changes about 1417 here.

PR #1422, opened 2024-10-16
Add new lora node support for extract adapter

xiaoyu-work:
## Describe your changes

Add new lora node support for extract adapter

### DIFF
--- a/olive/common/hf/peft.py
+++ b/olive/common/hf/peft.py
@@ -28,15 +28,18 @@ def make_export_compatible_peft(model: torch.nn.Module, merge_weights: bool = Fa
         - Rescale the lora_B weights with scaling factor, change scaling factor to 1 (int) so that it doesn't appear
           in the exported model (only works for torchscript export)
     """
-    if merge_weights:
-        return model.merge_and_unload()
-
     # if pytorch_model is PeftModel, we need to get the base model
     # otherwise, the model forward has signature (*args, **kwargs) and torch.onnx.export ignores the dummy_inputs
     if is_peft_model(model):
+        if merge_weights:
+            return model.merge_and_unload()
         model = model.get_base_model()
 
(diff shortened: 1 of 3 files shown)


olive/common/hf/peft.py
```diff
@@ -28,15 +28,13 @@ def make_export_compatible_peft(model: torch.nn.Module, merge_weights: bool = Fa
         - Rescale the lora_B weights with scaling factor, change scaling factor to 1 (int) so that it doesn't appear
           in the exported model (only works for torchscript export)
     """
-    if not is_peft_model(model):
-        return model
-
     if merge_weights:
```
jambayk, line 31: can you add this inside the condition at line 36? merge_weights is only valid for peft models.

olive/common/hf/peft.py
```diff
@@ -28,15 +28,13 @@ def make_export_compatible_peft(model: torch.nn.Module, merge_weights: bool = Fa
...
         return model.merge_and_unload()
 
     # if pytorch_model is PeftModel, we need to get the base model
     # otherwise, the model forward has signature (*args, **kwargs) and torch.onnx.export ignores the dummy_inputs
-    model = model.get_base_model()
+    if is_peft_model(model):
+        model = model.get_base_model()
 
     from peft.tuners.lora import LoraLayer
```
jambayk, line 39: also, this import needs to be in a try block. we can just return if import fails since it guarantees the model doesn't have loralayers

xiaoyu-work: If we just return here, this will never be triggered cause peft is not required for conversion. Is this expected?

jambayk: yes, that is expected. if we cannot import peft then the model cannot have the peft lora layers.

jambayk (approved):

PR #1424, opened 2024-10-16
Skip quantization that require calibration if model has adapters

jambayk:
## Describe your changes
If a model has unmerged adapters, we expect it to be used for MultiLoRA where the lora adapters can be switched. Quantization techniques that require calibration change the base weights as well as adapter weights together so we cannot just switch these quantized adapters. 

More work/research is needed to make a fully quantized model work with multi lora so we will skip any quantization that might break adapter switching.

### DIFF
--- a/olive/passes/onnx/common.py
+++ b/olive/passes/onnx/common.py
@@ -3,12 +3,14 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 import logging
+import re
 from pathlib import Path
-from typing import Dict, Optional, Union
+from typing import Dict, List, Optional, Union
 
 import onnx
 
 from olive.model import ONNXModelHandler
+from olive.passes.onnx.onnx_dag import OnnxDAG
 from olive.passes.pass_config import PassConfigParam
 from olive.resource_path import LocalFile, LocalFolder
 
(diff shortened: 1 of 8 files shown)


olive/passes/onnx/common.py
```diff
@@ -183,3 +185,29 @@ def model_proto_to_olive_model(
...
     return olive_model
+
+
+def get_lora_name_patterns() -> List[str]:
+    """Get the node name patterns for lora modules."""
+    return [
+        f".*[./]{name}[./]{matmul}$"
+        for name in ["default", "default_1", "lora_A", "lora_B"]
+        for matmul in ["MatMul", "MatMul_Q4"]
+    ]
```
shaahji, line 196: Aren't the nodes usually numbered i.e. <prefix>_0, <prefix_1, and so on. With the '$' at the end it might not match the numbered nodes.

shaahji, line 196: nitpick: This can be a constant set rather than a function.

shaahji (changes requested):

olive/passes/onnx/common.py
jambayk: I haven't found any those are numbered. this are the existing patterns we know of and it works (this method was just copied over from a different file we have been using actively.

jambayk: good idea. done!

shaahji (approved):

PR #1430, opened 2024-10-21
Save output model to output_dir

xiaoyu-work:
## Describe your changes

Save output model to output_dir


test/unit_test/engine/test_engine.py
```diff
@@ -321,17 +321,56 @@ def test_run_no_search(self, mock_local_system_init, tmp_path):
         assert output_node.model_config == onnx_model_config
         assert expected_metrics == output_node.metrics.value
 
-        output_model_dir = output_dir / "output_model"
```
jambayk, line 324: please update the tests in this too https://github.com/microsoft/Olive/blob/acc87632d11a7d9aa7117f50ccfbe2fec963189b/test/unit_test/cli/test_cli.py#L116

olive/engine/footprint.py
```diff
@@ -422,3 +429,30 @@ def _plot_pareto_frontier(self, ranks=None, save_path=None, is_show=True, save_f
 
         if is_show:
             fig.show()
+
+
+def get_best_candidate_node(
```
jambayk, line 434: could you add some comments here to explain the logic?

test/unit_test/engine/test_engine.py
jambayk: https://github.com/microsoft/Olive/blob/acc87632d11a7d9aa7117f50ccfbe2fec963189b/olive/cli/base.py#L412

olive/engine/engine.py
```diff
@@ -229,12 +229,12 @@ def run(
...
-                        output_dir/output_model/metrics.json: evaluation results of the output model
-                        output_dir/output_model/model_config.json: output model configuration
-                        output_dir/output_model/...: output model files
+                        output_dir/metrics.json: evaluation results of the output model
+                        output_dir/model_config.json: output model configuration
+                        output_dir/...: output model files
 
                     B. Multiple pass flows:
-                        output_dir/output_model/{pass_flow}/...: Same as A but for each pass flow
+                        output_dir/{pass_flow}/...: Same as A but for each pass flow
```
jambayk, line 237: since there is only one save_model call, I don't think this structure is correct? 
Please update the search and no search output model structure in this docstring after running and validating all scenarios.

test/unit_test/engine/test_engine.py
xiaoyu-work: curious why this test can pass

olive/engine/footprint.py
xiaoyu-work: will do

olive/engine/engine.py
xiaoyu-work: This case is same for both search and no search. Engine can still get a `output_footprints` and only one model will be saved to output folder. These are intermediate folders created by the engine. I can double check this and paste the folder structure later here.

test/unit_test/engine/test_engine.py
jambayk: the test passes since base.py was not updated and we just mock the run output. we might have to update one of the cli tests to not do the mock.

olive/engine/engine.py
xiaoyu-work: image
They have metrics file in each pass flow pass folder. Though I don't think those files are needed so we can keep the output folder clean. We can remove those intermediate folders later if needed.

test/unit_test/cli/test_cli.py
```diff
@@ -284,7 +282,7 @@ def test_quantize_command(mock_repo_exists, mock_tempdir, mock_run, algorithm_na
     mock_tempdir.return_value = tmpdir.resolve()
     mock_run.return_value = {}
 
-    workflow_output_dir = tmpdir / "output_model" / algorithm_name
+    workflow_output_dir = tmpdir / algorithm_name
```
jambayk, line 285: I don't think the algorithm name exists anymore? This was for the behavior where each pass flow saved it's own output model

olive/engine/engine.py
```diff
@@ -422,9 +429,6 @@ def run_no_search(
                     json.dump(signal.to_json(), f, indent=4)
                 logger.info("Saved evaluation results of output model to %s", results_path)
 
-            self.cache.save_model(model_id=model_ids[-1], output_dir=flow_output_dir, overwrite=True)
```
jambayk, line 425: is this is not present, then I don't think this `pass_flow` nesting is present? 
image

not sure if we should keep it or not

test/unit_test/cli/test_cli.py
xiaoyu-work: Good call. That's correct. Updated.

olive/engine/footprint.py
```diff
@@ -422,3 +429,46 @@ def _plot_pareto_frontier(self, ranks=None, save_path=None, is_show=True, save_f
...
```
jambayk, line 434: how does this work wrt to no-search model with multiple workflows? In these cases, we probably won't have any metrics for the output models. Does this just select the first output model? If so, is using multiple flows without evaluator not supported since the other flows are ignored? Maybe it's fine since Devang did not want pass_flows to be exposed to the user.

xiaoyu-work: For no search, there is one condition before calling this func in Engine: `if outputs and self.passes and not next(iter(outputs.values())).check_empty_nodes()`. No search has no node and the output is `MetricResult`. Engine will directly return the outputs.

jambayk: I don't think that's true? In line 334 you can see that no search also returns a output with the nodes for the final models for each workflow. so, this won't be True. 
Otherwise, it would mean the output model is never saved in no search mode.

xiaoyu-work: Yeah i just double checked and you are right. I don't remember in which case that the result is MetricResult but from what i tested, there will be only one node in pf footprints for no search case and that one will be the output one.

jambayk: with multiple flows, I think there will be multiple nodes? in that case, if there is no evaluator, does it choose the first output model? If so, is this the behavior we want?

jambayk: could you verify the output structure of the combination of the all the following scenarios?
- search/no-search(with/out accelerator)
- one/multiple accelerators
- one/multiple pass flows

xiaoyu-work: We won't cover the case for no search + multi pass flow as pass flow will be deprecated with no search soon. So the output structures are:

- search/no-search(with/out evaluator)
 - search will copy the best node model files to output folder.
 - no search will copy the last node files to the output folder (with/out evaluator)
- one/multiple accelerators
 - will copy the best node models to the output folder, with device-ep folder with run history json file.
- one/multiple pass flows
 - will copy the best node models to the output folder, with pass_flow folder (if has) for metric json file.

xiaoyu-work: Most of them are correct. I removed multi pass flow case for no search, and added final output model path description.

jambayk (dismissed):

olive/engine/footprint.py
jambayk: thanks! sorry for taking long to review. wanted to make sure all scenarios were covered since some the cli and demos depend on the output structure.

olive/cli/base.py
```diff
@@ -416,7 +416,7 @@ def save_output_model(config: Dict, output_model_dir: Union[str, Path]):
 
     This assumes a single accelerator workflow.
     """
-    run_output_path = Path(config["output_dir"]) / "output_model"
+    run_output_path = Path(config["output_dir"])
```
jambayk, line 419: btw without this `output_model` nesting, now the output of the cli would also have the footprint and other files copied over even though they mean nothing to the user of the cli. I think we might need an option to disable saving these files like you mentioned once before.

jambayk: or make it opt-in to save the extra files instead of opt-out. most users only care about the final output.

xiaoyu-work: Good call! I don't think users need those files. We should only copy model files here.

olive/cli/base.py
```diff
@@ -417,35 +424,32 @@ def save_output_model(config: Dict, output_model_dir: Union[str, Path]):
...
-        with model_config_file.open("w") as f:
-            json.dump(model_config, f, indent=4)
+    with model_config_path.open("r") as f:
+        model_config = json.load(f)
+
+    all_resources = find_all_resources(model_config)
+    for resource_key, resource_path in all_resources.items():
+        src_path = Path(resource_path.get_path()).resolve()
+        if src_path.is_dir():
+            hardlink_copy_dir(src_path, output_model_dir / src_path)
```
jambayk, line 441: I don't think this is correct. `output_model_dir / src_path = src_path` since `src_path` is a fully resolved named. you probably wanted `str_path.name` like in 448?

xiaoyu-work: Yeah true! I forgot to update this!

jambayk: please also check how the outputs for a graph capture for with `--use_ort_genai` and `--use_model_builder` look. the `additional_files` attributes might make this copy the additional files into the output directory, even though they refer to files already in the `model` subdir.

jambayk: Like I described in the previous comment, I think it will be easier to just make saving the footprint, etc opt-in in the workflow config. that way, even for cli, we can just use the final output dir and not need to do any of this copy and path update.

xiaoyu-work: My intention is copying additional files to output folder as they are also a part of the output model.

jambayk: I think the additional files get copied twice in this. Once is as part of the `model_path` resource copy into `model` subdir. Then they are again copied individually directly into the output_dir itself.

xiaoyu-work: Are all additional files stored in the model path? I kinda remember it can be everywhere depending on the pass who created it. Was this logic updated before?

jambayk: it is stored in model path normally. some passes like modelbuilder with metadata saves it in a different folder but that was because we weren't sure if should copy/hardlink the existing model files. But both the pass carry forward https://github.com/microsoft/Olive/blob/main/olive/passes/olive_pass.py#L274 and cache model save have always saved in model_path https://github.com/microsoft/Olive/blob/0db2d72b00a3aeff3f859ea693b63d69c6de08eb/olive/cache.py#L451 

Since the output of a workflow goes through the cache model save, the additional files are already in model_path resources regardless of what pass was there before.

xiaoyu-work: I added `no_artifacts` to the Engine config, so no other artifacts will be saved to the output path except for the model files.

jambayk: Thanks!

jambayk: Other than the CI failures, the changes look good to me

jambayk (approved):
Thanks!

olive/cli/base.py
xiaoyu-work: The tests were fixed.

PR #1436, opened 2024-10-24
SplitModel: Pass to split transformers model

jambayk:
## Describe your changes
Added two new passes that split a transformers model into multiple models. The split is done on the transformers layers.
- `CaptureSplitInfo`: Used the pytorch/hf model to assign split ids to the transformers layers and saves them as model attributes. The conversion pass or model builder pass adds the split assignments and model metadata.
- `SplitModel`: Uses the split_assignments metadata to break the model into splits.
  - If `include_all_nodes` is `True`: Nodes before first split or outside the splits are assigned to the first split. Nodes after the last split are assigned to the last split.
  - If `include_all_nodes` is `False`: Such nodes are ignored. This means the embedding, attention mask computation, final norm and lm heads don't appear in the splits. We expect the user to extract them separately.

`Cache.save_model` now supports saving composite models.
  

olive/passes/pytorch/capture_split_info.py
```diff
@@ -0,0 +1,78 @@
...
+            raise ValueError("layers_block_name is not set and could not be inferred. Please set it manually.")
+
+        num_layers = None
+        if isinstance(model, HfModelHandler):
+            # model attributes already has the hf model config loaded
+            num_layers = find_first_matched_value(model.model_attributes, NUM_HIDDEN_LAYER_NAMES)
+        if not num_layers:
+            # load the models and get the number of layers
+            loaded_model = model.load_model(cache_model=False)
+            layers = get_attr(loaded_model, layers_block_name)
```
devang-ml, line 60: We should provide option to user to split all layers, not just transformer layer block.

jambayk: done

olive/passes/olive_pass.py
```diff
@@ -246,7 +246,10 @@ def run(
             # assumption: the model attributes from passes, if any, are more important than
             # the input model attributes, we should not update/extend anymore outside of the pass run
             output_model.model_attributes = output_model.model_attributes or model.model_attributes
-            Pass._carry_forward_additional_files(model, output_model)
+            if not isinstance(output_model, CompositeModelHandler):
+                # not sure if we should carry forward additional files for composite model
+                # if we should, should we carry forward for each component or just the composite model
+                Pass._carry_forward_additional_files(model, output_model)
```
devang-ml, line 252: What are the additional files?

olive/passes/onnx/split.py
```diff
@@ -0,0 +1,280 @@
...
+from olive.passes.pass_config import PassConfigParam
+
+logger = logging.getLogger(__name__)
+
+
+class SplitModel(Pass):
+    @classmethod
+    def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        return {
+            "include_all_nodes": PassConfigParam(
```
devang-ml, line 29: It would make sense to make the decision of splitting transformer layers (i.e. layer block) only or ALL layers at the time of capturing split info.

olive/passes/olive_pass.py
jambayk: these are files like tokenizer files that get saved with the exported model so that it can be run with genai config. this logic carries forward these files with the output model.

olive/passes/onnx/split.py
jambayk: this is not straightforward. without a way to know how deep the split needs to be and how to handle nested layers, if we split all layers in the model, then it only gets split at a higher level. for example, the hierachy of layers in llama is 
- model
  - embed_tokens
  - layers
  - norm
- lm_head 

just splitting at the top level doesn't split the transformer layers since it's nested. this option is just to ensure that we get a list of models that can be chained together to be equivalent to the unsplit model.

devang-ml: Could we hoist "include_all_nodes" into CaptureSplitInfo pass instead of here in SplitModel pass?

jambayk: yes, I can do that. make it an option in the capture split info part and save it as additional model metadata.

olive/passes/olive_pass.py
devang-ml: We would need to carry forward these files for composite model also. How would we handle tokenizer files after SplitModel pass is run?

olive/passes/onnx/split.py
devang-ml: In that case, leave it as it is.

olive/passes/olive_pass.py
jambayk: It gets carried forward using the `additional_files` attribute in the composite model handler's `model_attribute`. 
image

I will update the comment to make it clear. the comment was before I made some changes to the code to make this work.

devang-ml: ok. thanks!

jambayk: done

devang-ml (approved):

PR #1440, opened 2024-10-25
Update olive_config and auto-opt to use precision, providers and accelerators

shaahji:
## Update olive_config and auto-opt to use precision, providers and accelerators

* Implement providers, accelerators and precisions for pass in olive_config.json.
  Each pass module config now also requires listing supported providers, accelerators
  and precisions.
* Update auto-opt cli to use a pre-defined order of passes when search is disabled.
  The list of passes is selected based on the user's choice of precision for
  the output model.

### DIFF
--- a/olive/cli/auto_opt.py
+++ b/olive/cli/auto_opt.py
@@ -4,8 +4,9 @@
 # --------------------------------------------------------------------------
 import tempfile
 from argparse import ArgumentParser
+from collections import OrderedDict
 from copy import deepcopy
-from typing import Dict, List
+from typing import Any, Dict, List
 
 from olive.cli.base import (
     BaseOliveCLICommand,
@@ -24,46 +25,7 @@ from olive.cli.base import (
     update_shared_cache_options,
 )
 from olive.common.utils import hardlink_copy_dir, set_nested_dict_value
-
-EVALUATE_TEMPLATE = {
-    "common_evaluator": {
-        "metrics": [
-            {
-                "name": "accuracy",
-                "type": "accuracy",
-                "sub_types": [
-                    {"name": "accuracy_score", "priority": 1, "goal": {"type": "max-degradation", "value": 0.1}},
-                ],
-                "data_config": "data_config",
-            },
-            {
-                "name": "latency",
-                "type": "latency",
-                "sub_types": [
-                    {"name": "avg", "priority": 2, "goal": {"type": "percent-min-improvement", "value": 1}},
-                ],
-                "data_config": "data_config",
-                "user_config": {"io_bind": True},
-            },
-        ]
-    }
-}
-
-TEMPLATE = {
-    "input_model": {"type": "HfModel"},
-    "auto_optimizer_config": {},
-    "search_strategy": {"execution_order": "joint", "search_algorithm": "tpe", "num_samples": 5, "seed": 0},
-    "systems": {
-        "local_system": {
-            "type": "LocalSystem",
-            "accelerators": [{"device": "cpu", "execution_providers": ["CPUExecutionProvider"]}],
-        }
-    },
-    "host": "local_system",
-    "evaluators": EVALUATE_TEMPLATE,
-    "evaluator": "common_evaluator",
-    "target": "local_system",
-}
+from olive.package_config import OlivePackageConfig
 
 
 class AutoOptCommand(BaseOliveCLICommand):
@@ -122,7 +84,7 @@ class AutoOptCommand(BaseOliveCLICommand):
             "--precision",
             type=str,
             default="fp32",
-            choices=["fp16", "fp32", "int4", "int8"],
+            choices=["fp4", "fp8", "fp16", "fp32", "int4", "int8", "int16", "int32", "nf4"],
             help=(
                 "The output precision of the optimized model. If not specified, "
                 "the default precision is fp32 for cpu and fp16 for gpu"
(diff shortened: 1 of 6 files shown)


olive/olive_config.json
```diff
@@ -1,106 +1,332 @@
...
+            "module_path": "olive.passes.onnx.extract_adapters.ExtractAdapters",
+            "supported_providers": [ "*" ],
+            "supported_accelerators": [ "*" ],
+            "supported_precisions": [ "*" ]
         },
-        "DynamicToFixedShape": { "module_path": "olive.passes.onnx.dynamic_to_fixed_shape.DynamicToFixedShape" },
-        "ExtractAdapters": { "module_path": "olive.passes.onnx.extract_adapters.ExtractAdapters" },
         "ModelBuilder": {
             "module_path": "olive.passes.onnx.model_builder.ModelBuilder",
+            "supported_providers": [ "cpu", "cuda" ],
```
devang-ml, line 23: Let's use consistent string for provider ... CPUExecutionProvider in all places.

olive/hardware/constants.py
```diff
@@ -30,5 +30,5 @@
         "OpenVINOExecutionProvider",
         "JsExecutionProvider",
     ],
-    "npu": ["QNNExecutionProvider"],
+    "npu": ["QNNExecutionProvider", "VitisAIExecutionProvider"],
```
xiaoyu-work, line 33: Not related to this PR but OpenVINOEP supports npu as well.

olive/olive_config.json
```diff
@@ -1,108 +1,344 @@
 {
     "passes": {
         "AppendPrePostProcessingOps": {
-            "module_path": "olive.passes.onnx.append_pre_post_processing_ops.AppendPrePostProcessingOps"
+            "module_path": "olive.passes.onnx.append_pre_post_processing_ops.AppendPrePostProcessingOps",
+            "supported_providers": [ "*" ],
```
xiaoyu-work, line 5: Any possibility to make those `*` as default value? So no need to add `*` for all passes.

shaahji: We could make "empty list" be equal to "ALL". I couldn't come up with a reason (or an example) where we would want an empty list to mean "None" but decided to use "*" to be explicit.

I can change that if there is a strong opinion among devs. I just preferred to be explicit rather than overloading the meaning of empty list.

devang-ml: Explicit is fine.

olive/cli/auto_opt.py
```diff
@@ -137,6 +99,34 @@ def register_subcommand(parser: ArgumentParser):
             ),
         )
 
+        # DynamicToFixedShape options
+        sub_parser.add_argument(
+            "--dynamic-to-fixed-shape-dim-param",
+            type=str,
+            nargs="*",
+            default=None,
+            help="Symbolic parameter names to use for dynamic to fixed shape pass.",
```
devang-ml, line 108: Please make it clear that this for DML EP as the provider.

olive/cli/auto_opt.py
```diff
@@ -137,6 +99,34 @@ def register_subcommand(parser: ArgumentParser):
...
+        )
+        sub_parser.add_argument(
+            "--dynamic-to-fixed-shape-dim-value",
+            type=int,
+            nargs="*",
+            default=None,
+            help="Symbolic parameter values to use for dynamic to fixed shape pass.",
```
devang-ml, line 115: same as above.

olive/cli/auto_opt.py
```diff
@@ -137,6 +99,34 @@ def register_subcommand(parser: ArgumentParser):
...
+        )
+
+        # MixedPrecisionOverrides options
+        sub_parser.add_argument(
+            "--mixed-precision-overrides-config",
+            type=str,
+            nargs="*",
+            default=None,
+            help=(
+                "Dictionary of name to precision. Has to be even number of entreis with even "
```
devang-ml, line 125: Is this only for FP16 as precision?

devang-ml: Do we need to expose this via CLI?
CLI is not meant to cover all advanced scenarios.

olive/cli/auto_opt.py
```diff
@@ -163,24 +152,37 @@ def run(self):
...
+        olive_config = OlivePackageConfig.load_default_config()
 
-        excluded_passes = ["OrtPerfTuning"]
-        if self.args.use_model_builder:
-            excluded_passes.append("OnnxConversion")
-        else:
-            excluded_passes.append("ModelBuilder")
+        if self.args.provider == "DmlExecutionProvider":
+            # Force the device to gpu for Direct ML provider
+            self.args.device = "gpu"
```
devang-ml, line 159: Why?

devang-ml: https://devblogs.microsoft.com/directx/introducing-neural-processor-unit-npu-support-in-directml-developer-preview/

olive/cli/auto_opt.py
devang-ml: I see, they are required. Let's clarify which precision requires the config.

olive/cli/auto_opt.py
```diff
@@ -215,8 +220,140 @@ def _get_data_config(self) -> List[Dict]:
...
+            ):
+                del passes_config[pass_name]
+
+        if "to_fixed_shape" in passes_config and not (
+            self.args.dynamic_to_fixed_shape_dim_param and self.args.dynamic_to_fixed_shape_dim_value
+        ):
+            raise ValueError("dim-param and dim-value are required for DynamicToFixedShape.")
+
+        if "conversion" not in passes_config and "model_builder" not in passes_config:
+            raise ValueError("Cannot export an onnx model with combination of provided options.")
```
devang-ml, line 281: What if input model is an ONNX model

shaahji: Needed to handle the "metadata_only" scenario!

xiaoyu-work (approved):

PR #1456, opened 2024-11-05
Model Splitting: Add cost model, Simplify split logic

jambayk:
## Describe your changes
- In addition to `num_splits`, `CaptureSplitInfo` pass now supports cost model. The cost model is a csv containing the cost per component of the model. Currently, the only cost supported is the number of bytes. 
- The split decision is made based on the `max_memory` config parameter. This config parameter is temporary and will be moved into the accelerator spec. The required changes are bigger than the scope of this PR so it will be handled in a follow up PR.
- Added `generate-cost-model` CLI to generate cost models for huggingface models. Also added some pre-generated cost models for phi and llama models under `assets/cost_models`.
- `include_all_nodes` option has been removed from `SplitModel` pass. It always includes all nodes. This also made the logic simpler.
- OnnxDAG:
  - Updated logic to handle unused inputs/outputs for nodes that are left as `""`. Found in contrib operators commonly.
  - Handle unnamed nodes. Node names are optional in the onnx spec.

## (Optional) Issue link
Fixes:
#1459


olive/passes/onnx/model_builder.py
```diff
@@ -197,6 +197,16 @@ def _run_for_config(
...
+                replacements = {
+                    "self_attn": "attn",
+                    "self_attention": "attn",
+                    "dense_4h_to_h": "down_proj",
+                    "dense_h_to_4h": "gate_up_proj",
+                    "dense": "o_proj",
+                    "query_key_value": "qkv_proj",
+                }
+                for old, new in replacements.items():
+                    split_assignment_str = split_assignment_str.replace(f".{old}", f".{new}")
```
devang-ml, line 209: Why do you need to do this replacements?

jambayk: The modules are named differently in the model builder. they standardized the names regardless of the source model.

jambayk: Removed this changed and added a note instead

devang-ml (approved):

PR #1457, opened 2024-11-05
Auto-opt: Add model split options, Fix logic

jambayk:
- Add options for model splitting to the `auto-opt` cli. Only used for no-search mode since evaluating split model is undefined.
- Transformers optimizer pass config now correctly handles fp16 precision.
- MixedPrecision pass is removed since it is not relevant. We normally use the transformers optimizer pass with fp16 or OrtFloat32toFloat16 pass for fp16 conversion. 
- Dml EP does not require static shapes
- `mixed_precision_overrides` is optional and is applied before the quantization pass which uses it. It is not connected to the mixedprecision pass.

### DIFF
--- a/olive/cli/auto_opt.py
+++ b/olive/cli/auto_opt.py
@@ -121,6 +121,21 @@ class AutoOptCommand(BaseOliveCLICommand):
             ),
         )
 
+        # Split options
+        split_group = sub_parser.add_mutually_exclusive_group(required=False)
+        split_group.add_argument(
+            "--num-splits",
+            type=int,
+            help="Number of splits to use for model splitting. Input model must be an HfModel.",
+        )
+        split_group.add_argument(
+            "--cost-model",
+            type=str,
+            help="Path to the cost model file to use for model splitting. Mutually exclusive with num-splits.",
+        )
+        # TODO(jambayk): Move this to device options
+        sub_parser.add_argument("--device-memory", type=int, help="Device memory in bytes to use for model splitting.")
+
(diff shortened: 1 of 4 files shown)


olive/cli/auto_opt.py
```diff
@@ -327,19 +354,23 @@ def _get_passes_config(self, passes_config: Dict[str, Any], olive_config: OliveP
     },
     "passes": OrderedDict(
         [
+            ("capture_split_info", {"type": "CaptureSplitInfo"}),
             ("conversion", {"type": "OnnxConversion"}),
             ("model_builder", {"type": "ModelBuilder", "precision": "fp32", "metadata_only": False}),
-            ("transformer_optimizer", {"type": "OrtTransformersOptimization"}),
+            (
+                "transformer_optimizer",
+                {"type": "OrtTransformersOptimization", "opt_level": 0, "use_fp16": False, "keep_io_types": False},
```
devang-ml, line 362: why not do opt_level 99?

devang-ml (approved):

olive/cli/auto_opt.py
jambayk: opt_level>0 applies ort runtime graph optimizations to the model graph but this is not required since it can be done during session loading. This type of graph optimization has multiple rules such as operator fusion (matmul + add, etc) and renaming nodes that might break downstream passes. So, I turned it off like we do in most of our examples.

jambayk:
Closing. Will open a new PR into main with rebased changed.

Issue #1478, opened 2024-11-11
[FR]: Supply dynamic_shapes in ONNX export when `dynamo=True`


justinchuby:
### Proposal Summary

Supply dynamic_shapes in ONNX export when `dynamo=True`

### What component(s) does this request affect?

- [ ] OliveModels
- [ ] OliveSystems
- [ ] OliveEvaluator
- [ ] Metrics
- [ ] Engine
- [x] Passes
- [ ] Other

PR #1479, opened 2024-11-11
Set to use `dynamo=True` only with PyTorch 2.6 temporarily

justinchuby:
## Describe your changes

For the Olive release, in the ONNX export pass, since when `dynamo=True` the dynamic_shapes argument is not provided properly, we temporarily change the check to require torch 2.6 to enable the new `dynamo=True` logic pass. This way when a user has torch 2.5 Olive will still use the old `dynamo_export` logic and function without errors.

## (Optional) Issue link

- https://github.com/microsoft/Olive/issues/1478


shaahji (approved):
Model exports succeeding.

PR #1482, opened 2024-11-12
Use ONNX Rewriter and IR to simplify the mnb_to_qdq pass

justinchuby:

### DIFF
--- a/olive/passes/onnx/mnb_to_qdq.py
+++ b/olive/passes/onnx/mnb_to_qdq.py
@@ -69,28 +69,49 @@ class MatMulNBitsToQDQ(Pass):
         # 2. Repacking
         ir_model = ir.serde.deserialize_model(model.load_model())
 
-        def mat_mul_n_bits_pattern(op, input_A, qweight, qscales, qzeros, g_idx, bias):
-            return op.MatMulNBits(input_A, qweight, qscales, qzeros, g_idx, bias)
+        def mat_mul_n_bits_pattern(op, input_A, q_weight, q_scales, q_zeros, g_idx, bias):
+            # bias is an optional input
+            return op.MatMulNBits(
+                input_A,
+                q_weight,
+                q_scales,
+                q_zeros,
+                g_idx,
+                bias,
+                _outputs=["mat_mul_n_bits_out"],  # Bind the output to the name "mat_mul_n_bits_out"
+            )
 
         def _is_initializer(context, value: ir.Value) -> bool:
             graph: ir.Graph = context.graph
             return value in graph.initializers.values()
 
-        def mat_mul_n_bits_pattern_check(context, input_A, qweight, qscales, qzeros, g_idx, bias) -> bool:
-            if not _is_initializer(context, qweight):
+        def mat_mul_n_bits_pattern_check(context, *, q_weight, g_idx, mat_mul_n_bits_out: ir.Value, **_) -> bool:
+            if not _is_initializer(context, q_weight):
                 return False
-            node: ir.Node = _get_node(input_A)
-            block_size = node.attributes["block_size"].value
-            K = node.attributes["K"].value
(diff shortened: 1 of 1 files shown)


olive/passes/onnx/mnb_to_qdq.py
```diff
@@ -62,257 +64,124 @@ def _run_for_config(
...
-                )
+            # DequantizeLinear -> Transpose -> MatMul -> Add (optional)
+            dq = op.DequantizeLinear(
+                q_weight,
+                q_scales,
+                q_zeros,
+                block_size=None if is_per_axis else block_size,
+                # for some reason block_wise and per-axis appear to use swapped axis
+                # flip the axis if it is per-axis
+                axis=config["use_transpose_op"] or is_per_axis,
```
jambayk, line 132: still need the ^ to flip the axis. but it could be simplified to `int(config["use_transpose_op"] ^ is_per_axis)`

justinchuby: We can just take a boolean and treat it as an integer, or is config["use_transpose_op"] or is_per_axis not boolean?

jambayk: if it does the cast to int automatically then int is not required. but the operator is still ^ and not or

justinchuby: Oh ^ is xor sorry

justinchuby: Fixed

olive/passes/onnx/mnb_to_qdq.py
```diff
@@ -62,257 +64,125 @@ def _run_for_config(
...
-            node_attributes = dag.get_node_attributes(node_name)
-            K = node_attributes["K"]  # noqa: N806
-            N = node_attributes["N"]  # noqa: N806
-            block_size = node_attributes["block_size"]
-            num_k_blocks = math.ceil(K / block_size)
+        def _is_initializer(context, value: ir.Value) -> bool:
+            graph: ir.Graph = context.graph
+            return value in graph.initializers.values()
+
+        def mat_mul_n_bits_pattern_check(context, *, q_weight, g_idx, mat_mul_n_bits_out: ir.Value, **_) -> bool:
```
jambayk, line 88: does q_weight here match for the input right before g_idx or it is whatever it is in the mat_mul_n_bits_pattern signature? The input begore g_idx is qzero and can be optional.  we want to match the second input,

olive/passes/onnx/mnb_to_qdq.py
```diff
@@ -62,257 +64,125 @@ def _run_for_config(
...
+                node.replace_input_with(3, input_3)
+                ir_model.graph.register_initializer(input_3)
+
+            # Clear the meta data
+            del node.meta["needs_repacking"]
+            del node.meta["K"]
+            del node.meta["N"]
+
+        # TODO(justinchuby): Register and remove initializers
+        ir_model.opset_imports[""] = max(21, ir_model.opset_imports[""])
```
justinchuby, line 183: TODO: Use a more robust version conversion process

olive/passes/onnx/mnb_to_qdq.py
justinchuby: @gramalingam

gramalingam: The inputs of the pattern-function (mat_mul_n_bits_pattern) are bound to values in the graph, and these values are passed in as keyword-arguments to the rewrite function here. So, the order here doesn't really matter, though I usually just copy-paste and use the same argument list for both.

jambayk:
Closing for now since this has become stale is not a priority work item.

jambayk:
closing since this has gone stale. will revisit in the future

PR #1511, opened 2024-12-03
Support `conversion_dtype` in ONNXConversion pass and `dynamic_shapes` on torch.onnx.export(..., dynamo=True)

titaiwangms:


titaiwangms:
@jambayk Could you point me to the necessary files/docs that I need to update in order to introduce `dynamic_shapes`? It's a new mechanism to mark dynamic in torch that onnx exporter has to follow in PyTorch 2.0. Another concern is that different from `dynamic_axes` only needs ints and strings, `dynamic_shapes` needs `torch.export.Dim` which does not work very well with config.json

```python
    (1) dynamic_axes = {"x": {0: "my_custom_axis_name_1"}, "y": {1: "my_custom_axis_name_2"}}
    (2) dynamic_axes = {"x": [0], "y": [1]}

    (1) dynamic_shapes = {"x": {0: torch.export.Dim("my_custom_axis_name_1")}, "y": {1: torch.export.Dim("my_custom_axis_name_2")}}
    (2) dynamic_shapes = {"x": {0: torch.export.Dim("x_dim_0")}, "y": {1: torch.export.Dim("y_dim_1")}}names
```

For testing, the correct file to add the test seems to be: https://github.com/microsoft/Olive/blob/main/test/unit_test/passes/onnx/test_conversion.py. I guess just follow `test_onnx_conversion_with_past_key_values` to create a mock onnx export?

olive/common/utils.py
```diff
@@ -253,6 +255,26 @@ def tensor_data_to_device(data, device: str):
         return data
 
 
+def tensor_data_to_dtype(data, dtype):
+    if dtype is None:
```
jambayk, line 259: can you move the torch import into this function. We wanted to only keep stdlib imports on the top level for this file.

olive/passes/onnx/conversion.py
```diff
@@ -236,8 +238,10 @@ def _export_pytorch_model(
                     input_names=io_config.input_names,
                     output_names=io_config.output_names,
                     dynamic_axes=io_config.dynamic_axes,
+                    dynamic_shapes=io_config.dynamic_shapes,
                     dynamo=True,
                     fallback=True,
+                    report=False,  # suppress the debug report
```
devang-ml, line 244: How about enabling debug report for verbosity level DEBUG?

olive/passes/onnx/conversion.py
```diff
@@ -160,6 +160,7 @@ def _export_pytorch_model(
         device: Union[str, torch.device],
         torch_dtype: Optional[torch.dtype] = None,
         tempdir: Optional[Union[Path, str]] = None,
+        dynamic: bool = True,
```
jambayk, line 163: if you think the option to turn dynamic on and off is useful, I think we can make it into a config option and add it to the `_default_config` at line 50. This way we don't have to depend on model_attributes and it is instead directly provided to the pass config through `config["dynamic"]`

olive/passes/onnx/conversion.py
```diff
@@ -553,3 +562,49 @@ def _run_for_config(
...
+
+    # dynamic_shapes follows input format, which could be nested
+    def _from_tuple_to_dim(data: Union[Dict, List, Tuple, Any]) -> Union[Dict, List, Tuple, Any]:
+        if isinstance(data, dict):
+            for key, value in data.items():
+                data[key] = _from_tuple_to_dim(value)
+        elif isinstance(data, list):
+            for i in range(len(data)):
+                data[i] = _from_tuple_to_dim(data[i])
+        elif isinstance(data, tuple):
```
jambayk, line 589: I think you might have to treat tuple/list as the same here. In yaml and json, there is no tuple data type so it we have to use save and provide tuples using list.

jambayk: does the exporter care whether to original type wrapping the dynamic shapes are tuple or list? That might be a problem if so since we cannot differentiate between them in json/yaml.

titaiwangms: The idea is that the nested `dynamic_shapes` should follow the nested `input`, for example:

```python
        input = (
            torch.ones(5),
            (torch.zeros(5), torch.ones(5)),
            {"a": torch.zeros(5), "b": torch.ones(5)},
            torch.ones(4),
        )
        dim = torch.export.Dim("dim", min=3)
        dynamic_shapes = (
            {0: dim},
            ({0: dim}, {0: dim}),  # torch detects mismatches if we change this to list
            {"a": {0: dim}, "b": {0: dim}},
            None,
        )
```

When `model input` has tuple or list, the `dynamic_shapes` should follow.

titaiwangms:
@jambayk @devang-ml This PR is ready for review.

olive/common/hf/model_io.py
```diff
@@ -124,6 +124,9 @@ def get_model_io_config(model_name: str, task: str, model: "PreTrainedModel", **
         for axis, axis_name in value.items():
             if axis_name == "past_sequence_length + 1":
                 value[axis] = "past_sequence_length + sequence_length"
+    # NOTE: Due to the complexity of dynamic_shapes, we don't provide it here.
```
jambayk, line 127: is the torch exporter able to combine dynamic axes and dynamic shapes? 

In the kv cache config, you add dynamic shapes for the kv cache, but we don't have dynamic shapes for the other inputs since it's not added here. Would that be a valid configuration?

test/unit_test/passes/onnx/test_conversion.py
```diff
@@ -178,3 +180,79 @@ def mock_onnx_export_func(*args, **kwargs):
     p = create_pass_from_dict(OnnxConversion, {}, disable_search=True)
     _ = p.run(input_model, str(output_folder))
     assert "past_key_values" in dummy_inputs  # pylint: disable=unsupported-membership-test
+
```
jambayk, line 183: not directly related to this PR but should we also parameterize dynamo export in `test_onnx_conversion_pass` at line 24?

olive/common/hf/model_io.py
titaiwangms: > is the torch exporter able to combine dynamic axes and dynamic shapes?
> 

Yes, kind of. We have a "do our best" mechanism to convert `dynamic_axes` to `dynamic_shapes` in dynamo exporter for torchscript users to have a quick way of "trying it out" feature that to sell it. But there is no guarantee that the conversion would 100% succeed though. Especially in `dynamic_shapes` we need users to provide specific min/max boundary of each dynamic dim.

> In the kv cache config, you add dynamic shapes for the kv cache, but we don't have dynamic shapes for the other inputs since it's not added here. Would that be a valid configuration?

Good catch on this. I am confused by the logics around `kv_cache` actually. I know there are two automation abouth `kv_cache`. I think one is from Optimum that Olive generates io_config from it if it's not provided, and HF model id is used. And the other one is that when `kv_cache` is set to True, Olive auto-gen `kv_cache` for users. 

My idea was that for the first approach, we don't generate `dynamic_shapes`, but only let dynamo exporter to trigger **from dynamic axes to dynamic shapes**, since Optimum does not have `dynamic_shapes` anyway. For the second one, in order to follow this promise to users (generate `kv_cache` for them), I wrote the conversion here. I didn't notice that could be conflict. Is it that `get_dynamic_shapes()` actually needs an initial version blocks it? If so, that should be fine? I only wanted to support non-HF auto cases. 

however, if HF cases are important, we can figure something out. I just didn't want to crash the export with broken `dynamic_shapes`, which is very likely to be the case, if we are going to generate it from Optimum config, as we need to auto-set up min and max of dynamic dims.

jambayk: the automatic kv cache addition applies to both user-provided io_config and hf model io_config from optimum. So, my concern with regards to the second case since we are missing dynamic shapes for the input_names coming from optimum created io_config. It only has dynamic shapes for the kv cache inputs. 

HF cases are most important for us since most models of interests are some huggingface causal lm models. Hopefully the conversion for dynamic axis to dynamic shape for the optimum produced io_config is simpler since there is no nesting structure for those.

test/unit_test/passes/onnx/test_conversion.py
titaiwangms: Done. But awkwardly, dynamo exporter fails to export hf_model because the nested structure of past_key_values. It also involves mismatches input_names, which kind of says it's impossible to turn Optimum config to dynamo exporter required format in automation. I will bring this to the team, and see how should we proceed this. But the test passes, because we have fallback system from dynamo_exporter to torchscript exporter. As long as the test passes torchscript exporter, it will pass dynamo_exporter.

jambayk: Do you think it's failing because of what we discussed in the other comment https://github.com/microsoft/Olive/pull/1511#discussion_r1879064104?

That part is still unaddressed. For the optimum produced io_config + automatic cache, it is still missing the dynamic shapes for the optimum part. 
If you also created dynamic shapes there instead of just skipping as is now, it might work? We already have the dynamic axes, so the dynamic shapes creation there should be straightforward too.

titaiwangms: It should be just CI torch version. https://github.com/microsoft/Olive/blob/dc6700880d4df013fcfe49f2b6da59c9955ca14f/olive/passes/onnx/conversion.py#L206

Fallback should cover the case.

jambayk: What about locally, does the dynamo exporter test with hf model work without the fallback? Because we definitely want that case to work with these dynamic shapes changes.

.azure_pipelines/job_templates/olive-setup-template.yaml
```diff
@@ -25,3 +25,6 @@ steps:
   - script: |
       pip install ${{ parameters.onnxruntime }}
     displayName: Install ${{ parameters.onnxruntime }}
+
+- script: pip install onnxscript
+  displayName: Install onnxscript
```
jambayk, line 30: since torch export is key olive feature and onnxscript is used in the new torch versions for it, I think we should just add it as a required dependency for olive once it has a stable release? 
For the time being, let's add it to https://github.com/microsoft/Olive/blob/main/test/requirements-test.txt instead?

titaiwangms:
Blocked and waiting for https://github.com/pytorch/pytorch/pull/143158

jambayk (dismissed):
Thanks!

test/unit_test/passes/onnx/test_conversion.py
titaiwangms: It works without fallback with this PR: https://github.com/pytorch/pytorch/pull/143158. 

However, there are some cases it won't work, as `dynamic_axes` creates no constraints, while `dynamic_shapes` examines the models while exporting them. We will fix them one by one when we see the user cases.

olive/common/hf/model_io.py
titaiwangms: kv cache dynamic shape automation has been deleted. We have decided to go with the automation inside exporter. Inl Olive, we can just generate `dynamic_axes` and feed it to exporter.

titaiwangms:
Will merge after https://github.com/pytorch/pytorch/pull/143158 is merged

olive/passes/onnx/conversion.py
```diff
@@ -228,19 +238,35 @@ def _export_pytorch_model(
...
+                        tmp_model_path,  # needed for fallback=True
+                        kwargs=dummy_kwargs,
+                        opset_version=config["target_opset"],
+                        input_names=io_config.input_names,
+                        output_names=io_config.output_names,
+                        dynamic_axes=io_config.dynamic_axes,
+                        dynamic_shapes=io_config.dynamic_shapes,
+                        dynamo=True,
+                        fallback=True,
+                        report=True,
```
titaiwangms, line 266: revert this

titaiwangms:
@jambayk Sorry, could you approve again? It's a minor fix on a parameter.

titaiwangms:
> Will merge after [pytorch/pytorch#143158](https://github.com/pytorch/pytorch/pull/143158) is merged

We are merging this to see how it works on benchmark with static shapes.

devang-ml (approved):

PR #1516, opened 2024-12-05
Add constant folding and onnxoptimizer to peephole optimizer

xiaoyu-work:
## Describe your changes

Add constant folding and remove initializer from input to peephole optimizer.

### DIFF
--- a/olive/passes/onnx/peephole_optimizer.py
+++ b/olive/passes/onnx/peephole_optimizer.py
@@ -307,10 +307,8 @@ class ModelOptimizer:
         return const_nodes
 
     def fold_constant(self):
-        model_copy = copy.deepcopy(self.model)
-
         while True:
-            const_nodes = self._get_constant_nodes(model_copy)
+            const_nodes = self._get_constant_nodes(self.model)
             if not const_nodes:
                 break
 
@@ -322,15 +320,15 @@ class ModelOptimizer:
                     continue
 
                 try:
-                    op_model = self._run_op(model_copy, node)
+                    op_model = self._run_op(self.model, node)
                     outputs = self._run_onnx_model(op_model)
                     for output_array, name in zip(outputs, node.output):
-                        if any(init.name == name for init in model_copy.graph.initializer):
+                        if any(init.name == name for init in self.model.graph.initializer):
                             continue
                         tensor = numpy_helper.from_array(output_array, name)
-                        model_copy.graph.initializer.append(tensor)
+                        self.model.graph.initializer.append(tensor)
                         vi = helper.make_tensor_value_info(name, tensor.data_type, tensor.dims)
-                        model_copy.graph.value_info.append(vi)
+                        self.model.graph.value_info.append(vi)
                     nodes_to_remove.append(node)
                 except Exception as e:
                     logger.warning("Failed to run %s op (name is %s): %s, skip...", node.op_type, node.name, e)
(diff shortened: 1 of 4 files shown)


olive/passes/onnx/peephole_optimizer.py
```diff
@@ -28,6 +31,19 @@ def __init__(self, source_model_path):
         self.source_model_path = str(source_model_path)
         self.model = onnx.load(self.source_model_path)
 
+    def remove_initializer_from_input(self) -> onnx.ModelProto:
```
jambayk, line 34: is there a use case for this transformation. the same name being input and initializer makes the input optional for most ort EPs. this can be intentional. 
For example, extract adapter pass adds initializers to use a default value for optional inputs. 
Maybe this can be moved into the graph surgery as a configurable option?

xiaoyu-work: Discussed offline. Move this to graph surgery pass.

olive/passes/onnx/peephole_optimizer.py
```diff
@@ -232,13 +235,142 @@ def fuse_reshape_operations(self):
...
+        }
+        const_nodes = []
+        for node in model.graph.node:
+            if all(node_input == "" or node_input in const_inputs for node_input in node.input):
+                const_nodes.append(node)
+                const_inputs.update(node.output)
+        return const_nodes
+
+    def fold_constant(self):
+        model_copy = copy.deepcopy(self.model)
```
jambayk, line 310: is there a reason to copy the model? what if the changes are done directly on the original model? self.model is not touched again until assignment with model_copy.

xiaoyu-work: This logic modifies the model by adding new initializers and removing nodes. Using a deep copy ensures that if any error occurs during this process, the original model remains unchanged.

jambayk: Are you worried about the input model being affected by this or something? There is no caching of the model proto like we do for loaded pytorch model so that's not an issue. 

If there is an error here, then the pass will fail. this behavior does not change whether we change the original model proto or a copy of it.

xiaoyu-work: That's fair, i originally thought it has try-catch block so the logic may fail, but the behavior should be same. I'll remove this copy part thanks!

olive/passes/onnx/peephole_optimizer.py
```diff
@@ -232,13 +235,139 @@ def fuse_reshape_operations(self):
...
+            initializer.name for initializer in model.graph.initializer if initializer.name not in dynamic_inputs
+        }
+        const_nodes = []
+        for node in model.graph.node:
+            if all(node_input == "" or node_input in const_inputs for node_input in node.input):
+                const_nodes.append(node)
+                const_inputs.update(node.output)
+        return const_nodes
+
+    def fold_constant(self):
```
jambayk, line 309: Btw, do you know how fast this optimization step takes? 

I am concerned the iterations through the model proto each time to find initializers, etc might be slow for large models. 
The OnnxDAG I implemented is able to create a dag representation of the model so that we can access the graph components directly.

This could be done later since there is already a TODO for this pass https://github.com/microsoft/Olive/blob/dc6700880d4df013fcfe49f2b6da59c9955ca14f/olive/passes/onnx/peephole_optimizer.py#L23 but it would become more efficient if we used the DAG to find the components we want.

jambayk (approved):

olive/passes/onnx/peephole_optimizer.py
xiaoyu-work: I didn't compare these 2, but we can check this when we do this migration. btw I'll remove all TODOs and create tech debt work items for each of them.

PR #1520, opened 2024-12-10
CaptureSplitInfo: Separate split for memory intensive module

jambayk:
## Describe your changes
Some modules such as `Embeddings` are memory intensive, i.e., they consume large amounts of memory but do little compute. For example, Embeddings is just a simple lookup or gather operation but holds a large matrix (`vocab X embedding_dim`). Such modules are probably better off in their own splits and run on a separate device like CPU. 
- `generate-cost-model` now calculates `FLOPs` for each module. Unknown or irrelevant module currently have `0` flops.
- `CaptureSplitInfo` identifies whether a module is memory intensive and creates a new split for it.
- Currently there is no threshold for the `parameters/flops` ratio or `bytes` for a module to be memory intensive since only embedding meets the requirements in relevant models. We can add and tune these thresholds in the future if needed.


olive/passes/pytorch/capture_split_info.py
```diff
@@ -153,28 +153,31 @@ def split_using_cost_model(
         split_idx = 0
         split_bytes = 0
         node_idx = 0
+        mem_intensive = False
         for name, _ in model.load_model(cache_model=False).named_modules():
-            if name not in module_to_bytes or name in modules_to_exclude:
+            if name not in module_to_cost or name in modules_to_exclude or name.lower().endswith("dropout"):
```
devang-ml, line 158: Any reason to skip dropout here?

jambayk: since dropouts are only used in train mode, they don't appear in the exported inference graph. So, I removed it to make the split mappings cleaner.

devang-ml (approved):

PR #1524, opened 2024-12-14
new notebook for optimizing popular SLMs

samuel100:
Added a notebook that allows users to select from a curated list of popular SLMs to optimize.


devang-ml (approved):
Nice!

Issue #1540, opened 2025-01-08
Cannot import user scripts outside the working directory if user_script is relative to config file


tezheng:
**Describe the bug**
Cannot import user scripts outside the working directory if user_script/script_dir is relative to config file.

**To Reproduce**
- Put `workflow.json` and `user_script.py` under `D:/project/`
- `cd D:/`, then `olive run --config ./project/workflow.json`

**Olive config**
```json
  "data_configs": [
    {
      "name": "quant_data_config",
      "user_script": "imagenet.py",
      "dataloader_config": {
        "type": "mobilenet_calibration_reader",
        "data_dir": "data/imagenet_subset/train",
        "batch_size": 1
      }
    }
  ]
```

**Olive logs**
```pwsh
pydantic.v1.error_wrappers.ValidationError: 6 validation errors for RunConfig
data_configs -> __root__
  D:\NPU\NPU-Examples\imagenet.py doesn't exist (type=value_error)
evaluators -> common_evaluator
  metric_data_config not found in {} (type=value_error)
engine
  Invalid evaluators (type=value_error)
passes -> conversion
  Invalid engine (type=value_error)
passes -> qnn_preprocess
  Invalid engine (type=value_error)
passes -> quantization
  Invalid engine (type=value_error)
```

**Other information**
 - OS: Any
 - Olive version: 0.7.1.1
 - ONNXRuntime package and version: onnxruntime-qnn==1.20.0
 - Transformers package version: 4.47.1

**Additional context**
Specifying `script_dir` does not help as well (which will be insert into sys.path) as existence checking of `script_dir` is done with `Path.resolve`

jambayk:
Hi, thanks for opening this issue. 

As mentioned in the linked PR, the path to user_script is relative to where the olive command is called from. This also applies to any path that is provided in the config such as `data_dir`. In order to make this work, the config needs to be updated to have `"user_script": "project/imagenet.py"`

PR #1542, opened 2025-01-08
Fix: User script import now respects `script_dir`

tezheng:
Prefer `find_spec` for user script imports; fallback to `spec_from_file_location` if not found.

## Describe your changes

The original implementation of `import_module_from_file` incorrectly
checked for the existence of user scripts in the current working
directory (`cwd`), regardless of the `script_dir` argument provided in `import_user_module`.

This PR corrects this behavior by implementing the following logic:

1.  Attempt to locate the module using `importlib.util.find_spec`.
2.  Fallback to using `importlib.util.spec_from_file_location` if `find_spec` fails

## (Optional) Issue link

https://github.com/microsoft/Olive/issues/1540

### DIFF
--- a/olive/common/import_lib.py
+++ b/olive/common/import_lib.py
@@ -11,10 +11,7 @@ from typing import Optional, Union
 
 @functools.lru_cache
 def import_module_from_file(module_path: Union[Path, str], module_name: Optional[str] = None):
-    module_path = Path(module_path).resolve()
-    if not module_path.exists():
-        raise ValueError(f"{module_path} doesn't exist")
-
+    module_path = Path(module_path)
     if module_name is None:
         if module_path.is_dir():
             module_name = module_path.name
@@ -24,7 +21,18 @@ def import_module_from_file(module_path: Union[Path, str], module_name: Optional
         else:
             module_name = module_path.stem
 
-    spec = importlib.util.spec_from_file_location(module_name, module_path)
+    # Try to find the module in sys.path
+    spec = importlib.util.find_spec(module_name)
+    if not spec:
+        # If not found, try to load the module from the file
+        module_path = module_path.resolve()
+        if not module_path.exists():
+            raise ValueError(f"{module_path} doesn't exist")
+
+        spec = importlib.util.spec_from_file_location(module_name, module_path)
+        if not spec:
+            raise ValueError(f"Could not load module at {module_path}")
+
     new_module = importlib.util.module_from_spec(spec)
     spec.loader.exec_module(new_module)
     return new_module
--- a/test/unit_test/common/test_import_lib.py
+++ b/test/unit_test/common/test_import_lib.py
@@ -5,6 +5,7 @@
 import os
 import shutil
 from pathlib import Path
+from tempfile import TemporaryDirectory
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


tezheng:
@microsoft-github-policy-service agree company="Microsoft"

jambayk:
could you provide an example scenario where the current implementation doesn't work? 

It was a design decision that the `user_script` location is independent of the `script_dir` location so the path to user_script must be absolute or relative to current dir for it to get resolved correctly. Our examples might have the `user_script` inside `script_dir` but it is not a requirement. 
The only relation between `script_dir` and `user_script` is that since `script_dir` is added to `sys.path`, `user_script` can import modules under `script_dir` directly.

tezheng:
> could you provide an example scenario where the current implementation doesn't work?

See below

> It was a design decision that the `user_script` location is independent of the `script_dir` location so the path to user_script must be absolute or relative to current dir (i.e. wherever the olive command is called from, not the config.json's parent dir) for it to get resolved correctly.

By "design" you mean this doc, right? [Custom Script](https://microsoft.github.io/Olive/extending/custom-scripts.html)

> Our examples might have the `user_script` inside `script_dir` but it is not a requirement. The only relation between `script_dir` and `user_script` is that since `script_dir` is added to `sys.path`, `user_script` can import modules under `script_dir` directly.

This is the origin implementation
```python
def import_module_from_file(module_path: Union[Path, str], module_name: Optional[str] = None):
    module_path = Path(module_path).resolve()
    if not module_path.exists():
        raise ValueError(f"{module_path} doesn't exist")
    ...
```
It seems this implementation specifically requires module_path (aka user_script) either an absolute path or locates in `cwd`, regardless of `script_dir`.

For example, a project structured like
```
- my_project
  - my_modules
    - file.py
  - model_a
    - workflow.json
```

workflow.json
```json
{
  "user_script": "file.py",
  "script_dir": "my_modules",
}
```

Executing below commands will cause `ValueError(f"{module_path} doesn't exist")`, since `cwd` is `my_project` but `file.py` isn't in the project folder.
```
cd my_project
olive run --config ./model_a/workflow.json
```

tezheng:
What's more, IMHO, I think maybe `Path(workflow.json).parent` should be always insert into `sys.path`, since `user_script` normally will be shipped alongside `workflow.json`, and its path is usually relative to `workflow.json`.

Do you think this makes any sense? @jambayk

olive/common/import_lib.py
jambayk, line 1: @tezheng commenting on this file so we can have a thread. 

Your proposal for allowing user_script paths relative to the directory containing the config json makes sense. It would help a lot with the portability of olive projects. 
However, apart from user_scripts, we also have many other resources like data dirs, script dirs, etc that have the same issue. Workflows can also be run on remote systems such as docker and azureml, both of which require resolvable paths for mounting. Due to this, we originally decided on paths that are absolute or relative to current dir so that every resource is independent of one another. 

After a discussion with the team, we decided that allowing paths to be relative to the working directory (directory containing config json) should also be supported. I have created a work item for this, and we can discuss offline on teams with you about the priority of this feature. This would require a more involved update of the entire codebase so that all resources have the same behavior (+ not have user script be a special case) and don't break compatibility with remote systems. 

FYI this is [how we infer resource types](https://github.com/microsoft/Olive/blob/dcc0cb9905e521db109d18e071db9955df7aaa4d/olive/resource_path.py#L122). This would need to be updated to check for paths relative to the working dir. There are also other `Path.resolve()` instance that we would need to update accordingly.

PR #1550, opened 2025-01-13
Set onnxoptimizer.optimize as default for peepholeoptimizer

xiaoyu-work:
## Describe your changes

Set onnxoptimizer.optimize as default for peepholeoptimizer

### DIFF
--- a/olive/passes/onnx/peephole_optimizer.py
+++ b/olive/passes/onnx/peephole_optimizer.py
@@ -4,7 +4,7 @@
 # --------------------------------------------------------------------------
 import logging
 from pathlib import Path
-from typing import Any, Dict, List
+from typing import Any, Dict
 
 import numpy as np
 import onnx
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/peephole_optimizer.py
```diff
@@ -277,16 +271,7 @@ def _run_for_config(
         peephole_optimizer.patch_unsupported_argmax_operator()
         peephole_optimizer.fuse_reshape_operations()
         peephole_optimizer.onnxscript_optimize()
```
jambayk, line 273: btw, should we move this to be the first optimization that is applied? I saw some discussion that this is needed make dynamo exported models similar to torchscript exported models. Otherwise, the earlier optimizations might fail since the are not compatible with dynamo exported models.

xiaoyu-work: Sure I moved onnxscript optimize to be the first one here. But I'm a little bit confused that why other optimizations is not compatible with dynamo exported models?

jambayk: dynamo exported models has the nodes stored in `model_proto.graph.functions` + node naming is different (there might be more difference) but the other optimizations don't deal with model functions. they just expect a simple model proto where the nodes are in `model_proto.graph.node`.

olive/passes/onnx/peephole_optimizer.py
```diff
@@ -239,32 +239,26 @@ def onnxscript_optimize(self):
...
 
-        onnxscript.optimizer.optimize(self.model)
+        self.model = onnxscript.optimizer.optimize(self.model)
+
+    def onnxoptimizer_optimize(self):
+        try:
+            from onnxoptimizer import optimize
+        except ImportError:
+            logger.warning(
+                "Please install onnxoptimizer to use the ONNX optimizer feature. Skip onnxoptimizer optimization."
```
devang-ml, line 249: The user does not know what ONNX optimizer feature is.

Just say something along the line install onnxoptimizer to enable further optimizations.

xiaoyu-work: Done!

PR #1553, opened 2025-01-15
Validate pass config before instantiating the pass

shaahji:
## Validate pass config before instantiating the pass

* Use of search point should be limited to engine logic only. Rest of the Olive implementation should receive a validated configuration to use.
* Validation is for complete configuration and not merely for a search point.
* Fixed a few issues related to use of BasePassConfig vs. FullPassConfig.
* Add local caching to OlivePackageConfig for loaded modules.
* Renamed a few variables in engine logic to be explicit about use of pass config vs. pass run config.

### DIFF
--- a/olive/engine/engine.py
+++ b/olive/engine/engine.py
@@ -26,6 +26,7 @@ from olive.hardware import AcceleratorSpec
 from olive.logging import enable_filelog
 from olive.model import ModelConfig
 from olive.package_config import OlivePackageConfig
+from olive.strategy.search_parameter import SearchParameter
 from olive.strategy.search_strategy import SearchStrategy, SearchStrategyConfig
 from olive.systems.common import SystemType
 from olive.systems.system_config import SystemConfig
@@ -52,7 +53,7 @@ class Engine:
         search_strategy: Optional[Union[Dict[str, Any], SearchStrategyConfig]] = None,
         host: Optional[Union[Dict[str, Any], "SystemConfig"]] = None,
         target: Optional[Union[Dict[str, Any], "SystemConfig"]] = None,
-        evaluator: Optional[Union[Dict[str, Any], "OliveEvaluatorConfig"]] = None,
+        evaluator: Optional[Union[Dict[str, Any], OliveEvaluatorConfig]] = None,
         cache_config: Optional[Union[Dict[str, Any], CacheConfig]] = None,
         plot_pareto_frontier: bool = False,
         no_artifacts: bool = False,
@@ -83,14 +84,9 @@ class Engine:
         self.skip_saving_artifacts = no_artifacts
         self.azureml_client_config = azureml_client_config
 
-        # dictionary of passes
-        self.pass_config = OrderedDict()
-
-        # {"pass_name": {"pass": pass, "host": host, "evaluator": evaluator}
-        self.passes = OrderedDict()
-        self.pass_flows = None
-        self.pass_flows_search_spaces = None
-
+        self.pass_run_configs: Dict[str, Dict[str, Any]] = OrderedDict()
+        self.pass_flows: List[List[str]] = []
+        self.search_spaces: List[List[Tuple[str, Dict[str, SearchParameter]]]] = []
         self.footprints = defaultdict(Footprint)
 
         self._initialized = False
@@ -111,15 +107,17 @@ class Engine:
         if self.target_config.type != SystemType.AzureML:
             if self.evaluator_config:
                 self.evaluator_config = self.cache.prepare_resources_for_local(self.evaluator_config)
-            for pass_config in self.pass_config.values():
-                if pass_config["evaluator"]:
-                    pass_config["evaluator"] = self.cache.prepare_resources_for_local(pass_config["evaluator"])
 
-        for pass_config in self.pass_config.values():
-            host_type = pass_config["host"].system_type if pass_config["host"] else self.host_config.type
-            if host_type == SystemType.AzureML:
-                continue
-            pass_config["config"] = self.cache.prepare_resources_for_local(pass_config["config"])
+            for pass_run_config in self.pass_run_configs.values():
+                if pass_run_config["evaluator"]:
+                    pass_run_config["evaluator"] = self.cache.prepare_resources_for_local(pass_run_config["evaluator"])
+
+        for pass_run_config in self.pass_run_configs.values():
+            host_type = pass_run_config["host"].system_type if pass_run_config["host"] else self.host_config.type
+            if host_type != SystemType.AzureML:
+                pass_run_config["input_config"] = self.cache.prepare_resources_for_local(
+                    pass_run_config["input_config"]
+                )
 
         self.set_pass_flows(self.pass_flows)
         self._initialized = True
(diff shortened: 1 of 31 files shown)


olive/workflows/run/config.py
```diff
@@ -270,6 +270,10 @@ def validate_pass_search(cls, v, values):
             )
         return v
 
+    @validator("pass_flows", pre=True)
+    def validate_pass_flows(cls, v, values):
```
jambayk, line 274: nit: this can also be achieved by using `default_factory=list` at line 143 similar to 135

olive/workflows/run/run.py
```diff
@@ -196,66 +189,32 @@ def run_engine(package_config: OlivePackageConfig, run_config: RunConfig):
         engine.target_config, skip_supported_eps_check=target_not_used, is_ep_required=is_ep_required
     )
 
-    pass_list = []
-    acc_list = []
-    if auto_optimizer_enabled:
```
jambayk, line 201: is this case deprecated now? When auto optimizer is enabled, run_config.pass_config is empty and AutoOptimizer generates them. 
Are we only going to do auto-opt through the cli and remove the built in one?

jambayk: if so, we would need to first fully remove all auto_optimizer related code and examples. otherwise, this case is broken

olive/engine/engine.py
```diff
@@ -718,35 +672,38 @@ def _run_passes(
...
+        pass_type_name = pass_run_config["type"]
+
+        logger.info("Running pass %s:%s %s", pass_name, pass_type_name, pass_search_point)
 
         # check whether the config is valid
-        if not p.validate_search_point(pass_search_point, accelerator_spec, with_fixed_value=True):
-            logger.warning("Invalid search point, prune")
-            output_model_config = INVALID_CONFIG
+        pass_cls: Type[Pass] = self.olive_config.import_pass_module(pass_run_config["type"])
+        pass_config = pass_cls.config_at_search_point(pass_search_point, accelerator_spec, pass_run_config["config"])
```
jambayk, line 690: I think `config_at_search_point` is not required anymore since it is equivalent to 
```
pass_config = {**pass_run_config["config"], **pass_search_point}. 
```

`pass_run_config["config"]` already has all config params - fixed and search. pass_search_point has the same keys as the search params.
This avoids the redundant, potentially expensive calls inside the method.

olive/passes/olive_pass.py
```diff
@@ -189,26 +184,41 @@ def default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConf
...
+        config = validate_config(config, config_class)
+        config = cls._resolve_config(config, default_config)
+        fixed_values, search_params = cls._init_fixed_and_search_params(config, default_config)
+        assert set(point.keys()) == set(search_params.keys()), "Search point is not in the search space."
+        return {**fixed_values, **search_params, **point}
 
-    def validate_search_point(
-        self, search_point: Dict[str, Any], accelerator_spec: AcceleratorSpec, with_fixed_value: bool = False
+    @classmethod
+    def validate_config(
```
jambayk, line 211: maybe we can simplify the signature to only include config and accelerator spec. and make the assumption that config is always a full config. 
It is only called by engine with `with_fixed_value=True` currently and with the new changes, the engine always has the full config.

olive/passes/onnx/model_builder.py
```diff
@@ -92,15 +92,21 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
...
+        accelerator_spec: AcceleratorSpec,
+        disable_search: Optional[bool] = False,
     ) -> bool:
-        if with_fixed_value:
-            search_point = self.config_at_search_point(search_point or {})
-        precision = search_point.get("precision")
+        if not super().validate_config(config, accelerator_spec, disable_search):
+            return False
+
+        config_cls, _ = cls.get_config_class(accelerator_spec, disable_search)
```
jambayk, line 105: I think this is unnecessary and expensive? especially if we go with the assumption I suggested in olive_pass.py? we can always access config as a dictionary with all parameters present.

olive/engine/engine.py
```diff
@@ -718,35 +672,38 @@ def _run_passes(
...
+            logger.warning("Invalid config, pruned.")
+            logger.debug(pass_config)
             # no need to record in footprint since there was no run and thus no valid/failed model
             # invalid configs are also not cached since the same config can be valid for other accelerator specs
             # a pass can be accelerator agnostic but still have accelerator specific invalid configs
             # this helps reusing cached models for different accelerator specs
-            return output_model_config, None
+            return INVALID_CONFIG, None
+
+        p: Pass = pass_cls(accelerator_spec, pass_config, self.get_host_device())
```
jambayk, line 700: I am not sure about instantiating the pass every time in the long run. with these changes, the pass is somewhere between being stateless and stateful + the costs of instantiating the pass each time.  

If the end goal is to make it fully stateless, then I don't mind it for the time being. There are somethings needed to make Pass stateless in the future:
- host_device: this can be removed since it's a concept 
- path_params: only used by docker system but this concept is also outdated. Docker system can be updated to automatically infer paths like aml system
- _user_module_loader: the pass run can create the user module loader itself
- _config_class: generate_config can return the config object holding the params (both fixed and search). This way we always have access to the config class to serialize the pass config.

olive/workflows/run/config.py
shaahji: This was a temporary change and is removed in follow up PR #1562

olive/workflows/run/run.py
shaahji: > Are we only going to do auto-opt through the cli and remove the built in one?

Temporarily, yes. A new implementation of auto-optimizer will be introduced soon. @devang-ml okay-ed it being broken for the time being.

olive/engine/engine.py
shaahji: This was a temporary change and is removed in follow up PR https://github.com/microsoft/Olive/pull/1562

olive/passes/onnx/model_builder.py
shaahji: I was moving away from accessing the config as a dictionary and instead use the it as a class object. Code is more readable and browsable with intellisense. The expensive part will be optimized away in follow up PR where
* remove dependency on accelerator_spec when building config class
* config class will be a class member and will not be created over and over again
* logic in default config that is dependent on accelerator spec will use search parameter

olive/engine/engine.py
shaahji: Ideally, the pass shouldn't even be instantiated in engine at all., It should be done in System::run. Except for the case of local, all other system implementation turns the pass back into a config to write on the wire. So, instead of passing the Pass object to the system, it should use the config as the argument. That would still mean we instantiate the pass for each pass run but that would be limited to local run only.

@devang-ml can chime in on this.

jambayk (approved):

jambayk:
As discussed offline, we will do the code improvements related to config class creation in follow up PRs.

PR #1562, opened 2025-01-21
Implement pass search

shaahji:
## Implement pass search

Reimplement search logic to include passes in search space.


test/unit_test/search/test_search_strategy.py
```diff
@@ -0,0 +1,844 @@
...
+                    ('PassB', Categorical([
+                            SearchSpace([('PB0_SP1', Categorical(['B01a', 'B01b']))], 2),
+                            SearchSpace([], 1)
+                        ])
+                    )], 3),
+                SearchSpace([('PassC', Categorical([SearchSpace([], 1), SearchSpace([], 1)]))], 2)
+            ]""",
+            )
+
+        assert actual == expected
```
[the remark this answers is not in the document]
shaahji: Fixed.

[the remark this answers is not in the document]
shaahji: Fixed.

olive/search/search_results.py
```diff
@@ -0,0 +1,125 @@
...
+
+        _, model_ids, _ = self._results[self._sorted_indicies[next_best_index]]
+        return next_best_index, self._sorted_indicies[next_best_index], model_ids
+
+    def _get_results_list(self, apply_goals: bool = False) -> Tuple[List[int], List[float]]:
+        """Return the results as a tuple of indicies and values.
+
+        Values are multiplied by the objective multiplier so that higher is better for all objectives.
+        """
+        # TODO(olivedev): Can same named objective have different values for higher_is_better?
```
xiaoyu-work, line 78: i don't think this is true

examples/test/local/test_bert_cuda_gpu.py
```diff
@@ -16,14 +16,14 @@ def setup():
 
 
 @pytest.mark.skip(reason="Disable failing tests")
-@pytest.mark.parametrize("search_algorithm", ["tpe"])
+@pytest.mark.parametrize("sampler", ["tpe"])
 @pytest.mark.parametrize("execution_order", ["joint", "pass-by-pass"])
 @pytest.mark.parametrize("system", ["local_system"])
 @pytest.mark.parametrize("olive_json", ["bert_cuda_gpu.json"])
```
xiaoyu-work, line 22: This test will fail if you updated the config file name. We can use `bert.py` you added for this test.

xiaoyu-work:
Can you also update documents related to your changes?

olive/search/search_parameter.py
```diff
@@ -0,0 +1,310 @@
...
+        default: SearchParameter = None,
+    ):
+        assert isinstance(parents, tuple), "parents must be a tuple"
+        for key in support:
+            assert isinstance(key, tuple), "support key must be a tuple"
+            assert len(key) == len(parents), "support key length must match the number of parents"
+
+        self.parents = parents
+        self.support = support
+        self.default = default or self.get_invalid_choice()
```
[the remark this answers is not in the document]
shaahji: Updated documentation for the class.

olive/search/search_parameter.py
```diff
@@ -0,0 +1,310 @@
...
+        """Return a categorical search parameter with the invalid choice."""
+        return Categorical([SpecialParamValue.INVALID])
+
+    @staticmethod
+    def get_ignored_choice():
+        """Return a categorical search parameter with the ignored choice."""
+        return Categorical([SpecialParamValue.IGNORED])
+
+
+class ConditionalDefault(Conditional):
```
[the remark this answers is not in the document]
shaahji: Yes! BTW, search parameter implementation is not new. This is carry forward from old implementation under search strategy.

olive/search/search_strategy.py
```diff
@@ -0,0 +1,315 @@
...
+    def _create_results(self) -> SearchResults:
+        """Create and return a search result."""
+        return SearchResults()
+
+    def _initialize_step(self) -> bool:
+        state = self._state[tuple(self._path)]
+
+        # TODO(shaahji): Two scenarios for pass-by-pass mode -
+        #   1. All search points are evaluated for each search space before moving down the tree.
+        #   2. Evaluting search points until we find a suitable candidate and move down.
```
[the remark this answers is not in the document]
shaahji: Not at the moment but it's an option so documenting it.

examples/test/local/test_bert_cuda_gpu.py
shaahji: Fixed

shaahji:
> Can you also update documents related to your changes?

Done!

examples/__init__.py
jambayk, line 1: why is this required?

examples/bert/bert.py
jambayk, line 1: please update the "BERT optimization with CUDA/TensorRT on GPU" section in the readme

examples/llama2/llama2.py
```diff
@@ -173,17 +173,26 @@ def get_general_config(args):
     gqa = "gqa" if args.use_gqa else "mha"
     config_name = f"llama2_{device}_{gqa}"
 
-    # add pass flows
+    # add pass names
     if not args.use_gptq:
-        template_json["pass_flows"] = [flow for flow in SUPPORTED_WORKFLOWS[device] if "gptq" not in flow[0]]
+        used_passes = [
```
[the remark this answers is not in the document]
jambayk: we need to either make the mapping specific or generate multiple workflows and run the separately.

examples/phi2/phi2.py
```diff
@@ -186,25 +186,23 @@ def main(raw_args=None):
             legacy_optimization_setting(template_json)
 
         # add pass flows
-        pass_flows = [[]]
+        used_passes = {}
```
jambayk, line 189: Just in general, I think using sets might be a bit risky since they are unordered.

examples/phi3/phi3.py
```diff
@@ -191,7 +191,10 @@ def use_passes(template_json, *passes):
     else:
         del template_json["data_configs"]
 
-    template_json["pass_flows"] = [passes]
+    for pass_name in set(template_json["passes"].keys()):
```
jambayk, line 194: I think instead of popping the unused passes, it might be better to create a new dict with the used passes. Popping assumes the order in `template_json["passes"]` is the same as in `passes`

examples/stable_diffusion/sd_utils/ort.py
```diff
@@ -23,7 +23,10 @@ def update_cuda_config(config_cuda: Dict):
     if version.parse(OrtVersion) < version.parse("1.17.0"):
         # disable skip_group_norm fusion since there is a shape inference bug which leads to invalid models
         config_cuda["passes"]["optimize_cuda"]["optimization_options"] = {"enable_skip_group_norm": False}
-    config_cuda["pass_flows"] = [["convert", "optimize_cuda"]]
+    used_passes = {"convert", "optimize_cuda"}
```
jambayk, line 26: this part in stable_diffusion.py also needs to be updated to only use `"convert", "optimize"`

olive/search/samplers/optuna_sampler.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(
+        self,
+        search_space: SearchSpace,
+        config: Optional[Union[Dict[str, Any], ConfigBase]] = None,
+    ):
+        super().__init__(search_space, config)
+
+        # Initialize the searcher
+        self._sampler = self._create_sampler()
+        # TODO(olivedev): There is no absolute direction to set.
```
jambayk, line 47: will this be investigated later? Since I still think we need directions for the signals to make sense and the sampler to choose the next option.

olive/search/samplers/optuna_sampler.py
```diff
@@ -0,0 +1,140 @@
...
+    @abstractmethod
+    def _create_sampler(self) -> optuna.samplers.BaseSampler:
+        """Create the sampler."""
+
+    def suggest(self) -> SearchPoint:
+        """Suggest a new configuration to try."""
+        if self.should_stop:
+            return None
+
+        while True:
```
jambayk, line 68: is this to avoid sampling the same point? If so, could you add a comment about it? 
does it do the same even with the current implementation of optuna sampler?

olive/search/search_strategy.py
```diff
@@ -0,0 +1,318 @@
...
+        if "execution_order" not in values:
+            raise ValueError("Invalid execution_order")
+        if v and values["execution_order"] != SearchStrategyExecutionOrder.JOINT:
+            logger.info("%s is only supported for joint execution order. Ignoring...", field.name)
+            return field.default
+        return v
+
+
+@dataclass
+class SearchWalkState:
```
jambayk, line 72: could you add some docstrings to describe what each class does/is used for? thanks!

olive/search/search_strategy.py
```diff
@@ -0,0 +1,318 @@
...
+    def __init__(self, config: Union[Dict[str, Any], SearchStrategyConfig]):
+        self.config: SearchStrategyConfig = validate_config(config, SearchStrategyConfig)
+
+        # Initialization variables
+        self._search_spaces: List[SearchSpace] = None
+        self._objectives: Dict[str, Dict[str, Any]] = None
+        self._init_model_id: str = None
+
+        # State variables
+        self._path: List[int] = None
```
jambayk, line 91: SearchWalkState also has a path attribute

olive/search/samplers/optuna_sampler.py
```diff
@@ -0,0 +1,140 @@
...
+                suggestions = param.get_support_with_args(parent_values)
+                max_length = max(len(support.get_support()) for support in param.support.values())
+                suggestions += param.default.get_support() * (max_length - len(suggestions))
+                suggestion_name += "___" + " ".join(str(v) for v in parent_values.values())
+
+            suggestion_len = len(suggestions)
+            suggestion_index = trial.suggest_categorical(suggestion_name, list(range(suggestion_len)))
+            suggestion = suggestions[suggestion_index]
+
+            if isinstance(suggestion, (SearchParameter, SearchSpace)):
```
jambayk, line 103: does this case happen? I thought this the if block at line 87 is the base case. so it should only result in a fixed value?

olive/search/samplers/optuna_sampler.py
```diff
@@ -0,0 +1,140 @@
...
+            for child_name, child_param in param.parameters:
+                child_index, suggestions_len, _ = self._get_search_point_values(
+                    prefix, child_name, child_param, trial, child_values
+                )
+                indicies_lengths.append((child_index, suggestions_len))
+            values[name] = (0, child_values)
+
+            spi = 0
+            for child_index, suggestions_len in reversed(indicies_lengths):
+                spi *= suggestions_len
```
jambayk, line 124: does this guarantee uniqueness for the search point index? behaves like a generic version of binary (some base_n) encoding?

olive/search/samplers/search_sampler.py
```diff
@@ -0,0 +1,69 @@
...
+                type_=int,
+                default_value=0,
+                description="Maximum number of samples to suggest. Search exhaustively if set to zero.",
+            ),
+        }
+
+    def __init__(
+        self,
+        search_space: SearchSpace,
+        config: Optional[Union[Dict[str, Any], ConfigBase]] = None,
```
jambayk, line 35: same comment related to optuna sampler. I think the sampler in general needs information about what metrics to expect and their directions to make informed decisions.

jambayk: nevermind, I saw objective is passed as part of `record_feedback_signal`. We can continue the discussion related to optuna later.

olive/search/search_space.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def get_sample_point_values(self, index: int) -> Dict[str, Tuple[int, Any]]:
+        assert index < self._length
+
+        values = OrderedDict()
+        for name, param in self._parameters:
+            index, values[name] = SearchSpace.get_suggestion(param, index, values)
+        return values
+
+    @staticmethod
```
jambayk, line 59: could you add some docstrings and comments to describe the functionalities and logic. It's a bit hard to follow

examples/__init__.py
shaahji: Without this file, VSCode doesn't find the tests in examples folders. Adding this file allows debugging of the tests in examples folder from within VSCode.

examples/phi2/phi2.py
shaahji: There are two ways to do it -
* Assume the order in the config is correct, in which case use of set is warranted because you will likely delete entries from config/passes.
* Code is providing the order in which to run the passes, in which case I am explicitly using an OrderedDict.

Both the strategies are used across the codebase. In some cases, the first is more meaningful in others it's the second. This specific use case here is the option 1 above because down below on line 218 we are explicitly iterating the config/passes and deleting the ones not used (the order of which ones are not being used doesn't matter).

examples/phi3/phi3.py
shaahji: There is an assumption one way or the other. Either we assume that the order in the template is correct or the order in the code is correct. I don't think there is a right or wrong way to do this. :)

examples/stable_diffusion/sd_utils/ort.py
shaahji: Done!

olive/search/samplers/optuna_sampler.py
shaahji: Yes! I will follow up with another PR to address the issue of objectives/goals specifically.

olive/search/samplers/optuna_sampler.py
shaahji: Documented.

olive/search/search_strategy.py
shaahji: Done!

olive/search/search_strategy.py
shaahji: Yes! `SearchWalkState.path` uniquely identifies the state (unused in code but good for debugging) and `SearchStrategy.path` represents the current active state.

olive/search/samplers/optuna_sampler.py
shaahji: `SearchSpace` is hierarchical so each choice can potentially be a `SearchParameter` or a `SearchSpace`. For example, in _pass-by-pass_ mode, each pass level `SearchSpace` contains `Categorical` with each choice being a `SearchSpace` built using `SearchParameter` for the corresponding pass.

For _pass-by-pass_ mode, [SearchSpace[Categorical[SearchSpace], Categorical[SearchSpace], ...], ...]
For _joint_ mode, SearchSpace[Categorical[SearchSpace], Categorical[SearchSpace], ...]

test_search_strategy.py::TestSearchStrategy::test_initialize is a good example to showcase the differences in search space across the two modes.

olive/search/samplers/optuna_sampler.py
shaahji: Yes, it is unique and there's a test specifically to test for this. See test_tpe_sampler.py::TestTPESampler::test_suggest

olive/search/search_space.py
shaahji: Done!

examples/bert/bert.py
shaahji: Done!

examples/llama2/llama2.py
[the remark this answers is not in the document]
shaahji: Done! Added CLI parameter "precision" and "quantize".

Will update the CI tests accordingly.

xiaoyu-work:
Without pass flow, how can we handle the pass search? Like pass: [[A,C],[B,C]]

.lintrunner.toml
```diff
@@ -117,37 +117,6 @@ init_command = [
   '--requirement=requirements-dev.txt'
 ]
 
-[[linter]]
```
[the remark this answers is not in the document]
shaahji: Had a conversation with @jambayk earlier on this. MYPY is slow and does only static type checking.
Given the number of external dependencies for Olive it tends to break in unpredictable ways. Introduction of __init__.py in examples folder generated a number of errors for which there's no clear solution. Given the number of other linters we use, this isn't very useful.

docs/architecture.md
```diff
@@ -139,18 +139,18 @@ The corresponding conceptual search space is the space of all possible parameter
 {“pass_id/space_name”: {“param_name”: param_value}}
 ```
 
-### SearchAlgorithm
```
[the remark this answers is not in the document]
shaahji: Done!

docs/source/how-to/configure-workflows/auto-opt.md
```diff
@@ -41,11 +41,9 @@ Here is a simple example of Auto Optimizer configuration, the item which is not
         "engine": {
```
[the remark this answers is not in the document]
shaahji: Not sure what you mean by flattened. `search_algorithm_config` is already removed in the change. I don't think this can be further reduced.

olive/search/search_space.py
```diff
@@ -0,0 +1,230 @@
...
+    Also, each intermediate node in the SearchSpace could also either be a SearchParameter
+    or a SearchSpace, generating its own possible choices.
+
+    Note that the length in this context is the same as the number of possible unique choices.
+
+    A "index" represent the index across all of the possible permutations of the leaf nodes.
+    An result of indexing is deterministic only if the length of all the intermediate nodes
+    in the tree are constant.
+
+    Catch: Conditional search paramter don't have a constant length since the length depend on
```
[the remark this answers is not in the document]
shaahji: Fixed this and many spelling errors.

olive/search/search_strategy.py
```diff
@@ -0,0 +1,345 @@
...
+                ):
+                    return None
+
+                # Check is any of the global stop criteria are met.
+                # NOTE: Search will run at least one step before stopping.
+                if self.should_stop:
+                    return None
+
+            # Try stepping down the tree, and if that fails, try stepping up. If both fails, we are done
+            if not self._step_down() and not self._step_up():
```
jambayk, line 252: this is a nice idea!
Just to be sure, in the pass-by-pass order, if the search doesn't stop when we reach the last pass, it tries again with the next best sample from the previous pass? 
What if we exhaust all options from the previous pass? I don't think it goes another step back? 

Not an issue either way since this behavior is already more than what we did previously.

jambayk (approved):
Thanks!

olive/search/search_strategy.py
shaahji: The logic always goes step forward first and then tries step back so it does walk back up if the tree if the current level fails. I will add a more elaborate test case to verify the logic.

PR #1571, opened 2025-01-24
Fix AutoAWQQuantizer GptqQuantizer supported ep and device

xiaoyu-work:
## Describe your changes

Fix AutoAWQQuantizer GptqQuantizer supported ep and device.

### DIFF
--- a/olive/olive_config.json
+++ b/olive/olive_config.json
@@ -214,29 +214,29 @@
         },
         "OpenVINOConversion": {
             "module_path": "olive.passes.openvino.conversion.OpenVINOConversion",
-            "supported_providers": [ "OpenVINOExecutionProvider" ],
-            "supported_accelerators": [ "cpu", "gpu", "npu" ],
+            "supported_providers": [ "*" ],
+            "supported_accelerators": [ "*" ],
             "supported_precisions": [ "*" ],
             "extra_dependencies": [ "openvino" ]
         },
         "OpenVINOQuantization": {
             "module_path": "olive.passes.openvino.quantization.OpenVINOQuantization",
-            "supported_providers": [ "OpenVINOExecutionProvider" ],
-            "supported_accelerators": [ "cpu", "gpu", "npu" ],
+            "supported_providers": [ "*" ],
+            "supported_accelerators": [ "*" ],
             "supported_precisions": [ "*" ],
             "extra_dependencies": [ "openvino" ]
         },
         "AutoAWQQuantizer": {
             "module_path": "olive.passes.pytorch.autoawq.AutoAWQQuantizer",
-            "supported_providers": [ "CPUExecutionProvider" ],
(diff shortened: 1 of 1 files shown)


olive/olive_config.json
```diff
@@ -228,15 +228,15 @@
         },
         "AutoAWQQuantizer": {
             "module_path": "olive.passes.pytorch.autoawq.AutoAWQQuantizer",
-            "supported_providers": [ "CPUExecutionProvider" ],
-            "supported_accelerators": [ "cpu" ],
+            "supported_providers": [ "CUDAExecutionProvider" ],
```
jambayk, line 231: maybe we should make it `*` for both providers and ep. the quantization happens on pytorch model and the exported model is compatible with all eps.

jambayk: their only requirement is that they require gpus to run. but that is a host machine requirement, not target provider or ep.

xiaoyu-work: Does this `supported_providers` mean the output model supports which ep? Same concept for `supported_accelerators`?

jambayk: yes. it's used by the auto-opt to filter out the passes based on the intended target ep

xiaoyu-work: I see. I can see some other passes like `QNNConversion` is also in this list. But QNN model doesn't not have any ep concept. If this is only used by auto-opt, and auto-opt is targeting onnx model, should those unrelated pass whose output model is not onnx model?

jambayk: @shaahji can comment more on it. I think in general, we can just use * for passes that don't deal with onnx models. that way if they are scheduled by the auto-opt, they don't get filtered out.

devang-ml: QNN model does not have EP concept but you don't want to run QNNConversion pass if the user did not select QNN_EP.

xiaoyu-work: @shaahji

shaahji: > Does this `supported_providers` mean the output model supports which ep? Same concept for `supported_accelerators`?

Yes, these fields dictate the target provider and accelerator for running the output model. We do not have a parallel support for host machine and depend on Pass.validate_config to make that decision.

xiaoyu-work: > I see. I can see some other passes like `QNNConversion` is also in this list. But QNN model doesn't not have any ep concept. If this is only used by auto-opt, and auto-opt is targeting onnx model, should those unrelated pass whose output model is not onnx model be removed here?

@shaahji what about this? i can see all passes needs to be registered here anyway?

shaahji: Yes, all passes needs to be registered in this config since this is the only way to create a pass by importing modules.

jambayk (approved):

PR #1585, opened 2025-01-29
Resolving objectives issue introduced with introduction of pass search

shaahji:
## Resolving objectives issue introduced with introduction of pass search

Olive allows multiple passes under the "passes" key in config where each entry can have dictate its own evaluator config i.e. the evaluation config to use for that specific pass. However, with pass search, this becomes an issue because each pass within the group can dictate conflicting objectives and goals. Circumventing the issue by collecting all the objectives across the group with the last one in the list winning if named the same. Also, handle the case where not all objectives are generated as part of the
post evaluation signal.

### DIFF
--- a/olive/search/samplers/optuna_sampler.py
+++ b/olive/search/samplers/optuna_sampler.py
@@ -2,6 +2,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+import sys
 from abc import abstractmethod
 from collections import OrderedDict
 from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Union
@@ -39,15 +40,15 @@ class OptunaSampler(SearchSampler):
         self,
         search_space: SearchSpace,
         config: Optional[Union[Dict[str, Any], ConfigBase]] = None,
+        objectives: Dict[str, Dict[str, Any]] = None,
     ):
-        super().__init__(search_space, config)
+        super().__init__(search_space, config, objectives)
 
         # Initialize the searcher
         self._sampler = self._create_sampler()
-        # TODO(olivedev): There is no absolute direction to set.
-        # directions = ["maximize" if hib else "minimize" for hib in self._higher_is_betters]
-        # self._study = optuna.create_study(directions=directions, sampler=self._sampler)
-        self._study = optuna.create_study(sampler=self._sampler)
+        directions = ["maximize" if self._higher_is_betters[name] else "minimize" for name in self._objectives]
+        self._study = optuna.create_study(directions=directions, sampler=self._sampler)
+
         self._num_samples_suggested = 0
         self._search_point_index_to_trail_id = {}
 
@@ -133,12 +134,18 @@ class OptunaSampler(SearchSampler):
         else:
             raise ValueError(f"Unsupported parameter type: {type(param)}")
 
-    def record_feedback_signal(
-        self, search_point_index: int, objectives: Dict[str, dict], signal: "MetricResult", should_prune: bool = False
-    ):
+    def record_feedback_signal(self, search_point_index: int, signal: "MetricResult", should_prune: bool = False):
         trial_id = self._search_point_index_to_trail_id[search_point_index]
         if should_prune:
             self._study.tell(trial_id, state=optuna.trial.TrialState.PRUNED)
         else:
-            values = [signal[objective].value for objective in objectives]
+            values = []
+            for name in self._objectives:
+                if name in signal:
+                    values.append(signal[name].value)
+                elif self._higher_is_betters[name]:
+                    values.append(-sys.maxsize - 1)
+                else:
+                    values.append(sys.maxsize)
+
             self._study.tell(trial_id, values)
--- a/olive/search/search_results.py
+++ b/olive/search/search_results.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 import sys
+from collections import OrderedDict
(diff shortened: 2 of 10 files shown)


olive/search/samplers/optuna_sampler.py
```diff
@@ -133,12 +134,18 @@ def _get_search_point_values(
...
         trial_id = self._search_point_index_to_trail_id[search_point_index]
         if should_prune:
             self._study.tell(trial_id, state=optuna.trial.TrialState.PRUNED)
         else:
-            values = [signal[objective].value for objective in objectives]
+            values = []
+            for name in self._objectives:
+                if name in signal:
+                    values.append(signal[name].value)
+                elif self._higher_is_betters[name]:
```
jambayk, line 146: this looks good for now, but I wonder how the search behaves when some metrics are missing since those are implied to be very bad by these values.

olive/search/search_results.py
```diff
@@ -31,16 +44,14 @@ def meets_goals(self, search_point_index: int) -> bool:
...
+        if not self._goals:
             return True  # if goals are not set, always return True
 
-        # multiplier for each objective and goals
-        multipliers = {
-            name: 1 if objective.get("higher_is_better", False) else -1 for name, objective in objectives.items()
-        }
-        return all((multipliers[obj] * result[obj].value) >= (multipliers[obj] * goal) for obj, goal in goals.items())
+        result, _ = self._results[search_point_index]
+        return all(
```
jambayk, line 51: what happens if the results is missing the value for a metric that has a goal coming from another pass' evaluator?

jambayk: might need an if condition in the loop here?

olive/search/samplers/optuna_sampler.py
shaahji: I don't think this matters in the bigger picture since this is only for the purpose of sorting. If a particular pass depends only a subset of the objectives, it wouldn't even care for the other metrics. 

In short, if two passes use different objectives with same priority, we need a way to compare these results. As long as, all results are sorted based on priority, search wouldn't be impacted.

olive/search/search_results.py
```diff
@@ -67,58 +78,48 @@ def get_next_best_result(self, start_index: int) -> Tuple[int, int, List[str]]:
...
-        if not all_objectives:
+        indices = []
+        if not self._objectives:
             # If no objectives, then use the indices of the valid results in no specific order
             indices = [spi for spi, entry in enumerate(self._results) if entry]
             return indices, values
 
         # NOTE: values array need to be packed but a simple loop thru' each entry could
-        # possibly create a zagged array if the number of objectives are different.
+        # possibly create a zagged array if the number of actual objectives in the signal
```
jambayk, line 94: typo: ragged?

olive/search/search_results.py
shaahji: The signal doesn't include any identification for which pass generated it so yes, that's still a problem.

Honestly, the whole concept of objectives, goals, and metrics, need a revisit with emphasis on search.

olive/search/search_results.py
shaahji: Misspelled Jagged or as you said, Ragged.
https://en.wikipedia.org/wiki/Jagged_array

olive/search/samplers/optuna_sampler.py
jambayk: sounds good! Like we discussed earlier, I personally don't see a case where we would compare two passes with different metrics, so it probably won't appear in normal use cases. the max/min method sets a rule for comparing them so it's okay for me. even if such cases appear, it might be used with early stopping anyways.

olive/search/search_results.py
jambayk: what if you add if name in results check so that it doesn't hit an index error atleast. if the passes used the same metric but different objectives, then the last one taking priority already enforces that one metric name can only have one objective.

olive/search/search_results.py
jambayk: oh, didn't know ragged and jagged were the same thing

olive/search/search_results.py
shaahji: At the `SearchStrategy` level, I can see a reason i.e. to conditionally discard search points that don't meet a goal.

Within the context of `SearchResult`, how does goals matter? As long as the results are sorted based on priority and the signal values, why does goals matter? `SearchResult` shoudn't even be accounting for any goals at all.

olive/search/search_results.py
shaahji: Fixed the spelling error.

olive/search/search_results.py
jambayk: since this method is used for early stopping herehttps://github.com/microsoft/Olive/blob/c22f68bf62e8588020d0f616871acfcc0f7f20ff/olive/search/search_strategy.py#L242, wouldn't that cause an index error if the metric is missing?

jambayk (approved):

olive/search/search_results.py
shaahji: > At the `SearchStrategy` level, I can see a reason i.e. to conditionally discard search points that don't meet a goal.
> 
> Within the context of `SearchResult`, how does goals matter? As long as the results are sorted based on priority and the signal values, why does goals matter? `SearchResult` shoudn't even be accounting for any goals at all.

My comment was in context to `apply_goals` argument in `SearchResult._get_results_list`.

I added the condition in prevent indexing error in the function.

jambayk: I see. The apply goals option was added so that from all of the candidate output models, the user can filter out the ones that didn't meet their goal. otherwise, the ranked models can have models that do really well on a high priority metric but don't meet the minimum requirements for another metric.

PR #1611, opened 2025-02-12
Add extract adapters support for dora and loha

xiaoyu-work:
## Describe your changes

Add extract adapters support for dora and loha.


olive/passes/onnx/extract_adapters.py
jambayk, line 1: Can you remove this condition in the peft export helper https://github.com/microsoft/Olive/blob/50f360aeacfb949abc0d845e4070922555f7c58a/olive/common/hf/peft.py#L48? 

Otherwise, the lora scaling factors appears in the graph and would need to be extracted also since it is not the same for all adapters.

olive/passes/onnx/extract_adapters.py
```diff
@@ -220,22 +145,287 @@ def _run_for_config(
...
+        return weights
+
+    def _extract_adapter(self, dag: OnnxDAG, config: Dict[str, Any], adapter_type: AdapterType = AdapterType.LORA):
+        """Extract adapter weights for either LoRA or Dora from an ONNX model.
+
+        LoRA:
+        output + default (lora_A) -> MatMul -> ...
+        output + default_1 (lora_B) -> MatMul -> ...
+
+        DoRA:
```
jambayk, line 225: I looked at the exported graph for a dora model. the lora matmuls appear twice in the computation. 
Also, the Add weight doesn't need to be extracted. It is just the transpose of the base weight and comes from this step https://github.com/huggingface/peft/blob/363c14e673a12d19f951609d06221962d5c3eb2a/src/peft/tuners/lora/dora.py#L78. This is constant for a model. 

What needs to be extracted is the weight in the Div node that depends on the lora weights. https://github.com/huggingface/peft/blob/363c14e673a12d19f951609d06221962d5c3eb2a/src/peft/tuners/lora/dora.py#L63

Image

jambayk: Ideally, during export, we can precompute https://github.com/huggingface/peft/blob/main/src/peft/tuners/lora/dora.py#L70-L86 so that `ConstantOfShape` -> `Reshape` doesn't appear in the graph and we just have the `mag_norm_scale` weight directly in the graph. But I think this can be done later if needed and is out of scope for this PR.

xiaoyu-work: Thanks for pointing this out! I double checked the graph and updated the code to use Div rather Add for now!

olive/passes/onnx/extract_adapters.py
xiaoyu-work: Removed.

olive/passes/onnx/extract_adapters.py
jambayk: could you confirm if there are five operators that are modified per adapter subgraph for dora? looking at the dora layer implementation, both inputs for div node are dependent on lora.

maybe it would be easier if we applied something like `quant_preprocess` or constant folding so that the subgraph from constantofshape -> Reshape gets folded into a single weight as output value is constant for each lora weight pair.

test/unit_test/passes/onnx/test_extract_adapters.py
```diff
@@ -154,60 +215,36 @@ def test_extract_adapters_as_initializers(tmp_path, input_model_info, model_type
     assert seen_weights == expected_weights
 
 
+@pytest.mark.parametrize("input_model_info", [AdapterType.LORA, AdapterType.DORA, AdapterType.LOHA], indirect=True)
 @pytest.mark.parametrize("model_type", ["float", "qdq", "int4"])
 @pytest.mark.parametrize("save_format", [el.value for el in WeightsFileFormat])
-def test_extract_adapters_as_inputs(tmp_path, input_model_info, save_format, model_type):
+def test_extract_adapters_as_inputs(tmp_path, save_format, model_type, input_model_info):
```
jambayk, line 221: maybe we should run the model using an inference session for at least the float model to confirm the adapter extracted model works correctly for different adapters. we don't have a test yet to check the correctly of the model when we switch adapters.

olive/passes/onnx/extract_adapters.py
xiaoyu-work: Yes, 5 extra ops for dora. It sounds good to do a constant folding to it. I will create a separate workitem to do this.

test/unit_test/passes/onnx/test_extract_adapters.py
xiaoyu-work: I can add a real model test for this after finishing the constant folding task.

PR #1615, opened 2025-02-14
Fix device handling and logits concatenation in OliveEvaluator

tezheng:
- Add exception handling for unsupported devices in `device_string_to_torch_device` method.
- Correct logits concatenation in `OnnxEvaluator` by using `logits_dict` instead of `logits`.
- Initialize `logits_dict` in `PyTorchEvaluator` to handle different result types.
- Update `_inference` method in `PyTorchEvaluator` to handle different result types and concatenate logits correctly.

### DIFF
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -196,7 +196,11 @@ class OliveEvaluator(ABC):
 class _OliveEvaluator(OliveEvaluator):
     @staticmethod
     def device_string_to_torch_device(device: Device):
-        return torch.device("cuda") if device == Device.GPU else torch.device(device)
+        try:
+            return torch.device("cuda") if device == Device.GPU else torch.device(device)
+        except (ValueError, TypeError, RuntimeError):
+            logger.warning("Device %s is not supported in torch, fallback to CPU instead.", device)
+            return torch.device("cpu")
 
     @classmethod
     def io_bind_enabled(cls, metric: Metric, inference_settings: Dict) -> bool:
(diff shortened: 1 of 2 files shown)


olive/evaluator/olive_evaluator.py
```diff
@@ -196,7 +196,11 @@ def compute_throughput(metric: Metric, latencies: Any) -> MetricResult:
 class _OliveEvaluator(OliveEvaluator):
     @staticmethod
     def device_string_to_torch_device(device: Device):
-        return torch.device("cuda") if device == Device.GPU else torch.device(device)
+        try:
```
jambayk, line 199: maybe we can just do `torch.device("cuda") if device == Device.GPU else torch.device("cpu")`. Otherwise, this would raise the warning for npu targets everytime. since this method is only used when evaluating pytorch models, I think it's save to just return cuda or cpu only.

tezheng: So we will only consider `cuda` and `cpu`, for now at least? Are we planning to support device like Apple `mps`?

jambayk: yes, there are no plans for other torch devices like mps for now.

olive/evaluator/olive_evaluator.py
```diff
@@ -748,15 +753,20 @@ def _inference(
...
-                if not isinstance(result, torch.Tensor) and getattr(result, "logits", None) is not None
-                else result.cpu()
-            )
+            if isinstance(result, torch.Tensor):
+                logits.append(result.cpu())
+            elif isinstance(result, (list, tuple)):
+                logits.append([r.cpu() for r in result])
+            elif isinstance(result, dict):
+                for k in result:
+                    logits_dict[k].append(result[k].cpu())
```
xiaoyu-work, line 762: How can we handle the case when result has logits attribute?

tezheng: You mean result like `{ "logits": tensor }`?

jambayk:
closing since it has become stale and dict issue was resolved in #701. Might still need the device solution.

PR #1647, opened 2025-02-27
[dynamo exporter] Fix a bug in dynamic_shapes related processing that makes regression on the benchmark

titaiwangms:
## Describe your changes

Sorry! Follow up #1639, found a bug from the benchmark that makes regression on static shapes.

When it was static shape (io_config.dynamic_shapes is None), `_validate_dynamic_shapes` returned only one value, which crashed the benchmark with static shapes.

### DIFF
--- a/test/unit_test/passes/onnx/test_conversion.py
+++ b/test/unit_test/passes/onnx/test_conversion.py
@@ -24,17 +24,21 @@ from olive.passes.onnx.conversion import OnnxConversion, OnnxOpVersionConversion
 
 @pytest.mark.skipif(sys.version_info > (3, 8), reason="Failed with Python 3.10, need to investigate.")
 @pytest.mark.parametrize(
-    ("input_model", "use_dynamo_exporter"),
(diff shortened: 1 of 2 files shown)


test/unit_test/passes/onnx/test_conversion.py
```diff
@@ -24,17 +24,21 @@
 
 @pytest.mark.skipif(sys.version_info > (3, 8), reason="Failed with Python 3.10, need to investigate.")
```
titaiwangms, line 25: It's better we can fix this issue, as this test is very important to validate the exporter (although I still test it on local).

jambayk: Oh, I didn't realize this test has been disabled. I created a work item to track this and will ask the team to prioritize this since it's an important test.

jambayk (approved):
Thanks! CI failures are unrelated to this PR. Will rerun them once fixed and merge this.

Issue #1632, opened 2025-02-21
[FR]: save model from pass OnnxStaticQuantization quant_preprocess = true


xieofxie:
### Proposal Summary

do we have a way to this? Sometimes I need to quantize from the model here and preprocess is not needed to be reruned multiple times.

We could do 
- Add a new pass to run this only
- Add a parameter to OnnxStaticQuantization  so only save the mode from this and no quantize

### What component(s) does this request affect?

- [ ] OliveModels
- [ ] OliveSystems
- [ ] OliveEvaluator
- [ ] Metrics
- [ ] Engine
- [ ] Passes
- [ ] Other

jambayk:
This is a good idea. Having to rerun this multiple times was a concern for us so we tried to do some caching with https://github.com/microsoft/Olive/blob/51d2c8a0c92fa2ba94e1eaea5661acb65bbd2ba3/olive/passes/onnx/quantization.py#L395 but that only persists during the lifetime of the pass and was meant for use with search. 

I think for a general use case, we could create a new pass to do quant preprocess. Providing a directory to save the preprocessed model is a possibility too but I am not sure if we want the user to have to manage that.

xieofxie:
Yes, a new pass is very useful. Because many parameters of OnnxStaticQuantization are based on the preprocessed model not the input model of the pass

PR #1651, opened 2025-02-28
add OnnxQuantizationPreprocess

xieofxie:
## Describe your changes

Solve issue https://github.com/microsoft/Olive/issues/1632

What else should I update?


olive/passes/onnx/quantization.py
```diff
@@ -526,36 +527,65 @@ def _run_for_config(
...
+    
+    @classmethod
+    def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        configs = get_external_data_config()
+        configs.pop("convert_attribute")
+        return {
+            **configs,
+        }
+    
+    def _run_for_config(
```
jambayk, line 585: The olive `model_proto_to_file` has some logic that saves the model even if the model is too large and the user has not enabled `save_as_external_data` https://github.com/microsoft/Olive/blob/ba91e143671b3362bcbf252199e3b91d9a9c3f0d/olive/passes/onnx/common.py#L101. Right now, the save would fail for large models unless the users enables it. 

My suggestion for a foolproof solution would be
- keep the `_quant_preprocess` logic the same as before
- Using a temporary directory similar to the quantization pass to save the preprocessed model
- load the preprocessed model and save to the output path using `model_proto_to_olive_model` here.

jambayk: Also, please get a full output model path with .onnx suffix using https://github.com/microsoft/Olive/blob/ba91e143671b3362bcbf252199e3b91d9a9c3f0d/olive/passes/onnx/quantization.py#L734

xieofxie: Now it works like a wrapper for `onnxruntime.quantization.preprocess.quant_pre_process` to give user more flexibility

olive/passes/onnx/quantization.py
```diff
@@ -542,6 +542,53 @@ def _quant_preprocess(self, model: ONNXModelHandler, output_model_path: Union[st
...
+                quant_pre_process(
+                    input_model_path=model.model_path,
+                    output_model_path=tmp_model_path,
+                    auto_merge=True,
+                    save_as_external_data=True,
+                    skip_optimization=config.skip_optimization,
+                    verbose=3,  # set verbose to 3 to get more information about the preprocessing
+                )
+            except Exception:
+                # quantization preprocessing will fail if the model is too large and `skip_optimization = False`
```
jambayk, line 581: this comment can be removed since this issue has been fixed in onnxruntime. We should have updated the comments in https://github.com/microsoft/Olive/blob/db8963b12d21e5558e09070a9d3ea8951b92e765/olive/passes/onnx/quantization.py#L525 when the issue got fixed

olive/passes/onnx/quantization.py
```diff
@@ -542,6 +542,53 @@ def _quant_preprocess(self, model: ONNXModelHandler, output_model_path: Union[st
         return ONNXModelHandler(LocalFile({"path": output_model_path}))
 
 
+class OnnxQuantizationPreprocess(Pass):
+    """ONNX Quantization Preprocess Pass. Same as OnnxQuantization quant_preprocess."""
+
+    @classmethod
+    def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        config = {
+            "skip_optimization": PassConfigParam(
```
jambayk, line 551: could you also expose this option https://github.com/microsoft/onnxruntime/blob/30c682547bdae3e09523614f8e38526e49ca8fbc/onnxruntime/python/tools/quantization/shape_inference.py#L29? 

From my experience, quant preprocess fails more when the ort symbolic shape infer doesn't support some contrib ops.

xieofxie: Sure, add all skip parameters

olive/passes/onnx/quantization.py
xieofxie: removed

PR #1660, opened 2025-03-04
Extract multiple adapters from PyTorch model

kunal-vaishnavi:
### Description

This PR adds the ability to extract multiple adapters from a PyTorch model using the Olive CLI and save them in the desired format.

Example usage:
```bash
olive extract-adapters -m microsoft/Phi-4-multimodal-instruct -o phi-4-mm -f onnx_adapter
```

### Motivation and Context

The Phi-4 multimodal model contains two sets of adapters that are pre-loaded into the PyTorch model when running `AutoModelForCausalLM.from_pretrained`. The model is not a PEFT model but is a Transformers model with PEFT's `LoraLinear` layers inserted. Other Olive CLI commands do not have the ability to extract multiple adapters from a PyTorch model.

### DIFF
--- a/olive/cli/extract_adapters.py
+++ b/olive/cli/extract_adapters.py
@@ -55,26 +55,6 @@ class ExtractAdaptersCommand(BaseOliveCLICommand):
             default="cache_dir",
             help="Cache dir to store temporary files in. Default is `cache_dir`.",
         )
-        # int4 quantization options for adapter weights
-        sub_parser.add_argument(
-            "--quantize_int4",
-            action="store_true",
-            help="Quantize the adapter weights to int4 using blockwise quantization.",
-        )
-        sub_parser.add_argument(
-            "--int4_block_size",
-            type=int,
-            default=32,
-            choices=[16, 32, 64, 128, 256],
-            help="Block size for int4 quantization of adapter weights. Default is 32.",
-        )
-        sub_parser.add_argument(
-            "--int4_quantization_mode",
-            type=str,
-            default="symmetric",
-            choices=["symmetric", "asymmetric"],
-            help="Quantization mode for int4 quantization of adapter weights. Default is symmetric.",
-        )
         add_logging_options(sub_parser)
         sub_parser.set_defaults(func=ExtractAdaptersCommand)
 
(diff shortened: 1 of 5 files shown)


test/unit_test/cli/test_cli.py
```diff
@@ -272,5 +272,27 @@ def test_quantize_command(mock_repo_exists, mock_run, algorithm_name, tmp_path):
     assert mock_run.call_count == 1
 
 
+@patch("huggingface_hub.repo_exists", return_value=True)
+def test_extract_adapters_command(mock_repo_exists, tmp_path):
```
jambayk, line 276: CI says it's missing a package `backoff`. you can add it to https://github.com/microsoft/Olive/blob/main/test/requirements-test.txt

Not sure how long the model download would take on the pipeline. Would it be possible to create a small test model for this?

test/unit_test/cli/test_cli.py
```diff
@@ -272,5 +272,76 @@ def test_quantize_command(mock_repo_exists, mock_run, algorithm_name, tmp_path):
...
+    vision_config = LoraConfig(
+        r=4,
+        lora_alpha=12.0,
+        target_modules=["q_proj", "k_proj", "v_proj", "o_proj", "gate_proj", "up_proj", "down_proj"],
+        lora_dropout=0.0,
+        task_type="CAUSAL_LM",
+    )
+    peft_model.add_adapter("vision", vision_config)
+
+    model_name_or_path = os.path.join(".", "test_peft")
```
jambayk, line 325: please use `model_name_or_path = tmp_path / test_peft` instead so that the test artifacts get cleaned up automatically. 

Same for [`--cache_dir`, str(tmp_path / cache_dir)`]

olive/cli/extract_adapters.py
```diff
@@ -0,0 +1,164 @@
...
+            "--dtype",
+            type=str,
+            default="float32",
+            choices=["float32", "float16"],
+            help="Data type to save LoRAs as. Default is float32.",
+        )
+        sub_parser.add_argument(
+            "--cache_dir",
+            type=str,
+            default="cache_dir",
```
jambayk, line 50: maybe use `transformers.utils.TRANSFORMERS_CACHE` if the user doesn't provide their own? This way it will just use the default hf cache.

test/unit_test/cli/test_cli.py
kunal-vaishnavi: Fixed

olive/cli/extract_adapters.py
kunal-vaishnavi: Fixed

jambayk (approved):
Thanks!

PR #1662, opened 2025-03-05
Allow running model optimization with recent OpenVINO versions and im…

asoldano:
…prove documentation

I've manually tested this with OpenVINO 2024.6.0.

### DIFF
--- a/examples/stable_diffusion/README.md
+++ b/examples/stable_diffusion/README.md
@@ -148,6 +148,14 @@ cd olive/examples/stable_diffusion
 pip install -r requirements-ov.txt
 ```
 
+##### Install onnxruntime
+
+This example requires onnxruntime and its openvino module to be installed:
+
+```bash
+pip install onnxruntime onnxruntime-openvino
+```
+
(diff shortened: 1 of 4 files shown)


asoldano:
@microsoft-github-policy-service agree

devang-ml (approved):

examples/stable_diffusion/README.md
```diff
@@ -148,6 +148,14 @@ cd olive/examples/stable_diffusion
 pip install -r requirements-ov.txt
 ```
 
+##### Install onnxruntime
+
+This example requires onnxruntime and its openvino module to be installed:
+
+```bash
+pip install onnxruntime onnxruntime-openvino
```
jambayk, line 156: is `onnxruntime` required? I think `onnxruntime-openvino` is a different flavor of the onnxruntime package and will install into the same namespace and override the onnxruntime installation.

asoldano: OK, I've checked, that can be removed. I've just updated the doc.

jambayk (approved):

PR #1668, opened 2025-03-07
StaticLLM: pass to create static shaped llms

jambayk:
New pass `StaticLLM` added to convert dynamic shaped llm into a static shaped llm for NPUs.
- Only GQA models supported currently to allow for long context through sliding window. Only the batch size and input sequence length are fixed. 
- Models without GQA are complicated and need more work. Will be added in the future if needed. 

Cache model save for composite model handlers has been improved.
- all component onnx model files are saved in the same directory
- hardlink_copy_file is used

### DIFF
--- a/docs/source/reference/pass.rst
+++ b/docs/source/reference/pass.rst
@@ -158,6 +158,12 @@ SplitModel
 ----------
 .. autoconfigclass:: olive.passes.SplitModel
 
+.. _static_llm:
+
+StaticLLM
+----------
+.. autoconfigclass:: olive.passes.StaticLLM
+
 .. _optimum_conversion:
 
 OptimumConversion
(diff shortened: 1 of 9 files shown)


devang-ml:
Do we need a separate pass from `DynamicToFixedShape` ?

jambayk:
> Do we need a separate pass from `DynamicToFixedShape` ?

@devang-ml the dynamictofixedshape is a simple and generic pass which just fixes the shapes for a single onnx model. However, the scenario for llms is much more complex:
- The embeddings and lm_head can be used for both prefill and token generation, so their shapes are kept dynamic
- The transformers layers could be split into multiple components
- The transformers model leads to two new models for the prefill and token generation. 

It is not easy to encode such information in a pass config or input model so I chose to create a specialized pass that has these assumptions. If we plan to support non gqa llms in the future, that would also require graph manipulations to create the required kv cache input/outputs. Making the logic even more complex and specialized.

devang-ml (approved):

PR #1672, opened 2025-03-10
Add support for new versions of transformers

shaahji:
## Add support for new versions of transformers
* Change of argument name in transformers>=v4.45
* Handle DynamicCache/EncoderDecoderCache in transformers>=4.48


olive/passes/onnx/conversion.py
```diff
@@ -411,6 +497,8 @@ def _convert_model_on_device(
         # get dummy inputs
         dummy_inputs = self._get_dummy_inputs(model, config)
         io_config = model.io_config
+
+        self._patch_model_if_necessary(pytorch_model)
```
jambayk, line 501: can you apply this inside `_export_pytorch_model` after https://github.com/microsoft/Olive/blob/0a2873c9bbb3875d09be7add6cc9bf59a6a27504/olive/passes/onnx/conversion.py#L203 so that the peft and quantized model patches are applied first? Otherwise, adapter models don't have a generic forward signature of `*args, **kwargs`

olive/passes/onnx/conversion.py
```diff
@@ -389,6 +392,89 @@ def _prepare_hf_model(
...
+        )
+        pkv_index = (
+            list(signature.parameters.keys()).index("past_key_values")
+            if "past_key_values" in signature.parameters
+            else None
+        )
+
+        @functools.wraps(orig_forward)
+        def patched_forward(*args, **kwargs):
+            from transformers.cache_utils import DynamicCache, EncoderDecoderCache
```
jambayk, line 424: I think there is an issue with these classes being `torch.nn.Module` children in transformers 4.48 so they had to be replaced with a custom module. 
If this is an issue, maybe we can add a warning log to upgrade transformers version? I don't think it's worthwhile doing the module replacement but we could let the user know if it's an issue.

olive/passes/onnx/conversion.py
```diff
@@ -389,6 +392,89 @@ def _prepare_hf_model(
...
+            elif logits_to_keep_index is not None:
+                args[logits_to_keep_index] = 0
+
+            if (
+                pkv_index
+                and pkv_index < len(args)  # pkv is in args
+                and isinstance(args[pkv_index], (list, tuple))
+                and isinstance(args[pkv_index][0], (list, tuple))
+            ):
+                logger.info("Stepping into first if clause")
```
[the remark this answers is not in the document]
shaahji: Removed. Those were for local debugging that got left in.

olive/passes/onnx/conversion.py
shaahji: Done!

olive/passes/onnx/conversion.py
shaahji: Wouldn't that be something that belongs in transformers? I don't think such a warning belongs in here.

jambayk: I thought there was an issue with exporting model using the cache class so optimum also did the following patch https://github.com/huggingface/optimum/blob/3adbe7c75e3c41c1a3b945cf085e74ece7f8e192/optimum/exporters/onnx/model_patcher.py#L204

But the torch.nn.Module inheritance is present in 4.49 and if you were able to get the export to work fine, then maybe it's not required.

olive/passes/onnx/conversion.py
```diff
@@ -191,6 +193,7 @@ def _export_pytorch_model(
             torch.cuda.empty_cache()
 
         logger.debug("Converting model on device %s with dtype %s.", device, torch_dtype)
+        OnnxConversion._patch_model_if_necessary(pytorch_model)
```
jambayk, line 196: please move this to after line 205 so that the forward signature of `pytorch_model` is as expected.

olive/passes/onnx/conversion.py
jambayk, line 1: can you also add a unit test to onnx conversion for `text-generation` hf model that checks that the expected i/o are present?

olive/passes/onnx/conversion.py
shaahji: I successfully exported llama and bert. Llama was failing before this patch and succeed after the change.

olive/passes/onnx/conversion.py
shaahji: Done!

jambayk (approved):

PR #1686, opened 2025-03-17
Update adetailer example

xiaoyu-work:
## Describe your changes

1. Updated adetailer example with PyTorch input model and onnx conversion pass.
2. Removed unused prepare onnx model file.
3. Verified output onnx model.
4. Updated pytorch evaluator with output postprocessing.


examples/adetailer/face_yolo_qnn.json
```diff
@@ -1,5 +1,11 @@
 {
-    "input_model": { "type": "ONNXModel", "model_path": "models/face/face_yolov9c.onnx" },
+    "input_model": {
+        "type": "PytorchModel",
```
devang-ml, line 3: You don't need ONNX conversion pass?

xiaoyu-work: Added!

devang-ml (dismissed):

olive/evaluator/olive_evaluator.py
```diff
@@ -715,6 +715,9 @@ def _inference(
         execution_providers: Union[str, List[str]] = None,
     ) -> Tuple[OliveModelOutput, Any]:
         session = model.prepare_session()
+        io_config = model.io_config
```
jambayk, line 718: `io_config` is not required and guaranteed to be present for pytorch models. You can update the post_process function to handle your specific scenario.

xiaoyu-work: onnx evaluator has this similar logic, this logic should be applied to pytorch evaluator as well. I can check if there is workaround if a pytorch model doesn't have io config.

jambayk: we do this for onnx model since the output of an onnx inference session run is always a dict. For pytorch model, its forward function can return a dict or a similar object as needed. The post processing func can be made generic to handle both onnx and pytorch model outputs like for https://github.com/microsoft/Olive/blob/4f1496e52610c20754278fc8eedbd5f3e2a73a39/olive/data/component/post_process_data.py#L71

xiaoyu-work: Done.

jambayk (approved):

PR #1695, opened 2025-03-20
Add GTE model example

xiaoyu-work:
## Describe your changes

Add GTE model example

### DIFF
--- a/examples/gte/config.json
+++ b/examples/gte/config.json
@@ -1,0 +1,40 @@
+{
+    "input_model": {
+        "type": "PytorchModel",
+        "model_path": "Alibaba-NLP/gte-large-en-v1.5",
+        "model_script": "user_script.py",
+        "io_config": "get_io_config",
+        "model_loader": "load_model"
+    },
+    "data_configs": [
+        {
+            "name": "c4",
+            "type": "HuggingfaceContainer",
+            "user_script": "user_script.py",
+            "load_dataset_config": {
+                "data_name": "allenai/c4",
+                "subset": "en",
+                "split": "validation",
+                "streaming": true
+            },
+            "pre_process_data_config": {
+                "type": "dataset_pre_process",
+                "max_samples": 128,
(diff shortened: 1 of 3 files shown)


examples/gte/config.json
```diff
@@ -0,0 +1,40 @@
+{
+    "input_model": {
+        "type": "PytorchModel",
```
jambayk, line 3: is there a reason you don't use an hf model. the loading logic and io config look similar to a regular hf model. the task is probably feature-extraction or something related.

examples/gte/config.json
```diff
@@ -0,0 +1,40 @@
...
+            "name": "c4",
+            "type": "HuggingfaceContainer",
+            "user_script": "user_script.py",
+            "load_dataset_config": {
+                "data_name": "allenai/c4",
+                "subset": "en",
+                "split": "validation",
+                "streaming": true
+            },
+            "pre_process_data_config": {
```
jambayk, line 20: same for here, I think we could add a new preprocess function for the new task https://github.com/microsoft/Olive/blob/007bbf4e2ccbb5ea64e245c961d4395ac6701181/olive/data/container/huggingface_container.py#L21. Most of the hf preprocess logic is already abstracted and this one also looks similar.

jambayk: oh an option in the existing default hf preprocess to make the "label" column name optional.

examples/gte/config.json
xiaoyu-work: This model has unusual config.json that the `architectures` is `NewModel` and `model_type` is `New`, which will cause error when fetching io config from optimum. But I can simplify this with only io config provided in the config file.

examples/gte/config.json
xiaoyu-work: Sound good to me! i can double check this and make a separate PR for this.

jambayk: sgtm! thanks!

PR #1699, opened 2025-03-22
Create genai_config for NPU llms, inference example

jambayk:
## Describe your changes
Fixes and updates:
  - `CompositeModelHandler` now has an optional `model_path` resource to store additional files. In practice, most of the composite models that are output of passes use the same directory for the component models. 
  - Abstracted out the common llm-pipeline logic from the passes. It also automatically updates the genai config with the correct model io and session options.
  - Fixed untie word embedding implementation in hf model wrapper.
  - Fixed the data type of the layernorm weight in the rms norm graph surgeries. In numpy 2.0+, the data type after multiplication with `sqrt_n` gets promoted to double. but we want float.
  - Additional files carry over now supports composite models as well. Removed unnecessary check for input model dir. It was a remnant from the original implementation.

Examples:
- Instructions and example inference script added
- Qwen 2.5 example added.

### DIFF
--- a/olive/passes/onnx/common.py
+++ b/olive/passes/onnx/common.py
@@ -2,16 +2,18 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+import json
 import logging
 import re
+from copy import deepcopy
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Type, Union
+from typing import Any, Callable, Dict, List, Optional, Type, Union
 
 import onnx
 from onnx import external_data_helper
 
 from olive.common.utils import hardlink_copy_file
-from olive.model import ONNXModelHandler
+from olive.model import CompositeModelHandler, ONNXModelHandler
 from olive.passes.onnx.onnx_dag import OnnxDAG
 from olive.passes.pass_config import BasePassConfig, PassConfigParam
 from olive.resource_path import LocalFile, LocalFolder
@@ -145,6 +147,7 @@ def model_proto_to_olive_model(
     check_model: bool = False,
     external_initializers_file_name: Optional[str] = None,
     constant_inputs_file_name: Optional[str] = None,
+    force_model_dir: bool = False,
 ) -> ONNXModelHandler:
     """Save the ONNX model to the specified path and return the ONNXModelHandler.
 
@@ -155,6 +158,8 @@ def model_proto_to_olive_model(
     :param check_model: If True, run onnx.checker.check_model on the model before returning.
     :param external_initializers_file_name: The name of the external initializers file.
     :param constant_inputs_file_name: The name of the constant inputs file.
+    :param force_model_dir: If True, use the parent directory of the output model path as the model directory
+        regardless of whether external data is used.
 
     :return: The ONNXModelHandler.
     """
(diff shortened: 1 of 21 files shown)


olive/passes/onnx/common.py
```diff
@@ -409,3 +414,123 @@ def fix_input_shapes(model_proto: onnx.ModelProto, input_names: List[str], input
...
+    llm_pipeline: List,
+    process_func: Callable,
+    output_dir: Union[str, Path],
+    decoder_config_extra: Optional[Dict[str, Any]] = None,
+    group_session_options: Optional[Dict[str, Any]] = None,
+) -> CompositeModelHandler:
+    """Process an LLM pipeline with the given function.
+
+    :param model_handler: The composite model with the pipeline.
+    :param llm_pipeline: The pipeline to process.
```
devang-ml, line 430: Do we have a comment anywhere describing what exactly the llm_pipeline is? What does it include? And how it is used between passes?

olive/passes/onnx/common.py
```diff
@@ -409,3 +414,123 @@ def fix_input_shapes(model_proto: onnx.ModelProto, input_names: List[str], input
...
+        if Path(file_path).name == "genai_config.json":
+            genai_config_path = file_path
+            break
+
+    if genai_config_path:
+        with open(genai_config_path) as f:
+            genai_config = json.load(f)
+
+        # update model_type
+        genai_config["model"]["type"] = "decoder-pipeline"
```
devang-ml, line 481: Is "decoder-pipeline" the only model type supported by genai ?

olive/passes/onnx/common.py
```diff
@@ -409,3 +414,123 @@ def fix_input_shapes(model_proto: onnx.ModelProto, input_names: List[str], input
...
+    lm_head_model_path = output_dir / "lm_head.onnx"
+    resave_model(component_models[llm_pipeline["lm_head"]].model_path, lm_head_model_path)
+    new_component_models["lm_head"] = ONNXModelHandler(model_path=output_dir, onnx_file_name=lm_head_model_path.name)
+    new_llm_pipeline["lm_head"] = "lm_head"
+
+    # create new model attributes
+    new_model_attributes = deepcopy(model.model_attributes) or {}
+    new_model_attributes["llm_pipeline"] = new_llm_pipeline
+
+    # update genai_config if it exists
```
devang-ml, line 469: You may want to refactor this into separate utility.

olive/passes/onnx/common.py
jambayk: Added some details about the "llm_pipeline" attribute under the staticllm pass docstring which creates the first model with this attribute.

olive/passes/onnx/common.py
jambayk: There are other model types such as phi3, llama, phi3v, phi3mm but those are for the directly supported model types with the dynamic shapes, etc. 
For our static llm pipeline models, this `decoder-pipeline` is the only model type. It is essentially a pipeline of models that are run one after another with some assumptions on the io.

olive/passes/onnx/common.py
jambayk: done. This is currently only used for llm pipeline models so the code is mostly just copied over but can be updated later to become generic if needed for other other new pipeline like model types.

PR #1730, opened 2025-04-07
add stable diffusion qdq

xieofxie:


examples/stable_diffusion/README.md
```diff
@@ -187,3 +187,35 @@ Inference will loop until the generated image. The result will be saved as `resu
...
+
+`python stable_diffusion.py --model_id stabilityai/sd-turbo --provider qdq --optimize --only_conversion`
+
+Then generate data:
+
+`python .\evaluation.py --save_data --model_id stabilityai/sd-turbo --num_inference_steps 1 --seed 0 --num_data 100 --guidance_scale 0`
+
+### Optimize
+
+`python stable_diffusion.py --model_id stabilityai/sd-turbo --provider qdq --optimize --clean_cache`
```
jambayk, line 207: do we want to clean cache here? Wouldn't it be faster to reuse the converted model from the first step?

examples/stable_diffusion/sd_utils/qdq.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+def update_qdq_config(config: Dict, submodel_name: str):
+    used_passes = {}
+    if sd_utils.config.only_conversion:
+        used_passes = { "convert" }
+    # TODO(hualxie): onnx or onnxruntime needs to fix this
+    elif submodel_name == "unet":
+        config["input_model"]["io_config"]["dynamic_axes"] = None
+        config["input_model"]["io_config"]["input_shapes"] = [
```
jambayk, line 28: is there a reason you only do this for unet but use the dynamic_shape_to_fixed shape for the other models?

examples/stable_diffusion/README.md
```diff
@@ -187,3 +187,35 @@ Inference will loop until the generated image. The result will be saved as `resu
...
+
+## Stable Diffusion Quantization encoded in QDQ format
+
+### Generate data for static quantization
+
+To get better result, we need to generate real data from original model instead of using random data for static quantization.
+
+First generate onnx unoptimized model:
+
+`python stable_diffusion.py --model_id stabilityai/sd-turbo --provider qdq --optimize --only_conversion`
```
xiaoyu-work, line 199: The `provider` definition in this script means which hardware backend to use. Can we use a different argument like `--format` for qdq?

examples/stable_diffusion/evaluation.py
```diff
@@ -0,0 +1,267 @@
...
+            logger.info(prompt)
+            if i < train_num:
+                pipeline.save_data_dir = script_dir / args.data_dir / "data" / prompt
+                os.makedirs(pipeline.save_data_dir, exist_ok=True)
+            else:
+                pipeline.save_data_dir = None
+            run_inference(pipeline, args, prompt, unoptimized_path)
+        get_clip_scores(prompts, unoptimized_path, clip_score_fn)
+        get_fid_scores(prompts, unoptimized_path, real_images)
+        #get_hpsv2_scores(unoptimized_path / "hpsv2", partial(run_inference, pipeline, args, pathIsFile=True))
```
xiaoyu-work, line 253: Still need this?

examples/stable_diffusion/README.md
xiaoyu-work, line 199: Generating `generate onnx unoptimized model` with `--optimize` option is confusing. Any reason we only need exported model without optimizing?

xieofxie: I need this to generate data for static quantization. See the modified OnnxStableDiffusionPipelineWithSave

examples/stable_diffusion/evaluation.py
xieofxie: I am running it for some results, will update soon. This needs 800 images so commented out. I will add a switch for it

examples/stable_diffusion/README.md
xieofxie: maybe I will rename it to cpu_qdq? as it is only a kind of optimization recipe

xiaoyu-work: Even for qdq, it can be run on all devices. Just your config's the default device is CPU. I would keep provider option for qdq and add `--format qdq` or `--qdq` for this recipe.

examples/stable_diffusion/sd_utils/qdq.py
xieofxie: oh.. yes it is fixed now. This part of code is old

examples/stable_diffusion/evaluation.py
xieofxie: updated result

examples/stable_diffusion/README.md
xieofxie: use --provider cpu --format qdq

examples/stable_diffusion/stable_diffusion.py
```diff
@@ -190,13 +190,18 @@ def update_config_with_provider(config: Dict, provider: str):
...
+        return update_qdq_config(config, submodel_name)
     else:
-        raise ValueError(f"Unsupported provider: {provider}")
+        raise ValueError(f"Unsupported provider: {provider} with format: {modelFormat}")
 
 
 def optimize(
     model_id: str,
     provider: str,
+    modelFormat: str,
```
jambayk, line 204: nit: snake case `model_format` to be consistent with the rest of the code

examples/stable_diffusion/stable_diffusion.py
```diff
@@ -190,13 +190,18 @@ def update_config_with_provider(config: Dict, provider: str):
         from sd_utils.ov import update_ov_config
 
         return update_ov_config(config)
+    elif provider == "cpu" and modelFormat == "qdq":
```
jambayk, line 193: in theory, the same qdq model can be run on cuda so it is not limited to cpu. there is nothing in the passes being run that makes the final model different on cpu/cuda. 

just a thought since it might be faster to run both the model inference and quantization calibration on cuda if available.

xieofxie: updated

examples/stable_diffusion/stable_diffusion.py
```diff
@@ -190,18 +190,18 @@ def update_config_with_provider(config: Dict, provider: str, modelFormat: str, s
         from sd_utils.ov import update_ov_config
 
         return update_ov_config(config)
-    elif provider == "cpu" and modelFormat == "qdq":
+    elif provider in ("cpu", "cuda", "qnn") and model_format == "qdq":
```
jambayk, line 193: the condition at 185 also takes over the cuda, qdq case

jambayk (approved):

PR #1732, opened 2025-04-08
Enable EP Context Encapsulation with OpenVINO models

kaunain-98:
## Describe your changes
Enable creation of static/dynamic models from huggingface to openvino models
Enable creation of static/dynamic models from openvino models
Enhance quantization capabilities for openvino models
Enable encapsulating openvino models with an ep context node within onnx models
Update documentation as required


olive/passes/openvino/conversion.py
```diff
@@ -19,22 +18,24 @@ class OpenVINOConversion(Pass):
     def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
         return {
             **get_user_script_data_config(),
-            "input": PassConfigParam(
+            "input_shapes": PassConfigParam(
```
jambayk, line 21: not sure if we should do it here but I have been thinking about using the pre-existing `io_config` and `dummy_input_func` functionalities in the pytorch and modelhandler. like we use them in the onnx conversion pass https://github.com/microsoft/Olive/blob/4f751e9d61a46fe1cf560a9b66d92adc21d05255/olive/passes/onnx/conversion.py#L499. 
This has the added benefit of automatically using optimum to get the io_config (input names, shapes, output names) and dummy input for hf models. we would not need to maintain a separate userscript just for this pass.

olive/passes/openvino/encapsulation.py
```diff
@@ -0,0 +1,226 @@
...
+        context_model_output_dir = Path(output_model_path) / (context_model_output)
+        save(model_def, context_model_output_dir)
+        
+        xml_file = input_dir.with_suffix(".xml")
+        bin_file = input_dir.with_suffix(".bin")
+        if not config.shared_cache :
+            shutil.copy2(xml_file, output_model_path)
+            shutil.copy2(bin_file, output_model_path)
+        
+        return ONNXModelHandler(model_path=output_model_path)
```
jambayk, line 226: just to be safe, can you also provide `onnx_file_name=f"{model_name}.onnx"` here in case the directory has multiple .onnx files?

olive/passes/openvino/conversion.py
```diff
@@ -58,8 +59,19 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
                 type_=str,
```
jambayk, line 59: should we remove this option and just use `model` and `ov_model` by default? `output_model` is confusing and we probably won't use this option.

jambayk:
please follow the instructions at https://github.com/microsoft/Olive/blob/main/CONTRIBUTING.md#linting to do the linting. thanks!

olive/passes/openvino/conversion.py
kaunain-98: I agree with you but the downside is that we would be reliant on optimum. Since we cannot guarantee optimum support (for all models at all times), It would be beneficial to have a way for users to create their own inputs

olive/passes/openvino/conversion.py
kaunain-98: ok. will remove this option

olive/passes/openvino/encapsulation.py
kaunain-98: ok will make the changes

olive/passes/openvino/conversion.py
jambayk: we allow users to provide their own io config and dummy inputs for both pytorch and hf models. optimum is only an optional dependency and will be used if the user hasn't provided their own.

olive/passes/openvino/conversion.py
```diff
@@ -79,32 +91,38 @@ def _run_for_config(
...
         extra_configs = config.extra_configs or {}
         args = {
             "input_model": input_model,
-            "input": input_shape,
             "example_input": example_input,
             **extra_configs,
         }
 
-        ov_model = ov.convert_model(**args)
+        if config.static:
```
xiaoyu-work, line 119: We pin openvino==2023.2.0, is this `static`/`dynamic` supported by 2023.2.0? 2025 released so we may also need to migrate to 2025 as well.

olive/passes/openvino/encapsulation.py
```diff
@@ -0,0 +1,226 @@
...
+            "target_device": PassConfigParam(
+                type_=Device,
+                default_value=accelerator_spec.accelerator_type.CPU,
+                required=False,
+                description=(
+                    "Device the encapsulated model should run on."
+                    "Available devices are cpu, gpu, npu."
+                ),
+            ),
+            "shared_cache": PassConfigParam(
```
xiaoyu-work, line 58: This term is used by Olive cache system already. For better understanding, can you use other names for this? Like `reuse_cache`?

olive/passes/openvino/encapsulation.py
```diff
@@ -0,0 +1,226 @@
...
+                required=False,
+                description="Name of the input OpenVINO model.",
+            ),
+            "ov_version": PassConfigParam(
+                type_=str,
+                default_value=None,
+                required=False,
+                description=(
+                    "Name of the OpenVINO version to override in model SDK version."
+                    "Requires a minimum version of OpenVino 2025.1"
```
xiaoyu-work, line 81: Is this a user input? Can you explain how this will be used? Can you also add this version format example into the description?

olive/passes/openvino/reshape.py
```diff
@@ -0,0 +1,141 @@
...
+from pathlib import Path
+from typing import Dict,Type, Union
+from olive.hardware.accelerator import AcceleratorSpec
+from olive.model import OpenVINOModelHandler
+from olive.passes import Pass
+from olive.passes.pass_config import BasePassConfig, PassConfigParam, get_user_script_data_config
+import shutil
+
+class OpenVINOReshape(Pass):
+    """Converts PyTorch, ONNX or TensorFlow Model to OpenVino Model."""
```
xiaoyu-work, line 14: Can you please update this?

xiaoyu-work:
Can you please add unit tests for new passes?

olive/passes/openvino/conversion.py
jambayk: we can do this later since these pass options already exist in the pass and is not really in scope for this PR.

olive/passes/openvino/conversion.py
kaunain-98: please migrate to latest version of OpenVINO 2025

olive/passes/openvino/encapsulation.py
kaunain-98: Yes this is a user input which can be added in the config files. We wanted to provided the functionality of encapsulating the openVINO model to the version present in the target system as it is likely for quantization and encapsulation to occur in a development eco-system and deployment on a target system. 
It would also allow end users to select the version they want to use (if multiple versions exist) given the benefits they brings between stability, functionality and performance
I have added an example to the documentation

test/unit_test/passes/openvino/test_openvino_encapsulation.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # execute
+    onnx_model = p.run(openvino_model, output_folder)
+
+    # assert
+    assert Path(onnx_model.model_path).exists()
+    assert (Path(onnx_model.model_path)).is_file()
+
+    # cleanup
+    shutil.rmtree(tmp_path)
```
jambayk, line 66: nit: rmtree is not required. the tmp_path directory should get cleaned up automatically

olive/passes/openvino/encapsulation.py
```diff
@@ -0,0 +1,237 @@
...
+    ) -> ONNXModelHandler:
+        try:
+            import openvino as ov
+        except ImportError:
+            raise ImportError("Please install olive-ai[openvino] to use OpenVINO model") from None
+
+        if config.reuse_cache:
+            output_model_path = model.model_path
+
+        if config.input_model:
```
jambayk, line 118: do we need this option? since this is the name of the .xml,.bin files in the input model handler, the name will always need to be the same as `model.model_config["model_name"]`

olive/passes/openvino/reshape.py
```diff
@@ -0,0 +1,146 @@
...
+        model: Union[OpenVINOModelHandler],
+        config: Type[BasePassConfig],
+        output_model_path: str,
+    ) -> OpenVINOModelHandler:
+        try:
+            import openvino as ov
+        except ImportError:
+            raise ImportError("Please install olive-ai[openvino] to use OpenVINO model") from None
+
+        if config.input_model:
```
jambayk, line 82: same here for the model_name being inferred from the `model`

olive/passes/openvino/reshape.py
```diff
@@ -0,0 +1,146 @@
...
+            elif not config.input_shapes:
+                msg = "Error! Missing input shapes"
+                raise Exception(msg) from None
+
+            else:
+                msg = f"Error! The number of inputs in model {len(loaded_model.inputs)}"
+                "do not match the number of entries in the input_shapes in config, {len(config.input_shapes)}"
+                raise Exception(msg) from None
+
+        else:
```
jambayk, line 132: there is no case for `reuse_cache=True` in this block. does the modified model not need to be saved again? since we are using the same folder and same `model_name` as the input model, would it not override the original .xml file?

olive/passes/openvino/encapsulation.py
```diff
@@ -0,0 +1,237 @@
...
+from olive.hardware.accelerator import AcceleratorSpec, Device
+from olive.model import ONNXModelHandler, OpenVINOModelHandler
+from olive.passes import Pass
+from olive.passes.pass_config import BasePassConfig, PassConfigParam
+
+
+class OpenVINOEncapsulation(Pass):
+    """Encapsulates OpenVINO models with onnx context nodes."""
+
+    openvino_to_onnx_dtype : ClassVar[dict] = {
```
xiaoyu-work, line 21: Nit: all upper-case variable name for static constant?

olive/passes/openvino/encapsulation.py
```diff
@@ -0,0 +1,237 @@
...
+        if config.reuse_cache:
+            output_model_path = model.model_path
+
+        if config.input_model:
+            model_name = config.input_model
+        else:
+            model_name = model.model_config["model_name"]
+
+        if config.ov_version:
+            ov_version = config.ov_version
```
xiaoyu-work, line 124: Do we need to validate the format of this str?

olive/passes/openvino/reshape.py
```diff
@@ -0,0 +1,146 @@
...
+            **get_user_script_data_config(),
+            "extra_configs": PassConfigParam(
+                type_=Dict,
+                default_value=None,
+                required=False,
+                description=(
+                    "Extra configurations for OpenVINO model conversion. extra_config can be set by "
+                    "passing a dictionary where key is the parameter name, and the value is the parameter value. "
+                    "Please check Conversion Parameters documentation for more details: "
+                    "https://docs.openvino.ai/2023.3/openvino_docs_OV_Converter_UG_Conversion_Options.html"
```
xiaoyu-work, line 29: Do we want to use latest `https://docs.openvino.ai/2025/openvino-workflow/model-preparation/conversion-parameters.html` instead?

docs/source/features/ihv-integration/openvino.md
```diff
@@ -31,7 +31,7 @@ Please refer to [OpenVINOConversion](openvino_conversion) for more details about
 ```json
 {
     "type": "OpenVINOConversion",
-    "input": [1, 3, 32, 32]
+    "input_shapes": [1, 3, 32, 32]
```
xiaoyu-work, line 34: I saw we are using "input_shapes": [[1, 3, 32, 32]] for testing, should this be updated to double braces? Any reason that the input shape becomes double braces from a list?

kaunain-98: it should always be double braces to accommodate multiple inputs, updating the docs

olive/passes/openvino/encapsulation.py
kaunain-98: since the user can override the ov version to one that may not be present in the system such as in cases where the model is being created for a deployment system. we cannot validate the specific version. By default we are fetching the current version from the current system. hence its on the user to provide the correct version if they choose to override it

olive/passes/openvino/reshape.py
```diff
@@ -0,0 +1,115 @@
...
+                )
+                raise ValueError(msg) from None
+
+        elif update_io_names:
+            ov.save_model(loaded_model, Path(output_model_path) / (f"{model_name}.xml"))
+
+        else:
+            model_name_path_dst = Path(output_model_path) / (f"{model_name}.xml")
+            weight_name_path_dst = Path(output_model_path) / (f"{model_name}.bin")
+            hardlink_copy_file(model_name_path, model_name_path_dst, follow_symlinks=True)
```
jambayk, line 112: curious about this part. since the model shapes have been updated, shouldn't the xml file be different and need to be saved from the new model? if we copy the both the xml and bin files, would that be just a copy of the input model?

kaunain-98: Model shapes would be updated only be updated if the model is a static models and would go via if static path.
The elif section is for dynamic models that require io name changes and therefore a new xml and bin is generated
The else section of the code essentially is for dynamic models that require no modifications. hence we can just reuse the input model xml and bin files

jambayk: I see. so the idea is that if the io names didn't change, the shape/dim names don't matter and we just copy over the input model?

kaunain-98: for dynamic models yes

jambayk (approved):

xiaoyu-work:
Closing as https://github.com/microsoft/Olive/pull/1747 got merged.

PR #1741, opened 2025-04-10
Update default conversion opset & add --static to conv cli

xiaoyu-work:
## Describe your changes
1. Update default conversion opset.2
2. add --static to capture-onnx-graph cli to export static input shape onnx model.

### DIFF
--- a/olive/cli/capture_onnx.py
+++ b/olive/cli/capture_onnx.py
@@ -31,8 +31,10 @@ class ModelBuilderAccuracyLevel(IntEnumBase):
 def parse_dim_dict(s):
     try:
         return {k: int(v) if v.isdigit() else v for k, v in (item.split("=") for item in s.split(","))}
-    except Exception:
-        raise argparse.ArgumentTypeError("Format must be key=value,... with positive integers as values")
+    except Exception as exc:
+        raise argparse.ArgumentTypeError(
+            "Format must be key=value,... with positive integers as values"
+        ) from exc
 
 
 class CaptureOnnxGraphCommand(BaseOliveCLICommand):
--- a/olive/cli/capture_onnx.py
+++ b/olive/cli/capture_onnx.py
@@ -2,6 +2,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+import argparse
 from argparse import ArgumentParser
 from copy import deepcopy
 from typing import Dict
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/conversion.py
```diff
@@ -113,8 +113,8 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassCon
                 default_value=True,
                 description="Whether to export the model with constant folding and redundancies elimination.",
             ),
-            "dynamic": PassConfigParam(
-                type_=bool, default_value=True, description="Whether to export the model with dynamic axes/shapes."
+            "static": PassConfigParam(
```
jambayk, line 116: is there a reason for this change? also the description needs to be updated.

olive/cli/capture_onnx.py
```diff
@@ -196,6 +199,7 @@ def _get_run_config(self, tempdir: str) -> Dict:
                     (("passes", "c", "torch_dtype"), self.args.torch_dtype),
                     (("passes", "c", "target_opset"), self.args.target_opset),
                     (("passes", "c", "use_dynamo_exporter"), self.args.use_dynamo_exporter),
+                    (("passes", "c", "static"), self.args.static),
```
jambayk, line 202: a possible issue with this is that model shapes would be fixed to the dummy inputs we get from optimum for hf models. for static shapes, I think we would want the user to decide the exact shape values.

jambayk: so maybe it's not worth exposing currently

xiaoyu-work: Customers reach out to me about how to convert models with static shape, while they are expecting a cli to do so instead of a config file (too expensive for them to learn how to write a config file). I'm thinking having default static shape might be okay for them they can check onnx model file for the shape themselves. While if they want different shape values, they may need a config file.

olive/passes/onnx/conversion.py
xiaoyu-work: I want to make this option matches cli option. `static` option for cli is more straightforward than `dynamic`.

olive/cli/capture_onnx.py
jambayk: I don't think the default shape coming out of optimum is useful for much. The batch size is 2 and tthe other dims are also small. 

It might be more useful to maybe expose some options to provide values for dimensions and we do dynamic to fixed shape pass after the conversion.

xiaoyu-work: Sounds good. I can add an optional option for this,

olive/passes/onnx/conversion.py
xiaoyu-work: Reverted this change as no need any more.

olive/cli/capture_onnx.py
```diff
@@ -174,6 +191,7 @@ def _get_run_config(self, tempdir: str) -> Dict:
         ]
         if self.args.use_model_builder:
             del config["passes"]["c"]
+            del config["passes"]["f"]
```
jambayk, line 194: we could fix shapes coming out of the model builder too if we want.

jambayk (approved):

PR #1746, opened 2025-04-11
Add MatMulAddGemm Graph Surgery

jambayk:
## Describe your changes
Add MatMulAddGemm Graph Surgery to transform Add Op followed by Matmul as Gemm op

Contributed by @gnedanur, #1745

### DIFF
--- a/docs/source/features/onnx-transformations.md
+++ b/docs/source/features/onnx-transformations.md
@@ -833,6 +833,67 @@ Transformed model graph:
             (p=2, axis=-1)
 ```
 
+### `MatMulAddToGemm`
+
+#### Description
+
+Replace MatMul + Add with Gemm.
+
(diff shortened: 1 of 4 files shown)


xiaoyu-work (approved):

justinchuby:
Just curious: Where is this needed? In pytorch export we have the exact opposite of this rule that converts reshape-gemm-reshape to matmul-add. I wonder if we should disable the rule?

gnedanur:
> Just curious: Where is this needed? In pytorch export we have the exact opposite of this rule that converts reshape-gemm-reshape to matmul-add. I wonder if we should disable the rule?

It depends on the model and dimensions of the tensors. In some cases, Reshapes are costly so Reshape-gemm-Reshape is more costly on NPU and in some cases it will be more optimal than Matmul + Add.

PR #1750, opened 2025-04-15
ONNX Script: Convert non-4D MatMul to Transpose-Conv-Transpose sequence 

yf711:
## Describe your changes
Add a transform pass (MatMul to Transpose-Conv-Transpose sequence) 
which is needed by TensorRT EP to meet certain hardware requirement.
(The final recipe requires multiple passes to work together. Will add more pass in the following PR to decouple unsqueeze/squeeze and upscale non-4D weight initializers to 4D)


olive/passes/onnx/onnxscript_dla_transforms.py
```diff
@@ -0,0 +1,111 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+import logging
+from pathlib import Path
+from typing import Dict, Type
+
+import onnx
+import onnxscript
```
xiaoyu-work, line 10: onnxscript is not a dependency of Olive. Better to move this import to `_run_for_config` and check import error.

olive/passes/onnx/onnxscript_dla_transforms.py
```diff
@@ -0,0 +1,111 @@
...
+from olive.model import ONNXModelHandler
+from olive.model.utils import resolve_onnx_path
+from olive.passes import Pass
+from olive.passes.onnx.common import get_external_data_config, model_proto_to_olive_model
+from olive.passes.pass_config import BasePassConfig, PassConfigParam
+
+logger = logging.getLogger(__name__)
+
+
+class MatMulToConvTransform(Pass):
```
xiaoyu-work, line 23: Can you add a docstring here to explain what does this pass do? That will be useful for auto-generated doc.

olive/passes/onnx/onnxscript_dla_transforms.py
jambayk: It was recently made a dependency since torch export now uses it underneath https://github.com/microsoft/Olive/blob/96133c857c29191a76bb66ac2339e7341b2d3341/requirements.txt#L3

olive/passes/onnx/onnxscript_dla_transforms.py
```diff
@@ -0,0 +1,111 @@
...
+        def matmul_pattern(op, input_a, input_b):
+            return op.MatMul(input_a, input_b)
+
+        def matmul_to_conv_replacement(op, input_a, input_b):
+            input_a_rank = len(input_a.shape)
+            input_b_rank = len(input_b.shape)
+            
+            # Only handle 2D�2D and 3D�2D cases
+            if not ((input_a_rank == 2 and input_b_rank == 2) or (input_a_rank == 3 and input_b_rank == 2)):
+                logger.info(f"Skipping MatMul with shapes {input_a.shape} x {input_b.shape}")
```
jambayk, line 43: Please use debug level for this log since we use info level for higher workflow level logs.

olive/passes/onnx/onnxscript_dla_transforms.py
jambayk, line 23: Maybe we could make this a graph surgery instead of a standalone pass? What do you think @xiaoyu-work 

https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/graph_surgeries.py#L1096

olive/passes/onnx/onnxscript_dla_transforms.py
```diff
@@ -0,0 +1,111 @@
...
+        
+        def matmul_pattern(op, input_a, input_b):
+            return op.MatMul(input_a, input_b)
+
+        def matmul_to_conv_replacement(op, input_a, input_b):
+            input_a_rank = len(input_a.shape)
+            input_b_rank = len(input_b.shape)
+            
+            # Only handle 2D�2D and 3D�2D cases
+            if not ((input_a_rank == 2 and input_b_rank == 2) or (input_a_rank == 3 and input_b_rank == 2)):
```
jambayk, line 42: Should we also limit second inputs to initializers for now? We might have matmuls whose weights are already quantized using DequantizeLinear. For such cases, we would need to modify the DequatizeLinear's inputs so that we have a valid `DQ->Conv` node group

olive/passes/onnx/onnxscript_dla_transforms.py
```diff
@@ -0,0 +1,111 @@
...
+            # Step 2: Transpose for Conv format
+            if input_a_rank == 2:
+                # [1, M, K, 1] to [1, K, M, 1]
+                transposed_a = op.Transpose(padded_a, perm=[0, 2, 1, 3])
+            elif input_a_rank == 3:
+                # [B, M, K, 1] to [B, K, M, 1]
+                transposed_a = op.Transpose(padded_a, perm=[0, 2, 1, 3])
+                
+            # [1, K, N, 1] to [N, K, 1, 1]
+            transposed_b = op.Transpose(padded_b, perm=[2, 1, 0, 3])
```
jambayk, line 72: Just curious. In the case that b is an initializer, do the unsqueeze and transpose operations on the weight get constant folded so that we have a new 4d initializer as the conv weight?

jambayk:
please follow the instructions at https://github.com/microsoft/Olive/blob/main/CONTRIBUTING.md#linting to apply most of the suggested lint changes automatically

olive/passes/onnx/onnxscript_dla_transforms.py
justinchuby: I wonder if it makes sense to create a olive configurable `Pass` for onnxscript rewrite rules. This way (1) users gets a uniform config experience with olive for composing passes (2) developers are able to leverage the onnxscript rewriter for easy rewrite rule authoring. Manipulating protobufs directly in newly added passes may create higher maintanance cost down the line.

justinchuby: At the same time, is this pass for ORT or is it optional for specific EPs? There is a possibility to include it in https://github.com/microsoft/onnxscript/blob/df26586a116f071b0cce9f005aedd4725ed2d3cb/onnxscript/rewriter/ort_fusions/_core.py#L31, which will be applied in https://github.com/microsoft/Olive/blob/96133c857c29191a76bb66ac2339e7341b2d3341/olive/passes/onnx/onnxscript_fusion.py#L39

olive/passes/onnx/onnxscript_dla_transforms.py
jambayk: or this would probably be done by the ort session graph optimizer during load time?

olive/passes/onnx/onnxscript_dla_transforms.py
yf711: Thanks for the feedback! To clarify, this pass is specifically designed for the TensorRT EP. It addresses certain hardware limitations, ensuring compatibility and optimal performance. 
While this pass is not intended as a general-purpose pass for all EPs, I plan to insert other generic passes to graph surgery

yf711: Sure, description has been added

olive/passes/onnx/onnxscript_dla_transforms.py
yf711: Applied

olive/passes/onnx/onnxscript_dla_transforms.py
yf711: Sure, this is the primary use case. I've applied filter on that.

olive/passes/onnx/onnxscript_dla_transforms.py
yf711: > Just curious. In the case that b is an initializer, do the unsqueeze and transpose operations on the weight get constant folded so that we have a new 4d initializer as the conv weight?

Yeah I think this would probably be the case

jambayk (approved):

olive/passes/onnx/onnxscript_dla_transforms.py
jambayk: just to be clear, do you think the constant folding will be done by the rewriter or the ort session during init?

olive/passes/onnx/tensorrt/trt_dla_transforms.py
```diff
@@ -0,0 +1,106 @@
...
+from olive.model import ONNXModelHandler
+from olive.model.utils import resolve_onnx_path
+from olive.passes import Pass
+from olive.passes.onnx.common import get_external_data_config, model_proto_to_olive_model
+from olive.passes.pass_config import BasePassConfig, PassConfigParam
+
+logger = logging.getLogger(__name__)
+
+
+class TrtMatMulToConvTransform(Pass):
```
jambayk, line 22: please add this pass in https://github.com/microsoft/Olive/blob/main/olive/olive_config.json

olive/passes/onnx/onnxscript_dla_transforms.py
yf711: I think it would be done during ort session init, as long as ort optimization level is higher than the basic

olive/passes/onnx/tensorrt/trt_dla_transforms.py
yf711: Thanks! The pass is added

olive/passes/onnx/tensorrt/trt_dla_transforms.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @classmethod
+    def _default_config(cls, accelerator_spec: AcceleratorSpec) -> Dict[str, PassConfigParam]:
+        return get_external_data_config()
+
+    def _run_for_config(
+        self, model: ONNXModelHandler, config: Type[BasePassConfig], output_model_path: str
+    ) -> ONNXModelHandler:
+        output_model_path = resolve_onnx_path(output_model_path, Path(model.model_path).name)
+        model_proto = model.load_model()
```
justinchuby, line 33: Use ir.load() to load the model

olive/passes/onnx/tensorrt/trt_dla_transforms.py
```diff
@@ -0,0 +1,106 @@
...
+
+        # Create a set of initializer names for quick lookup
+        initializer_names = {initializer.name for initializer in model_proto.graph.initializer}
```
justinchuby, line 36:
```suggestion
        initializer_names = set(model.graph.initializers)
```

olive/passes/onnx/tensorrt/trt_dla_transforms.py
```diff
@@ -0,0 +1,106 @@
...
+            is_input_b_initializer = input_b.name in initializer_names
+
+            return is_non_4d and is_input_b_initializer
+
+        matmul_to_conv_rule = pattern.RewriteRule(
+            matmul_pattern, matmul_to_conv_replacement, is_valid_for_transform, verbose=1
+        )
+
+        transformed_model_proto = onnxscript.rewriter.rewrite(
+            model_proto,
```
justinchuby, line 101:
```suggestion
            model,
```

Prefer an IR model.

olive/passes/onnx/tensorrt/trt_dla_transforms.py
justinchuby: Edit: NVM. I will update ONNXModelHandler and refactor if needed

justinchuby, line 33:
```suggestion
        model = ir.load(model.model_path)
```

Use ir.load() to load the model

olive/passes/onnx/tensorrt/trt_dla_transforms.py
```diff
@@ -0,0 +1,106 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+import logging
+from pathlib import Path
+from typing import Dict, Type
+
+import onnxscript
+from onnxscript.rewriter import pattern
```
justinchuby, line 10:
```suggestion
from onnxscript.rewriter import pattern
from onnxscript import ir
```

olive/passes/onnx/tensorrt/trt_dla_transforms.py
justinchuby: You may now use `model.load_ir_model()` from https://github.com/microsoft/Olive/pull/1771

jambayk (approved):

PR #1751, opened 2025-04-15
Use ONNX IR in graph surgeries

justinchuby:
## Describe your changes

Use ONNX IR in graph surgeries to replace some proto manipulation. This PR adds a `call_ir` in the base class that is called when `__call__` is not overriden. This way the changes can be incremental and compatible with existing usages of the `Surgeon` class.

Refactoring:
- Renamed the Surgeon class that operates on proto to ProtoSurgeon to make the intention clear. Developers can subclass ProtoSurgeon to operate on proto and subclass Surgion to transform graphs directly with IR.

### DIFF
--- a/olive/passes/onnx/graph_surgeries.py
+++ b/olive/passes/onnx/graph_surgeries.py
@@ -3,13 +3,11 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 
-# ruff: noqa: RUF012
-
 import inspect
 import logging
 import math
 from pathlib import Path
-from typing import Any, ClassVar, Dict, List, Optional, Sequence, Type
+from typing import Any, ClassVar, Dict, List, Mapping, Optional, Sequence, Type
 
 import numpy as np
 import onnx
@@ -160,7 +158,7 @@ class ReorderInputs(Surgeon):
 
 
 class ReplaceErfWithTanh(ProtoSurgeon):
-    DTYPE_MAP = {
+    DTYPE_MAP: Mapping = {
         TensorProto.FLOAT: np.float32,
         TensorProto.FLOAT16: np.float16,
         TensorProto.DOUBLE: np.float64,
--- a/olive/passes/onnx/graph_surgeries.py
+++ b/olive/passes/onnx/graph_surgeries.py
@@ -147,7 +147,7 @@ class ReorderInputs(Surgeon):
             raise ValueError("Invalid permutation: permutation must be a rearrangement of input indices.")
 
         reordered_inputs = [inputs[idx] for idx in self.permutation]
-        del model.graph.inputs[:]
+        model.graph.inputs.clear()
         model.graph.inputs.extend(reordered_inputs)
 
         return model
--- a/olive/passes/onnx/graph_surgeries.py
+++ b/olive/passes/onnx/graph_surgeries.py
@@ -29,6 +29,11 @@ logger = logging.getLogger(__name__)
 
 
 class Surgeon:
+    """Base class for surgeons that operate on the ONNX IR model."""
+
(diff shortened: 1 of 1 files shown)


olive/passes/onnx/graph_surgeries.py
```diff
@@ -35,7 +36,10 @@ def __init_subclass__(cls, **kwargs):
         super().__init_subclass__(**kwargs)
         Surgeon.registry[cls.__name__.lower()] = cls
 
-    def __call__(self, model: ModelProto):
+    def __call__(self, model: ModelProto) -> ModelProto:
+        return ir.to_proto(self.call(ir.from_proto(model)))
```
justinchuby, line 40:
```suggestion
        return ir.to_proto(self.call_ir(ir.from_proto(model)))
```

olive/passes/onnx/graph_surgeries.py
```diff
@@ -78,50 +82,23 @@ def __init__(self, old_names: List[str], new_names: List[str]):
...
-                if graph_input.name == old_name:
-                    model.graph.input[idx].name = new_name
-
-            for idx, graph_output in enumerate(model.graph.output):
-                if graph_output.name == old_name:
-                    model.graph.output[idx].name = new_name
-        return model
+    def call_ir(self, model: ir.Model) -> ir.Model:
+        replacement = dict(zip(self.old_names, self.new_names))
+        for input in model.graph.inputs:
```
justinchuby, line 87: Using ONNX IR is more efficient and simple: this now becomes an `O(#input)` operation.

olive/passes/onnx/graph_surgeries.py
```diff
@@ -147,28 +126,29 @@ class RemoveInitializerFromInputs(Surgeon):
...
+        inputs = list(model.graph.inputs)
         num_inputs = len(inputs)
 
         if sorted(self.permutation) != list(range(num_inputs)):
             raise ValueError("Invalid permutation: permutation must be a rearrangement of input indices.")
 
         reordered_inputs = [inputs[idx] for idx in self.permutation]
-        del model.graph.input[:]
-        model.graph.input.extend(reordered_inputs)
+        del model.graph.inputs[:]
```
justinchuby, line 150:
```suggestion
        model.graph.inputs.clear()
```

olive/passes/onnx/graph_surgeries.py
```diff
@@ -147,28 +126,29 @@ class RemoveInitializerFromInputs(Surgeon):
     def __init__(self):
         pass
 
-    def __call__(self, model: ModelProto):
-        initializer_names = {initializer.name for initializer in model.graph.initializer}
-        updated_inputs = [graph_input for graph_input in model.graph.input if graph_input.name not in initializer_names]
-        del model.graph.input[:]
-        model.graph.input.extend(updated_inputs)
+    def call_ir(self, model: ir.Model) -> ir.Model:
+        while model.graph.inputs[-1].name in model.graph.initializers:
```
xiaoyu-work, line 130: Nit: may not be a real case like all inputs are initializer then this will get index errors. Maybe some testing models can have this case?

xiaoyu-work:
Thanks for the change! This makes this pass cleaner. Can you add a documentation link about what is IR model and how to operate IR model in the docstring of this pass? So later when people want to add new surgeon to this pass they know what to refer.

olive/passes/onnx/graph_surgeries.py
justinchuby: That’s a valid point. Will check size first

justinchuby: Done

justinchuby:
> Thanks for the change! This makes this pass cleaner. Can you add a documentation link about what is IR model and how to operate IR model in the docstring of this pass? So later when people want to add new surgeon to this pass they know where to refer.

Added, thanks

xiaoyu-work (approved):

PR #1666, opened 2025-03-06
Simplify quantize command

devang-ml:
## Describe your changes
- Simplify quantize command implementation
- Enable static quantization

## Checklist before requesting a review
- [ ] Add unit tests for this change.
- [ ] Make sure all tests can pass.
- [ ] Update documents if necessary.
- [ ] Lint and apply fixes to your code by running `lintrunner -a`
- [ ] Is this a user-facing change? If yes, give a description of this change to be included in the release notes.
- [ ] Is this PR including examples changes? If yes, please remember to update [example documentation](https://github.com/microsoft/Olive/blob/main/docs/source/examples.md) in a follow-up PR.

PR #1756, opened 2025-04-17
Allow cli to take capture-onnx-graph output model

xiaoyu-work:
## Describe your changes

1. Print CLI output model save path (not sure why this was removed before).
2. Add `input_cols` option to dataset for cli for text-classification task.
3. Update error message when missing dataset, otherwise it will say the precision is not supported.
4. Save `hf_task` to `model_attributes` from Conversion pass, this will be used for later pass.
5. Correct model name and task for dataset config when input model is `onnxmodel`.

### DIFF
--- a/olive/cli/base.py
+++ b/olive/cli/base.py
@@ -39,7 +39,9 @@ class BaseOliveCLICommand(ABC):
             run_config = self._get_run_config(tempdir)
             if self.args.save_config_file:
                 self._save_config_file(run_config)
-            return olive_run(run_config)
+            output = olive_run(run_config)
+            print(f"Model is saved at {self.args.output_path}")
+            return output
 
     @staticmethod
     def _save_config_file(config: dict):
@@ -505,6 +507,12 @@ def add_dataset_options(sub_parser, required=True, include_train=True, include_e
         default=1,
         help="Batch size.",
     )
+    dataset_group.add_argument(
+        "--input_cols",
+        type=str,
+        nargs="+",
+        help="List of input column names. Provide one or more names separated by space. Example: --input_cols sentence1 sentence2",
+    )
 
     return dataset_group, text_group
 
@@ -529,6 +537,7 @@ def update_dataset_options(args, config):
         ((*preprocess_key, "chat_template"), args.use_chat_template),
         ((*preprocess_key, "max_seq_len"), args.max_seq_len),
         ((*preprocess_key, "add_special_tokens"), args.add_special_tokens),
+        ((*preprocess_key, "input_cols"), args.input_cols),
         ((*preprocess_key, "max_samples"), args.max_samples),
         ((*dataloader_key, "batch_size"), args.batch_size),
     ]
--- a/olive/cli/base.py
+++ b/olive/cli/base.py
@@ -40,7 +40,10 @@ class BaseOliveCLICommand(ABC):
             if self.args.save_config_file:
                 self._save_config_file(run_config)
             output = olive_run(run_config)
-            print(f"Model is saved at {self.args.output_path}")
+            if output is None:
+                print("No output model produced. Please check the log for details.")
+            else:
+                print(f"Model is saved at {self.args.output_path}")
             return output
 
     @staticmethod
--- a/olive/cli/quantize.py
+++ b/olive/cli/quantize.py
@@ -126,8 +126,8 @@ class QuantizeCommand(BaseOliveCLICommand):
             ):
                 if not self._check_data_name_arg(pinfo):
                     raise ValueError(
-                        f"Dataset is required for pass {r['pass_type']} but no dataset is provided. "
-                        "Please provide a dataset using --data_name option."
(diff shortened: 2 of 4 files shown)


olive/cli/quantize.py
```diff
@@ -123,9 +123,14 @@ def _get_pass_list(self, precision, algo, impl, is_hf_model):
                 and (algo is None or algo in pinfo.supported_algorithms)
                 and (precision is None or precision in pinfo.supported_precisions)
                 and (not self.args.use_qdq_encoding or "qdq" in pinfo.supported_quantization_encodings)
-                and self._check_data_name_arg(pinfo)
```
jambayk, line 126: I think this condition was added to make passes requiring data config to be optional. With this change, data config is not optional?

jambayk: Or maybe the optional logic is not valid anymore since we got rid of pass flows?

xiaoyu-work: I move this check inside this if condition in order to throw a specific data config error. What do you mean?

jambayk: previously it did not fail if there is no data config and just chose other valid passes. At least at the beginning, @shaahji can confirm is it's true. 
Maybe the recent refactor in #1666 changed this behavior. 

if the issue is that there is no valid pass at the end, we could throw the error at the end of the loop and just debug log why the pass requiring data configs were ignored.

shaahji: I raised similar [issue](https://github.com/microsoft/Olive/pull/1666#discussion_r1985785085) on #1666 as well.
See @devang-ml's response on the same issue.

I don't think this change modifies the behavior except for reporting an "explicit" error for cases where a pass requires a data config but none was provided. In previous implementation, it would have reported that pass list came out empty which would be less helpful to the user.

To keep the previous behavior intact, you could report this as a warning.

xiaoyu-work: I see. I don't mean to change this behavior, my goal is to remind the user that a specific pass requires a dataset but no dataset passing to the cli, instead of throw not supporting error.

devang-ml: In summary, presence of data name influences which quantization pass is picked - static or dynamic ?

olive/cli/quantize.py
```diff
@@ -123,9 +123,14 @@ def _get_pass_list(self, precision, algo, impl, is_hf_model):
...
             ):
-                pass_list.append(r["pass_type"])
+                if not self._check_data_name_arg(pinfo):
+                    raise ValueError(
+                        f"Dataset is required for pass {r['pass_type']} but no dataset is provided. "
+                        "Please provide a dataset using --data_name option."
+                    )
```
devang-ml, line 131: Not a good idea. See my comments above.

olive/cli/base.py
```diff
@@ -506,6 +508,12 @@ def add_dataset_options(sub_parser, required=True, include_train=True, include_e
         default=1,
         help="Batch size.",
     )
+    dataset_group.add_argument(
+        "--input_cols",
```
devang-ml, line 512: Auto-opt already has --input_cols option (which should be cleaned).

olive/cli/quantize.py
```diff
@@ -123,13 +123,19 @@ def _get_pass_list(self, precision, algo, impl, is_hf_model):
                 and (algo is None or algo in pinfo.supported_algorithms)
                 and (precision is None or precision in pinfo.supported_precisions)
                 and (not self.args.use_qdq_encoding or "qdq" in pinfo.supported_quantization_encodings)
-                and self._check_data_name_arg(pinfo)
             ):
-                pass_list.append(r["pass_type"])
+                if not self._check_data_name_arg(pinfo):
+                    raise ValueError(
```
jambayk, line 128: doesn't look like the original comment was resolved? this still raised value issue so it still requires data config?

jambayk: might be better to log a warning here saying this pass was not selected since dataset is required but not provided. 

in theory, two passes might pass all of the other conditions but one requires data and the other doesn't

xiaoyu-work: Yes, I've synced with Devang offline. The main purpose for this PR is correcting the error message. For example, for a specific quantization cli argument combination, like int8 + ort +qdq, cli should automatically pick up onnx static quantization, while static quantization needs a dataset, if the user doesn't provide the dataset, current behavior is throw an error "INT8 ORT QDQ is not supported". This is a wrong message cause Olive supports this combination. So for this change, if a specific combination is supported, current error is needing a dataset for this combination. We need to 1. do not tell user which pass we use cause this cli is designed to be a black box. 2. Warning log won't work cause a dataset required pass will be added to the pass list then Olive workflow will fail cause it needs a dataset.

jambayk: it won't be added though. what i meant is skip adding the pass to the list but log a warning. if no pass is added at the end, then we raise an error. 
the current combinations might have your logic and the suggested one be the same. but i theory, there could be two combinations that are same for everything except for the `need dataset` part

jambayk (approved):

PR #1762, opened 2025-04-17
Move generative flag down to PyTorch models

shaahji:
## Move generative flag down to PyTorch models

Contrary to PyTorch models, ONNX models, implement different code paths for generative and non-generative models (specifically, for evaluation). For PyTorch models, the same handlers and evaluators implement the evaluation logic and needs to distinguish between the API (run vs. generate) invoked for inference. So, moved the flag from base OliveHandler down to PyTorchBaseHandler.


olive/model/handler/pytorch.py
```diff
@@ -25,6 +25,23 @@
 class PyTorchModelHandlerBase(OliveModelHandler, DummyInputsMixin, PytorchKvCacheMixin):  # pylint: disable=too-many-ancestors
     """Base class for PyTorch model handler."""
 
+    def __init__(
+        self,
+        model_file_format: ModelFileFormat,
+        model_path: OLIVE_RESOURCE_ANNOTATIONS = None,
+        model_attributes: Optional[Dict[str, Any]] = None,
+        io_config: Union[Dict[str, Any], "IoConfig", str, Callable] = None,
+        generative: bool = True,
```
jambayk, line 34: maybe we can default to false? In the hf model, if the task starts with `text-generation`, it can be set to true.

devang-ml (approved):

PR #1775, opened 2025-04-21
[DO NOT MERGE] Copy of 1764

jambayk:


examples/bert/bert_common.py
```diff
@@ -0,0 +1,206 @@
+from dataclasses import dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING, Optional, Union
+
+import numpy as np
+import torch
+from torch.utils.data import DataLoader
+from tqdm.auto import tqdm
```
xiaoyu-work, line 8: Do we want this? If so, this should be added to requirements.txt

examples/bert/bert_common.py
```diff
@@ -0,0 +1,206 @@
+from dataclasses import dataclass
```
xiaoyu-work, line 1: This should be added to requirements.txt

jambayk: its already in the requirements

examples/bert/bert_common.py
jambayk: I think this gets installed as part of other olive dependencies

xiaoyu-work (approved):

devang-ml (approved):

PR #1778, opened 2025-04-21
Update pytest config for readability

justinchuby:
Currently in CI the pytest errors return full functions in the error stack, which I think is overly verbose and not readable. I updated the ini options taken from https://github.com/microsoft/onnxscript/blob/main/pyproject.toml to make the pytest errors more readable.


jambayk (approved):
Thanks!

PR #1809, opened 2025-04-30
Add OnnxHqqQuantization Pass

xiaoyu-work:
## Describe your changes

Add OnnxHqqQuantization Pass. This quantization only supports 4 bits quantization.


olive/passes/onnx/hqq_quantization.py
```diff
@@ -0,0 +1,342 @@
...
+
+        clean_initializers(loaded_model.graph, loaded_model)
+        # topologically sort the graph at the end since previous optimizations may have broken it
+        model.topological_sort()
+        # model._check_init is not needed since it's only meant for float8 quantization
+
+        # save the model to the output path and return the model
+        return model_proto_to_olive_model(loaded_model, output_model_path, config)
+
+    def _process_subgraph(self, graph_stack: list[onnx.GraphProto]):
```
devang-ml, line 88: Use OnnxDAG to process the graph for now. See how surgery uses it.
OnnxDAG is more efficient, and it already has topological sort utility.

olive/model/handler/onnx.py
```diff
@@ -111,6 +111,21 @@ def load_ir_model(self) -> ir.Model:
         """Load the model as an ONNX IR model object."""
         return ir.load(self.model_path)
 
+    def set_opset_import(self, domain, version):
```
jambayk, line 114: the onnx model handler does not persist the loaded model so this change will not be there in a new loaded model. I think we can do this in the onnxdag itself.

xiaoyu-work: Will update to dag instead.

olive/passes/onnx/hqq_quantization.py
xiaoyu-work: Done

olive/passes/onnx/hqq_quantization.py
```diff
@@ -0,0 +1,316 @@
...
+
+                if len(quantized_nodes) == 1 and quantized_nodes[0].op_type == str(OpType.MatMulNBits):
+                    for initializer in initializers:
+                        dag.add_initializer(initializer, graph_idx)
+
+                    dag.add_node(quantized_nodes[0], graph_idx)
+                    for consumer in dag.get_consumers(node_name):
+                        dag.replace_node_input(consumer, node.proto.output[0], quantized_nodes[0].output[0])
+
+                    dag.remove_node(node_name)
```
jambayk, line 114: this might fail if the node produces a model output like the lm head. It needs to be handled especially where you need to remove the graph output first and the rename the node's output to the original output name.

test/unit_test/passes/onnx/test_hqq_quantization.py
```diff
@@ -0,0 +1,130 @@
...
+        matmul_node = onnx.helper.make_node(
+            str(OpType.MatMul), inputs=[input_name, weight_name], outputs=[matmul_output_name], name="MatMul_Node"
+        )
+
+        # Create ReLU node
+        relu_node = onnx.helper.make_node("Relu", inputs=[matmul_output_name], outputs=[output_name], name="Relu_Node")
+
+        # Create graph with both nodes
+        graph_def = onnx.helper.make_graph(
+            nodes=[matmul_node, relu_node],
```
jambayk, line 51: please add another matmul node which directly leads to the final output of the model. the lm head and classification heads are all matmuls.

olive/passes/onnx/hqq_quantization.py
jambayk: https://github.com/microsoft/Olive/blob/da01b3737ccca4304e83a87683724f046f75c900/olive/passes/onnx/mnb_to_qdq.py#L304

xiaoyu-work: I see. I will add this case support.

jambayk: Thanks! I will try to abstract this node replacement logic in a follow up since it's a common use case.

olive/olive_config.json
```diff
@@ -164,6 +164,13 @@
             "supported_algorithms": [ "rtn", "hqq" ],
             "supported_quantization_encodings": [ "qdq" ]
         },
+        "OnnxHqqQuantization": {
+            "module_path": "olive.passes.onnx.hqq_quantization.OnnxHqqQuantization",
+            "supported_providers": [ "CPUExecutionProvider", "CUDAExecutionProvider", "DmlExecutionProvider" ],
```
jambayk, line 169: should we limit the supported eps? this is because of the matmulnbits op? I think webgpu also supports this operator now. for other eps, we could run the matmulnbits to qdq pass and the model would become compatible

olive/passes/onnx/hqq_quantization.py
```diff
@@ -0,0 +1,332 @@
...
+    def _run_for_config(
+        self, model: ONNXModelHandler, config: type[BasePassConfig], output_model_path: str
+    ) -> ONNXModelHandler:
+        if model_has_adapters(model.model_path):
+            logger.info(
+                "HQQ quantization is not supported for models with adapters. Returning the model without quantization."
+            )
+            return model
+        output_model_path = resolve_onnx_path(output_model_path, Path(model.model_path).name)
+        model_copy = deepcopy(model.load_model())
```
jambayk, line 67: nit: since model.load_model() does not cache the loaded model in the handler, deepcopy is not required

olive/passes/onnx/hqq_quantization.py
```diff
@@ -0,0 +1,332 @@
...
+        dag = OnnxDAG(model_copy)
+        dag.set_opset_import(MSFT_DOMAIN, 1)
+        dag = self._process_graph(dag, config.block_size, config.axis, config.nodes_to_exclude, config.nodes_to_include)
+        dag.clean_unused_initializers()
```
jambayk, line 71: this is probably safer but there is no need to manually clean unused initializers. Since their consumer nodes are removed, during dag.update() the initializers without consumers are not added to the new graph.

jambayk: https://github.com/microsoft/Olive/blob/ebf8954df52d9cf68f157b6697dc497e4bee7ae4/olive/passes/onnx/onnx_dag.py#L896

olive/passes/onnx/hqq_quantization.py
```diff
@@ -0,0 +1,332 @@
...
+        ordered_nodes = dag.topological_sort()
+
+        for node_name in ordered_nodes:
+            node = dag.get_node(node_name)
+
+            if node.proto.name in nodes_to_exclude:
+                logger.debug("exclude to quantize %s as specified by nodes_to_exclude...", node.proto.name)
+                continue
+
+            elif (node.proto.name in nodes_to_include) or (node.op_type == str(OpType.MatMul)):
```
jambayk, line 97: the `or` condition feels a bit weird since it assumes `nodes_to_include` only has matmul nodes but that is not guaranteed.

olive/olive_config.json
xiaoyu-work: Yes i used the same supported ep as OnnxMatMul4Quantizer using. What do you mean by limiting the ep here?

olive/passes/onnx/hqq_quantization.py
xiaoyu-work: Updated

olive/passes/onnx/hqq_quantization.py
xiaoyu-work: Good to know! Updated.

olive/olive_config.json
jambayk: I mean that the list of supported providers is limited. but not an issue since it's only used for auto-opt and the matmul4quantizer has the same as you said.

olive/passes/onnx/hqq_quantization.py
jambayk: I think we have to always make sure the op_type is matmul? just thinking of a case where the node name is in nodes_to_include list but it is not a matmul node.

olive/passes/onnx/hqq_quantization.py
```diff
@@ -0,0 +1,332 @@
...
+
+        nodes_to_exclude = nodes_to_exclude or []
+        nodes_to_include = nodes_to_include or []
+
+        ordered_nodes = dag.topological_sort()
+
+        for node_name in ordered_nodes:
+            node = dag.get_node(node_name)
+
+            if node.proto.name in nodes_to_exclude:
```
jambayk, line 93: nit: node.proto.name is the same as `node_name`

olive/passes/onnx/hqq_quantization.py
xiaoyu-work: That's a valid point. Do you know what does normally `nodes_to_include` mean? It means MatMul && nodes_to_include or MatMul + nodes_to_include? If later, we can nodes_to_include this here. If former, we need a check here.

jambayk: it means the first one here since other ops are not supported anyways. maybe the original implementation in ort was just making this assumption.

jambayk (approved):

PR #1813, opened 2025-05-01
OnnxModelHandler: Remove OnnxGraphMixin

jambayk:
## Describe your changes
- Remove the `OnnxGraphMixin` from `OnnxModelHandler`. We want to use the `OnnxDAG` for any graph manipulations. The model handler should only be used to configure models and load them. Don't want to accidently modify the graph for pass input models when creating output models.
- Some bug fixes on `OnnxDAG.rename_node_output` and `ModelBuilder` with split assignments.


olive/passes/onnx/mixed_precision.py
```diff
@@ -44,20 +44,23 @@ def _run_for_config(
         and set parameters for float16 conversion automatically.
         """
         from onnxruntime.transformers.float16 import float_to_float16_max_diff
+        from onnxruntime.transformers.onnx_model import OnnxModel
+
+        onnx_model = OnnxModel(model.load_model())
```
xiaoyu-work, line 49: Should we use our own dag here?

jambayk: I think we can look into this. This implementation was copied from the ort repo so the onnxmodel replacement works right now. Migration to onnxdag is a work item for most of the repo so we could try then.

PR #1818, opened 2025-05-05
Support k_quant quantization in Olive

jiafatom:
## Describe your changes

As titled.

### DIFF
--- a/olive/constants.py
+++ b/olive/constants.py
@@ -67,6 +67,8 @@ class QuantAlgorithm(CaseInsensitiveEnum):
     RTN = "rtn"
     SPINQUANT = "spinquant"
     QUAROT = "quarot"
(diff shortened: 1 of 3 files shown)


xiaoyu-work:
@jiafatom Our path is to migrate or implement quantization algorithms into our own pass, reducing reliance on external dependencies. For example, we just added OnnxHqqQuantization pass and do not call onnxruntime matmulnbits quantizer for hqq any more: https://github.com/microsoft/Olive/pull/1809. As for k quant, we should do the same. Can you add K quant pass separately?

jambayk:
Closing since PR has become stale and we plan to use SelectiveMixedPrecision pass + other onnx/quantizer passes to achieve the same effect.

PR #1819, opened 2025-05-05
Add reuse cache feature

kaunain-98:
This feature helps reduce the storage footprint when executing olive, current implementation causes copies of the model to be generated in each pass. Therefore we can overwrite the previous pass models and reuse the config files / encoder / decoder models


olive/passes/openvino/encapsulation.py
```diff
@@ -195,33 +204,35 @@ def _run_for_config(
         if not os.path.exists(output_model_path):
             os.makedirs(output_model_path)
 
-        model_name_path_dst = Path(output_model_path) / (f"{model_name}.xml")
-        weight_name_path_dst = Path(output_model_path) / (f"{model_name}.bin")
-        hardlink_copy_file(model_name_path, model_name_path_dst, follow_symlinks=True)
```
jambayk, line 200: so the hardlink did not work? normally it works fine. there are duplicate files but they should all point to the same underlying file in the filesystem.

olive/passes/openvino/io_update.py
```diff
@@ -55,8 +62,36 @@ def _run_for_config(
...
+        if config.reuse_cache:
+            output_model_path = model.model_path
+
+        model_name = model.model_config["model_name"]
+        model_name_path = Path(model.model_path) / (f"{model_name}.xml")
+        weight_name_path = Path(model.model_path) / (f"{model_name}.bin")
+
+        update_io_names = self._run_pass(model, config, output_model_path)
+
+        if config.reuse_cache and (config.static or update_io_names):
```
jambayk, line 74: overwriting the original files makes the olive cache invalid. since the input model point to the same file but the contents are not the same anymore. So, if you run the workflow again, olive will assume the cache is valid when it is not.

olive/passes/openvino/quantization.py
```diff
@@ -197,14 +204,31 @@ class OpenVINOQuantization(OpenVINOQuantizationBase):
     def _run_for_config(
         self, model: OpenVINOModelHandler, config: type[BasePassConfig], output_model_path: str
     ) -> OpenVINOModelHandler:
+        if config.reuse_cache:
```
jambayk, line 207: is there any storage gains from reusing cache in quantization? this is not the same like encapsulation or io update where the .bin files remain the same for both input and output. here, you would expect the .bin file contents to be different too?

kaunain-98: regardless of any pass (except the first pass) if the output we are overwriting the ov model if the output is ov model else writing the new file in same folder. This way we always keep only one copy of the ov model and reduce storage footprint

olive/passes/openvino/io_update.py
kaunain-98: yes, if its required to rerun the entire flow ignoring the present cache then we would prefer to set "clean_cache": true in the config file

olive/passes/openvino/encapsulation.py
kaunain-98: the hardlink uses shutils.copy function underneath and this creates copies of the files, we observed that the storage was being utilized

jambayk: This is unexpected. Did you check storage using something like file explorer on windows? Looks like Windows file explorer counts them as separate when counting the size used even though they don't actually take up extra space. So, I think it still works fine. If you turn debug logging on, the `hardlink_copy_file` utils tells you if it failed to hardlink. shutils.copy is only used when hardlink fails.

I am just concerned that the reuse cache for anything other than the encapsulation pass does not ensure that the olive cache for the passes is valid. But since this cache reuse behavior is isolated to the openvino passes, maybe it's okay for now.

kaunain-98: we saw it using the file explorer. We have encountered LLMs like llama to be using over 10GB (1 copy of tokenizer/detokernizer/ model per pass). we are open to relook at a later stage

olive/passes/openvino/quantization.py
xiaoyu-work: This is why we designed a caching system for Olive. We treat the Olive workflow as a sequence of Pass modules rather than a traditional workflow module. As expected, Olive saves multiple intermediate models in the cache throughout the workflow.

olive/passes/openvino/encapsulation.py
kaunain-98: this is interesting as we ran into errors due to no disk space

olive/passes/openvino/quantization.py
kaunain-98: we have added reuse_cache as an optional parameter with default set to false, this should allow users to use the default flow of olive, the param can be used only for large models where storage space is also a constraint

olive/passes/openvino/encapsulation.py
jambayk: do you know if the out of disk error happened during the quantization stage? In which case, I guess it's expected since the quantization has its own weights which are different from the input model.

olive/passes/openvino/quantization.py
xiaoyu-work: I'm worried that using this option might break the caching system, causing the second run to fail. We need to find a way to make these two options mutually exclusive.

xiaoyu-work: btw since you add this options to all examples. Can you add `clean_cache` to those configs as well?

kaunain-98: it doesnt look like clean_cache works. i have added that to the config files but do not see the .olive_cache folder being cleared

jambayk (approved):
Approving since the reuse cache is isolated to the openvino passes.

PR #1824, opened 2025-05-07
Unify example cache folder

xiaoyu-work:
## Describe your changes

These 2 examples have user scripts that will use `cache` to cache dataset. Using the same cache folder for workflow.

### DIFF
--- a/examples/resnet/openvino/resnet_context_ov_static.json
+++ b/examples/resnet/openvino/resnet_context_ov_static.json
@@ -39,5 +39,6 @@
     "search_strategy": false,
     "host": "local_system",
     "evaluate_input_model": false,
+    "cache_dir": "cache",
     "output_dir": "models/resnet_context_ov_static"
 }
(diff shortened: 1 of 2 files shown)


examples/resnet/openvino/resnet_context_ov_static.json
```diff
@@ -39,5 +39,6 @@
     "search_strategy": false,
     "host": "local_system",
     "evaluate_input_model": false,
+    "cache_dir": "cache",
```
jambayk, line 42: the examples are already in different folders so it probably won't help here. unless they are run from the same directory. but that wouldn't work since the example refers to the imagenet.py which is in the same directory as the config.json

jambayk: the user could change this to an absolute path if they want the runs to share the cache.

xiaoyu-work: I mean specifically for this example. I want to unify olive cache folder and dataset cache: https://github.com/microsoft/Olive/blob/main/examples/resnet/openvino/imagenet.py#L60.

jambayk: Oh, I see. Misunderstood the description.

jambayk, line 42: should we update this cache_dir or the assumed cache dir in the user script?

shaahji (approved):

PR #1832, opened 2025-05-07
[dynamo exporter] Fix dynamic shapes with DynamicCache

titaiwangms:
## Describe your changes

transformers introduces "Cache" objects, which is a custom class that contains kv_cache. However, in the current exporter, this becomes an exception of the pre-condition: input should have the same tree spec as the dynamic_shapes. For example, although DynamicCache is now traceable in exporter, it does not make sense to dynamic_shapes. Therefore, we can't use input tree spec to unflatten dynamic_shapes anymore. This PR deletes that, and simply iterates the input tree spec and turn any tuple to list to overcome the format mismatch caused by config.json.


justinchuby:
Feel free to merge this before the IR change

justinchuby (approved):

titaiwangms:
@jambayk I think CI is being weird. Would you help me merge it?

titaiwangms:
> onnxscript fusion test failed: https://github.com/microsoft/Olive/blob/main/test/unit_test/passes/onnx/test_onnxscript_fusion.py#L45. How will this test change onnxscript fusion behavior?

Thanks! It looks like TreeSpec is still wrong. I will need another fix on this.

justinchuby (approved):

PR #1860, opened 2025-05-15
Temporary patch for WinML support to auto register execution providers

shaahji:
## Temporary patch for WinML support to auto register execution providers

ORT WinML is ORTv1.22.0 + a few cherry picked PRs and so version check alone can't be used for auto registering the execution providers (required for auto ep selection feature in WinML).

### DIFF
--- a/olive/common/ort_inference.py
+++ b/olive/common/ort_inference.py
@@ -71,9 +71,7 @@ def get_ort_execution_provider_device_policy(policy: str):
     return None
 
 
-def initialize_inference_session_options_for_winml(
-    sess_options, device, providers, provider_options, provider_selection_policy=None
-):
+def initialize_ort_session_options(sess_options, device, providers, provider_options, provider_selection_policy=None):
     import onnxruntime as ort
 
     providers = providers or []
@@ -174,9 +172,11 @@ def get_ort_inference_session(
     for idx, provider in enumerate(providers):
         if provider in ["CUDAExecutionProvider", "DmlExecutionProvider"] and device_id is not None:
             provider_options[idx]["device_id"] = str(device_id)
-        elif provider == "QNNExecutionProvider":
+        elif provider == "QNNExecutionProvider" and "backend_path" not in provider_options[idx]:
             # add backend_path for QNNExecutionProvider
-            provider_options[idx]["backend_path"] = "QnnHtp.dll"
+            # WinML should not use provider options
+            # provider_options[idx]["backend_path"] = "QnnHtp.dll"
+            pass
     logger.debug("Normalized providers: %s, provider_options: %s", providers, provider_options)
 
     # dml specific settings
@@ -185,7 +185,7 @@ def get_ort_inference_session(
 
     sess_kwargs = {}
     if is_winml_installation():
-        initialize_inference_session_options_for_winml(
+        initialize_ort_session_options(
             sess_options, device, providers, provider_options, inference_settings.get("provider_selection_policy")
         )
     else:
--- a/olive/workflows/run/run.py
+++ b/olive/workflows/run/run.py
@@ -213,6 +213,29 @@ def set_olive_config_for_aml_system(olive_config: dict):
     AzureMLSystem.olive_config = olive_config
 
 
+def _setup_for_winml():
+    # pylint: disable=protected-access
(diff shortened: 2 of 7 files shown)


olive/workflows/run/run.py
```diff
@@ -213,13 +213,42 @@ def set_olive_config_for_aml_system(olive_config: dict):
...
+                extra_providers = {ep_device.ep_name for ep_device in ort.get_ep_devices()} - set(providers)
+                return providers + list(extra_providers)
+
+            ort.get_available_providers = get_available_providers_winml
+    except Exception as e:
+        logger.warning("An error occurred during WinML setup: %s", e)
+        logger.debug("Exception details:", exc_info=True)
+
+
 def run(
```
xieofxie, line 239: it is only limited to run.. but we have custom code that not using run

olive/workflows/run/run.py
```diff
@@ -213,13 +213,42 @@ def set_olive_config_for_aml_system(olive_config: dict):
...
 
+def _setup_for_winml():
+    # pylint: disable=protected-access
+    # TODO(olivedev): Remove post ORT release >1.22.0
+    try:
+        import onnxruntime as ort
+        from packaging import version
+
+        if version.parse(ort.__version__) >= version.parse("1.22.0"):
+            ort.winml.register_execution_providers()
```
xieofxie, line 224:
```
    import sys
    from .winml import register_execution_providers
    this = sys.modules[__name__]
    register_execution_providers(this)
```

An error occurred during WinML setup: register_execution_providers() missing 1 required positional argument: 'ort'

xieofxie: but if I add it directly, it will be `An error occurred during WinML setup: module 'olive.workflows.run.run' has no attribute 'register_execution_provider_library'`

xieofxie: so I guess it only workable in win-ort?

olive/common/ort_inference.py
```diff
@@ -174,7 +172,7 @@ def get_ort_inference_session(
     for idx, provider in enumerate(providers):
         if provider in ["CUDAExecutionProvider", "DmlExecutionProvider"] and device_id is not None:
             provider_options[idx]["device_id"] = str(device_id)
-        elif provider == "QNNExecutionProvider":
+        elif provider == "QNNExecutionProvider" and "backend_path" not in provider_options[idx]:
```
xieofxie, line 175: it will not work, getting the same error. I think it is because it is set after this. So perhaps do not set it when it is win-ort

olive/workflows/run/run.py
xieofxie: or at least I should pass in the module of win-ort not `this`

xieofxie: just debugged, please remove this line `ort.winml.register_execution_providers()`

olive/workflows/run/run.py
xieofxie: so maybe replace all of `get_available_providers`

olive/common/ort_inference.py
xieofxie: backend_path is set up inside ort.InferenceSession

shaahji: Then why is this even a problem. InferenceSession can just override whatever is provided from here.

olive/workflows/run/run.py
shaahji: > it is only limited to run.. but we have custom code that not using run

What *custom code*? And, what other code paths are you executing. `run` is the entry point for all config related runs. How/why are you bypassing this call?

olive/workflows/run/run.py
```diff
@@ -213,13 +213,42 @@ def set_olive_config_for_aml_system(olive_config: dict):
...
 
+def _setup_for_winml():
+    # pylint: disable=protected-access
+    # TODO(olivedev): Remove post ORT release >1.22.0
+    try:
+        import onnxruntime as ort
+        from packaging import version
+
+        if version.parse(ort.__version__) >= version.parse("1.22.0"):
+            ort.winml.register_execution_providers(sys.modules["onnxruntime"])
```
xieofxie, line 224: also check under `is_winml_installation`?

PR #1875, opened 2025-05-27
Rollback WinML specific changes and make the logic depend on ORT version

shaahji:
## Rollback WinML specific changes and make the logic depend on ORT version

Olive doesn't explicitly support AutoEP features but WinML does. ORT v1.22+ requires EPs to be pre-registered and use alternative API to query for supported EPs. Branching the logic based on version instead to keep things independent of package releases that embed Olive as a dependent module.

### DIFF
--- a/olive/common/ort_inference.py
+++ b/olive/common/ort_inference.py
@@ -28,21 +28,16 @@ class OrtSessionFallbackError(Exception):
     """Raised when the onnxruntime fallback happens."""
 
 
-def is_winml_installation() -> bool:
+def ort_supports_ep_devices() -> bool:
     from onnxruntime import __version__ as OrtVersion
 
-    if version.parse(OrtVersion) >= version.parse("1.22.0"):
-        try:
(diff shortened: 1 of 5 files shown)


olive/common/ort_inference.py
```diff
@@ -28,21 +28,16 @@ class OrtSessionFallbackError(Exception):
     """Raised when the onnxruntime fallback happens."""
 
 
-def is_winml_installation() -> bool:
+def ort_supports_ep_devices() -> bool:
```
jambayk, line 31: can you trigger the nightly pipeline so that we can verify this works for versions higher than 1.22? Otherwise, we should just remove all winml logic and not keep this version specific logic @devang-ml?

devang-ml: let's trigger nightly pipeline run, if not already done so.

jambayk: triggered. The examples are running fine, but the unit tests are failing

PR #1877, opened 2025-05-29
add sd xl qdq

xieofxie:
## Describe your changes

Original
CLIP Scores avg: 31.575813
FID: 174.414032

New
CLIP Scores avg: 31.725784
FID: 177.423035
Average train error 444.848633
Average test error 421.371246

My setup for the dynamic shape to fixed issue:
[freeze.txt](https://github.com/user-attachments/files/20561537/freeze.txt)


examples/stable_diffusion_xl/stable_diffusion_xl.py
```diff
@@ -285,12 +299,30 @@ def run_inference(
...
+    config: dict, provider: str, is_fp16: bool, only_conversion: bool, submodel_name: str, model_format: str
+) -> dict:
     used_passes = {}
-    if provider == "dml":
+    if only_conversion:
+        used_passes = {"convert"}
+        config["evaluator"] = None
+    elif model_format == "qdq":
+        if submodel_name in ("text_encoder", "text_encoder_2"):
+            # TODO(anyone): dynamic_shape_to_fixed is not working
```
xieofxie, line 316: weird

examples/stable_diffusion_xl/config_text_encoder.json
```diff
@@ -110,6 +116,46 @@
...
             "use_gpu": true,
             "keep_io_types": true
+        },
+        "dynamic_shape_to_fixed": {
+            "type": "DynamicToFixedShape",
+            "dim_param": [ "batch_size", "sequence_length" ],
+            "dim_value": [ 1, 77 ]
+        },
+        "surgery": { "type": "GraphSurgeries", "surgeries": [ { "surgeon": "ReplaceAttentionMaskValue" } ] },
+        "optimize_qdq": {
```
jambayk, line 126: Same comment as https://github.com/microsoft/Olive/pull/1893#discussion_r2124453201. I don't think this is required. gelu fusion is done by the quant_preprocess step in the quantization pass. opset 17 has layernorm in expored model.

the sd config also uses this so we can keep it for now. but i think it's worth checking if this is pass is needed and remove to simplify the recipe if it's redundant.

examples/stable_diffusion_xl/stable_diffusion_xl.py
jambayk: could you give more information about this issue? does it not work with new onnxruntime version? maybe the input names or shapes are not matching?

examples/stable_diffusion_xl/config_text_encoder.json
xieofxie: Hi, I don't understand for these
- for `enable_gelu`: I see Olive uses quant_pre_process https://github.com/microsoft/Olive/blob/76aee46daacc09f98854e4452490a5aaebb857eb/olive/passes/onnx/quantization.py#L512 which is default `ORT_ENABLE_BASIC` https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/shape_inference.py and from doc, BASIC does not have enable_gelu, https://onnxruntime.ai/docs/performance/model-optimizations/graph-optimizations.html only Extended has
- for `enable_layer_norm`: why update the opset will change the structure of the model? From what I search, https://github.com/search?q=repo%3Amicrosoft%2Fonnxruntime%20enable_layer_norm&type=code it is a function called `fuse_layer_norm`

xieofxie: from what I search in onnxruntime https://github.com/search?q=repo%3Amicrosoft%2Fonnxruntime+enable_gelu+path%3A%2F%5Eonnxruntime%5C%2Fpython%5C%2Ftools%5C%2Ftransformers%5C%2F%2F&type=code I think it only applies for certain types? So I think they have no effect for different reason, not quant_preprocess step or opset version

xieofxie: If we remove this step, the model will be different because even we set them to false, the onnxruntime optimizer still do some optimization for the model https://github.com/microsoft/onnxruntime/blob/242cb4398a042221895b982c59f5069a491ffb49/onnxruntime/python/tools/transformers/onnx_model_bert.py. I don't have qnn device but I could provide some data later

jambayk: - `enable_layernorm`: LayerNorm was added as an onnx operator in opset 17. so with 17+, the model gets exported with layernorm operator directly so the fusion is not needed. https://onnx.ai/onnx/operators/onnx__LayerNormalization.html
- `enable_gelu`: the regular gelu fusion was moved to `L1` when the opset is 20 https://github.com/microsoft/onnxruntime/pull/21332 but the docs were not updated

jambayk: but looks like there are other optimizations done too so we can keep this pass. It would be useful for us to investigate more and see what extra optimizations are applied by this pass so that we can do it ourselves. transformers optimizer pass is not maintained much anymore, and it has too many options and behavior we cannot control.

jambayk (approved):

PR #1888, opened 2025-05-30
Remove "generative" flag from input_model and "is_generative" command line option

devang-ml:
## Describe your changes
Remove "generative" flag from input_model and "is_generative" command line option.

Check GenerationConfig to see if HF model is generative or not.

### DIFF
--- a/olive/model/handler/hf.py
+++ b/olive/model/handler/hf.py
@@ -28,7 +28,7 @@ logger = logging.getLogger(__name__)
 @model_handler_registry("HFModel")
 class HfModelHandler(PyTorchModelHandlerBase, MLFlowTransformersMixin, HfMixin):  # pylint: disable=too-many-ancestors
     resource_keys: tuple[str, ...] = ("model_path", "adapter_path")
-    json_config_keys: tuple[str, ...] = ("task", "load_kwargs", "generative")
+    json_config_keys: tuple[str, ...] = ("task", "load_kwargs")
 
     def __init__(
         self,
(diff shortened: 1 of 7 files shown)


test/unit_test/cli/test_base.py
jambayk, line 1: this test needs to be updated as well https://github.com/microsoft/Olive/blob/56cf848ef3a41751015b0ef45c49ca60ac37f73e/test/unit_test/cli/test_cli.py#L135. Was added earlier today

devang-ml: I am doing merge with main right now.

olive/model/handler/hf.py
```diff
@@ -215,8 +203,11 @@ def run_session(
         inputs: Union[dict[str, Any], list[Any], tuple[Any, ...]] = None,
         **kwargs: dict[str, Any],
     ) -> Any:
+        is_generative = False
+        if self.get_hf_generation_config() is not None:
```
jambayk, line 207: we could also look at the task like before `self.task.startswith("text-generation")`. It might be expensive to keep loading generation config for each session run.

jambayk (approved):

olive/model/handler/pytorch.py
```diff
@@ -58,9 +56,9 @@ def run_session(
         **kwargs: dict[str, Any],
     ) -> Any:
         if isinstance(inputs, dict):
-            results = session.generate(**inputs, **kwargs) if self.generative else session(**inputs, **kwargs)
+            results = session(**inputs, **kwargs)
         else:
-            results = session.generate(inputs, **kwargs) if self.generative else session(inputs, **kwargs)
+            results = session(inputs, **kwargs)
         return results
```
shaahji, line 62: With this change, aren't we dropping support for non-hf generative pytorch models?

Or, is the consideration that this code path gets used only for latency computation and so use of `forward` call over `generate` call is justified.

PR #1797, opened 2025-04-27
Refined QDQ recipes of BERT/CLIP/VIT for QC and AMD.

tezheng:
## Checklist before requesting a review
- [ ] Add unit tests for this change.
- [ ] Make sure all tests can pass.
- [ ] Update documents if necessary.
- [x] Lint and apply fixes to your code by running `lintrunner -a`
- [ ] Is this a user-facing change? If yes, give a description of this change to be included in the release notes.
- [x] Is this PR including examples changes? If yes, please remember to update [example documentation](https://github.com/microsoft/Olive/blob/main/docs/source/examples.md) in a follow-up PR.

PR #1893, opened 2025-06-03
update bert vit clip qdq to align with Model Lab

xieofxie:

### DIFF
--- a/examples/clip/qdq/clip_script.py
+++ b/examples/clip/qdq/clip_script.py
@@ -139,7 +139,9 @@ def eval_similarity_degrad(output, targets, batch_size=1024):
 
     preds = output.preds
     scores = [
+        # pylint: disable=E1102
         F.cosine_similarity(preds[i : i + batch_size], targets[i : i + batch_size])
+        # pylint: enable=E1102
         for i in range(0, preds.size(0), batch_size)
     ]
     return {"percentage": f"{100.0 - torch.mean(torch.cat(scores)) * 100.0:.2f}"}


xieofxie:
Part of this pr https://github.com/microsoft/Olive/pull/1797 to align with AITK https://github.com/microsoft/windows-ai-studio-templates/tree/dev/model_lab_configs/huggingface

examples/bert/bert_ptq_qdq.json
```diff
@@ -52,16 +52,36 @@
...
         },
-        "surgery": { "type": "GraphSurgeries", "surgeries": [ { "surgeon": "ReplaceAttentionMaskValue" } ] },
+        "surgery": {
+            "type": "GraphSurgeries",
+            "surgeries": [
+                { "surgeon": "ReplaceAttentionMaskValue", "replacement": -100.0 },
+                { "surgeon": "MatMulAddToGemm" }
+            ]
+        },
+        "transformer_optimizer": {
```
jambayk, line 68: I don't think this is required. gelu fusion is done by the quant_preprocess step in the quantization pass. opset 20 has layernorm in expored model.

examples/clip/qdq/laion_clip_text_b32_qdq.json
jambayk, line 1: what about the exisiting qdq config jsons? we should remove them?

examples/vit/vit_qdq.json
```diff
@@ -49,21 +49,27 @@
...
-        "conversion": {
-            "device": "cpu",
-            "type": "OnnxConversion",
-            "target_opset": 17,
-            "save_as_external_data": true,
-            "all_tensors_to_one_file": true,
-            "use_dynamo_exporter": false
+        "conversion": { "type": "OnnxConversion", "target_opset": 20 },
+        "surgery": { "type": "GraphSurgeries", "surgeries": [ { "surgeon": "MatMulAddToGemm" } ] },
+        "transformer_optimizer": {
```
jambayk, line 54: same here. gelu and layernorm fusion are already done without this pass.

examples/clip/qdq/laion_clip_text_b32_qdq.json
xieofxie: They are different. These separate the text model and vision model, while existing qdq configs works on the whole model

examples/vit/vit_qdq.json
xieofxie: [vit.zip](https://github.com/user-attachments/files/20586897/vit.zip)
image
As discussed in another pr, this step itself will do optimizations. so the results are different and model with it has better performance

examples/bert/bert_ptq_qdq.json
xieofxie: [bert.zip](https://github.com/user-attachments/files/20586945/bert.zip)
Same. models and performance are different. with this pass is better 
image

jambayk: The latency difference seems to be within range for it to be just regular variance. but it's not an issue since adding the transformers optimizer pass doesn't hurt. I just think that this pass is complicated with lots of fusions so it's almost like a black box to end users.

examples/vit/vit_qdq.json
jambayk: thanks for checking. Looks like there are other optimization than gelu and layernorm applied.

PR #1919, opened 2025-06-17
Refactor hqq quantization pass

xiaoyu-work:
## Describe your changes:

1. Remove `HqqQuantizer`.
1. Add onnxruntime import check.
1. Remove duplicated unit tests.

### DIFF
--- a/olive/passes/onnx/hqq_quantization.py
+++ b/olive/passes/onnx/hqq_quantization.py
@@ -174,7 +174,7 @@ class OnnxHqqQuantization(Pass):
             b_array_torch = b_array_torch.cuda()
 
         quant_weight_torch, scales_torch, zero_points_torch = self._quantize_internal(
-            b_array_torch.T, group_size=block_size, axis=axis
+            b_array_torch, group_size=block_size, axis=axis
         )
         quant_weight_torch = quant_weight_torch.contiguous()
         scales_torch = scales_torch.contiguous()
--- a/olive/passes/onnx/hqq_quantization.py
+++ b/olive/passes/onnx/hqq_quantization.py
@@ -24,9 +24,6 @@ from olive.passes.pass_config import BasePassConfig, PassConfigParam
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


olive/passes/onnx/hqq_quantization.py
```diff
@@ -61,17 +64,17 @@ def _run_for_config(
                 "HQQ quantization is not supported for models with adapters. Returning the model without quantization."
             )
             return model
+        self.block_size = config.block_size
```
jambayk: i think it's better if we pass these values as method params like before. passes are supposed to be stateless in theory (even though it allows setting attributes) with all info in the config.

olive/passes/onnx/rtn_quantization.py
```diff
@@ -248,8 +246,11 @@ def _qbits_block_quant(self, fp32weight: npt.ArrayLike) -> tuple[np.ndarray, np.
...
         if qbits == 8:
-            from onnxruntime import __version__ as OrtVersion
-            from packaging import version
+            try:
+                from onnxruntime import __version__ as OrtVersion
+                from packaging import version
+            except ImportError as e:
+                raise ImportError("onnxruntime is required for RTN quantization.") from e
 
             if version.parse(OrtVersion) > version.parse("1.21.0"):
```
shaahji: Ambiguity here! Check is w.r.t. version >1.21.0 so anything <= 1.21.0 should raise an error and not 1.21.1.

olive/passes/onnx/hqq_quantization.py
xiaoyu-work: sounds good to me. updated!

olive/passes/onnx/rtn_quantization.py
xiaoyu-work: True! i checked 1.21.1 it doesn't support either. updated!

jambayk (approved):

shaahji (approved):

shaahji (approved):

PR #1921, opened 2025-06-17
Add AddOliveMetadata pass to example recipes

ashrit-ms:
## Describe your changes
* Added `AddOliveMetadata` pass to P0 example recipe files
* Improved the readability and structure of recipe files to ensure consistent indentation and readability

### DIFF
--- a/examples/adetailer/face_yolo_qnn.json
+++ b/examples/adetailer/face_yolo_qnn.json
@@ -9,14 +9,7 @@
     "systems": {
         "qnn_system": {
             "type": "LocalSystem",
-            "accelerators": [
-                {
-                    "device": "npu",
-                    "execution_providers": [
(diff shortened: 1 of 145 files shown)


ashrit-ms:
> @devang-ml i have some concerns about this change. i don't think we need to add metadata pass for all our examples. either we should create a branch specifically for this testing, or if we do want this metadata, this metadata should be added for each pass automatically.

I couldn't find a way to trigger the AddOliveMetadata pass automatically but if there is then I can make that change. We would ideally like to have both options, if user decides to add metadata via config then that should take precedence over the automatic addition.

ashrit-ms:
Spoke to @devang-ml offline and decided on adding the AddOliveMetadata pass only to the P0 model recipes.

xiaoyu-work (approved):

PR #1927, opened 2025-06-22
Llama3.1 8B example recipe for QNN, VitisAI and OpenVino

chapsiru:
Update Llama 3.1 8B example recipe for QNN, VitisAI and OpenVino


examples/llama3_1/qdq_config_vitis_ai.json
jambayk, line 1: thanks for adding the example. Since the config is same as the phi3_5 qdq and the readme also points to it, this and the qnn_config.json are not needed. we are trying to reuse the same configs to minimize duplicate configs and easy updates.

chapsiru: Sure, I will remove it

Issue #1936, opened 2025-06-26
Cache is not calculated correctly


xieofxie:
**Describe the bug**

Cache is not calculated correctly for some models.

i.e. two different models will use the same cache which will lead to wrong results

**To Reproduce**
Steps to reproduce the behavior.

https://github.com/microsoft/Olive/tree/hualxie/clip_cache_issue

Run examples/clip/qdq openai_clip_text_b32_qdq.json  and openai_clip_text_b16_qdq.json

They use same model cache

[2025-06-26 11:45:33,330] [INFO] [engine.py:723:_run_pass] Loaded model from cache: 6b9ad787
[2025-06-26 11:45:33,332] [INFO] [engine.py:323:run_accelerator] Save footprint to D:\Olive-pr\examples\clip\qdq\models\openai\clip_b32\text\footprints.json.

[2025-06-26 11:45:58,879] [INFO] [engine.py:723:_run_pass] Loaded model from cache: 6b9ad787
[2025-06-26 11:45:58,882] [INFO] [engine.py:323:run_accelerator] Save footprint to D:\Olive-pr\examples\clip\qdq\models\openai\clip_b16\text\footprints.json.

**Expected behavior**

Should use different one

**Olive config**
Add Olive configurations here.

**Olive logs**
Add logs here.

**Other information**
 - OS: [e.g. Windows, Linux]
 - Olive version: [e.g. 0.4.0 or main]
 - ONNXRuntime package and version: [e.g. onnxruntime-gpu: 1.16.1]
 - Transformers package version: [e.g. transformers 4.44.1]

**Additional context**

xiaoyu-work:
Closed as the fix PR had been merged.

PR #1941, opened 2025-07-01
Fix input model cache issue

xiaoyu-work:
## Describe your changes

Fix input model cache issue by including model_path in the hash calculation, to handle cases where multiple config files share the same model script.

## (Optional) Issue link
https://github.com/microsoft/Olive/issues/1936

### DIFF
--- a/olive/model/config/model_config.py
+++ b/olive/model/config/model_config.py
@@ -57,6 +57,7 @@ class ModelConfig(NestedConfig):
 
     def get_model_identifier(self):
         model_path = self.config.get("model_path")
+        hash_list = []
         if model_path:
             model_path_resource_path = create_resource_path(model_path)
             if (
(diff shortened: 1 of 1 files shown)


olive/model/config/model_config.py
```diff
@@ -77,10 +78,12 @@ def get_model_identifier(self):
 
             if model_path_resource_path.is_azureml_resource():
                 return model_path_resource_path.get_path()
+            hash_list.append(model_path)
```
jambayk: instead of doing a special case here, maybe _get_file_hash can be updated to handle the case where the value is a string? that make it more generic

xiaoyu-work: "adapter_path", "model_script", "script_dir" may not be a valid case if they are a str in the config but there is no such file or folder. Our workflow will ignore it in the config but if we check it here that will be added to hash calculation.

jambayk: if they are not valid, the model would be invalid too

jambayk: also, adapter path could also be a hf hub id so it's a string

jambayk (approved):

Issue #1194, opened 2024-06-12
Failed on optimization step of Unet - Intel Meteor Lake Processor


hooroobaby:
**Describe the bug**
I am using Intel Meteor Lake architecture, Core Ultra 7 165H Processor.

Under `Olive\examples\directml\stable_diffusion_xl` I ran the command `python stable_diffusion_xl.py --model_id stabilityai/sdxl-turbo --optimize --clean_cache `, and then I encountered error while Unet optimizing step.

The error is: `onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : C:\a\_work\1\s\onnxruntime\core\providers\dml\DmlExecutionProvider\src\DmlCommandRecorder.cpp(371)\onnxruntime_pybind11_state.pyd!00007FFDFEB251FE: (caller: 00007FFDFEB0BE04) Exception(1) tid(13d8c) 887A0006 The GPU will not respond to more commands, most likely because of an invalid command passed by the calling application.`

**Expected behavior**

Also currently in stable_diffusion_xl.py I only see support for CUDA and DML, I want to inquire if you have any suggestions for me if I want to use *OpenvinoExecutionProvider* as the provider. Thank you

**Other information**
 - OS: Windows
 - Olive version: 0.7.0
 - ONNXRuntime package and version: onnxruntime-directml  1.18.0

**Additional context**

```

Optimizing unet
[2024-06-11 17:15:01,391] [INFO] [run.py:138:run_engine] Running workflow default_workflow
[2024-06-11 17:15:01,401] [INFO] [engine.py:986:save_olive_config] Saved Olive config to cache\default_workflow\olive_config.json
[2024-06-11 17:15:01,403] [INFO] [accelerator_creator.py:224:create_accelerators] Running workflow on accelerator specs: gpu-dml
[2024-06-11 17:15:01,403] [INFO] [engine.py:109:initialize] Using cache directory: cache\default_workflow
[2024-06-11 17:15:01,404] [INFO] [engine.py:265:run] Running Olive on accelerator: gpu-dml
[2024-06-11 17:15:01,405] [INFO] [engine.py:1085:_create_system] Creating target system ...
[2024-06-11 17:15:01,405] [INFO] [engine.py:1088:_create_system] Target system created in 0.000000 seconds
[2024-06-11 17:15:01,405] [INFO] [engine.py:1097:_create_system] Creating host system ...
[...]

Issue #1170, opened 2024-05-22
[Bug]: Optimization of Unet fails - AMD RDNA3.5 Strix Point Processor


woonyee28:
**Describe the bug**

For context, I am using AMD RDNA3.5 architecture, Strix Point Processor. 

Under `Olive\examples\stable_diffusion` I ran the following command.
```python stable_diffusion.py --model_id stabilityai/stable-diffusion-2-1 --optimize --clean_cache ```
I encountered error while I was optimizing unet.
Same error was found when I ran ```python stable_diffusion_xl.py --model_id stabilityai/sdxl-turbo --optimize --clean_cache ``` under `Olive\examples\directml\stable_diffusion_xl`

The error is: `onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : C:\a\_work\1\s\onnxruntime\core\providers\dml\DmlExecutionProvider\src\DmlCommandRecorder.cpp(371)\onnxruntime_pybind11_state.pyd!00007FFA29BD2C5E: (caller: 00007FFA29BB9864) Exception(1) tid(35c4) 887A0006 The GPU will not respond to more commands, most likely because of an invalid command passed by the calling application.`

I have scrolled through the issue list. Multiple approaches have been taken. For example:
1. Downgrading Python to 3.10, previously I was using Python 3.12. (Suggested from this issue: https://github.com/microsoft/Olive/issues/1023)
2. Set `"save_as_external_data": true` (Suggested from the same issue as above: https://github.com/microsoft/Olive/issues/1023)
3. Set --temp-dir . (Suggested from the same issue as above: https://github.com/microsoft/Olive/issues/1023)
None of these works.

I found the exact same problem faced by this issue: https://github.com/microsoft/Olive/issues/517

**Other information**
 - OS: Windows
 - Olive version: 0.6.0 (git clone main branch on 21 May Singapore Time)
 - ONNXRuntime package and version: onnxruntime-directml 1.18.0

For full error log:
```
Optimizing unet
[2024-05-22 15:44:40,729] [INFO] [run.py:279:run] Loading Olive module configuration from: C:\Users\wy-te\AppData\Local\Programs\Python\Python310\lib\site-packages\olive\olive_config.json
[2024-05-22 15:44:40,734] [DEBUG] [olive_evaluator.py:1153:validate_metrics] No priority is specified, but only one sub type  metric is specified. Use rank 1 for single for this metric.
[...]

Issue #1165, opened 2024-05-19
Get Error while optimizing SDXL of DirectML example


klin2024:
**Describe the bug**
Get Error while optimizing SDXL

**To Reproduce**
Command: python stable_diffusion_xl.py --optimize

**Expected behavior**
Optimize successfully.

**Olive logs**

Optimizing unet
[2024-05-19 05:11:53,655] [INFO] [run.py:279:run] Loading Olive module configuration from: C:\Users\AMD\miniconda3\envs\sdxl_turbo\lib\site-packages\olive\olive_config.json
[2024-05-19 05:11:53,665] [INFO] [accelerator_creator.py:224:create_accelerators] Running workflow on accelerator specs: gpu-dml
[2024-05-19 05:11:53,666] [INFO] [engine.py:107:initialize] Using cache directory: cache
[2024-05-19 05:11:53,666] [INFO] [engine.py:263:run] Running Olive on accelerator: gpu-dml
[2024-05-19 05:11:53,666] [INFO] [engine.py:1075:_create_system] Creating target system ...
[2024-05-19 05:11:53,667] [INFO] [engine.py:1078:_create_system] Target system created in 0.000997 seconds
[2024-05-19 05:11:53,667] [INFO] [engine.py:1087:_create_system] Creating host system ...
[2024-05-19 05:11:53,667] [INFO] [engine.py:1090:_create_system] Host system created in 0.000000 seconds
[2024-05-19 05:11:53,692] [INFO] [engine.py:865:_run_pass] Running pass convert:OnnxConversion
[2024-05-19 05:18:58,157] [INFO] [engine.py:952:_run_pass] Pass convert:OnnxConversion finished in 424.462611 seconds
[2024-05-19 05:18:58,168] [INFO] [engine.py:865:_run_pass] Running pass optimize:OrtTransformersOptimization
fusion: 100%|█████████████████████████████████████████████████████████████████████████| 18/18 [07:59<00:00, 26.61s/it]
Traceback (most recent call last):
  File "D:\olive_0516\Olive\examples\directml\stable_diffusion_xl\stable_diffusion_xl.py", line 637, in <module>
    main()
  File "D:\olive_0516\Olive\examples\directml\stable_diffusion_xl\stable_diffusion_xl.py", line 603, in main
    optimize(
  File "D:\olive_0516\Olive\examples\directml\stable_diffusion_xl\stable_diffusion_xl.py", line 369, in optimize
    olive_run(olive_config)
[...]

Issue #1164, opened 2024-05-18
"num_images" doesn't work for the example of directml stable_diffusion_xl.


peterer0625:
**Describe the bug**
"num_images" doesn't work for the example of directml stable_diffusion_xl.

**To Reproduce**
Command: python stable_diffusion_xl.py --num_image 2 --prompt="great wall" --num_inference_steps=4 

**Expected behavior**
Generate two images. However, I only saw one image.

Jay19751103:
hi Olive Team
    It looks that   while images_saved < num_images: removed in run_inference_loop
    or just use batch_size to replace it on SDXL? 
    anyone can fix this?

PR #1947, opened 2025-07-04
SDXL example bugfixes and enhancements

lostdisc:
Bugfixes:
- Fixed bug where dml provider was being changed to cpu
- Avoid running costly ByteSize() check if save_as_external_data is already enabled.  Related to [Issue 1165](https://github.com/microsoft/Olive/issues/1165).
- Enable save_as_external_data for optimization pass in config_unet.json, since unet always exceeds the 2GB Protobuf limit.  Addresses [Issue 1165](https://github.com/microsoft/Olive/issues/1165).
- Fixed bug where vestigial unoptimized model.onnx.data files were still in optimized model folders
- Fixed error/crash in interactive GUI mode near start of generation
- Fixed visual glitch where interactive GUI progress bar stopped short of full
- Removed unused vestigial function run_refiner_inference_loop().  This functionality was already being handled by code in run_inference_loop().

Enhancements:
- Added `--skip_evaluation` option, since some systems (like mine) can complete all operations EXCEPT evaluation.  A workaround for [Issue 510](https://github.com/microsoft/Olive/issues/510), [Issue 1170](https://github.com/microsoft/Olive/issues/1170), and [Issue 1194](https://github.com/microsoft/Olive/issues/1194).
- Added `--single_local_file` option for optimizing a local .safetensors model file
- Support `--num_images` larger than `--batch_size`.  This should be the intended behavior, based on the relevant [Readme](https://github.com/microsoft/Olive/blob/main/examples/directml/stable_diffusion_xl/README.md) saying "The minimum number of inferences will be ceil(num_images / batch_size)".  Fixes [Issue 1164](https://github.com/microsoft/Olive/issues/1164).
- Added optional `--image_width` and `--image_height` flags
- In interactive GUI, enable pressing Enter key in input text field to start generation
- Move interactive GUI controls from bottom to top, so they stay onscreen if image window is taller than screen

Documentation updates:
- Updated SDXL example link in examples.md to point to correct folder
- Updated SDXL example's DirectML readme with info on new options/flags

I got all this into a working state months ago on my machine, but didn't have time to split it into logically-separated, digestible commits until now.  Hope this format is satisfactory!  Let me know if anything else is needed from me, as I'm basically new to Github.


examples/stable_diffusion_xl/stable_diffusion_xl.py
```diff
@@ -43,93 +44,75 @@ def run_inference_loop(
...
+    while images_saved < num_images:
+        print(f"\nInference Batch Start (batch size = {batch_size}).")
+        if base_images is None:
+            result = run_base_batch()
+        else:
+            result = run_refined_batch()
 
-    print("Inference Batch End.")
+        for image_index in range(batch_size):
+            if images_saved < num_images:
```
shaahji: nitpick: Wouldn't this just be same as `for image_index in range(len(result.images)):` ?

examples/stable_diffusion_xl/stable_diffusion_xl.py
```diff
@@ -416,9 +428,12 @@ def optimize(
...
 
+        if skip_evaluation:
+            olive_config["evaluator"] = None
+
         olive_run(olive_config)
 
-        footprints_file_path = Path(__file__).resolve().parent / "footprints" / submodel_name / "footprints.json"
+        footprints_file_path = script_dir / "footprints" / submodel_name / "footprints.json"
         with footprints_file_path.open("r") as footprint_file:
             footprints = json.load(footprint_file)
```
shaahji, line 439: `olive_run` now returns an object that has the API to query for output data like footprints and models. I would recommend using the API rather than hardcoding the paths.

examples/stable_diffusion_xl/stable_diffusion_xl.py
lostdisc: Looks like it works the same, yes.  I've made the suggested change.

examples/stable_diffusion_xl/stable_diffusion_xl.py
```diff
@@ -182,34 +166,35 @@
...
-    gui_images = []
-    for row in range(image_rows):
-        for col in range(image_cols):
-            label = tk.Label(window, width=image_size, height=image_size, background="black")
-            gui_images.append(label)
-            label.place(x=col * image_size, y=row * image_size)
+    prompt_textbox = tk.Entry(window)
+    prompt_textbox.insert(tk.END, prompt)
+    prompt_textbox.place(x=0, y=0, width=window_width - button_width, height=button_height)
+    prompt_textbox.bind('<Return>', on_generate_click)
```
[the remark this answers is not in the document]
lostdisc: I've made this change, and others suggested by lintrunner.

examples/stable_diffusion_xl/stable_diffusion_xl.py
xiaoyu-work: https://microsoft.github.io/Olive/how-to/python_api.html#accessing-optimization-results

lostdisc: Thanks for the doc link!  That can probably be arranged, but would take me more study/testing to figure out.  (May be a week or weeks, depending on life.)

PR #1949, opened 2025-07-07
Quantization: Native QuantLinear, HfQuantizer, GPTQ Pass

jambayk:
## Describe your changes
Native hf quantization support:
- New QuantLinear module. Weight packing is the same as autogptq for easier integration with downstream passes such as torch export and model builder.
- Integration with transformers model saving and loading
- New Gptq pass with modular implementation. There is no bass quantization pass but it can easily be abstracted to add additional algorithms such as AWQ (on the way).


olive/common/quant/hf_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+    @property
+    def is_trainable(self) -> bool:
+        # TODO(jambayk): investigate what this means (peft, scale+bias, etc.?)
+        # need to support peft, scale+bias comes for free since everything is in torch
+        return False
+
+
+def replace_matching_submodules(
+    module: nn.Module,
```
devang-ml, line 177: Move to hf/utils.py ?

jambayk: I would like to restructure the `olive.common.hf` and `olive.common.quant` files but am not currently sure what it should be. Will take this into consideration when I do the restructure. Thanks!

PR #1953, opened 2025-07-08
Add support to run workflow in docker system

xiaoyu-work:
## Describe your changes

This is the first PR to refactor the docker system.
1. Add support to run workflow in docker system.
2. Update Phi3.5 model support.

The follow-up PR:
1. Remove docker system support for standalone pass run.
2. Update documentation.
3. Update tests.


olive/systems/docker/constants.py
```diff
@@ -0,0 +1,3 @@
+from pathlib import Path
+
+CONTAINER_ROOT_PATH = Path("/olive-ws/")
```
shaahji, line 6: Please add some documentation for this constant. The name is a misnomer and can mean multiple different things.
* Is it directory where olive gets cloned? or
* Is it the "home directory" within the container and olive get cloned as a sub-directory within it? or
* Is it the docker container's home directory?

I can't think of a better name for this constant but a little documentation around it will go a long way. If nothing else, it will state your intentions clearly.

olive/systems/docker/docker_system.py
```diff
@@ -103,6 +108,69 @@ def __init__(
...
                     raise
 
+    def submit_workflow(self, run_config: RunConfig):
+        with tempfile.TemporaryDirectory() as tempdir:
+            self._run_workflow(run_config, tempdir)
+
+    def _ensure_directory(self, path: str, default: str) -> str:
+        dir_path = Path(path or default)
+        dir_path.mkdir(parents=True, exist_ok=True)
+        return str(dir_path)
```
shaahji, line 118: Recommend moving it to utils/common.py.

Also, the decision to use `path` vs. `default` should be caller's not this implementation. Within this implementation it could mean two different things -

* `mkdir(path)` if path is valid but doesn't exist
* `mkdir(default) if path is valid but it doesn't exist

The API will be cleaner if the caller explicitly passed one or the other.

If you like to keep the current implementation move it into the _run_workflow so it's invisible outside.

olive/systems/olive_system.py
```diff
@@ -40,6 +43,20 @@ def __init__(
...
+
+        if self.is_dev:
+            logger.warning(
+                "Dev mode is only enabled for CI pipeline! "
+                "It will overwrite the Olive package in AML computer with latest code."
+            )
+            cur_dir = Path(__file__).resolve().parent
+            project_folder = cur_dir.parents[1]
+            copy_dir(project_folder, target_path / "olive", ignore=shutil.ignore_patterns("__pycache__"))
+
```
shaahji: Is this any different from common/utils.py::copy_dir?
May be move it to the same file if the implementations do something different.

shaahji (changes requested):

shaahji:
Food for thought - We should consider adding a separate "virtual" filesystem module to Olive that can do the mount/unmount on demand rather than having to deal with these on a case-by-case basis. The virtual filesystem can be responsible for ensuring the files/folders get correctly mapped without copying files and instead using links. Module can also extend functionality with functions like copy_dir, copy_files, dir_exits, file_exits, etc.

olive/systems/docker/docker_system.py
xiaoyu-work: This is internal use for docker system. But this may be removed in my follow-up PR. If run config is initialized and validated, it should have default value, so there will be no none check here.

olive/systems/docker/constants.py
xiaoyu-work: Sure will add more documentation here.

olive/systems/olive_system.py
xiaoyu-work: I just realized that i moved this func from aml_system to olive_system cause i use this func in docker_system at some point but later i removed this from docker system. Since AML system will be deprecated soon, this func will be removed. I'll just move this back to aml_system for now.

xiaoyu-work:
> Food for thought - We should consider adding a separate "virtual" filesystem module to Olive that can do the mount/unmount on demand rather than having to deal with these on a case-by-case basis. The virtual filesystem can be responsible for ensuring the files/folders get correctly mapped without copying files and instead using links. Module can also extend functionality with functions like copy_dir, copy_files, dir_exits, file_exits, etc.

If we remove aml system, the only use case for this would be docker system i guess.

olive/systems/docker/constants.py
xiaoyu-work: i probably will remove this constant in my next PR. Instead, user will provide the `WORKDIR` as a part of the docker system config.

shaahji: That would require passing the constant to all the necessary functions. You just took the parameter out from most of those functions.

xiaoyu-work: Yeah, previously the logic is kinda like Olive will manage how and where the folders and files mount to. For new docker system, my expectation is user will provide and manage the Dockerfile, thus they can decide how and where the olive will run and where to store the output in the container.

shaahji (approved):

PR #1963, opened 2025-07-11
Fix "Provider xxx has already been registered."

fangyangci:
We observed that when connecting remotely to a device with a GPU, the `ort.get_ep_devices()` function returns duplicate GPU EP devices.

Execute the following code.
```
import onnxruntime as ort

ep_devices = ort.get_ep_devices()
for ep_device in ep_devices:
    print(f"{id(ep_device.device)} {ep_device.ep_name} {ep_device.device.type} {ep_device.device.device_id} {ep_device.device.vendor_id} {ep_device.device.vendor}")
```

The output contains duplicate GPU EP devices.
```
2157111435568 CPUExecutionProvider OrtHardwareDeviceType.CPU 0 32902 GenuineIntel
2157111435568 DmlExecutionProvider OrtHardwareDeviceType.GPU 9988 4318 NVIDIA
2157111435568 DmlExecutionProvider OrtHardwareDeviceType.GPU 9988 4318 NVIDIA
2157111435568 NvTensorRTRTXExecutionProvider OrtHardwareDeviceType.GPU 9988 4318 NVIDIA
2157111435568 NvTensorRTRTXExecutionProvider OrtHardwareDeviceType.GPU 9988 4318 NVIDIA
```

This result in the following error when creating session:

```
File "C:\Users\xx\AppData\Local\miniconda3\envs\xx\Lib\site-packages\onnxruntime\capi\onnxruntime_inference_collection.py", line 570, in _create_inference_session
    sess.initialize_session(providers, provider_options, disabled_optimizers)
onnxruntime.capi.onnxruntime_pybind11_state.Fail: [ONNXRuntimeError] : 1 : FAIL : Provider NvTensorRTRTXExecutionProvider has already been registered.
```
https://github.com/microsoft/win-onnxruntime/blob/e46c0d86b9cd15fce1dd6d53fdf03a22edce1ff7/onnxruntime/core/framework/execution_providers.h#L77

Additionally, this error can also occur if your device has two identical graphics cards installed.
So we try to add only one device for each EP to resolve this error.

FYI:
https://microsoft.visualstudio.com/OS/_workitems/edit/57991508

### DIFF
--- a/olive/common/ort_inference.py
+++ b/olive/common/ort_inference.py
@@ -78,7 +78,14 @@ def initialize_inference_session_options(
     provider_options = provider_options or []
     provider_options_by_ep = dict(zip(providers, provider_options))
     ort_device_type = get_ort_hardware_device_type(device)
-    for ep_device in ort.get_ep_devices():
+
+    # Remove duplicate EP devices to avoid conflicts
+    # In remote connection scenarios, ort.get_ep_devices() may return duplicate
+    # ep_device.ep_name and ep_device.device.type combinations, but we can only
+    # call add_provider_for_devices once for each unique EP device combination.
+    # Use a dictionary to deduplicate by (ep_name, device_type) key.
+    unique_ep_devices = { (ep.ep_name, ep.device.type): ep for ep in ort.get_ep_devices() }
+    for ep_device in unique_ep_devices.values():
         if ep_device.device.type == ort_device_type and ep_device.ep_name in provider_options_by_ep:
             sess_options.add_provider_for_devices([ep_device], provider_options_by_ep.get(ep_device.ep_name) or {})
 
--- a/olive/common/ort_inference.py
+++ b/olive/common/ort_inference.py
@@ -84,6 +84,7 @@ def initialize_inference_session_options(
     # ep_device.ep_name and ep_device.device.type combinations, but we can only
     # call add_provider_for_devices once for each unique EP device combination.
     # Use a dictionary to deduplicate by (ep_name, device_type) key.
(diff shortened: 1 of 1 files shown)


xieofxie:
how about the case for multiple graph cards? The unique check should consider more information

shaahji:
In my opinion, there shouldn't be any change on Olive front to fix the underlying issue. If the intended behavior is to expect unique devices, then onnxruntime API should do so. Adding a filter in Olive seems like an unnecessary Band-Aid that only hides the underlying issue rather than fixing it.

xieofxie:
> In my opinion, there shouldn't be any change on Olive front to fix the underlying issue. If the intended behavior is to expect unique devices, then onnxruntime API should do so. Adding a filter in Olive seems like an unnecessary Band-Aid that only hides the underlying issue rather than fixing it.

Currently we don't have an estimation about when this will be fixed in onnxruntime. How about we add a `if onnxruntime version <= xxx` for this patch, then
- if new version still does not fix it ,we will patch Olive for the new version
- if new version fixes it, Olive will automatically support it

fangyangci:
> In my opinion, there shouldn't be any change on Olive front to fix the underlying issue. If the intended behavior is to expect unique devices, then onnxruntime API should do so. Adding a filter in Olive seems like an unnecessary Band-Aid that only hides the underlying issue rather than fixing it.

I restrict the patch to the existing version of ONNX Runtime to ensure compatibility. This allows future working versions to integrate seamlessly without requiring additional changes.

shaahji (approved):

PR #1974, opened 2025-07-14
Fix AddOliveMetadata pass to preserve external data files

ashrit-ms:
## Describe your changes
Refactored the AddOliveMetadata ONNX pass to ensure all external data files (e.g., .bin, .xml) are preserved when adding metadata to ONNX models.

Key changes:
- Always copy entire input directory to output location to preserve all external files alongside the main ONNX model
- Simplified pass logic by treating all models as directory-based
- Load ONNX models without external data to preserve file structure
- Return directory-based ONNXModelHandler for consistent behavior

Added unit tests:
- test_add_metadata_with_external_data_files: Verifies external file preservation for directory-based models with mock .bin/.xml
- test_add_metadata_single_file_to_directory_conversion: Ensures correct metadata handling for single-file ONNX models

This fix addresses issues where external data files were lost during metadata addition, ensuring model integrity for ONNX/OpenVINO models with external dependencies.

### DIFF
--- a/olive/passes/onnx/add_metadata.py
+++ b/olive/passes/onnx/add_metadata.py
@@ -4,13 +4,12 @@
 # --------------------------------------------------------------------------
 
 import logging
-from typing import Union
 
 import onnx
 
 from olive.common.utils import hardlink_copy_dir
 from olive.hardware import AcceleratorSpec
-from olive.model import CompositeModelHandler, ONNXModelHandler
+from olive.model import ONNXModelHandler
 from olive.passes import Pass
 from olive.passes.onnx.common import get_external_data_config
 from olive.passes.pass_config import BasePassConfig, PassConfigParam
@@ -133,22 +132,15 @@ class AddOliveMetadata(Pass):
         import shutil
         from pathlib import Path
 
-        input_model_path = Path(model.model_path)
+        input_onnx_file = Path(model.model_path)
         output_path = Path(output_model_path)
 
-        # Determine input model directory and ONNX file
-        if input_model_path.is_dir():
-            input_model_dir = input_model_path
-            if model.onnx_file_name:
-                input_onnx_file = input_model_path / model.onnx_file_name
-            else:
-                onnx_files = list(input_model_path.glob("*.onnx"))
-                if not onnx_files:
-                    raise ValueError(f"No ONNX file found in model directory: {input_model_path}")
-                input_onnx_file = onnx_files[0]
-        else:
-            input_model_dir = input_model_path.parent
-            input_onnx_file = input_model_path
+        # Determine input model directory
+        if input_onnx_file.is_dir():
+            # This shouldn't happen since model.model_path should resolve to the ONNX file
+            raise ValueError(f"Model path resolved to directory, expected ONNX file: {input_onnx_file}")
+
+        input_model_dir = input_onnx_file.parent
 
         # Load ONNX model without external data to preserve file structure
         onnx_model = onnx.load_model(str(input_onnx_file), load_external_data=False)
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/add_metadata.py
```diff
@@ -129,18 +128,33 @@ def _run_for_config(
...
+        # Determine input model directory and ONNX file
+        if input_model_path.is_dir():
+            input_model_dir = input_model_path
+            if model.onnx_file_name:
+                input_onnx_file = input_model_path / model.onnx_file_name
+            else:
+                onnx_files = list(input_model_path.glob("*.onnx"))
+                if not onnx_files:
+                    raise ValueError(f"No ONNX file found in model directory: {input_model_path}")
+                input_onnx_file = onnx_files[0]
```
shaahji: Will this work for multi-modal scenario?

It would be safer to use only if a single modal is found. In other cases throw a relevant error and ask the user to explicitly provider a model file name.

olive/passes/onnx/add_metadata.py
```diff
@@ -151,20 +165,25 @@ def _run_for_config(
...
+        else:
+            output_dir = output_path
 
-        # Set graph name (always required)
-        onnx_model = self._set_graph_name(onnx_model, graph_name)
+        # Copy entire input directory to preserve external files
+        output_dir.parent.mkdir(parents=True, exist_ok=True)
+        if output_dir.exists():
+            shutil.rmtree(output_dir)
+        shutil.copytree(input_model_dir, output_dir)
```
shaahji: User hard symlink instead of making duplicate copies of files?

olive/passes/onnx/add_metadata.py
ashrit-ms: Just want to confirm if you referring to multi mod'e'l scenarios or multi mod'a'l scenarios?

For multi model scenarios, Olive automatically applies the pass to each component of a composite model.

olive/passes/onnx/add_metadata.py
```diff
@@ -151,20 +159,25 @@ def _run_for_config(
...
+            output_dir = output_path.parent / output_path.stem
+        else:
+            output_dir = output_path
 
-        # Set graph name (always required)
-        onnx_model = self._set_graph_name(onnx_model, graph_name)
+        # Copy entire input directory to preserve external files using hardlinks
+        output_dir.parent.mkdir(parents=True, exist_ok=True)
+        if output_dir.exists():
+            shutil.rmtree(output_dir)
```
shaahji: Is this really a good idea to delete the output directory, especially this late in the pass logic implementation? There could be files in this directory that were created as part of this pass before this statement executes. You would blindly be deleting your own work.

Also, what happens if this directory was a hardlink itself?

olive/passes/onnx/add_metadata.py
```diff
@@ -151,20 +159,25 @@ def _run_for_config(
...
+        hardlink_copy_dir(input_model_dir, output_dir)
```
shaahji, line 174: With use of hardlink, both _input_model_dir_ and _output_dir_ will basically be referring to the same directory on disk. If you write anything to this directory, you are basically modifying the contents of _input_model_dir_ directory also. I don't think that's the intent here. Please check your logic or make your intent clear with comments.

olive/passes/onnx/add_metadata.py
ashrit-ms: Fixed. We no longer delete the output_dir.

olive/passes/onnx/add_metadata.py
ashrit-ms: Fixed.

PR #1980, opened 2025-07-16
remove version check when add_provider_for_devices

fangyangci:
## 
Based on @skottmckay 's explanation, it is reasonable to filter out duplicate EPs. Given this, the version detection logic appears to be redundant and can be safely removed. 

### DIFF
--- a/olive/common/ort_inference.py
+++ b/olive/common/ort_inference.py
@@ -80,24 +80,17 @@ def initialize_inference_session_options(
     provider_options_by_ep = dict(zip(providers, provider_options))
     ort_device_type = get_ort_hardware_device_type(device)
 
-    # NOTE: In the current latest version 1.22.0:
     # ort.get_ep_devices may return ep_devices with the same ep_name and device, for example when connecting remotely or when there are multiple graph cards.
     # However, in onnxruntime, each EP name can only be added once. See: https://github.com/microsoft/onnxruntime/blob/fb0f6c652be5db0a3182c424a995efecf792d41c/onnxruntime/core/framework/execution_providers.h#L75
-    if version.parse(ort.__version__) < version.parse("1.22.0"):
-        added_ep_names = set()
-        for ep_device in ort.get_ep_devices():
-            if (
-                ep_device.device.type == ort_device_type
-                and ep_device.ep_name in provider_options_by_ep
-                and ep_device.ep_name not in added_ep_names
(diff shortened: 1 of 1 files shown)


olive/common/ort_inference.py
```diff
@@ -80,24 +80,17 @@ def initialize_inference_session_options(
...
-            ):
-                added_ep_names.add(ep_device.ep_name)
-                sess_options.add_provider_for_devices([ep_device], provider_options_by_ep.get(ep_device.ep_name) or {})
-    else:
-        # In ort v1.22.0 and later, we can use add_provider_for_devices to add providers with options.
-        for ep_device in ort.get_ep_devices():
-            if ep_device.device.type == ort_device_type and ep_device.ep_name in provider_options_by_ep:
-                sess_options.add_provider_for_devices([ep_device], provider_options_by_ep.get(ep_device.ep_name) or {})
+    added_ep_names = set()
+    for ep_device in ort.get_ep_devices():
```
devang-ml, line 86: Is `get_ep_devices()` available in ALL versions of ORT released in last one year or so?

shaahji:
AFAIK, the `ort.get_ep_devices()` wasn't introduced until after 1.22.0 so I don't think this will work for older releases. You might want to confirm with either 1.22.0 or a slightly older release.

xieofxie:
> AFAIK, the `ort.get_ep_devices()` wasn't introduced until after 1.22.0 so I don't think this will work for older releases. You might want to confirm with either 1.22.0 or a slightly older release.

Hi, this is part of revert of this pr https://github.com/microsoft/Olive/pull/1963/files it looks previously there is no check for `ort.get_ep_devices()`

shaahji:
> > AFAIK, the `ort.get_ep_devices()` wasn't introduced until after 1.22.0 so I don't think this will work for older releases. You might want to confirm with either 1.22.0 or a slightly older release.
> 
> Hi, this is part of revert of this pr https://github.com/microsoft/Olive/pull/1963/files it looks previously there is no check for `ort.get_ep_devices()`

I guess, I missed the version check in the previous commit/review. This code will not work for `ort <= 1.22.0` since the `ort.get_ep_devices()` API doesn't exist. In your previous commit, only the else part was valid. 

I recommend creating a local environment with both versions (<=1.22.0 & >1.22.0) and trying it out to confirm.

PR #1984, opened 2025-07-18
add get_ep_devices

chinazhangchao:

### DIFF
--- a/olive/systems/local.py
+++ b/olive/systems/local.py
@@ -49,6 +49,10 @@ class LocalSystem(OliveSystem):
         """Get the available execution providers."""
         import onnxruntime as ort
 
+        if hasattr(ort, "get_ep_devices"):
+            ep_devices = ort.get_ep_devices()
+            return [ep_device.ep_name for ep_device in ep_devices]
+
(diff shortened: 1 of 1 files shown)


olive/__init__.py
```diff
@@ -16,6 +16,22 @@
...
+    import onnxruntime as ort
+
+    # pylint: disable=protected-access
+    ort._get_available_providers = ort.get_available_providers
+
+    def get_available_providers_winml():
+        # pylint: disable=protected-access
+        providers = ort._get_available_providers()
+        extra_providers = {ep_device.ep_name for ep_device in ort.get_ep_devices()} - set(providers)
+        return providers + list(extra_providers)
```
xieofxie, line 29: one pitfall for this: the get_available_providers has a fixed order, we should keep the order https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/get_execution_providers.cc
I think we had better filter from get_all_providers https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/onnxruntime_pybind_state.cc#L1618

PR #1990, opened 2025-07-19
Refactor docker system

xiaoyu-work:
## Describe your changes

Refactor docker system:
1. If user specific docker system as host, the whole workflow will run in the docker contianer.
2. Remove pass run and evaluation run support for docker system.
3. No need to add workflow_host for docker system.
4. Remove managed docker system support. User needs to provide his own Dockerfile.
5. Remove unused docker integration test.
6. Update unit tests and docs.


olive/systems/accelerator_creator.py
```diff
@@ -56,12 +56,11 @@ def normalize(self) -> "SystemConfig":
...
                     self._fill_device()
             else:
-                # for AzureML and Docker System
+                # for AzureML
                 for accelerator in self.system_config.config.accelerators:
                     if not accelerator.device or (not accelerator.execution_providers and self.is_ep_required):
                         raise ValueError(
-                            "AzureML and Docker system requires device and execution providers to be specified "
-                            "explicitly."
+                            "AzureML system requires device and execution providers to be specified explicitly."
```
jambayk, line 63: I think this message still needs to refer to docker system? docker system still meets the condition here and we need the ep to be given.

xiaoyu-work: If the host is docker system, the workflow will be submitted to docker container to run, and this `normalize` is called when we run the workflow: https://github.com/microsoft/Olive/blob/main/olive/workflows/run/run.py#L192. The host is replaced with local host already.

jambayk: so in the original calling environment, the code will never reach here?

xiaoyu-work: Yes, it will never reach here.

shaahji (approved):

PR #2000, opened 2025-07-22
Add QuickGeluToSigmoid surgery

xiaoyu-work:
## Describe your changes

Add QuickGeluToSigmoid surgery. Lower QuickGelu to Mul + Sigmoid + Mul: `QuickGelu(x) = x * sigmoid(alpha * x)`

### DIFF
--- a/docs/source/features/onnx-transformations.md
+++ b/docs/source/features/onnx-transformations.md
@@ -745,6 +745,87 @@ graph {
 }
 ```
 
+### `QuickGeluToSigmoid`
+
+#### Description
+
+Replaces QuickGelu operators with equivalent standard ONNX operators. This surgery converts a single QuickGelu node into a subgraph composed of Mul, Sigmoid, and Mul operations.
+
+The QuickGelu function is mathematically defined as: `QuickGelu(x) = x * sigmoid(alpha * x)`, where alpha defaults to 1.702.
+
+Reference: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/graph/contrib_ops/contrib_defs.cc
+
+#### Example
+
+Initial model graph:
+
+```
+graph {
+  input: "input"
+  node {
+    op_type: "QuickGelu"
+    input: ["input"]
+    output: ["quickgelu_output"]
+    name: "QuickGeluNode"
+  }
+  output: "quickgelu_output"
+}
+```
+
+After applying:
+
+```json
+{
+    "type": "GraphSurgeries",
+    "surgeries": [
+        {
+            "surgeon": "QuickGeluToSigmoid"
+        }
+    ]
+}
+```
+
+Transformed model graph:
+
+```
+graph {
+  input: "input"
+  initializer: "QuickGeluNode_alpha" (FLOAT, value: 1.702)
+  node {
+    op_type: "Mul"
+    input: ["input", "QuickGeluNode_alpha"]
+    output: ["QuickGeluNode_mul1_output"]
+    name: "QuickGeluNode_mul1"
+  }
+  node {
+    op_type: "Sigmoid"
+    input: ["QuickGeluNode_mul1_output"]
+    output: ["QuickGeluNode_sigmoid_output"]
+    name: "QuickGeluNode_sigmoid"
+  }
+  node {
+    op_type: "Mul"
+    input: ["input", "QuickGeluNode_sigmoid_output"]
+    output: ["QuickGeluNode_mul2_output"]
+    name: "QuickGeluNode_mul2"
+  }
+  output: "QuickGeluNode_mul2_output"
+}
+```
+
+Pattern transformation:
+
+```
+Original pattern:
+[Input] --> QuickGelu --> [Output]
+
+Replaced pattern:
+[Input] --> Mul --> Sigmoid --> Mul --> [Output]
+             |                    ^
+             |                    |
+             +--------------------+
+             (alpha=1.702)
+```
 
 ### `RMSNormToL2Norm`
 
--- a/docs/source/features/onnx-transformations.md
+++ b/docs/source/features/onnx-transformations.md
@@ -759,7 +759,7 @@ Reference: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/g
 
 Initial model graph:
 
-```
+```proto
 graph {
   input: "input"
   node {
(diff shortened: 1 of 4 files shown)


olive/passes/onnx/graph_surgeries.py
```diff
@@ -447,6 +448,51 @@ def __call__(self, model: ModelProto):
...
+        # Create alpha constant
+        alpha = op.Constant(value_float=self.ALPHA)
+
+        # Compute: x * sigmoid(alpha * x)
+        alpha_x = op.Mul(alpha, x)
+        sigmoid_alpha_x = op.Sigmoid(alpha_x)
+        return op.Mul(x, sigmoid_alpha_x)
+
+    def call_ir(self, model: ir.Model) -> ir.Model:
+        from onnxscript import rewriter
```
justinchuby: Maybe import rewriter on the top? Since pattern was imported there as well

docs/source/features/onnx-transformations.md
```diff
@@ -745,6 +745,87 @@ graph {
...
+
+The QuickGelu function is mathematically defined as: `QuickGelu(x) = x * sigmoid(alpha * x)`, where alpha defaults to 1.702.
+
+Reference: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/graph/contrib_ops/contrib_defs.cc
+
+#### Example
+
+Initial model graph:
+
+```
```
justinchuby:
```suggestion
```json
```

olive/passes/onnx/graph_surgeries.py
```diff
@@ -447,6 +448,51 @@ def __call__(self, model: ModelProto):
...
+
+    def __init__(self):
+        super().__init__()
+        self._rule = pattern.RewriteRule(
+            self._target_pattern,
+            self._replacement_pattern,
+        )
+
+    def _target_pattern(self, op, x):
+        return op.QuickGelu(x)
```
justinchuby: Do you need to specify the domain here?

docs/source/features/onnx-transformations.md
xiaoyu-work: This is not a json

olive/passes/onnx/graph_surgeries.py
xiaoyu-work: Done!

olive/passes/onnx/graph_surgeries.py
xiaoyu-work: Done!

docs/source/features/onnx-transformations.md
justinchuby:
```suggestion
```txtpb
```

maybe this?

olive/passes/onnx/graph_surgeries.py
```diff
@@ -447,6 +452,49 @@ def __call__(self, model: ModelProto):
...
+            self._target_pattern,
+            self._replacement_pattern,
+        )
+
+    def _target_pattern(self, op, x):
+        return op.QuickGelu(x, _domain=MSFT_DOMAIN)
+
+    def _replacement_pattern(self, op, x):
+        # Create alpha constant
+        alpha = op.Constant(value_float=self.ALPHA)
```
justinchuby: You may also want to cast alpha to the same type as x with CastLike

justinchuby: Or if we assume that the type of x is known, we can do

```
alpha = op.Constant(value=ir.tensor(self.ALPHA, dtype=x.dtype))
```

maybe use an if condition on whether x.dtype is not None?

xiaoyu-work: I assume x always has dtype. Any case that x.dtype could be None?

justinchuby: Any intermediate value may not have a dtype in an onnx graph, that is unless some previous passes allows for additional guarantees on the graph (for example if we know there is a shape inference pass that will guarantee dtypes).

xiaoyu-work: Not sure about the Nivida use case of this Pass but i added a check here.

justinchuby: For model builder models there is always types. maybe an assert will do?

xiaoyu-work: Yeah they primarily want to handle model builder model.

justinchuby (approved):

olive/passes/onnx/graph_surgeries.py
```diff
@@ -447,6 +452,50 @@ def __call__(self, model: ModelProto):
...
+            self._replacement_pattern,
+        )
+
+    def _target_pattern(self, op, x):
+        return op.QuickGelu(x, _domain=MSFT_DOMAIN)
+
+    def _replacement_pattern(self, op, x):
+        # Create alpha constant
+        x_dtype = x.dtype if x.dtype is not None else ir.DataType.FLOAT
+        alpha = op.Constant(value=ir.tensor(self.ALPHA, dtype=x_dtype))
```
shaahji, line 486: What if the graph already has this node?

test/unit_test/passes/onnx/test_graph_surgeries.py
shaahji, line 1: It would be worth adding a test where the alpha node is already in the graph.

olive/passes/onnx/graph_surgeries.py
xiaoyu-work: The node name seems random. Can you confirm this? @justinchuby

justinchuby: Yes. All the new nodes will take (not random but) unique names. If you worry about duplicated constants, I would suggest a separate dedup pass (https://github.com/onnx/ir-py/blob/main/src/onnx_ir/passes/common/initializer_deduplication.py)

xiaoyu-work: If op.Constant guarantee creating a unique name, this case won't happen. Same to other mul and Sigmoid node.

shaahji: Please add a work item to implement a dedplication run post all surgeries.

xiaoyu-work: Done!

shaahji (approved):

PR #2004, opened 2025-07-24
[NVMO] Add nodes_to_exclude and save_external_data support

vishalpandya1990:
## Describe your changes

- Adding a config option for nodes_to_exclude. 
- Saving the onnx data in external file during onnx.save call.
- In NVMO's Phi3 example readme, adding a note for the user to try out different onnxruntime/onnxruntime-genai packages in case of any issue (e.g. versions compatibility) with default installed versions of these components.

### DIFF
--- a/.azure_pipelines/job_templates/huggingface-login-template.yaml
+++ b/.azure_pipelines/job_templates/huggingface-login-template.yaml
@@ -2,5 +2,5 @@ parameters:
   hf_token: 'huggingface_token'
 
 steps:
-- script: huggingface-cli login --token ${{ parameters.hf_token }}
+- script: hf auth login --token ${{ parameters.hf_token }}
   displayName: 'Hugging Face Login'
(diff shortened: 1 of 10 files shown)


olive/passes/onnx/nvmo_quantization.py
```diff
@@ -475,7 +487,14 @@ def _run_for_config(
 
             output_model_path = resolve_onnx_path(output_model_path, Path(model.model_path).name)
 
-            onnx.save(converted_model_proto, output_model_path)
+            onnx.save(
```
jambayk: this part is overriden since model_proto_to_olive_model does the same thing. You can provide the `save_as_external_data` related options to the utils as a dict in the third argument https://github.com/microsoft/Olive/blob/fcb643219c67091db9e0d07082c5266d52ee6290/olive/passes/onnx/common.py#L156

```
model_proto_to_olive_model(
     converted_model_proto, 
    output_model_path, 
    { "save_as_external_data": True}
)
```
It handles the external data file name creation and the rest are defaults.

vishalpandya1990: Updated.

jambayk (approved):

PR #2008, opened 2025-07-25
Add more options/params to Olive from model builder

selenayang888:
Since model builder has new run parameters, I added more options to Olive based on these new extra parameters from model builder.

### DIFF
--- a/.azure_pipelines/job_templates/huggingface-login-template.yaml
+++ b/.azure_pipelines/job_templates/huggingface-login-template.yaml
@@ -2,5 +2,5 @@ parameters:
   hf_token: 'huggingface_token'
 
 steps:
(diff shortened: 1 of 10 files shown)


olive/passes/onnx/model_builder.py
```diff
@@ -223,6 +223,22 @@ def _run_for_config(
         if config.int4_is_symmetric is not None:
             extra_args["int4_is_symmetric"] = config.int4_is_symmetric
 
+        # Use 8-bit quantization for MoE layers
+        if config.use_8bits_moe is not None:
```
jambayk, line 247: thse options also need to be added to `default_config` at line 49

jambayk (approved):

PR #2017, opened 2025-07-28
Intel GPU model and general fixes

kaunain-98:
1. Shift task from ov_quant_config to extra_config for GPU models in config files
2. Pass task paramater to optimum optimum_intel
3. reuse_cache fixe to explicitly release lock on ov model
4. Remove throwing explicit error messages from python packages

### DIFF
--- a/.azure_pipelines/job_templates/olive-test-cpu-template.yaml
+++ b/.azure_pipelines/job_templates/olive-test-cpu-template.yaml
@@ -35,34 +35,18 @@ jobs:
     parameters:
       hf_token: $(hf_token)
 
-  - ${{ if and(eq(variables.WINDOWS, 'True'), eq(variables.testType, 'multiple_ep')) }}:
-    - script: |
-          call python -m pip install pytest
-          call curl --output openvino_toolkit.zip https://storage.openvinotoolkit.org/repositories/openvino/packages/2023.0.1/windows/w_openvino_toolkit_windows_2023.0.1.11005.fa1c41994f3_x86_64.zip
-          call 7z x openvino_toolkit.zip
-          call w_openvino_toolkit_windows_2023.0.1.11005.fa1c41994f3_x86_64\\setupvars.bat
-          call python -m pip install numpy psutil coverage
-          call coverage run --source=$(Build.SourcesDirectory)/olive -m pytest -v -s -p no:warnings --disable-warnings --log-cli-level=WARNING --junitxml=$(Build.SourcesDirectory)/logs/test-TestOlive.xml $(Build.SourcesDirectory)/test/$(testType) --basetemp $(PYTEST_BASETEMP)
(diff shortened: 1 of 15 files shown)


olive/passes/openvino/quantization.py
```diff
@@ -268,6 +268,8 @@
         output_model_name = f"{model_name}_quant.xml"
         output_model_path = Path(output_model_path) / output_model_name
         ov.save_model(quantized_model, output_model=output_model_path)
+        del quantized_model
+        del loaded_model
```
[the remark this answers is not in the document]
kaunain-98: openvino sometimes keeps a lock on the file until explicit released

olive/passes/openvino/quantization.py
```diff
@@ -386,3 +388,5 @@
...
```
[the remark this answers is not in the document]
kaunain-98: openvino sometimes keeps a lock on the file until explicit released

olive/passes/openvino/optimum_intel.py
```diff
@@ -73,13 +73,13 @@ def infer_task(
 ):
     try:
         from optimum.exporters import TasksManager
-    except Exception as e:
-        raise ImportError("Unable to import optimum packages:", e) from None
+    except Exception:
+        raise ImportError("Unable to import optimum packages") from None
```
xiaoyu-work: Adding a solution in error message?

kaunain-98: We prefer obfuscated throwing explicit error messages from every package and in turn provides a simple, traceable, generic error message.

xiaoyu-work: How do you expect users to debug and resolve the issue if the original error is completely hidden?

kaunain-98: if this is MSFT standard we can add it back, since it wasnt raised for any of previous commits we wanted to be consistent

xiaoyu-work: I'm okay with previous commit as it will print the error clearly:
```
Traceback (most recent call last):
  File "C:\Users\xiaoyuzhang\test.py", line 4, in <module>
    raise ImportError("Unable to import optimum packages:", e) from None
ImportError: ('Unable to import optimum packages:', ModuleNotFoundError("No module named 'optimum'"))
```
While this change will hide the actual error for the user:
```
Traceback (most recent call last):
  File "C:\Users\xiaoyuzhang\test.py", line 4, in <module>
    raise ImportError("Unable to import optimum packages") from None
ImportError: Unable to import optimum packages
```

kaunain-98: ok will revert the error changes

xiaoyu-work: Thanks!

PR #2052, opened 2025-08-04
Register EP libraries in ORT

jambayk:
## Describe your changes
- New ORT design has EPs as libraries that can be registered on the core runtime. 
- New API is used only for Windows + ORT 1.23+. Linux support is not well defined and the apis don't seem to work.
- Provide a way for users to provide the EP as tuple where the second element is the local path to the library. This is optional and the ep can still be provided as a single string.
- If the ORT package was built with additional eps such as cuda, vitisai or qnn, we automatically register them using the name of the dll which will be in the `onnxruntime/capi` directories with the other ort libaries.
- Windows ML case is not supported yet since it's not clear what is required. 

### DIFF
--- a/olive/common/ort_inference.py
+++ b/olive/common/ort_inference.py
@@ -71,11 +71,10 @@ def get_ort_available_providers():
         return ort.get_available_providers()
 
     # only return registered EPs since session with be created with SessionOptions.add_provider_for_devices()
-    providers = []
-    for ep_device in ort.get_ep_devices():
-        if ep_device.ep_name not in providers:
-            providers.append(ep_device.ep_name)
-    return providers
+    # this is ordered by priority
+    all_providers = ort.get_all_providers()
+    available_provider_set = {ep_device.ep_name for ep_device in ort.get_ep_devices()}
+    return [ep_name for ep_name in all_providers if ep_name in available_provider_set]
 
 
 def get_ort_hardware_device_type(device: Union["Device", str]):


olive/common/ort_inference.py
```diff
@@ -37,6 +38,43 @@ def ort_supports_ep_devices() -> bool:
...
+
+
+def maybe_register_ep_libraries(ep_paths: dict[str, str]):
+    """Register execution provider libraries if onnxruntime supports it."""
+    if not ort_supports_ep_devices():
+        return
+
+    import onnxruntime as ort
+
+    ep_paths = maybe_add_winml_ep_paths(ep_paths)
```
chinazhangchao: According to the [doc](https://learn.microsoft.com/en-us/windows/ai/new-windows-ml/tutorial?tabs=python#acquiring-the-model-and-preprocessing) as following, I think `maybe_add_winml_ep_paths` function should be called in a separate process.
```
# Known issue: import winrt.runtime will cause the TensorRTRTX execution provider to fail registration.
# As a workaround, please run pywinrt related code in a separate thread.
```

olive/common/ort_inference.py
```diff
@@ -37,6 +38,43 @@ def ort_supports_ep_devices() -> bool:
     return version.parse(OrtVersion).release >= version.parse("1.23.0").release
 
 
+def maybe_add_winml_ep_paths(ep_paths: dict[str, str]) -> dict[str, str]:
+    """Get the Windows ML EP paths."""
+    try:
+        from onnxruntime import winml as _  # noqa: F401 # pylint: disable=unused-import
```
xieofxie: this will not work. onnxruntime-winml GA is same as onnxruntime-dml.

We should check if we are in Windows and we support `register_execution_provider_library`, then try this. And if we met error, we may be advice user to use onnxruntime-winml in Windows with proper app sdk version and Windows version etc.

xieofxie:
AITK has patched Olive from our side, so maybe we could wait for Winml GA to make the final changes

olive/common/ort_inference.py
```diff
@@ -31,48 +32,81 @@ class OrtSessionFallbackError(Exception):
...
+    """Get the available providers for ONNXRuntime."""
+    import onnxruntime as ort
+
+    if not ort_supports_ep_devices():
+        # what ORT was built with. Session will be created directly with InferenceSession(model_path, providers=["ProviderName"])
+        return ort.get_available_providers()
+
+    # only return registered EPs since session with be created with SessionOptions.add_provider_for_devices()
+    providers = []
+    for ep_device in ort.get_ep_devices():
```
xieofxie: maybe 
```
all_providers = ort.get_all_providers()
        available_providers_set = set( [ep_device.ep_name for ep_device in ort.get_ep_devices()])
        return [provider for provider in all_providers if provider in available_providers_set]
``` 

to keep order as previous

jambayk: updated. thanks!

PR #2056, opened 2025-08-06
Always register all available providers

jambayk:
## Describe your changes
- To keep behavior consistent with ORT <1.23, register all EPs that the package was built with.

### DIFF
--- a/olive/common/ort_inference.py
+++ b/olive/common/ort_inference.py
@@ -44,17 +44,19 @@ def maybe_register_ep_libraries(ep_paths: dict[str, str]):
     import onnxruntime as ort
 
     # providers that ort was built with such as CUDA, QNN, VitisAI but need registration
-    additional_providers = set(ort.get_available_providers()) - {"CPUExecutionProvider", "DmlExecutionProvider"}
+    for provider in set(ort.get_available_providers()):
+        if ep_paths.get(provider) is None:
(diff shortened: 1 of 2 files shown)


AnanyaA-9:
Thank you @jambayk , when we run the config after removing local_system, target and host, registration successfully happens.
However, given these changes to config:
- Added only the EP name to inference_settings in metrics under evaluator
- Removed local_system, target and host
we see this error: `The onnxruntime fallback happens. VitisAIExecutionProvider is not in the session providers ['CPUExecutionProvider']. session._enable_fallback = True`

It seems to be because:
- In ⁠[initialize_inference_session_options()](https://github.com/microsoft/Olive/blob/8ab01c5c869634f2586f2299fca30e7f6ab1a4c8/olive/common/ort_inference.py#L106-L130), when acceleratorSpec is not specified in config, device defaults to NPU. 
- At this ⁠[line](https://github.com/microsoft/Olive/blob/8ab01c5c869634f2586f2299fca30e7f6ab1a4c8/olive/common/ort_inference.py#L121), ort_device_type is CPU and doesn't match ep_device.device.type which is NPU for VitisAI EP
- So the ⁠[condition](https://github.com/microsoft/Olive/blob/8ab01c5c869634f2586f2299fca30e7f6ab1a4c8/olive/common/ort_inference.py#L120-L123) isn't satisfied and add_provider_for_devices() isn't called.

We would like to confirm that this is expected behaviour.

cc @VishalX

PR #2057, opened 2025-08-06
Onnx Conversion: Make QuantLinear export aware of mode

jambayk:
## Describe your changes
- torch 2.8 is released and now supports dynamo export for quantized linear modules as well
- But the dynamo export code path is triggered even under torchscript export so need to add an additional condition for export mode.

### DIFF
--- a/olive/common/hf/quant.py
+++ b/olive/common/hf/quant.py
@@ -78,11 +78,12 @@ def get_olive_qlinear_cls(quantization_config):
 
 @torch.no_grad()
 def _replace_qlinear_modules(
-    model: torch.nn.Module, mapping: dict[str, tuple[Callable, Callable]], desc: str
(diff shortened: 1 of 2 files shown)


xiaoyu-work (dismissed):

olive/common/hf/quant.py
```diff
@@ -161,9 +162,10 @@ def forward(
         group_size: int,
         in_features: int,
         out_features: int,
+        dynamo: bool = False,
     ):
         if torch.onnx.is_in_onnx_export():
```
justinchuby, line 167: Instead, you may do 

```
if torch.onnx.is_in_onnx_export() and torch.compiler.is_exporting()
```

to avoid having to pass the dynamo parameter around. See https://docs.pytorch.org/docs/2.7/generated/torch.compiler.is_exporting.html

PR #2064, opened 2025-08-08
Add QAIRT MHA2SHA transformation support

qti-kromero:
## Describe your changes

- Add MHA2SHA pass for QAIRT SDK that enables better performance on Qualcomm NPU


test/unit_test/passes/onnx/test_qairt_mha2sha.py
jambayk, line 1: we recently moved this folder to `test/passes/onnx`. could you merge main and move this to the new folder?

olive/passes/onnx/qairt/mha2sha.py
```diff
@@ -0,0 +1,106 @@
...
+from olive.hardware import AcceleratorSpec
+from olive.model import CompositeModelHandler, ONNXModelHandler
+from olive.passes.olive_pass import Pass
+from olive.passes.onnx.common import get_external_data_config
+from olive.passes.pass_config import BasePassConfig, PassConfigParam
+
+logger = logging.getLogger(__name__)
+
+
+class QairtMHA2SHA(Pass):
```
jambayk, line 19: can this be applied before the model is split? so that the logic is simpler and we only deal with the pre-split single model?

jambayk: or the underlying tool does some node rename that causes the splitmodel pass to not work correctly?

jambayk: we could also run this pass before the staticllm pass. there is nothing special about the split model before the staticllm pass so this pass can just operate on the split components independently https://github.com/microsoft/Olive/blob/9cb9dc4f47cee0deeac3d7d7dad3d5c2f9ffcd67/olive/passes/olive_pass.py#L229. unless you require the models from the staticllm pass because of the fixed shapes

qti-kromero: theoretically it could be done pre-split but in practice we're unsure if this would be compatible with MSFT's splitting algorithm as noted above - we have our own splitting algorithm that we are looking to potentially integrate here and we perform this transformation after splitting within our APIs

qti-kromero: > we could also run this pass before the staticllm pass. there is nothing special about the split model before the staticllm pass so this pass can just operate on the split components independently
> 
> https://github.com/microsoft/Olive/blob/9cb9dc4f47cee0deeac3d7d7dad3d5c2f9ffcd67/olive/passes/olive_pass.py#L229
> 
> . unless you require the models from the staticllm pass because of the fixed shapes

I see - let me look into this

jambayk: post split is fine too. I am thinking we do it before the static llm pass so that we don't need to worry about dealing with composite models inside the pass. the pass runner will iterate over the components separately and it's not an issue since the splits are not tied to each other with shared resources unlike the output of staticllm pass

jambayk: I think there might be some of the tokenizer files etc that are copied into the split sub directories this way but I can do a follow up to clean up that behavior.

olive/olive_config.json
```diff
@@ -506,6 +506,14 @@
             "supported_algorithms": [  ],
             "supported_quantization_encodings": [  ]
         },
+        "QairtMHA2SHA": {
```
jambayk: can you move it further up in alphabetical order? thanks!

olive/passes/onnx/qairt/mha2sha.py
```diff
@@ -0,0 +1,68 @@
...
+
+        qairt_onnx_model = OnnxModel.load(model_path=model.model_path)
+        try:
+            qairt_onnx_model.mha2sha_v2(**(config.mha2sha_kwargs if config.mha2sha_kwargs is not None else {}))
+        except AttributeError:
+            # Backwards compatibility with older definitions of OnnxModel in <= QAIRT 2.37
+            logger.warning("MHA2SHA V2 is not available for this SDK version, defaulting to MHA2SHA V1")
+            qairt_onnx_model.mha2sha(**(config.mha2sha_kwargs if config.mha2sha_kwargs is not None else {}))
+
+        qairt_onnx_model.export(output_model_path, prefix=model.onnx_file_name.removesuffix(".onnx"))
```
jambayk: `model.onnx_file_name` can be None if it doesn't have external data. 

instead you can use 
```
from pathlib import Path 

qairt_onnx_model.export(output_model_path, prefix=Path(model.model_path).stem)
```

olive/olive_config.json
qti-kromero: I see - I had erroneously thought this was alpha order by module path

PR #2107, opened 2025-08-18
Add extra_options to ModelBuilder pass

xiaoyu-work:
## Describe your changes

Add extra_options to ModelBuilder pass

### DIFF
--- a/olive/cli/capture_onnx.py
+++ b/olive/cli/capture_onnx.py
@@ -149,7 +149,7 @@ class CaptureOnnxGraphCommand(BaseOliveCLICommand):
             ),
         )
         mb_group.add_argument(
-            "--extra_options",
+            "--extra_mb_options",
             type=str,
             required=False,
             help="Extra key-value pairs options to pass to the model builder. e.g., 'int4_is_symmetric=true,int4_op_types_to_quantize=MatMul/Gemm'.",
@@ -200,9 +200,12 @@ class CaptureOnnxGraphCommand(BaseOliveCLICommand):
                     (("passes", "m", "enable_cuda_graph"), self.args.enable_cuda_graph),
                 ]
             )
-            if self.args.extra_options:
+            if self.args.extra_mb_options:
                 to_replace.append(
-                    (("passes", "m", "extra_options"), self._parse_extra_options(self.args.extra_options.split(",")))
+                    (
+                        ("passes", "m", "extra_options"),
+                        BaseOliveCLICommand._parse_extra_options(self.args.extra_mb_options.split(",")),
+                    )
                 )
             if self.args.int4_block_size is not None:
                 to_replace.append((("passes", "m", "int4_block_size"), self.args.int4_block_size))
@@ -244,18 +247,6 @@ class CaptureOnnxGraphCommand(BaseOliveCLICommand):
         update_shared_cache_options(config, self.args)
 
         return config
-
-    @staticmethod
-    def _parse_extra_options(kv_items):
-        kv_pairs = {}
-
-        if kv_items:
-            for kv_str in kv_items:
-                kv = kv_str.split("=")
-                kv_pairs[kv[0].strip()] = kv[1].strip()
-
-        print(f"Extra options: {kv_pairs}")
-        return kv_pairs
 
 
 TEMPLATE = {
--- a/olive/cli/capture_onnx.py
+++ b/olive/cli/capture_onnx.py
@@ -148,6 +148,12 @@ class CaptureOnnxGraphCommand(BaseOliveCLICommand):
                 "for the CUDA graph to be used correctly."
             ),
         )
+        mb_group.add_argument(
+            "--extra_options",
+            type=str,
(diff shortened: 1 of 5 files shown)


olive/passes/onnx/model_builder.py
```diff
@@ -229,6 +234,10 @@ def _run_for_config(
             if model.adapter_path:
                 extra_args["adapter_path"] = model.adapter_path
 
+        # Add extra options support for model builder
+        if config.extra_options:
```
jambayk: if there is overlap between the extra_options and the same options exposed through the config, which one should take precedence?

olive/passes/onnx/model_builder.py
```diff
@@ -229,6 +234,10 @@ def _run_for_config(
...
+            extra_args.update(check_extra_options(config.extra_options))
+
```
jambayk: in line 245, I think we also need to ignore "extra_args"

olive/passes/onnx/model_builder.py
devang-ml: Good point. 
extra_options should be preferred IMO.

olive/cli/capture_onnx.py
```diff
@@ -148,6 +148,12 @@ def register_subcommand(parser: ArgumentParser):
                 "for the CUDA graph to be used correctly."
             ),
         )
+        mb_group.add_argument(
+            "--extra_options",
```
devang-ml: For cli let's use --extra-mb-options as argument name, in case other passes also use extra options in future.

devang-ml: Also, we should add this to `optimize` CLI as well.

devang-ml (dismissed):

olive/passes/onnx/model_builder.py
xiaoyu-work: Updated

olive/passes/onnx/model_builder.py
xiaoyu-work: Added

olive/cli/capture_onnx.py
xiaoyu-work: Added to `optimize` cli

test/passes/onnx/test_model_builder.py
```diff
@@ -16,7 +16,11 @@
 def test_model_builder(tmp_path, metadata_only):
     input_model = make_local_tiny_llama(tmp_path / "input_model", "onnx" if metadata_only else "hf")
 
-    p = create_pass_from_dict(ModelBuilder, {"precision": "fp32", "metadata_only": metadata_only}, disable_search=True)
+    p = create_pass_from_dict(
+        ModelBuilder,
+        {"precision": "fp32", "metadata_only": metadata_only, "extra_options": {"int4_is_symmetric": "true"}},
```
jambayk: since we are not parsing the extra options dict, the value cannot be string

jambayk: I think we should call the check_extra_options like before the recent commit. The cli would also just directly provide string as values.

xiaoyu-work: Makes sense. Will add our own check_extra_options. Thanks!

olive/passes/onnx/model_builder.py
```diff
@@ -318,3 +327,50 @@ def _run_for_config(
...
+        if "exclude_lm_head" in kv_pairs and "include_hidden_states" in kv_pairs:
+            # 'exclude_lm_head' is for when 'hidden_states' are outputted and 'logits' are not outputted
+            # 'include_hidden_states' is for when 'hidden_states' are outputted and 'logits' are outputted
+            raise ValueError(
+                "Both 'exclude_lm_head' and 'include_hidden_states' cannot be used together. Please use only one of them at once."
+            )
+
+        # NvTensorRtRtx EP requires Opset 21, so force use_qdq which controls it.
+        if self.accelerator_spec.execution_provider == ExecutionProvider.NvTensorRTRTXExecutionProvider:
+            kv_pairs["use_qdq"] = True
```
devang-ml: What is the relationship between Opset 21 and QDQ?

jambayk: I talked with Kunal about this before and he said this is not necessary anymore. Model builder uses opset 21 always. This was added by the Nvidia team before to force the opset to be 21 as a side effect.

xiaoyu-work: Yes i synced with Kunal also. I had a PR to remove this ep check so we can use this method directly https://github.com/microsoft/onnxruntime-genai/pull/1692

test/passes/onnx/test_model_builder.py
jambayk: the value should be bool here? the pass doesn't parse the values

xiaoyu-work: Good catch! Updated.

devang-ml (approved):

PR #2137, opened 2025-09-04
Quantization: lm_head quantization, k_quant_down heuristic, No zero offset while packing

jambayk:
## Describe your changes
- `Gptq` and olive quant checkpoint now supports lm_head quantization
- `SelectiveMixedPrecision` has a new heuristic called `k_quant_down` which is similar to `k_quant_mixed` but only takes the `down_proj` linears and `lm_head`.
- `QuantLinear` packing now doesn't do the gptq style -1 offset to the zero points while packing. This causes issues with asymmetric quantization when any zero point value is 0 as the values are expected to be unsigned int. It also now has a `from_tensor` method that can be used to create a new instance from already quantized weights.

### DIFF
--- a/olive/common/quant/linear.py
+++ b/olive/common/quant/linear.py
@@ -2,8 +2,9 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
+from __future__ import annotations
+
 import math
-from typing import Optional
 
 import torch
 import torch.nn as nn
@@ -30,7 +31,7 @@ class QuantLinear(nn.Module):
         symmetric: bool = True,
         group_size: int = -1,
         bias: bool = True,
-        device: Optional[torch.device] = None,
+        device: torch.device | None = None,
         dtype: torch.dtype = torch.float32,
     ):
         """Initialize QuantLinear layer.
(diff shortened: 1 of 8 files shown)


olive/evaluator/olive_evaluator.py
```diff
@@ -986,9 +986,12 @@ def evaluate(
         from lm_eval import simple_evaluate
         from lm_eval.api.registry import get_model
         from lm_eval.tasks import TaskManager
+        from lm_eval.utils import setup_logging
 
         import olive.evaluator.lmeval_ort  # noqa: F401 # pylint: disable=unused-import
 
+        setup_logging("ERROR")
```
devang-ml, line 993: ??

olive/common/quant/linear.py
```diff
@@ -181,7 +220,7 @@ def get_unpacked_params(self) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]
 
         """
         qweight = self._unpack_from_int32(self.qweight, (self.in_features, self.out_features), axis=0)
-        qzeros = self._unpack_from_int32(self.qzeros, self.scales.shape, axis=1) + 1
+        qzeros = self._unpack_from_int32(self.qzeros, self.scales.shape, axis=1)
```
devang-ml, line 223: Is this intentional?

jambayk: yes, this is to remove the -/+1 in the zero points that was done during packing/unpacking before. It was based on autogptq packing but leads to issues with asymmetric quantization when some values are 0 since it underflows.

olive/evaluator/olive_evaluator.py
jambayk: this is to suppress the verbose logging from lm_eval. Their log level setting requires a string value.

devang-ml (approved):

PR #2174, opened 2025-09-18
Fix inc quantization bugs

xiaoyu-work:
## Describe your changes
Fix inc quantization bugs


olive/common/utils.py
```diff
@@ -34,13 +34,13 @@ class IntEnumBase(IntEnum):
         pass
 
 else:
-    from enum import Enum
+    from enum import Enum, IntEnum
```
jambayk: if intenumbase is just the same as intenum, should we just remove intenumbase entirely?

xiaoyu-work: Makes sense. I'll update the PR.

PR #2178, opened 2025-09-23
Adding Estimate NPU Latency pass and unit test

alinah-amd:
### **Overview**
Added the EstimateNPULatency pass under `olive/onnx/vitis_ai/estimate_latency.py`
EstimateNPULatency makes use of the NPU Perf Estimator tool to predict computational performance of workloads given a set of parameters.

This is an **analysis pass** and does **not** transform the graph at all. Used for performance analysis only.

### **Installation**
To install (if not installed through requirements.txt), run the following:
`pip install [placeholder for wheel]`

Confirm python version installed is >= 3.10 for compatibility.

If perf estimator package is not installed, the following warning will show and the pass will simply be bypassed:
c5adac46-8376-49c1-a9c0-f78d12d2af9d

### **Usage**
**Inputs**
EstimateNPULatency takes in both the model in the form of an `OnnxModelHandler` object and a list of optional parameters in the form of a dict of `PassConfigParams` (consistent with all other passes).
 
**Optional Parameters**
To pass in optional parameters, list parameter name and parameter value as key-value pairs in the json file. See example:
 
a3d72e0e-0717-4f5d-a590-31333b6a81bb

1. target_device: Target device type. This is used to provide default config specific to that device type. Currently, only Strix is supported but will modify to include other devices in future.
- Type: `str`
- Default:  `stx`
- Allowed Values: `["stx"]`

**Adding Pass to Config File**
Should ideally be run as the last pass and listed last in the `<model>.json` file. For example:

4b0dce1b-a906-46c4-9528-8f94b4a0876e

**Output**
Generates a `concise_summary` directory within the run directory with the following files:

557c58ea-7593-429b-9abf-d73ab9d58069

`{model_name}_concise_summary.txt` will display the following info on roofline latency, total compute ops, and what conclusion can be drawn on performance bottleneck (whether it is DDR Bandwidth bound or Compute bound):

1b2d13c4-4b2a-4775-929b-f429c27e1757

`{model_name}_concise_summary.csv` will display the same info but specific to per op. Ops are listed in descending order of latency:

528e3263-39e9-4a71-8db8-545325ff6f05

### **Known Passing Tests**
**Resnet w/ Perf Estimator**
Refer to Olive Recipes Repo

**MobileNet w/ Perf Estimator**
Refer to Olive Recipes Repo

**Unit Test**
- Run `python -m pytest test/unit_test/passes/vitis_ai/test_estimate_latency.py`
- This unit test runs a dummy model with the estimate latency pass (calls it directly instead of end to end flow) and asserts that a concise_summary directory with dummy csv and txt files are generated


olive/passes/onnx/vitis_ai/estimate_npu_latency.py
```diff
@@ -0,0 +1,85 @@
...
+                input_model_path,
+                EstimatorSettings.target_device,
+            )
+
+            run_perf_estimate(EstimatorSettings)
+
+            logger.info("Finish running perf estimator pass")
+
+        # return the original model
+        output_model_path = Path(resolve_onnx_path(output_model_path, Path(model.model_path).name))
```
jambayk: since we are returning the input model, you can just return the `model` input directly. 

although I think this might be better as an evaluator than as a pass. An example of a special evaluator https://github.com/microsoft/Olive/blob/be1ed76d08357e5a55a522c0f59e4cbfaa3e7741/olive/evaluator/olive_evaluator.py#L966 and example use

dekasthiti: If I've understood the design intent of an [Olive evaluator](https://test-olive.readthedocs.io/en/latest/_modules/olive/evaluator/olive_evaluator.html#OliveEvaluator), it is to measure the accuracy / latency of a workload by running it on a target device and for Olive to use the metrics to make online decisions on how best to optimize the workload's execution on the target device. In other words, we need the evaluator to monitor and measure the run on the device. 

On the other hand, the purpose of the estimator in this PR is to estimate the roofline latency of a workload on a model of the device and /not to/ measure the latency by running on the device. As such, I am not sure the estimator can be implemented as an evaluator.

Some important differences between a latency evaluator & latency estimator are: 
1. A latency evaluator runs the workload on the device and measures the time taken for the run, while an estimator models the target device and computes the time the workload might take to run on this target device. 
2. A latency evaluator will do some warmup runs before running n iterations of the workload and return average, min, max, percentile metrics of the run. A latency estimator will project the roofline latency of the workload based its model of the target device.

The intended use of the estimator here is to provide a way for a user to experiment with different quantization techniques and to get a sense of where the latency might land. As such, can it qualify as an analysis pass that analyzes the workload and projects some stats?

jambayk: Yes, your points make sense to me. After I saw the detailed report the pass generated, i had a similar thought that it doesn't look like a typical Olive evaluation metric, but I forgot to update the comment here. thanks!

jambayk: bringing this comment back up since it got buried with my other comment about pass vs evaluator: since we are returning the input model, you can just return the model input directly.

olive/passes/onnx/vitis_ai/estimate_npu_latency.py
```diff
@@ -0,0 +1,85 @@
+#
```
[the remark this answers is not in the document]
alinah-amd: Done thank you

olive/passes/onnx/vitis_ai/estimate_npu_latency.py
alinah-amd: Fixed. thank you

olive/passes/onnx/vitis_ai/estimate_npu_latency.py
```diff
@@ -0,0 +1,74 @@
...
+    def _run_for_config(
+        self, model: ONNXModelHandler, config: BasePassConfig, output_model_path: str
+    ) -> ONNXModelHandler:
+        perf_installed = True
+        try:
+            from estimator.config import EstimatorSettings
+            from estimator.run import run_perf_estimate
+        except ImportError:
+            perf_installed = False
+            logger.warning("Estimator module not found. Skipping EstimateNPULatency pass.")
```
jambayk: I think instead of raising a warning, it might be better to fail with import a helpful import error which tells what package to install. since olive caches runs, to rerun the pass with the dependency installed, they would have to clean the cache or delete the cached run that skipped the estimation.

olive/olive_config.json
```diff
@@ -70,6 +70,14 @@
             "supported_quantization_encodings": [  ],
             "run_on_target": true
         },
+        "EstimateNPULatency": {
+            "module_path": "olive.passes.onnx.vitis_ai.estimate_npu_latency.EstimateNPULatency",
+            "supported_providers": [ "*" ],
+            "supported_accelerators": [ "*" ],
+            "supported_precisions": [ "*" ],
+            "supported_algorithms": [  ],
+            "supported_quantization_encodings": [  ]
```
jambayk: could you add a `"module_dependencies"` option like under the autoawqquantizer pass for the package required to run this estimation?

test/passes/onnx/test_estimate_npu_latency.py
```diff
@@ -0,0 +1,37 @@
...
+
+from olive.passes.olive_pass import create_pass_from_dict
+from olive.passes.onnx.vitis_ai.estimate_npu_latency import EstimateNPULatency
+from test.utils import get_onnx_model
+
+
+class TestEstimateNPULatency:
+    """Test cases for EstimateNPULatency pass."""
+
+    def test_estimate_latency_basic(self, tmp_path):
```
jambayk, line 18: please also add the required dependency to the requirements-test.txt under `test`

olive/passes/onnx/vitis_ai/estimate_npu_latency.py
alinah-amd: Can you clarify on "clean cache or delete cached run"? I do notice I have to delete cache everytime I need to rerun. Is there a flag for running with delete cache option?

test/passes/onnx/test_estimate_npu_latency.py
alinah-amd: done

olive/passes/onnx/vitis_ai/estimate_npu_latency.py
jambayk: there is a workflow option called "clean_cache" that cleans the full cache. My concern is that the user might have multiple passes and this pass. if the dependency of this pass is missing so the profiling is skipped, the cache will still have an instance for this run. so when the user installs the missing dependency, unless they clean the full cache or manually delete this pass run cache, they won't be able to rerun the pass. that's why i suggest that the pass raise an import error for missing dependency so that it's run doesn't get cached in such cases.

test/requirements-test.txt
```diff
@@ -29,6 +29,7 @@ optimum[openvino]>=1.17.0, <=1.24
 optuna
 pandas
 peft
+perf-estimator
```
jambayk, line 31: there is no package called perf-estimator on pypi

olive/passes/onnx/vitis_ai/estimate_npu_latency.py
alinah-amd: I see. The added exception message should address that then.

test/requirements-test.txt
alinah-amd: Yes, we are working to push the package into pypi. Will update once that is done.

olive/passes/onnx/vitis_ai/estimate_npu_latency.py
alinah-amd: Actually could you comment on how to do the "clean_cache" workflow option? I am looking for it under flags but do not see any

jambayk:
Closing for now since the PR has become stale

PR #2183, opened 2025-09-25
Implicitly add olive version to saved onnx model proto

shaahji:


shaahji:
> This is little bit similar as this pass https://github.com/microsoft/Olive/blob/main/olive/passes/onnx/add_metadata.py#L63

The implemented pass is doing a lot more work than what's need here. Inserting the Olive version needs to be implicit (i.e. user shouldn't have to do anything like adding a pass to the workflow). The only common code is probably 3 lines that queries the version.

xiaoyu-work (approved):

PR #2185, opened 2025-10-01
Update extract adapter pass to match torch dynamo export

xiaoyu-work:
## Describe your changes

We will only focus on cases when the onnx model is exported from torch dynamo. If the optimize is True (which is default) for torch dynamo export, the lora weights will be folded. So:

1. we check where the initializer comes from for each initializer.
2. If it is from a LORA weight, rename it back to lora weight, and save it.
3. DORA weights might be folded to GEMM, we decompose this node to MatMul so it can be quantized by Olive quantization pass.

### DIFF
--- a/test/passes/onnx/test_extract_adapters.py
+++ b/test/passes/onnx/test_extract_adapters.py
@@ -140,11 +140,6 @@ def input_model_info_torchscript_fixture(tmp_path_factory, request):
     rtn_quantizer = create_pass_from_dict(OnnxBlockWiseRtnQuantization, {}, disable_search=True)
     olive_int4_onnx_model = rtn_quantizer.run(olive_onnx_model, str(tmp_path / "int4-onnx"))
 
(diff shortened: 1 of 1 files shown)


olive/passes/onnx/common.py
```diff
@@ -447,61 +446,52 @@ def resave_model(
...
-    f".*{pattern}$"
-    for pattern in [
-        "default/Div",
-        "default/default/MatMul",
-        "default/default_1/MatMul",
-        "default/default/MatMul_Q4",
-        "default/default_1/MatMul_Q4",
-    ]
-]
+LORA_NAME_PATTERNS = ["lora_A", "lora_B"]
```
jambayk: is it okay to remove the `default_*` patterns? would this break for models exported using the old torch_script export method?

xiaoyu-work: This pass will be designed to support dynamo export model only.

PR #2194, opened 2025-10-07
Remove redundant log for run cli

xiaoyu-work:
## Describe your changes

Logs from WorkflowRunCommand is duplicated from Engine log.

### DIFF
--- a/olive/cli/run.py
+++ b/olive/cli/run.py
@@ -65,9 +65,5 @@ class WorkflowRunCommand(BaseOliveCLICommand):
 
         if self.args.list_required_packages is True:
             print("Required packages listed!")
-        elif workflow_output.has_output_model():
-            print(f"Model is saved at {self.args.output_path}")
-        else:
-            print("No output model produced. Please check the log for details.")
 
         return workflow_output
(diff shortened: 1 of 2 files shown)


olive/cli/run.py
```diff
@@ -65,9 +65,5 @@ def run(self):
 
         if self.args.list_required_packages is True:
             print("Required packages listed!")
-        elif workflow_output.has_output_model():
```
jambayk, line 68: is this redundant? the engine run log is always present even for other cli commands.
this is a print which will not be affected by the log levels

xiaoyu-work: Currently if anyone run Olive with config file, Olive will print 2 model saving log, 1st is [this](https://github.com/microsoft/Olive/blob/main/olive/engine/engine.py#L255) and 2nd is "Model is saved at None" which is from this line because config file run doesn't have `args.output_path`. Removing this won't affect other CLIs cause other CLIs have the same log somewhere else: https://github.com/microsoft/Olive/blob/main/olive/cli/base.py#L45-L48

PR #2225, opened 2025-10-20
Add built in feature extraction data preprocess

xiaoyu-work:
## Describe your changes

Current default huggingface pre process is designed for classification task. Add new feature extraction data preprocess so no label will be returned from dataloader.

### DIFF
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -56,7 +56,7 @@ class OliveEvaluator(ABC):
         super().__init__()
 
     @staticmethod
-    def _unpack_batch_for_accuracy(batch):
+    def unpack_batch_for_accuracy(batch):
         """Unpack batch for accuracy evaluation. Requires (data, label) format."""
         if not isinstance(batch, (tuple, list)) or len(batch) != 2:
             raise ValueError(
@@ -66,7 +66,7 @@ class OliveEvaluator(ABC):
         return batch[0], batch[1]
 
     @staticmethod
-    def _extract_input_data(batch):
+    def extract_input_data(batch):
         """Extract input data from batch, ignoring labels if present."""
         return batch[0] if isinstance(batch, (tuple, list)) and len(batch) == 2 else batch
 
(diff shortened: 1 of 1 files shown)


olive/data/component/dataset.py
```diff
@@ -39,8 +39,10 @@ def __len__(self):
 
     def __getitem__(self, index):
         data = {k: v for k, v in self.data[index].items() if k != self.label_col}
-        label = self.data[index][self.label_col]
-        return data, label
+        if self.label_col is not None:
+            label = self.data[index][self.label_col]
+            return data, label
+        return data
```
jambayk: the `__getitem__` for the Dataset object is expected to return a tuple of the input and the label. downstream users such as the latency evaluator and quantization calibrators also have the same assumption. So this should also return a dummy label like in line 79. I think you could return None too but I am not fully sure if the dataloader can accept none when batching the dataset

xiaoyu-work: I'll update the PR to create a separate dataset class for feature extraction case only.

jambayk: rather than a new dataset class, I think returning a dummy label would be better? most passes and the evaluator interface with the dataloader expecting a tuple of data and label per batch.

xiaoyu-work: I intestinally want to differentiate classification dataset and non classification dataset. Always returning a label causes a lot of issue and confusion for those non-classification pass and task.

jambayk: what about the dataloader used by the quantization passes and evaluator. can you test that using a feature extraction data container doesn't break them? for evaluator, maybe it's fine if the accuracy metric doesn't handle it correctly but the latency evaluator should.

evaluator:
accuracy: https://github.com/microsoft/Olive/blob/main/olive/evaluator/olive_evaluator.py#L418
latency: https://github.com/microsoft/Olive/blob/main/olive/evaluator/olive_evaluator.py#L474

static quantization seems fine since it handles no tuple return from dataloader:
https://github.com/microsoft/Olive/blob/f1544e86d73fe3ed4b93a99dd8611a4e70233e49/olive/data/component/dataloader.py#L91

jambayk (approved):

Issue #2223, opened 2025-10-19
[Bug] ONNX Quantization Broken Since v0.9.2 Due to Incorrect Weight Data Saving


happyme531:
**Describe the bug**
ONNX model quantization using algorithms like HQQ and RTN appears to be broken since version 0.9.2, and the issue persists on the `main` branch.

In version 0.9.1, the quantization process correctly produces a single, valid `model.onnx` file. However, starting from version 0.9.2, the process generates a model with incorrectly large external data (`model.onnx` + `model.onnx.data`).

Consequently, tools like `onnxslim` or even `onnx.load()` fail to load the model, throwing an `onnx.onnx_cpp2py_export.checker.ValidationError` because they cannot locate the tensor data. This regression makes the quantization feature unusable for large models that are saved with external data.

The issue has been observed with both `HQQ` and `RTN` quantization algorithms.

**To Reproduce**
1. Install `olive-ai` version 0.9.2 or build from the `main` branch.
2. Get a large ONNX model.
3. Run the `olive quantize` command:
   ```bash
   olive quantize -m ./path/to/your_model.onnx --algorithm hqq --precision int4 --output_path ./quantized_model.onnx
   ```
4. Attempt to load or inspect the generated model, which is typically located at `./quantized_model.onnx/model/<original_model_name>.onnx`.
   ```bash
   # For example, using onnxslim
   onnxslim --inspect ./quantized_model.onnx/model/<original_model_name>.onnx
   ```
5. Observe the `onnx.onnx_cpp2py_export.checker.ValidationError` as shown in the logs below.

**Expected behavior**
The quantization process should produce a valid, self-contained ONNX model. If external data is used, the paths within the `.onnx` file should be relative to the model file itself, allowing it to be correctly loaded by standard ONNX tools.

The behavior should be consistent with version 0.9.1, where a single, valid ONNX file was successfully generated.

**Olive config**
The configuration is provided via command-line arguments. No JSON config file was used.
```bash
olive quantize -m ./onnx/onnx_models/language_model.onnx --algorithm hqq --precision int4 --output_path ./onnx/onnx_models_quantized/language_model.onnx --log_level 1
```

**Olive logs**

<details>
<summary><b>✅ Working Log on Olive v0.9.1</b></summary>

```
> olive quantize -m ./onnx/onnx_models/language_model.onnx --algorithm hqq --precision int4 --output_path ./onnx/onnx_models_quantized/language_model.onnx --log_level 1
[...]

PR #2227, opened 2025-10-23
Load external ONNX tensor data before transforming models

happyme531:
Ensure every ONNX pass hydrates external tensor data after loading the IR model. This avoids writing broken references when saving quantized or converted models with external weights, restoring the behavior seen in v0.9.1.

## Describe your changes
- load external tensor data into the in-memory IR model for conversion, HQQ and RTN quantization
- hydrate external data before running ONNXScript fusions so the passes can safely rewrite the graph
- restores the ability to emit valid quantized models with external weights after the IR migration

## (Optional) Issue link
Fixes https://github.com/microsoft/Olive/issues/2223

### DIFF
--- a/olive/passes/onnx/conversion.py
+++ b/olive/passes/onnx/conversion.py
@@ -655,6 +655,7 @@ class OnnxOpVersionConversion(Pass):
     ) -> ONNXModelHandler:
         output_model_path = resolve_onnx_path(output_model_path)
         model_ir = model.load_ir_model()
+        ir.external_data.load_to_model(model_ir)
         version_converter.convert_version(model_ir, config.target_opset, fallback=True)
         return ir_model_to_olive_model(model_ir, output_model_path, config)
 
--- a/olive/passes/onnx/hqq_quantization.py
+++ b/olive/passes/onnx/hqq_quantization.py
@@ -72,6 +72,7 @@ class OnnxHqqQuantization(Pass):
             return model
         output_model_path = resolve_onnx_path(output_model_path, Path(model.model_path).name)
         ir_model = model.load_ir_model()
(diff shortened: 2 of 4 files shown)


happyme531:
> Thanks for the PR! Did you verify this change work for your case?

Yeah, but I haven't added a test yet.

xiaoyu-work:
> > Thanks for the PR! Did you verify this change work for your case?
> 
> Yeah, but I haven't added a test yet.

I tried to repro your error with `Qwen/Qwen2.5-0.5B-Instruct`:
```
olive capture-onnx-graph -m Qwen/Qwen2.5-0.5B-Instruct --o onnx_model
olive quantize -m onnx_model --algorithm hqq --precision int4 --o quantized_model
```

Then load with `model = onnx.load("model.onnx")` or run `onnxslim --inspect --model-check --verbose model.onnx`, but I didn't get any error.

Can you tell me more details about how you exported the onnx model? From the error 
```
onnx.onnx_cpp2py_export.checker.ValidationError: Data of TensorProto ( tensor name: /backbone/layers.0/self_attn/rotary_emb/Constant_attr::value) should be stored in PSALM/onnx/onnx_models_quantized/language_model.onnx/model/_backbone_layers.0_self_attn_rotary_emb_Constant_attr__value, but it doesn't exist or is not accessible.
```
looks like something wrong with the constant. Your model seems an edge case. There is one assumption, but I need to verify it. I have 2 solutions here: can you share your .onnx model file to me, so i can repro the issue? That will be easier to investigate the root cause. Or i have a potential fix here https://github.com/microsoft/Olive/tree/xiaoyu/ir. If you can you try to install Olive with this branch and verify if this change works, that is fine also.

PR #2235, opened 2025-10-30
Add custom onnx model name support for output dir

xiaoyu-work:
## Describe your changes
Add custom onnx model name support for output dir

### DIFF
--- a/olive/resource_path.py
+++ b/olive/resource_path.py
@@ -238,14 +238,34 @@ class LocalResourcePath(ResourcePath):
         else:
             new_path_name = self.config.path.name
         new_path = dir_path / new_path_name
-        _overwrite_helper(new_path, overwrite)
+
         # is the resource a file or a folder
         is_file = Path(self.config.path).is_file()
+
+        # Handle overwrite logic based on actual destination
+        if is_file or not flatten:
+            # Actual destination is new_path
+            _overwrite_helper(new_path, overwrite)
+        else:
+            # flatten=True and is folder: destination is dir_path
+            # Check for conflicts with individual files in source
+            if overwrite:
+                for item in Path(self.config.path).iterdir():
+                    target = dir_path / item.name
+                    if target.exists():
+                        if target.is_file():
+                            target.unlink()
+                        else:
+                            shutil.rmtree(target)
+
         # copy the resource to the new path
         if is_file:
             shutil.copy(self.config.path, new_path)
         else:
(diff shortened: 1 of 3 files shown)


olive/passes/onnx/common.py
```diff
@@ -121,10 +121,16 @@ def model_proto_to_file(
 
     :return: True if the model has external data, False otherwise.
     """
+    import shutil
+
     output_path = Path(output_path)
     if output_path.exists():
-        logger.debug("Deleting existing onnx file: %s", output_path)
-        output_path.unlink()
+        if output_path.is_dir():
```
jambayk: does this case ever happen? by this point, we assume the name has already been normalized to become a .onnx file name?

olive/resource_path.py
```diff
@@ -235,7 +237,7 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
             new_path_name = Path(name).with_suffix(self.config.path.suffix).name
         else:
             new_path_name = self.config.path.name
-        new_path = dir_path / new_path_name
+        new_path = dir_path if flatten else dir_path / new_path_name
```
jambayk: should we check flatten is only provided for the `not is_file` scenario? otherwise, when it's a file, we are asking to copy the file into a file with the same name as `dir_path`

olive/passes/onnx/common.py
xiaoyu-work: Good catch. That's probably true. I will revert this change.

olive/resource_path.py
xiaoyu-work: Ideally for current case there is no case when both are true, but i'll add a check here. Thanks for catching this!

olive/resource_path.py
```diff
@@ -237,14 +239,13 @@ def save_to_dir(self, dir_path: Union[Path, str], name: str = None, overwrite: b
             new_path_name = self.config.path.name
         new_path = dir_path / new_path_name
         _overwrite_helper(new_path, overwrite)
```
jambayk: i think the is_file check and flatten conditions needs to be handled before the `_overwrite_helper` call so that we clean up the destination directory?

but this is interesting. should we be cleaning the dir_path if we are flattening or we just call `copy_dir` with `dirs_exist_ok=True` instead

xiaoyu-work: i don't think we should clean up the dir_path if flatten. The dir_path may have engine artifacts.

jambayk (approved):

PR #2238, opened 2025-11-03
Quantization: Embeddings quantization, new packing format, Rtn quantizer

jambayk:
## Describe your changes
- New `QuantEmbedding` module added to do input embedding quantization
  - export to `GatherBlockQuantized` with torch script and dynamo mode supported. model builder doesn't support it since it requires change in the builder script
  - If a quantization pass is responsible for quantizing both the embeds and lm_head of a model with original tied weights, it can keep them tied in the pytorch model. During export, torch script duplicates the shared qweight because of reshape and dynamo keeps the reshape on the MatMulNBits (need to test if it's more efficient for the reshape to be on the GatherBlockQuantized)
- New `Rtn` pass that can be composed on top of other quantization passes. For example, gptq on the transformer layers and then rtn on the embedding and lm head to take advantage of weight tieing
- New quantized model checkpoint format. Moved to the same format used by the MatMulNBits and GatherBlockQuantized. Now there is no overhead for unpacking and repacking the weights during export, so it is very fast.
  - Enforce the same restrictions on block size and weight shapes required by the contrib ops for compatibility.
  - We also enforce that that the quantization dim is divisible by the block size. This makes the packing logic easier as we don't have to worry about paddings and gives compatibility between 3D qweight for Linears and 2D qweight for Embeddings.
- Updated the autogptq and autoawq checkpoint export to make the quantization parameters 2D as per the latest specs for the contrib operators.

### DIFF
--- a/olive/common/quant/hf_utils.py
+++ b/olive/common/quant/hf_utils.py
@@ -281,6 +281,7 @@ def _repoint_buffer(src: nn.Module, dst: nn.Module, name: str):
     )
 
     # tie the buffers
+    # pylint: disable=W0212
     dst._buffers[name] = src_buf
     dst._non_persistent_buffers_set.add(name)
 


xiaoyu-work:
We are trying to improve our test coverage. Can you please add unit test for new files created in this PR?

xiaoyu-work (approved):

PR #2248, opened 2025-11-10
Add CLI command to benchmark models using lm-eval

shaahji:

### DIFF
--- a/olive/cli/api.py
+++ b/olive/cli/api.py
@@ -6,6 +6,7 @@ import inspect
 from argparse import ArgumentParser, Namespace
 from typing import Any
 
+from olive.cli.benchmark import BenchmarkCommand
 from olive.cli.capture_onnx import CaptureOnnxGraphCommand
 from olive.cli.convert_adapters import ConvertAdaptersCommand
 from olive.cli.extract_adapters import ExtractAdaptersCommand
@@ -236,7 +237,7 @@ def tune_session_params(model_name_or_path: str, **kwargs) -> WorkflowOutput:
     return _run_unified_command(SessionParamsTuningCommand, **kwargs)
 
 
-def generate_cost_model(model_name_or_path: str, **kwargs) -> None:
+def generate_cost_model(model_name_or_path: str, **kwargs) -> WorkflowOutput:
     """Generate a cost model for model splitting (HuggingFace models only).
 
     Args:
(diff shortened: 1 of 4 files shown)


olive/cli/benchmark.py
```diff
@@ -0,0 +1,132 @@
+# -------------------------------------------------------------------------
```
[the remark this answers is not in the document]
shaahji: Resolved.

olive/cli/api.py
```diff
@@ -236,42 +237,66 @@ def tune_session_params(model_name_or_path: str, **kwargs) -> WorkflowOutput:
     return _run_unified_command(SessionParamsTuningCommand, **kwargs)
 
 
-def generate_cost_model(model_name_or_path: str, **kwargs) -> None:
+def generate_cost_model(model_name_or_path: str, **kwargs) -> WorkflowOutput:
```
xiaoyu-work, line 240: For these cli, there is no output model so we use None as python api return. Any reason that we need to return WorkflowOutput here?

olive/cli/benchmark.py
```diff
@@ -0,0 +1,132 @@
...
+    get_input_model_config,
+    update_shared_cache_options,
+)
+from olive.common.utils import set_nested_dict_value
+
+
+class BenchmarkCommand(BaseOliveCLICommand):
+    @staticmethod
+    def register_subcommand(parser: ArgumentParser):
+        sub_parser = parser.add_parser("benchmark", help="Evaluate the model using lm-eval.")
```
xiaoyu-work, line 23: Is this benchmark cli for lm-eval specifically? I mean, if in the future, we want to support other benchmark tools, but want to keep using this CLI. Then `lm-eval` should be an argument then. Like `--tools lm-eval`, But not a block for this PR for now as we can change this later.

olive/cli/api.py
shaahji: I ran into some issues in the recent past when adding these cli commands for the service. Just trying to be consistent with this change. Value of it can be dealt with on the caller side (whether a None is legal or not).

olive/cli/benchmark.py
shaahji: You are correct. But requiring additional argument now is unnecessary. We can introduce this if and when we have another evaluation tool.

xiaoyu-work (approved):

PR #2262, opened 2025-11-17
Add size_on_disk API to OliveModelHandler and corresponding metric

shaahji:


olive/model/handler/onnx.py
```diff
@@ -76,6 +76,12 @@ def model_path(self) -> str:
         model_path = super().model_path
         return get_onnx_file_path(model_path, self.onnx_file_name) if model_path else None
 
+    @property
+    def size_on_disk(self) -> int:
+        """Compute size of the model on disk."""
+        model = self.load_model()
```
xiaoyu-work, line 82: Do we need to count external data size?

shaahji: Yes! Logic needs to account for weights as well.

PR #2278, opened 2025-12-08
Remove gidx input from MatMulNBits graph surgery

rM-planet:
## Describe your changes
Adding a graph surgery which will remove group index input from the MatMulNBit nodes, only if the group indexes are sorted. 

##Motivation: 
MLAS matmulnbits kernel expects g_idx to not be passed if the node was quantized with default column-wise grouping for block-wise quantization. If g_idx is passed, it runs the unpacked compute kernel i.e dequantizes everything to fp32 and triggers a floating point matmul which significantly degrades the runtime performance.

##Impact:
Significant performance improvement for phi4 14b model without accuracy drop.

### DIFF
--- a/olive/passes/onnx/graph_surgeries.py
+++ b/olive/passes/onnx/graph_surgeries.py
@@ -1427,6 +1427,39 @@ class ReplaceAttentionMaskValue(ProtoSurgeon):
         return None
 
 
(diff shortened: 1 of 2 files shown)


devang-ml:
Could you please add a unit test? Thanks!

gtonpe:
When are the pending Olive CI tests expected to complete?

PR #2279, opened 2025-12-08
Add diffusers model handler

xiaoyu-work:
## Describe your changes

Add diffusers model handler. It is used to load diffusers model for diffusers pass (sd lora etc.)

### DIFF
--- a/olive/model/handler/diffusers.py
+++ b/olive/model/handler/diffusers.py
@@ -79,6 +79,35 @@ class DiffusersModelHandler(OliveModelHandler):
         return self.get_resource("adapter_path")
 
     @property
+    def size_on_disk(self) -> int:
+        """Compute size of the model on disk.
+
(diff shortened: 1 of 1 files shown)


olive/model/handler/diffusers.py
```diff
@@ -0,0 +1,222 @@
...
+        # Check for Flux
+        if "flux" in model_path_lower:
+            return DiffusersModelType.FLUX
+
+        # Try to detect from model config
+        try:
+            from diffusers import FluxTransformer2DModel
+
+            FluxTransformer2DModel.load_config(self.model_path, subfolder="transformer")
+            return DiffusersModelType.FLUX
```
shaahji, line 129: Don't we want to update `self.model_type` also here to avoid repeated detection?

xiaoyu-work: Makes sense. Upadated!

PR #2313, opened 2026-01-23
Move WORKDIR to storage disk in GPU CI & Use dynamo export for tests

xiaoyu-work:
## Describe your changes

1. Remove optimum-intel version pin from tests, so we can test the latest transformers.
2. Use dynamo export by default in our tests, since torchscript is no longer been maintained.
3. Fix a couple of tests for latest transformers and dynamo export change. Skip a couple of tests that need further investigations.
4. Move GPU agent work dir and docker work dir to storage disk, otherwise the main disk will be out of space.
5. Add `register_dynamic_cache_export_support` to fix dynamic cache registration issue.
6. Remove default 13 target_opset in tests.

### DIFF
--- a/olive/data/template.py
+++ b/olive/data/template.py
@@ -39,14 +39,14 @@ def huggingface_data_config_template(model_name, task, **kwargs) -> DataConfig:
     **kwargs: dict
         Additional arguments:
         - olive.data.component.load_dataset_config.huggingface_dataset
-            - `data_name`: str, data name in huggingface dataset, e.g.: "glue", "squad"
+            - `data_name`: str, data name in huggingface dataset, e.g.: "nyu-mll/glue", "squad"
             - `subset`: str, subset of data, e.g.: "train", "validation", "test"
             - `split`: str, split of data, e.g.: "train", "validation", "test"
             - `data_files`: str | list | dict, path to source data file(s).
             e.g.
                 load_dataset_config={
                     "params": {
-                        "data_name": "glue",
+                        "data_name": "nyu-mll/glue",
                         "subset": "train",
                         "split": "train",
                         "data_files": "whatever.pt"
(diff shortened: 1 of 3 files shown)


.azure_pipelines/job_templates/olive-test-linux-gpu-template.yaml
```diff
@@ -24,6 +24,30 @@ jobs:
...
+  - script: |
+      set -euxo pipefail
+
+      # Move agent work directory to /mnt/storage via symlink
+      AGENT_ROOT=$(dirname "$(Agent.BuildDirectory)")
+      sudo mkdir -p /mnt/storage/vss_work
+      sudo chown -R $USER:$USER /mnt/storage/vss_work
+      sudo cp -a "$AGENT_ROOT"/* /mnt/storage/vss_work/ 2>/dev/null || true
+      sudo rm -rf "$AGENT_ROOT"
+      sudo ln -sf /mnt/storage/vss_work "$AGENT_ROOT"
```
shaahji, line 36: Do we need to move the agent as well? This isn't something we maintain. It's part of the default VM image so we might have other issues because we are changing it to be a symlink now (if someone isn't following symlink, by default, it might fail silently).

.azure_pipelines/job_templates/olive-test-linux-gpu-template.yaml
shaahji, line 1: We should probably set environment variables like HF cache, and pip cache also to move to the storage.

.azure_pipelines/job_templates/olive-test-linux-gpu-template.yaml
xiaoyu-work: I tried to move docker downloads and artifacts to storage disk only but still got out of space. so I thought it might need to move the agent stuff as well. but you are right we don't know what configuration it has

shaahji: Not for this PR itself but we should get an understanding of what's already on the disk. May be printing entire directory structure to the file with file size will help get a better understanding and hopefully better decision making rather than blindly moving stuff around (especially when we can't justify the choice).

May be it's a task for February first week. :)

PR #2321, opened 2026-02-03
Add OnnxModel to list of supported model types for benchmarking

shaahji:
## Add OnnxModel support for benchmarking

Add OnnxModel to list of supported model types for benchmarking

### DIFF
--- a/olive/cli/benchmark.py
+++ b/olive/cli/benchmark.py
@@ -81,7 +81,8 @@ class BenchmarkCommand(BaseOliveCLICommand):
         assert input_model_config["type"].lower() in {
             "hfmodel",
             "pytorchmodel",
-        }, "Only HfModel and PyTorchModel are supported in benchmark command."
+            "onnxmodel",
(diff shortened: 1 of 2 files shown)


olive/cli/benchmark.py
```diff
@@ -81,6 +81,7 @@ def _get_run_config(self, tempdir: str) -> dict:
         assert input_model_config["type"].lower() in {
             "hfmodel",
             "pytorchmodel",
+            "onnxmodel",
         }, "Only HfModel and PyTorchModel are supported in benchmark command."
```
devang-ml: Update the msg.
Is there a unit test for this cli?
May be something like this - https://github.com/microsoft/Olive/blob/8e3a769a395d7d239631cdb067d3b1b64244ac90/test/cli/test_cli.py#L496

shaahji: Fixed comment and added a new test to explicitly verify support for onnx model.

PR #2324, opened 2026-02-06
Quantization: Add AutoClip Pass

jambayk:
## Describe your changes
- Add `AutoClip` pass which conducts an automatic search for optimal clipping values for linear layers that will be quantized later.
- Based on [`BitDistiller`](https://arxiv.org/abs/2402.10631) paper.
- Model quality metrics will be provided in a follow up paper which adds the self-distillation part of the paper.
- Gptq pass and quant_utils refactored to share common code. These will also be useful later when adding AWQ pass which is similar to autoclip but does symmetric clipping and also optimizes scales. 

### DIFF
--- a/olive/olive_config.json
+++ b/olive/olive_config.json
@@ -28,6 +28,15 @@
             "dataset": "dataset_optional",
             "module_dependencies": [ "autoawq" ]
         },
+        "AutoClip": {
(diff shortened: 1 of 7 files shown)


xiaoyu-work (approved):

xiaoyu-work:
Can you also update this the pass doc with this new pass?

PR #2325, opened 2026-02-08
Support weight sharing in QNN GPU

vjatoth-qti:


olive/passes/onnx/static_llm.py
jambayk, line 1: @vjatoth-qti the test pipeline failures are unrelated to the PR but please fix the linting errors. thanks!

olive/passes/onnx/static_llm.py
```diff
@@ -56,6 +57,14 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> dict[str, PassCon
...
                 description="Input length of the context model.",
             ),
+            "context_lengths": PassConfigParam(
+                type_=list[int],
+                default_value=None,
+                description=(
+                    "List of context lengths to generate static models QNN_GPU."
+                    "If None or empty, falls back to single 'context_length'."
+                ),
+            ),
```
qti-mattsinc, line 79: @vjatoth-qti, would the intended use of this in a recipe look like this?

```json
        "st": {
            "type": "StaticLLM",
            "batch_size": 1,
            "context_lengths": [1, 64]
        }
```

Do we ever expect len(context_lengths) > 2? If not, it seems like we could follow the NPU strategy for StaticLLM, where `"context_length": x` always implies a hybrid AR1 + ARx model, and we wouldn't need a new "context_lengths" key.

olive/passes/onnx/context_binary.py
```diff
@@ -243,6 +243,8 @@ def _generate_context_binary(
         if execution_provider == ExecutionProvider.QNNExecutionProvider:
             if str(device).lower() == "gpu":
                 provider_options["backend_path"] = "libQnnGpu.so" if platform.system() == "Linux" else "QnnGpu.dll"
+                if share_ep_contexts:
+                    provider_options["enable_gpu_weight_sharing"] = "1"
```
qti-mattsinc: Is `enable_gpu_weight_sharing` a supported provider option? The corresponding PR to onnxruntime-qnn (https://github.com/onnxruntime/onnxruntime-qnn/pull/67) does not add it as one. #Resolved

olive/passes/onnx/static_llm.py
unnim-qti: resolved

olive/passes/onnx/static_llm.py
```diff
@@ -182,57 +188,139 @@ def process_context_iterator(component_models, llm_pipeline, output_dir):
...
+            new_model_attributes = deepcopy(model.model_attributes) or {}
+            handler = ONNXModelHandler(
+                model_path=output_model_dir,
+                onnx_file_name=output_model_file.name,
+                model_attributes=new_model_attributes,
+            )
+
+            # Store handler + a logical component name (e.g., ctx_128)
+            generated_handlers[ctx_len] = handler
+            generated_names[ctx_len] = f"ctx_{ctx_len}"
```
qti-mattsinc: This naming scheme won't be compatible with the context binary generator pass, which expects "context" and "iterator" as the decoder pipeline components.

https://github.com/microsoft/Olive/blob/1db66b8558172efc22c7375ac797b9efa21577ab/olive/passes/onnx/context_binary.py#L142-L164

I'd suggest we stick with the existing naming scheme used in the flow for NPU. Otherwise, we should at least update the context binary pass to work with these alternative component names.

olive/passes/onnx/static_llm.py
```diff
@@ -56,6 +57,11 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> dict[str, PassCon
                 default_value=64,
                 description="Input length of the context model.",
             ),
+            "weight_sharing": PassConfigParam(
+                type_=bool,
+                default_value=False,
```
qti-mattsinc: Why default to false? I'd imagine a weight sharing export is almost always the desired scenario. Is there some backwards-compatibility concern with defaulting to true?

olive/passes/onnx/static_llm.py
unnim-qti: resolved

olive/passes/onnx/static_llm.py
unnim-qti: resolved

olive/passes/onnx/common.py
```diff
@@ -792,40 +793,46 @@ def update_llm_pipeline_genai_config(
...
         return model
 
     with open(genai_config_path) as f:
         genai_config = json.load(f)
-
     # update model_type
     genai_config["model"]["type"] = "decoder-pipeline"
 
-    # Update the provider_options list
     provider_option = {"qnn": {"backend_type": "gpu"}}
```
qti-mattsinc: Although it's not necessarily related to this PR, we are going to update the provider options soon to include the new "enable_dx12_shared_memory_allocator" flag for KV$ zero-copy. @unnim-qti / @vjatoth-qti, do you think we should fold it into this PR or add it in a follow up?

```suggestion
    provider_option = {"qnn": {"backend_type": "gpu", "enable_dx12_shared_memory_allocator": "1"}}
```

vjatoth-qti: Yes, would prefer adding it to this PR and update the PR description accordingly.

olive/cache.py
```diff
@@ -541,9 +541,11 @@ def _save_model(
                 output_file = output_dir
                 actual_output_dir = output_dir.parent
             else:
-                # Otherwise, create model.onnx in the directory
+                # Otherwise, create model.onnx in the directory.
+                # Preserve the source onnx_file_name stem (e.g. model_ctx) so the output
+                # filename matches what genai_config.json references.
                 actual_output_dir = output_dir
-                model_file_name = "model"
+                model_file_name = Path(onnx_file_name).stem if has_additional_files and onnx_file_name else "model"
```
qti-mattsinc, line 603: Have we tried using `resave_model` in `update_llm_pipeline_genai_config_gpu` analogously to how it's used in the NPU path, instead of changing this file?

unnim-qti: Hi Matthew,

Based on my understanding, the resave function stores files in the destination cache. In the QNN GPU workflow, when the context_iterator_models split is disabled, a single ONNX model is generated instead of a composite model, unlike the HTP workflow.

In this scenario, the _save_model function in cache.py saves the ONNX file to the destination output directory with the hardcoded name "model.onnx". However, this is inconsistent with the naming convention used in genai_config, where the model is named **"model_ctx"** after the ctx_gen pass.

To keep the naming consistent with genai_config, I updated cache.py so that the file is saved as **"model_ctx"** instead of model. This aligns the output produced with the expected GenAI configuration naming convention.

qti-mattsinc (approved):

unnim-qti (approved):

vjatoth-qti:
@jambayk , can you have a look at this change for review/merge? Thanks.

unnim-qti (approved):

jambayk (approved):

PR #2348, opened 2026-03-05
Add SimplifiedLayerNormToRMSNorm surgery

unnim-qti:


olive/passes/onnx/graph_surgeries.py
```diff
@@ -850,6 +850,219 @@ def get_rmsnorm_nodes(pow_node: str, dag: OnnxDAG) -> list[str] | None:
         return rmsnorm_nodes if len(rmsnorm_nodes) >= (len(pattern) - 1) else []
 
 
+class SimplifiedLayerNormToRMSNorm(ProtoSurgeon):
```
jambayk, line 853: could you please add a corresponding test like in https://github.com/microsoft/Olive/blob/main/test/passes/onnx/test_graph_surgeries.py#L523? Thanks!

justinchuby:
Why not make this a rewrite rule, out of curiosity?

unnim-qti:
> Why not make this a rewrite rule, out of curiosity?

I considered a rewrite rule, but this transform isn’t purely structural. 
- It has opset‑dependent ReduceMean semantics, handles two operator variants and requires conditional rewiring of an optional second output with ValueInfo preservation, which is cumbersome to express in a pattern rewrite.
-  A ProtoSurgeon was therefore a cleaner and safer fit.

justinchuby:
Handling value infos and optional output handling should be straightforward with the onnxscript rewriter. I wonder if you saw any blockers?

unnim-qti:
I didn’t  see any hard blockers in the rewriter, but I followed the existing pattern SimplifiedLayerNormToL2Norm, which use ProtoSurgeon to handle variant‑specific wiring, and optional outputs explicitly.  Given that precedent and the comparable complexity here, I kept this transform in the ProtoSurgeon style for the consistency.

unnim-qti:
> Handling value infos and optional output handling should be straightforward with the onnxscript rewriter. I wonder if you saw any blockers?

can you review the changes and initiate the pending checks.

olive/passes/onnx/graph_surgeries.py
```diff
@@ -850,6 +850,221 @@ def get_rmsnorm_nodes(pow_node: str, dag: OnnxDAG) -> list[str] | None:
...
+            dag.add_node(mean_node, graph_idx)
+
+            # ---------------------------
+            # Step 3: Add epsilon
+            # ---------------------------
+            eps_value = 1e-06
+            add_eps_name = self.create_new_name(node_name, op_type, "AddEps")
+            add_eps_out = f"{add_eps_name}_out"
+
+            eps_const = numpy_helper.from_array(np.array([eps_value], dtype=np.float32), name=f"{add_eps_name}_const")
```
qti-mattsinc: @unnim-qti, this causes an error with fp16 activations, since `eps_const` is always fp32 in this surgery:

```
Type Error: Type parameter (T) of Optype (Add) bound to different types (tensor(float16) and tensor(float) in node (/model/layers.0/input_layernorm/LayerNorm_AddEps).
```

We should be able to get the datatype for epsilon from the type of `ln_input`.

Similar situation with `pow_const`, although Pow is allowed to have inputs of different types, so it's not a hard error unlike for Add.

unnim-qti: @qti-mattsinc , Thanks for pointing this out — I’ve updated the surgery to assign epsilon (and pow_const) to match the ln_input datatype, which resolves the FP16 activation type mismatch in the Add node.

unnim-qti:
@jambayk , Could you please re‑initiate the rebuild? The docs build failed due to a 404 Client Error.

jambayk (approved):

PR #2406, opened 2026-04-09
feat: add MobiusModelBuilder Olive pass for mobius-backed ONNX export

justinchuby:
## Summary

Adds a new Olive pass (`MobiusModelBuilder`) that wraps mobius `build()` to produce ONNX models from HuggingFace model IDs.

- Single-component models (LLMs) → `ONNXModelHandler`
- Multi-component models (VLMs, encoder-decoders) → `CompositeModelHandler`
- EP auto-detected from Olive accelerator spec (cpu/cuda/dml/webgpu)
- Precision: fp32 (default), fp16, bf16
- Registered in `olive_config.json` as `MobiusModelBuilder`
- Example pipeline config: `examples/gemma4/gemma4_int4_cuda.json`
- 14 unit tests covering single/multi-component, EP detection, and error cases

## Validated: Gemma4 INT4 Quantization Pipeline

Successfully tested `MobiusModelBuilder` → `OnnxBlockWiseRtnQuantization` on `google/gemma-4-E2B-it`:

### Quantized ops per component

| Component | Total nodes | MatMulNBits | GatherBlockQuantized | Other ops |
|-----------|-------------|-------------|---------------------|-----------|
| **decoder** | 1,277 | 277 | 39 | 961 |
| **audio** | 1,465 | 135 | 0 | 1,330 |
| **vision** | 1,488 | 114 | 66 | 1,308 |
| **embedding** | 24 | 0 | 1 | 23 |

### Weight quantization coverage

| Component | Quantized (UINT8/INT4) | Non-quantized (FP16) | % quantized by size |
|-----------|----------------------|---------------------|-------------------|
| **decoder** | 316 tensors (2.4G elements) | 585 tensors (71M elements) | 97% |
| **audio** | 135 tensors (154M elements) | 768 tensors (2.8M elements) | 98% |
| **embedding** | 1 tensor (201M elements) | 4 tensors (3.1M elements) | 98% |
| **vision** | 147 tensors (90M elements) | 698 tensors (1.7M elements) | 98% |

### Output structure (2.8GB total, down from ~5GB fp16)

```
models/gemma4-e2b-int4-cuda/
├── decoder.onnx      (853K) + decoder.onnx.data   (2.4G)
├── audio.onnx        (1.2M) + audio.onnx.data     (152M)
├── embedding.onnx    (8.5K) + embedding.onnx.data  (199M)
├── vision.onnx       (1.2M) + vision.onnx.data     (89M)
└── model_config.json
```

### Pipeline timing

| Pass | Time |
|------|------|
| `MobiusModelBuilder` (fp16 build) | 77s |
| `OnnxBlockWiseRtnQuantization` (int4) | 129s |
| **Total** | **~3.5 min** |


examples/gemma4/gemma4_fp32_cpu.json
jambayk, line 1: we don't keep examples in this repo anymore. can you create an accompanying PR in [microsoft/olive-recipes](https://github.com/microsoft/olive-recipes/tree/main)?

olive/passes/onnx/mobius_model_builder.py
```diff
@@ -0,0 +1,196 @@
...
+                type_=str,
+                required=False,
+                default_value=None,
+                description=(
+                    "Override the mobius execution provider (cpu, cuda, dml, webgpu). "
+                    "When None (default), the EP is auto-detected from the Olive "
+                    "accelerator spec."
+                ),
+            ),
+            "trust_remote_code": PassConfigParam(
```
jambayk: the trust_remote_code value can be access from the hfmodelhandler directly using `model.get_load_kwargs().get("trust_remote_code", None)`

examples/gemma4/gemma4_fp32_cpu.json
justinchuby: I am just using this PR to iterate the files. Could you comment on whether there are errors or changes needed? I will move the files over once stable.

olive/passes/onnx/mobius_model_builder.py
```diff
@@ -0,0 +1,196 @@
...
+                required=False,
+                default_value=Precision.FP32,
+                description=(
+                    "Model weight / compute precision. One of: fp32, fp16, bf16. "
+                    "Defaults to fp32. For INT4 quantization, run an Olive "
+                    "quantization pass (e.g. OnnxMatMulNBits) after this pass."
+                ),
+            ),
+            "execution_provider": PassConfigParam(
+                type_=str,
```
jambayk: we could create an enum of the supported eps for automatic validation like in https://github.com/microsoft/Olive/blob/8b1957e3cfc8bda51fa2d32106736b6ac3fa5d17/olive/passes/pytorch/autoawq.py#L27. 
unless you think the options might keep growing and it would be hard to keep it in sync across versions

olive/passes/onnx/mobius_model_builder.py
```diff
@@ -0,0 +1,196 @@
...
+            if not onnx_path.exists():
+                raise RuntimeError(
+                    f"MobiusModelBuilder: expected output file not found: {onnx_path}. "
+                    "mobius.build() may have failed silently or saved to an unexpected path."
+                )
+            return ONNXModelHandler(
+                model_path=str(output_dir),
+                onnx_file_name="model.onnx",
+                model_attributes={
+                    "mobius_package_keys": package_keys,
```
jambayk, line 197: the model attributes require "additional_files" for the files other than the onnx file and external data like https://github.com/microsoft/Olive/blob/8b1957e3cfc8bda51fa2d32106736b6ac3fa5d17/olive/passes/onnx/model_builder.py#L332 
this is required so that we can carry over these files when a new pass is run on the output model

olive/passes/onnx/mobius_model_builder.py
justinchuby: @copilot please fix

olive/passes/onnx/mobius_model_builder.py
justinchuby: @copilot please fix

examples/gemma4/gemma4_fp32_cpu.json
jambayk: some comments:
- have never tried adding an unused field like "comment" in the config. don't remember if the config accepts unknown field.
- engine level nesting is optional. most examples we have now don't use it for cleaner configs
- cpu ep is the default the system doesn't need to be added explicitly

examples/gemma4/gemma4_int4_cuda.json
```diff
@@ -0,0 +1,15 @@
...
+    "input_model": { "type": "HfModel", "model_path": "google/gemma-4-E2B-it", "task": "text-generation" },
+    "systems": {
+        "local_system": {
+            "type": "LocalSystem",
+            "accelerators": [ { "device": "gpu", "execution_providers": [ "CUDAExecutionProvider" ] } ]
+        }
+    },
+    "passes": {
+        "mobius_build": { "type": "MobiusModelBuilder", "precision": "fp16" },
+        "int4_quantize": { "type": "GptqQuantizer", "bits": 4, "group_size": 128, "sym": true }
```
jambayk: the quantization pass works in a pytorch model and should be run before mobius

justinchuby: Is there a pass I can use to quantize after the model? I would like to use that as a an example for now

jambayk: you can use the rtn pass

```json
{ 
      "type": "rtn", 
      "bits": 4, 
      "sym": false, 
      "group_size": 32,
      "embeds": true,
      "lm_head": true
  }
```

olive/cli/optimize.py
```diff
@@ -3,7 +3,6 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 
-# ruff: noqa: T201
```
jambayk: nit: maybe we could do this removal in a different PR? these make the PR seem bigger than it actually is.

examples/gemma4/gemma4_int4_cuda.json
jambayk: oh, i misread the comment. you can use the blockwise quantizer pass:

```json
 {
      "type": "OnnxBlockWiseRtnQuantization",
      "block_size": 128,
      "is_symmetric": true,
      "accuracy_level": 4,
      "save_as_external_data": true
  }
  ```

Comment (justinchuby) [reply examples/gemma4/gemma4_int4_cuda.json]: Thanks. And it will process all components together?

File: olive/cli/optimize.py (hunk 5 above)
Comment (justinchuby) [reply olive/cli/optimize.py]: sorry I ran lintrunner in the wrong repo 😅 Looks like Olive's ruff version can be updated

File: examples/gemma4/gemma4_int4_cuda.json (hunk 4 above)
Comment (jambayk) [reply examples/gemma4/gemma4_int4_cuda.json]: if the model is composite it should run the quantizer on each component and return a new composite model.

File: olive/passes/onnx/mobius_model_builder.py (hunk 2 above)
Comment (justinchuby) [reply olive/passes/onnx/mobius_model_builder.py]: Done.

**Comment by xiaoyu-work** [comment]
will mobius generate genai_config.json and related files for ort genai? Also, does mobius support customized naming for different component for multi components model? I can see all component models are named as "model.onnx"

**Comment by justinchuby** [comment]
@xiaoyu-work The models are `<component_name>/model.onnx`. Is it good enough or are there cases when we want a flat layout / renamed models?

File: olive/passes/onnx/mobius_model_builder.py [hunk 6]
```diff
@@ -0,0 +1,250 @@
...
+    models that produce ``model``, ``vision``, and ``embedding`` sub-graphs) the
+    pass returns a :class:`~olive.model.handler.composite.CompositeModelHandler`
+    whose components are individual :class:`~olive.model.ONNXModelHandler` objects.
+    Single-component models return a plain :class:`~olive.model.ONNXModelHandler`.
+
+    Requires ``mobius-ai`` to be installed::
+
+        pip install mobius-ai
+
+    See https://github.com/microsoft/mobius
```
devang-ml: Incorrect URL

olive/passes/onnx/mobius_model_builder.py
```diff
@@ -0,0 +1,250 @@
...
+                required=False,
+                default_value=Precision.FP32,
+                description=(
+                    "Model weight / compute precision. One of: fp32, fp16, bf16. "
+                    "Defaults to fp32. For INT4 quantization, run an Olive "
+                    "quantization pass (e.g. OnnxMatMulNBits) after this pass."
+                ),
+            ),
+            "execution_provider": PassConfigParam(
+                type_=MobiusModelBuilder.MobiusEP,
```
devang-ml: We should not have pass level execution provider user choices.
The Olive engine is running a series of passes and user selects the EP for a given engine run.

devang-ml: See another relevant comment below `_run_for_config`

olive/passes/onnx/mobius_model_builder.py
```diff
@@ -0,0 +1,250 @@
...
+            raise ImportError(
+                "mobius-ai is required to run MobiusModelBuilder. Install with: pip install mobius-ai"
+            ) from exc
+
+        if not isinstance(model, HfModelHandler):
+            raise ValueError(f"MobiusModelBuilder requires an HfModelHandler input, got {type(model).__name__}.")
+
+        # Resolve EP: explicit config override > accelerator spec > fallback to cpu.
+        ep_str: str = config.execution_provider or self.EP_MAP.get(self.accelerator_spec.execution_provider, "cpu")
+
```
devang-ml, line 138: Olive Engine expects a pass to raise an error if accelerator spec provided EP is not supported by the pass. Alternatively return on modified input model in cases where input model type is same as output model type.

justinchuby:
@copilot Update https://github.com/microsoft/olive-recipes/pull/381 according to the changes

olive/passes/onnx/mobius_model_builder.py
justinchuby: Fixed

titaiwangms (commented):
Nice work! A few suggestions from our experience building the Ministral-3 VLM recipe on top of mobius:

**1. Per-component VLM quantization guidance**
For VLMs, downstream quantization needs differ per component: decoder → INT4, vision → INT8, embedding → none (ORT has a Gather rank-1 bug with GatherBlockQuantized that prevents embedding quantization). It would be helpful to document how recipe authors should wire per-component quantization after `CompositeModelHandler`. Even a brief note in the docstring would save people time.

**2. context_length default may be too small**
`write_ort_genai_config()` defaults context_length to 4096, but models like Ministral-3 need 262144 for full YaRN RoPE cache. Consider reading `max_position_embeddings` from the HF config automatically, or exposing it as a pass config parameter.

**3. Minimum mobius version check (minor)**
If a user has an older mobius without support for newer model types (e.g., Pixtral), they would get confusing errors from `mobius.build()`. A version check or clearer error message would help.

**4. Migration example (nice-to-have)**
Existing recipes manually call `mobius.build()` + `write_ort_genai_config()`. A brief before/after example showing how to replace that with `MobiusModelBuilder` pass would help adoption.

Overall this is clean and well-tested — excited to migrate our recipe to use this once it lands!

justinchuby:
@xiaoyu-work how should we handle per component quantization with Olive?

justinchuby:
@xiaoyu-work Thanks for the question! 

**genai_config.json generation**: Yes, we've just added support for this. The pass now calls \`_write_genai_config()\` which generates all ORT GenAI artifacts (genai_config.json, tokenizer files, processor configs) and includes them in the model handler's \`additional_files\`. This happens automatically when \`runtime='ort-genai'\` (the default).

**Multi-component naming**: Correct, mobius uses the flat \`<component_name>/model.onnx\` layout (e.g., \`decoder/model.onnx\`, \`vision_encoder/model.onnx\`). This is what ORT GenAI expects. The pass returns a \`CompositeModelHandler\` with each component's directory as the model path, so the structure is preserved.

justinchuby:
@titaiwangms Great feedback! Addressing each point:

**1. Per-component VLM quantization guidance**
This is important. The \`CompositeModelHandler\` structure makes this straightforward — each component is a standalone \`ONNXModelHandler\` with its own model path, so recipe authors can wire per-component quantization passes. We'll add a docstring note with a brief example (e.g., decoder → INT4, vision → INT8, embedding → skip).

**2. context_length / max_position_embeddings**
Good catch! We've verified that \`write_ort_genai_config()\` already handles this automatically — it reads \`max_position_embeddings\` from the HF config and takes \`max(context_length, max_position_embeddings)\`. So Ministral-3 and similar models with large rope should already work without manual intervention. No config parameter needed!

**3. Minimum mobius version check**
Valid concern. For now, users hitting old mobius versions will see mobius's own errors. We can add a version check in a follow-up if needed.

**4. Migration example**
Will add a docstring example showing the before/after. Quick version:

\`\`\`
# Before: manual mobius.build()
from mobius import build
pkg = build('model-id', load_weights=True)
pkg.save('output')
write_ort_genai_config(pkg, 'output', hf_model_id='model-id')

# After: MobiusModelBuilder pass
# Just put {'type': 'MobiusModelBuilder', 'precision': 'fp32'} in your Olive config!
\`\`\`

Thanks for the thorough review — these notes will help adoption!

justinchuby:
@xiaoyu-work @justinchuby Regarding per-component quantization:

With \`CompositeModelHandler\`, you can wire different passes per component by targeting them in the pipeline config. Here's the pattern:

\`\`\`json
{
  "passes": {
    "mobius": {"type": "MobiusModelBuilder", "precision": "fp16"},
    "quantize_decoder": {
      "type": "OnnxBlockWiseRtnQuantization",
      "bits": 4,
      "nodes_to_exclude": [...]
    }
  }
}
\`\`\`

Then the Olive engine will:
1. Run MobiusModelBuilder → CompositeModelHandler with components [decoder, vision, embedding, audio]
2. Run quantize_decoder on each component separately

For **selective per-component quantization** (e.g., skip embedding), you'd use \`nodes_to_exclude\` or create separate configs per component. We can document this pattern in the pass docstring to make it clearer.

The architecture here is that each pass is component-agnostic; the engine applies it to all components in the composite. For more granular control, recipes can create separate Olive config variants (one per component, or one per quantization strategy).

olive/passes/onnx/mobius_model_builder.py
```diff
@@ -0,0 +1,262 @@
...
+        NONE = "none"
+        ORT_GENAI = "ort-genai"
+
+    class MobiusEP(StrEnumBase):
+        """Execution providers supported by mobius."""
+
+        DEFAULT = "default"
+        CPU = "cpu"
+        CUDA = "cuda"
+        DML = "dml"
```
xiaoyu-work: do we still need dml？

olive/passes/onnx/mobius_model_builder.py
```diff
@@ -0,0 +1,262 @@
...
+        WEBGPU = "webgpu"
+        TRT_RTX = "trt-rtx"
+        ONNX_STANDARD = "onnx-standard"
+
+    # Maps Olive ExecutionProvider enum values to mobius EP names.
+    EP_MAP: ClassVar[dict[ExecutionProvider, str]] = {
```
xiaoyu-work, line 74: Looks like similar to https://github.com/microsoft/Olive/blob/main/olive/hardware/constants.py, or add this map there

olive/olive_config.json
```diff
@@ -205,6 +205,15 @@
...
             "supported_quantization_encodings": [  ]
         },
+        "MobiusModelBuilder": {
+            "module_path": "olive.passes.onnx.mobius_model_builder.MobiusModelBuilder",
+            "supported_providers": [ "*" ],
+            "supported_accelerators": [ "*" ],
+            "supported_precisions": [ "fp32", "fp16", "bf16" ],
+            "supported_algorithms": [  ],
+            "supported_quantization_encodings": [  ],
+            "extra_dependencies": [ "mobius-ai", "onnx-ir" ]
```
xiaoyu-work: No need onnx-ir here. onnx-ir is olive dependency

olive/olive_config.json
```diff
@@ -682,6 +691,8 @@
         "inc": [ "neural-compressor" ],
         "lora": [ "accelerate>=0.30.0", "peft", "scipy" ],
         "diffusers": [ "accelerate>=0.30.0", "peft", "diffusers" ],
+        "mobius-ai": [ "mobius-ai" ],
+        "onnx-ir": [ "onnx-ir" ],
```
xiaoyu-work: same as above

olive/passes/onnx/mobius_model_builder.py
justinchuby: Removed

olive/passes/onnx/mobius_model_builder.py
justinchuby: Moved

olive/olive_config.json
justinchuby: Removed

olive/olive_config.json
justinchuby: Removed

xiaoyu-work (approved):

PR #2418, opened 2026-04-16
Fix optimize CLI to set system EP and device

jambayk:
## Describe your changes

Fix the `olive optimize` CLI to properly configure the system execution provider and device in the generated workflow config.

- `_update_system_config` now creates the `local_system` with the specified execution provider and optional device. Previously only the QNN AOT case was handled, leaving the system config empty.
- When model builder is used as the exporter, the `OnnxFloatToFloat16` pass is skipped since model builder already produces the model in fp16.
- `test_optimize_cli_pass_list` now verifies the system EP and device are correctly set in the generated config for all test cases.

  - `olive optimize` now correctly sets the target system execution provider and device.

## (Optional) Issue link

### DIFF
--- a/olive/cli/optimize.py
+++ b/olive/cli/optimize.py
@@ -294,6 +294,19 @@ class OptimizeCommand(BaseOliveCLICommand):
         """Update system configuration based on provider and device."""
         provider = ExecutionProvider(self.args.provider)
 
+        accelerator = {"execution_providers": [provider.value]}
+        if self.args.device:
+            accelerator["device"] = self.args.device
+        if self.args.memory is not None:
+            accelerator["memory"] = self.args.memory
+
+        config["systems"]["local_system"] = {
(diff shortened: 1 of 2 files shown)


devang-ml (dismissed):

jambayk:
Addressed the review comments:

1. **Test for fp16 + model_builder**: Added a test case with `--precision fp16 --provider CUDAExecutionProvider` using the default model_builder exporter, verifying `OnnxFloatToFloat16` is not in the pass list. (4d5861a3)

2. **Memory units**: The `--memory` arg in the optimize CLI uses `type=int` with help text "in MB", and the value is passed consistently as MB to both the accelerator config and pass configs (e.g., CaptureSplitInfo at line 468). This is a pre-existing convention in the optimize command, not something introduced by this PR. No conversion needed.

xiaoyu-work (approved):

PR #2444, opened 2026-05-06
Add WER and RTFx speech evaluation metrics to Olive evaluator

jiafatom:
## Summary

Add Word Error Rate (WER) and Real-Time Factor (RTFx) as built-in accuracy sub-types in the Olive evaluator framework, enabling speech (ASR) model evaluation.

- **WER**: Measures transcription accuracy (lower is better)
- **RTFx**: Measures inference speed relative to audio duration (higher is better, e.g., RTFx=5 means 5x faster than real-time)

## Changes

- **`olive/evaluator/metric.py`** — Add `WER` and `RTFX` to `AccuracySubType` enum
- **`olive/evaluator/accuracy.py`** — Add `WordErrorRate` and `RealTimeFactor` classes
- **`olive/evaluator/olive_evaluator.py`** — Add `_inference_text()` path in OnnxEvaluator and PyTorchEvaluator for string-based predictions with timing/audio tracking
- **`olive/data/component/pre_process_data.py`** — Add `speech_transcription_pre_process` data component with Olive-style sampling (`--limit`/`--seed`)
- **`olive/data/container/huggingface_container.py`** — Add `speech-transcription` task type
- **`olive/olive_config.json`** — Add `speech` extra dependencies (`jiwer`, `librosa`, `soundfile`)
- **`olive/evaluator/examples/`** — Example config and script for speech evaluation

## Testing

Tested with Nemotron speech streaming model on LibriSpeech test.clean:
- 64 samples: **WER = 2.87%**, RTFx = 5.44
- 5 samples: WER = 5.00%, RTFx = 5.56

Unit tests added for WordErrorRate and RealTimeFactor (7 tests, all passing).

## Usage

```json
{
  "metrics": [{
    "name": "speech_eval",
    "type": "accuracy",
    "sub_types": [
      {"name": "wer", "higher_is_better": false},
      {"name": "rtfx", "higher_is_better": true}
    ]
  }]
}
```


olive/evaluator/examples/speech_eval_script.py
```diff
@@ -0,0 +1,96 @@
+# -------------------------------------------------------------------------
```
xiaoyu-work: can you move these examples to docs somewhere? and update the evaluator https://github.com/microsoft/Olive/blob/main/docs/source/how-to/configure-workflows/metrics-configuration.md. All files in this evaluator folder will be packaged into Olive package we don't want example files in released package

olive/evaluator/olive_evaluator.py
```diff
@@ -496,9 +518,110 @@ def _evaluate_onnx_accuracy(
...
+            inference_output, targets = self._inference_text(
+                model, metric, dataloader, post_func, device, execution_providers
+            )
+        else:
+            inference_output, targets = self._inference(
+                model, metric, dataloader, post_func, device, execution_providers
+            )
         return OliveEvaluator.compute_accuracy(metric, inference_output, targets)
 
+    def _inference_text(
```
xiaoyu-work, line 555: do we need to check if the onnx model is actual an audio model first?

jiafatom: Let's hold it for now?

it's tricky:

For HfModelHandler/PyTorch models: There's a task field (e.g., "automatic-speech-recognition", "text-generation") that could be
checked.

For ONNXModelHandler: There's no task field. You'd have to rely on:

 - io_config["input_names"] — audio models typically have inputs like "audio_signal", "input_features", etc.
 - model_attributes — a dict that could contain model metadata, but it's optional and user-provided.

The challenge: ONNX models (which is the primary use case for Olive evaluation) don't carry a standardized "task type" field.
Input names vary by model (Whisper uses "encoder_input_ids", Nemotron uses custom genai inputs, etc.), so heuristics would be
fragile.

Bottom line: There's no reliable way to detect "audio model" for ONNX models in the current Olive architecture. The cleanest
approach would be relying on the data config task type ("speech-transcription") or the metric sub-type itself (WER/RTFx) as the
signal — which is what we already do. I think the reviewer's suggestion is fine to skip as you said.

docs/source/how-to/configure-workflows/metrics-configuration.md
```diff
@@ -128,3 +128,85 @@ If you have multiple metrics to evaluate, you can configure them in the followin
...
+        ]
+    }]
+}
+```
+
+```{Note}
+The GenAI path requires `onnxruntime-genai` and `soundfile` to be installed.
+Olive auto-detects the model type (English-only vs multilingual) from the model config
+and sets the appropriate decoder prompt tokens.
+```
```
xiaoyu-work: No need this part?

docs/source/how-to/configure-workflows/metrics-configuration.md
```diff
@@ -128,3 +128,81 @@ If you have multiple metrics to evaluate, you can configure them in the followin
...
+
+```{Note}
+The GenAI path requires `onnxruntime-genai` and `soundfile` to be installed.
+Olive auto-detects the model type from `genai_config.json`:
+- **`whisper`** (offline): Uses `create_multimodal_processor()` with automatic 30s audio chunking.
+- **`nemotron_speech`** (streaming): Uses `og.StreamingProcessor` with stateful chunked inference and silence padding for right-context flushing.
+
+For other model types, use the `custom` metric type with an `evaluate_func`.
+```
```
xiaoyu-work: is this duplicated?

jiafatom: removed

xiaoyu-work (approved):

PR #2451, opened 2026-05-08
feat: add ort-multimodal evaluator for fast multimodal benchmarks

justinchuby:
## Summary

Add `LMEvalORTMultimodalEvaluator` (`ort-multimodal` backend) that uses direct ORT InferenceSession for multimodal GenAI packages. This is **15× faster** than the `ortgenai` backend for text-only benchmarks on multimodal models.

## Problem

The `ortgenai` evaluator loads all sub-models (decoder + vision + audio + embedding) and creates a new `og.Generator` per model call, adding significant overhead for text-only benchmarks like MMLU Pro.

The built-in `ort` evaluator can't handle models with heterogeneous KV cache head dimensions (e.g. Gemma4 with head_dim=256 for sliding attention and head_dim=512 for full attention) because its `Prefill` class allocates uniform IOBinding buffers.

## Solution

The new `ort-multimodal` evaluator:
- Reads `genai_config.json` to locate decoder and embedding ONNX files
- Loads only the decoder and embedding sessions (skips vision/audio)
- Runs embedding model to convert `input_ids` → `inputs_embeds`
- Runs decoder with per-layer empty KV cache buffers (supporting heterogeneous head_dim)
- Returns full logits `[batch, seq_len, vocab]` in a single forward pass

## Benchmark

Gemma4 E2B-IT MMLU Pro (12,032 questions, CUDA EP, H200 GPU):

| Backend | Accuracy | Throughput | vs ortgenai |
|---------|---------|-----------|------------|
| PyTorch HF | 22.6% | 185 req/s | 19.5× |
| **ort-multimodal** | **22.6%** | **142 req/s** | **15×** |
| ortgenai | 22.6% | 9.5 req/s | 1× |

Full analysis: [onnxruntime/mobius#285](https://github.com/onnxruntime/mobius/issues/285)

## Usage

```python
import olive.evaluator.lmeval_ort  # registers backends

from lm_eval.api.registry import get_model
model = get_model('ort-multimodal')(
    pretrained='/path/to/genai/package',
    ep='CUDAExecutionProvider',
    max_length=2048,
)
```

Or via Olive config:
```json
{"type": "LMEvaluator", "tasks": ["leaderboard_mmlu_pro"], "model_class": "ort-multimodal"}
```

### DIFF
--- a/olive/evaluator/lmeval_ort.py
+++ b/olive/evaluator/lmeval_ort.py
@@ -460,6 +460,151 @@ class Prefill:
         self._batch_size = batch_size
 
 
+@register_model("ort-multimodal")
+class LMEvalORTMultimodalEvaluator(LMEvalOnnxBase):
+    """Evaluate a multimodal ONNX model using direct ORT InferenceSession.
+
+    Designed for ORT GenAI multimodal packages (e.g. Gemma4) that have separate
+    decoder and embedding ONNX models. Uses direct session.run() instead of
+    GenAI's Generator API, avoiding the overhead of loading all sub-models
+    and creating Generator objects per call.
+
+    Supports models with heterogeneous KV cache head dimensions (e.g. Gemma4
+    with head_dim=256 for sliding attention and head_dim=512 for full attention),
+    which the standard 'ort' backend cannot handle.
+    """
+
+    def __init__(
+        self,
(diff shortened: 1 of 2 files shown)


justinchuby:
Not generalized

PR #2457, opened 2026-05-08
feat: add components_to_skip to OnnxBlockWiseRtnQuantization

titaiwangms:
## Summary

Add optional `components_to_skip` parameter (list of component names) to the `OnnxBlockWiseRtnQuantization` pass. When a composite model component's name appears in this list, its model files are copied to the output path unchanged instead of being quantized. Non-listed components are quantized as usual. Default `None` quantizes all components (fully backward compatible).

**Depends on #2456** (MobiusBuilder `components_to_export`). This PR's diff is scoped to RTN changes only — MobiusBuilder changes are in the base branch.

## Motivation

When quantizing large multi-component VLMs (e.g. Mistral-3 with decoder + vision_encoder + embedding), users may want to skip RTN quantization for specific components such as the embedding model — which may need to stay in higher precision (e.g. due to a `GatherBlockQuantized` bug with certain inputs).

## Changes

- **`olive/passes/onnx/rtn_quantization.py`**:
  - `from typing import Optional` (removed unused `List`)
  - Add `components_to_skip: list` `PassConfigParam` (default `None`)
  - Override `run()` to intercept `CompositeModelHandler` processing: copy skipped components via `shutil.copytree` (handles file vs directory `model_path`), quantize the rest via recursive `self.run()`
  - **`onnx_file_name` fallback**: use `getattr(component_model, "onnx_file_name", None) or "model.onnx"` to handle handlers constructed without an explicit file name
  - Mirror base class `_initialize()` guard and post-processing (`model_attributes` propagation + `_carry_forward_additional_files` per component)
  - Non-composite (single) models are unaffected

- **`test/passes/onnx/test_rtn_quantization.py`**:
  - New `TestRTNQuantizationComponentsToSkip` class with 5 tests
  - Skipped component passes through unchanged (MatMul present, no MatMulNBits)
  - Non-skipped components are quantized
  - `components_to_skip=None` quantizes all (backward compat)
  - Single-model (non-composite) unaffected
  - Unknown component name logs a warning (does not raise)
  - Removed redundant `AcceleratorSpec` reimport; wrapped long lines (120-char limit); split compound assertions

## Testing

All 16 tests pass:
```
python3 -m pytest test/passes/onnx/test_rtn_quantization.py -v
```

### DIFF
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -331,6 +331,7 @@ class Pass(ABC):
     ) -> OliveModelHandler:
         """Copy a component model to ``output_path`` without processing it."""
         from olive.model import ONNXModelHandler
+        from olive.model.utils.onnx_utils import resolve_onnx_path
         from olive.passes.onnx.common import resave_model
 
         if not isinstance(component_model, ONNXModelHandler):
@@ -340,7 +341,12 @@ class Pass(ABC):
             )
 
         onnx_file_name = Path(component_model.model_path).name
-        destination = output_path / onnx_file_name
+        # `output_path` follows the same convention `run()` relies on for processed components: it is
+        # already the final .onnx file path when it carries a ".onnx" suffix (e.g. ModelBuilder's flat
+        # "encoder.onnx"/"decoder.onnx" naming for multi-file models), and a directory to nest the
+        # component's file under otherwise. Treating it as an unconditional directory here would produce
+        # a nested "encoder.onnx/encoder.onnx" layout for the flat-naming case.
+        destination = Path(resolve_onnx_path(str(output_path), onnx_file_name))
         if Path(component_model.model_path).resolve() == destination.resolve():
             # Re-running the pass in place: the component is already at the destination. Copying it onto
             # itself would delete the file (resave_model unlinks the destination first), so do nothing.
@@ -354,8 +360,8 @@ class Pass(ABC):
             # names (e.g. "weights.bin") and multiple external data files are handled correctly.
             resave_model(component_model.model_path, destination)
         return ONNXModelHandler(
-            model_path=str(output_path),
-            onnx_file_name=onnx_file_name,
+            model_path=str(destination.parent),
+            onnx_file_name=destination.name,
             model_attributes=component_model.model_attributes,
         )
 
--- a/olive/passes/olive_pass.py
+++ b/olive/passes/olive_pass.py
@@ -4,9 +4,10 @@
 # --------------------------------------------------------------------------
 import inspect
 import logging
+import os
 import shutil
 from abc import ABC, abstractmethod
-from pathlib import Path
+from pathlib import Path, PureWindowsPath
 from typing import Any, Callable, ClassVar, Optional, Union, get_args
 
 from pydantic import BaseModel, ValidationError, create_model
@@ -81,10 +82,20 @@ class Pass(ABC):
         if hasattr(self.config, "user_script") and hasattr(self.config, "script_dir"):
             self._user_module_loader = UserModuleLoader(self.config.user_script, self.config.script_dir)
 
+        default_config = self.default_config(accelerator_spec)
+
+        # "components_to_skip" is handled by the generic per-component loop in run(), which only runs
+        # for passes that let the composite model be split into its components.
+        assert not (self._accepts_composite_model and "components_to_skip" in default_config), (
+            f"{self.__class__.__name__} cannot declare 'components_to_skip' because it sets"
+            " _accepts_composite_model=True: it processes the composite model as a whole, so there is no"
+            " per-component loop to skip components in."
+        )
+
         # Params that are paths [(param_name, required)]
         self.path_params = [
             (param, param_config.required, param_config.category)
-            for param, param_config in self.default_config(accelerator_spec).items()
+            for param, param_config in default_config.items()
             if param_config.category in (ParamCategory.PATH, ParamCategory.DATA)
         ]
 
@@ -228,17 +239,7 @@ class Pass(ABC):
                 model_attributes=model.model_attributes,
             )
         elif isinstance(model, CompositeModelHandler) and not self._accepts_composite_model:
-            components = []
-            component_names = []
-            # all components will be saved in the same parent directory
-            model_dir = Path(output_model_path).with_suffix("")
-            model_dir.mkdir(parents=True, exist_ok=True)
-            for component_name, component_model in model.get_model_components():
-                component_output_path = model_dir / component_name
-                output_model_component = self.run(component_model, str(component_output_path))
-                components.append(output_model_component)
-                component_names.append(component_name)
-            output_model = CompositeModelHandler(components, component_names, model_path=model_dir)
+            output_model = self._run_composite(model, output_model_path)
         else:
             output_model = self._run_for_config(model, self.config, output_model_path)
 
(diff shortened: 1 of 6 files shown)


titaiwangms:
## Review (synthesized from 4 reviewers: readability / code / critical / deep)

Convergent findings flagged with **[N reviewers]**. Severity follows worst-case among reviewers.

### Major

**1. `copytree(src.parent, ...)` can copy unrelated siblings** **[code]**

In the skip branch of `run()`: when `component_model.model_path` points to an `.onnx` file (not a directory), the code does `src_dir = src.parent` and copies the whole parent. If multiple components share a parent directory (flat composite layout), the skip silently bloats/corrupts the output by dragging in siblings. Fix: copy the ONNX file + its `.data` sidecar + listed `additional_files` explicitly, not the entire parent dir.

**2. Path comparison without `.resolve()` risks deleting the source** **[critical]**

```py
if src_dir != component_output_path:
    if component_output_path.exists():
        shutil.rmtree(str(component_output_path))
    shutil.copytree(str(src_dir), str(component_output_path))
```

Lexical `Path` compare. If absolute and relative forms point to the same dir (very plausible in stacked passes, symlinked workspaces, or case-insensitive FS), the inequality check passes, then `rmtree(component_output_path)` deletes the actual source before `copytree` reads from it. Fix: compare `.resolve()`'d paths, and reject overlapping src/dst.

**3. Non-atomic `rmtree`+`copytree` on large external-data dirs** **[critical]**

A crash, signal, or disk-full between `rmtree` and the end of `copytree` leaves the output unusable. Fix: stage to a sibling dir, validate, then atomic-rename.

**4. SKIP branch silently lies about handler type** **[deep]**

```py
output_component = ONNXModelHandler(
    model_path=str(component_output_path),
    onnx_file_name=onnx_file_name,
    model_attributes=component_model.model_attributes,
)
```

It unconditionally constructs `ONNXModelHandler`, but `CompositeModelHandler` permits any `OliveModelHandler` (nested composite, distributed ONNX, PyTorch). Latent today since Mobius produces only ONNX components, but worth at least an `isinstance(component_model, ONNXModelHandler)` assert with a clear error, plus a docstring constraint.

**5. `onnx_file_name` fallback hardcodes `"model.onnx"`** **[code]**

If `src.is_file()` and the input handler has no explicit `onnx_file_name`, the output handler points to `model.onnx` even if the source filename was different (e.g. `decoder.onnx`). Fix:

```py
if src.is_file():
    onnx_file_name = getattr(component_model, "onnx_file_name", None) or src.name
else:
    onnx_file_name = getattr(component_model, "onnx_file_name", None) or "model.onnx"
```

**6. `_initialized` guard duplicates a private base-class protocol** **[readability]**

The override copies the `_initialized` / `_initialize()` dance from `Pass.run()` (olive_pass.py:209-211). The comment is honest about it, but the coupling is fragile — if `Pass.run()` ever adds a second guard or renames the attribute, this silently diverges with no test catching the drift. Long-term: add `_ensure_initialized()` to the base class so overrides can call it without copying. Short-term: at least add a class-level comment that this assumes `_initialize()` is a no-op or idempotent.

### Minor

- `model.get_model_components()` is called twice (once for the unknown-name warning, once for the loop). The property has side-effects merging parent attrs into children — idempotent today, but stylistically off and the base only iterates once. Materialize once: `pairs = list(model.get_model_components())`.
- **Inconsistent strictness with #2456:** unknown component name → **raise** in #2456, **warn** in #2457. Pick one and document.
- `Path(output_model_path).with_suffix("")` strips dotted-dir names like `model.v1` (inherited footgun from base, not a regression).
- Docstring should explain why `_accepts_composite_model = True` wasn't used instead — short answer: `_run_for_config` doesn't receive the component name, so a per-component skip check has no name to key on. Worth saying explicitly.
- Add regression tests for: (a) non-`model.onnx` filenames, (b) file-based components in shared parent dirs.

### Praise

- Recursing via `self.run(component_model, ...)` for the non-skipped branch correctly inherits all base-class per-component behavior — clean design that avoids re-implementing `_run_for_config` invocation.
- Deep-trace confirmed that `model_attributes` propagation and `_carry_forward_additional_files` are correctly preserved with no double-carry on additional files (the helper checks `if not output_filepath.exists()` before copying).
- `_initialize()` idempotency: verified safe — base `_initialize()` is a no-op and not overridden on this pass; the recursive `super().run()` call will see `_initialized = True` and skip.

### Note

This PR is stacked on #2456 — please rebase / re-target to `main` once #2456 lands.

> _Posted by GitHub Copilot CLI on behalf of @titaiwangms._

titaiwangms:
## Rebase + review summary

Rebased onto the updated #2456 branch (base was 101 commits behind main; the underlying `rtn_quantization.py` only had one small unrelated upstream commit to reconcile — #2539's `_rename_output_unless_graph_output` guard, untouched here). Squashed into one clean commit.

### Review finding fixed
- **Major**: the skipped-component file-copy path only handled the default `model.onnx.data` external-data sidecar filename. A component using a non-default `external_data_name` (e.g. `weights.bin`, which Olive itself supports via `save_as_external_data`/`external_data_name`) would have its external-data file silently dropped, producing a broken ONNX model on load. Fixed by discovering the actual referenced external-data file(s) from the ONNX graph itself via the existing `get_external_data_file_names()` helper (`olive/passes/onnx/common.py`), instead of assuming a hardcoded suffix. Added a regression test that reproduces the bug and verifies the fix (round-trips a model saved with `location="weights.bin"` and confirms it loads successfully after being copied through the skip path).

Tests: 18/18 pass in `test_rtn_quantization.py`. Lint clean (only repo-wide `CPY001` false positives).

olive/passes/onnx/rtn_quantization.py
```diff
@@ -69,9 +97,160 @@ def _default_config(cls, accelerator_spec: AcceleratorSpec) -> dict[str, PassCon
...
+
+    @staticmethod
+    def _validate_contained(path: Path, root: Path) -> Path:
+        """Return ``path`` resolved, ensuring it stays inside ``root`` (defense in depth)."""
+        resolved = path.resolve()
+        if not resolved.is_relative_to(root.resolve()):
+            raise ValueError(f"Refusing to write outside the output directory {str(root)!r}: {str(path)!r}")
+        return resolved
+
+    def run(self, model, output_model_path: str):
```
xiaoyu-work: Do we need components_to_skip for other ONNX quantization passes? I'm a little concerned that overriding run here may cause inconsistencies if we change the parent class's run method in the future.

titaiwangms: Moved to base

olive/passes/olive_pass.py
```diff
@@ -249,6 +252,143 @@ def run(self, model: OliveModelHandler, output_model_path: str) -> OliveModelHan
...
+                component_name,
+            )
+        else:
+            # resave_model discovers external data files from the ONNX graph itself, so arbitrary `location`
+            # names (e.g. "weights.bin") and multiple external data files are handled correctly.
+            resave_model(component_model.model_path, destination)
+        return ONNXModelHandler(
+            model_path=str(output_path),
+            onnx_file_name=onnx_file_name,
+            model_attributes=component_model.model_attributes,
```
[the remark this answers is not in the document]
titaiwangms: Confirmed — real bug, not just for the RTN case but any pass using `components_to_skip` on a ModelBuilder-style multi-file composite (component names like `encoder.onnx`/`decoder.onnx`). `_copy_component_unchanged` always treated `output_path` as a directory and appended the file name again, so a skipped `encoder.onnx` component landed at `.../encoder.onnx/encoder.onnx` instead of flat `.../encoder.onnx`, breaking the flat multi-file layout `ModelBuilder` relies on.

Fixed in ca0ce52c: `_copy_component_unchanged` now uses `resolve_onnx_path` (the same helper regular ONNX passes use) to determine whether `output_path` is already the final `.onnx` file (flat naming) or a directory to nest under — matching exactly what the non-skip `self.run()` path already does. Added a regression test (`test_skipped_component_keeps_flat_layout_when_component_name_has_onnx_suffix`) that reproduces the nested-directory bug against the old code and passes against the fix.

titaiwangms:
Included by https://github.com/microsoft/Olive/pull/2480

PR #2491, opened 2026-06-01
Add OnnxMoEQuantization pass (com.microsoft::MoE → QMoE)

justinchuby:
## Problem

Mobius and other ONNX exporters emit MoE blocks as fused `com.microsoft::MoE` nodes whose per-expert FC1/FC2 weights are 3-D fp16/bf16/fp32 initializers. The existing weight-quantization passes (`OnnxKQuantQuantization`, `OnnxBlockWiseRtnQuantization`, `OnnxBnb4Quantization`) only target standalone `MatMul` nodes, so:

- The per-expert weights — which are typically ~80% of total parameters in a MoE model — stay at the model's compute dtype after running any of those passes.
- For Gemma 4 26B-A4B, this means int4 quantization yields only ~6% file-size reduction (~48 GB → ~45 GB).

The correct target for these weights is `com.microsoft::QMoE`, which is now supported by both the CUDA and (experimental) CPU kernels in ORT main after [microsoft/onnxruntime#28467](https://github.com/microsoft/onnxruntime/pull/28467). There is no existing pass in either Olive or ORT that performs the `MoE → QMoE` rewrite.

## Solution

New `OnnxMoEQuantization` pass that walks the graph and rewrites every `com.microsoft::MoE` node:

1. Validates that `fc1_experts_weights` and `fc2_experts_weights` are 3-D static initializers.
2. For each expert, calls ORT's pybind `quantize_matmul_{4,8}bits` to produce per-expert quantized weights and symmetric fp16 scales, then CUTLASS-prepacks them via `pack_weights_for_cuda_mixed_gemm` so the QMoE kernels can read them directly.
3. Stacks the per-expert tensors along axis 0 and registers them as new initializers (`uint8` weights + `fp16` scales).
4. Replaces the MoE node with a QMoE node, carrying over all original routing/activation attributes plus `expert_weight_bits`, optional `block_size`, and `quant_type='int'`.
5. Drops the orphan fp16 weight initializers.

## Configuration

| Param              | Default | Notes                                                                 |
|--------------------|--------:|-----------------------------------------------------------------------|
| `bits`             | 4       | 4 (int4) or 8 (int8)                                                  |
| `block_size`       | 0       | 0 = per-row scales; otherwise must be a power of two ≥ 16             |
| `nodes_to_exclude` | None    | MoE node names to leave unquantized                                   |
| `force_arch`       | 80      | CUTLASS prepacking target SM (80 = Ampere, 90 = Hopper)               |

## Example

```json
{
  "input_model": {"type": "OnnxModel", "model_path": "decoder/model.onnx"},
  "passes": {
    "moe_quant": {
      "type": "OnnxMoEQuantization",
      "bits": 4,
      "block_size": 0
    }
  }
}
```

## Runtime requirement

The pass uses `pack_weights_for_cuda_mixed_gemm` from ORT, which is only exported when ORT is built with `USE_CUDA`. A descriptive `RuntimeError` is raised at run time when the binding is unavailable, telling the user to install `onnxruntime-gpu >= 1.28` (or a recent nightly).

## Limitations / out-of-scope

- **fc3 inputs** (3-fold MoE variants) trigger a warning-skip per node.
- **Symmetric integer quantization only** (matching the QMoE kernel's preferred layout). `fp4` / `fp8` / `wfp4afp8` `quant_type`s are left for a follow-up.
- **No calibration-aware quantization** (GPTQ / AWQ). This pass is pure RTN.

## Tests

5 unit tests in `test/passes/onnx/test_moe_quantization.py`:

| Test | Coverage |
|---|---|
| `test_moe_to_qmoe_conversion` | End-to-end MoE→QMoE with int4 + per-row scales; checks node replacement, input slot order, attribute preservation, and uint8/fp16 shape/dtype |
| `test_moe_to_qmoe_blockwise` | block_size=16 produces 3-D scales `[E, N, K/block_size]` and emits the `block_size` attribute |
| `test_moe_to_qmoe_skip_when_not_initializer` | Dynamic-weight MoE nodes are skipped with a warning, not converted |
| `test_invalid_bits_rejected` | bits ∉ {4, 8} fails fast |
| `test_invalid_block_size_rejected` | non-power-of-two block_size fails fast |

The CUTLASS prepack helper is patched with a pass-through stub during tests so CPU-only CI can exercise the graph transform. All 5 pass locally + the existing 416 `test/passes/onnx` tests still pass + `lintrunner` is clean.

### DIFF
--- a/olive/olive_config.json
+++ b/olive/olive_config.json
@@ -283,6 +283,14 @@
             "supported_algorithms": [ "kquant" ],
             "supported_quantization_encodings": [  ]
         },
+        "OnnxMoEQuantization": {
+            "module_path": "olive.passes.onnx.moe_quantization.OnnxMoEQuantization",
+            "supported_providers": [ "CUDAExecutionProvider" ],
+            "supported_accelerators": [ "gpu" ],
+            "supported_precisions": [ "int4", "int8" ],
+            "supported_algorithms": [ "rtn" ],
+            "supported_quantization_encodings": [  ]
+        },
         "OnnxBnb4Quantization": {
             "module_path": "olive.passes.onnx.bnb_quantization.OnnxBnb4Quantization",
             "supported_providers": [ "CPUExecutionProvider" ],
--- a/olive/passes/onnx/moe_quantization.py
+++ b/olive/passes/onnx/moe_quantization.py
@@ -312,9 +312,19 @@ def _get_input(node: ir.Node, idx: int) -> ir.Value:
 
 
 def _maybe_input(node: ir.Node, idx: int) -> ir.Value | None:
+    """Return the optional input at ``idx``, treating empty / missing slots as absent.
+
+    ONNX represents an unset optional input either as a slot past the end of the
+    inputs list, or as an in-place placeholder with an empty name. ``onnx_ir``
+    typically maps the latter to ``None``, but defensively handle the
+    ``ir.Value(name="")`` case too so callers can rely on ``None`` meaning absent.
+    """
     if idx >= len(node.inputs):
         return None
-    return node.inputs[idx]
+    value = node.inputs[idx]
+    if value is None or not value.name:
+        return None
+    return value
 
 
 def _require_initializer(value: ir.Value, what: str, initializers: dict[str, ir.Value]) -> np.ndarray:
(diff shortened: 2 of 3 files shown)


justinchuby:
Follow-ups filed:

- **microsoft/onnxruntime#28748** — proposed fix on the ORT side: `com.microsoft::QMoE` should prepack int4/int8 weights in `PrePack()`, like `com.microsoft::MatMulNBits` already does. That would let this pass quantize on a CPU-only host with no special steps. The current asymmetry is most likely an oversight inherited from the TensorRT-LLM-flavoured fpA_intB code, since `MatMulNBits` does it correctly.
- **microsoft/Olive#2492** — tracks a numpy + optional-cupy port of `pack_weights_for_cuda_mixed_gemm` for this pass, so it can run on CPU-only hosts without waiting for the ORT-side fix to ship.

justinchuby:
Sent the proposed ORT-side fix as [microsoft/onnxruntime#28749](https://github.com/microsoft/onnxruntime/pull/28749). Once that lands and ships in a release, this pass can drop the `pack_weights_for_cuda_mixed_gemm` dependency entirely — the Olive pass would only need `quantize_matmul_{4,8}bits` (which is in every ORT build, CPU or CUDA).

titaiwangms:
Related: #2583 proposes resuming `jambayk/moe-quant` to add MoE-aware weight quantization at the PyTorch/HF level (storage-only `QuantTensor` representation for fused-3D and `ModuleList`-style expert weights), as a complement to this ONNX-graph-level MoE work. Cross-linking for context in case an implementing agent picks either up.

PR #2496, opened 2026-06-04
Update capture onnx graph cli exporting logic

xiaoyu-work:
## Describe your changes
1. mobius exporter always default to genai.
2. remove import error catch from mb pass, it can swallow the real error.

### DIFF
--- a/test/passes/onnx/test_mobius_model_builder.py
+++ b/test/passes/onnx/test_mobius_model_builder.py
@@ -133,7 +133,6 @@ def test_default_config_params():
     )
     config = MobiusBuilder._default_config(accelerator_spec)  # pylint: disable=protected-access
     assert "precision" in config
-    assert "runtime" not in config
     assert "execution_provider" not in config
     assert "trust_remote_code" not in config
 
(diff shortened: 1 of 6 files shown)


test/passes/onnx/test_mobius_model_builder.py
```diff
@@ -127,13 +127,12 @@ def __exit__(self, *args):
 
 
 def test_default_config_params():
-    """MobiusBuilder must declare precision and runtime, and must not declare execution_provider or trust_remote_code."""
+    """MobiusBuilder must declare precision, and must not declare execution_provider or trust_remote_code."""
```
justinchuby, line 130: Why was execution_provider or trust_remote_code not declarable again? Does this match MB?

justinchuby (dismissed):

test/passes/onnx/test_mobius_model_builder.py
xiaoyu-work: Yes, I don’t think MB has a concept of an execution provider or “trust remote code.”?

justinchuby: It should? Is it not the --ep option?

In mobius we do have a toggle for trust remote code, I think. You could default it to true if that's preferable.

justinchuby (approved):

PR #2528, opened 2026-06-18
Fix genai_config.json discovery for nested multi-component model layouts

jiafatom:
## Description

The evaluator previously assumed `genai_config.json` is in the same directory as the ONNX model file (`Path(model.model_path).parent`). For nested multi-component model layouts (e.g., `models/decoder/model.onnx` where `genai_config.json` lives at `models/`), this caused evaluation to fail because the config file was not found.

This is the case for models exported by `MobiusBuilder` (gemma4, etc.), which creates a nested layout:
```
models/
├── genai_config.json
├── decoder/model.onnx
├── vision_encoder/model.onnx
├── audio_encoder/model.onnx
└── embedding/model.onnx
```

When the evaluator loads the decoder (`models/decoder/model.onnx`), `Path(model.model_path).parent` resolves to `models/decoder/`, missing the `genai_config.json` at `models/`.

## Changes

- Added `_find_genai_config()` static helper that searches upward (up to 3 levels) from the ONNX file's parent directory to find `genai_config.json`
- Added `_get_genai_model_dir()` static helper that returns the root model directory containing `genai_config.json` (falls back to ONNX file's parent if not found)
- Updated all 7 occurrences of `Path(model.model_path).parent` used for genai config discovery to use these new helpers:
  - `_load_genai_config()` — already fixed to use `_find_genai_config()`
  - `_inference_text_genai()` — uses `_get_genai_model_dir()`
  - `_inference_text_genai_streaming()` — uses `_get_genai_model_dir()`
  - `LMEvaluator.evaluate()` — uses `_get_genai_model_dir()`
  - `MTEBEvaluator.evaluate()` — uses `_find_genai_config()` and `_get_genai_model_dir()`

## Motivation

This fix enables evaluation of multimodal models with nested component layouts (e.g., gemma4-E2B exported by MobiusBuilder) using Olive's built-in evaluator.

### DIFF
--- a/olive/evaluator/olive_evaluator.py
+++ b/olive/evaluator/olive_evaluator.py
@@ -584,9 +584,14 @@ class OnnxEvaluator(_OliveEvaluator, OnnxEvaluatorMixin):
 
     @staticmethod
     def _load_genai_config(model: ONNXModelHandler) -> Optional[dict]:
-        """Load genai_config.json from the model directory, or return None if not found."""
-        genai_config_path = Path(model.model_path).parent / "genai_config.json"
-        if not genai_config_path.exists():
+        """Load genai_config.json from the model directory, or return None if not found.
+
+        Searches upward from the ONNX file's parent directory to support nested
+        multi-component model layouts (e.g. ``models/decoder/model.onnx`` where
+        ``genai_config.json`` lives at ``models/``).
+        """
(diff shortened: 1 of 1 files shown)


jiafatom:
Use https://github.com/microsoft/Olive/pull/2529

PR #558, opened 2023-09-13
📄 Docs for data_config settings

trajepl:
## Describe your changes
Docs for data_config settings: how to configure data

## Checklist before requesting a review
- [ ] Add unit tests for this change.
- [ ] Make sure all tests can pass.
- [ ] Update documents if necessary.
- [ ] Format your code by running `pre-commit run --all-files`
- [ ] Is this a user-facing change? If yes, give a description of this change to be included in the release notes.

PR #515, opened 2023-08-25
🗺️ Align hf dataset label with model output

trajepl:
## Describe your changes
In some cases, the model output label is mismatched with original huggingface dataset.
E.g. the deberta-base-mnli output id2label is:
image
But the dataset label is:
image

This PR is used to fill this gap.

## Checklist before requesting a review
- [ ] Add unit tests for this change.
- [ ] Make sure all tests can pass.
- [ ] Update documents if necessary.
- [ ] Format your code by running `pre-commit run --all-files`
- [ ] Is this a user-facing change? If yes, give a description of this change to be included in the release notes.

PR #2531, opened 2026-06-19
Add multimodal evaluation and component-aware quantization support

DelwinKim:
## Describe your changes

> [!IMPORTANT]
> The evaluator and ORT-GenAI package-handling changes (i.e. just the `LMMSEvaluator` integration) have been isolated in draft PR [#2615](https://github.com/microsoft/Olive/pull/2615), from [`DelwinKim/Olive:t-delwinkim/lmms-evaluator-only`](https://github.com/DelwinKim/Olive/tree/t-delwinkim/lmms-evaluator-only) at [`08a6a2b2`](https://github.com/DelwinKim/Olive/commit/08a6a2b2684d69d91061b5135eb0acd97bfacd73). The remaining quantization and calibration features should be reviewed as separate changes.
> Results are tracked in [microsoft/Olive#2613](https://github.com/microsoft/Olive/issues/2613).

This draft adds an end-to-end **multimodal evaluation and quantization workflow** to Olive. It integrates the [lmms-eval](https://github.com/EvolvingLMMs-Lab/lmms-eval) benchmark harness, makes Mobius multi-component ORT-GenAI packages directly evaluable, and adds component-aware quantization/calibration support for vision-language and omni models.

The PR branch intentionally serves as a single reproduction branch (for reproducing results) for the relevant Olive changes, including the multimodal evaluator, sensitivity-based mixed precision, and HQQ fixes, etc. 

### Validated environment

| Component | Version / revision |
|---|---|
| Python | `3.12.9` |
| Olive | `olive-ai==0.11.0.dev0`; `t-delwinkim/lmms-ort-evaluator` at [`ed853908`](https://github.com/DelwinKim/Olive/commit/ed853908cff873e9f87e8b40840bc80841c15efc) |
| Mobius | `mobius-ai==0.1.0`; published integration branch [`t-delwinkim/multimodal-quant-integration-20260803`](https://github.com/onnxruntime/mobius/tree/t-delwinkim/multimodal-quant-integration-20260803) at [`d369f992`](https://github.com/onnxruntime/mobius/commit/d369f9926b920d0a47b59209b2bd22e1fe936831) |
| lmms-eval | `lmms-eval==0.7.2`; evaluation-fix revision [`8a252a73`](https://github.com/DelwinKim/lmms-eval/commit/8a252a731ec2cba2ba2b148141d57e36d1f29bed) |
| ONNX Runtime | `onnxruntime-gpu==1.29.0`; source revision [`761185d2`](https://github.com/microsoft/onnxruntime/commit/761185d2239fcc0af8fbf46c7761751b0460ca3d) |
| ORT-GenAI | `onnxruntime-genai-cuda==0.15.0.dev0`; source revision [`9f7a24fb`](https://github.com/microsoft/onnxruntime-genai/commit/9f7a24fb1c2e0a313582db8ef3cf5d999ed6d29a); ONNX Runtime Extensions [`319f8d2c`](https://github.com/microsoft/onnxruntime-extensions/commit/319f8d2c02c4e2f013129bea52e70e75ba79bbcf) |
| Build/evaluation stack | `torch==2.12.0`; `transformers==5.11.0`; `datasets==5.0.0`; `onnxscript==0.7.1`; `protobuf==6.33.6` |

All benchmark results in the updated sample-results comment come from deployed ONNX/ORT-GenAI packages evaluated through `LMMSEvaluator`, not from PyTorch proxy evaluations.

This is the current public reproduction environment. Historical pre-integration result rows retain their recorded package/build provenance; updating this table does not relabel those artifacts as having been built by the current Mobius revision.

Post-review validation covered 463 affected Olive tests and 50 focused lmms-eval tests, changed-file lint/format checks, Olive import/CLI checks, and fresh deployed-package probes for external task resolution, package-native audio preprocessing, PNG/JPEG85 image profiles, MMLU loglikelihood, AI2D direct extraction, and limit-aware OCRBench. The exact fast-test workflow also passed against `onnxruntime-genai==0.15.1` (4 tests and 13 subtests), while preserving compatibility with the pinned 0.15.0 development runtime. The pinned lmms-eval revision also includes the exact Google-prompt `fleurs_en_google_asr` task as a built-in task. Machine-local artifacts remain outside this PR.

### Reported quantization variants

| Label | Exact evaluated variant |
|---|---|
| FP16 | Original Hugging Face checkpoint exported by `MobiusBuilder(precision="fp16")`; no weight quantization |
| KQ-all | Post-export ONNX K-Quant W4A16 group-32 asymmetric on every eligible MatMul |
| KQ-dec | Post-export ONNX K-Quant W4A16 group-32 asymmetric on decoder projections; LM head/embedding/encoders remain floating point |
| RTN-all | Post-export ONNX blockwise RTN W4A16 group-32 asymmetric on every eligible MatMul |
| RTN-dec | Post-export ONNX blockwise RTN W4A16 group-32 asymmetric on decoder projections; LM head/embedding/encoders remain floating point |
| HQQ | Post-export ONNX HQQ W4A16 with exact/glob node scoping and corrected `MatMulNBits` layout; replaced weights are removed from root and nested graph scopes |
| Olive GPTQ | PyTorch GPTQ W4A16 group-32 asymmetric decoder-only; `desc_act=false`, 1% Hessian damping |
| QAT W4A16-CT | Google `gemma-4-E2B-it-qat-w4a16-ct` compressed-tensors QAT checkpoint: symmetric W4A16 group-32 Linear weights with BF16 activations; vision/audio/projectors/embeddings/LM head remain BF16; no additional PTQ pass |
| Q4_0 GGUF | Google's QAT Q4_0 decoder imported as block-32 INT4 `MatMulNBits`; encoders/embeddings/LM head remain floating point |
| Q4_0 + INT8 embeds | Q4_0 decoder plus ONNX RTN W8A16 group-32 asymmetric token/per-layer embeddings and embedding projection |

Standalone W8 RTN and symmetric Google-policy RTN results are intentionally excluded. W8 packages appear only as source controls in the matched INT8-embedding experiment.

### 1. `LMMSEvaluator`

A new evaluator (`"type": "LMMSEvaluator"`) runs official lmms-eval tasks inside an Olive workflow.

Supported input handlers:

- **`ONNXModelHandler`** pointing into a complete ORT-GenAI package (`genai_config.json` plus decoder/vision/audio/embedding graphs). This dispatches to the new `ortgenai_mm` adapter.
- **`HfModelHandler`** for native Hugging Face multimodal evaluation. The lmms-eval model wrapper is inferred from `model_type` or selected explicitly with `model_class`.

The evaluator:

- implements image and audio `generate_until`;
- provides text `loglikelihood` for compatible tasks;
- handles `og.Images` and `og.Audios`;
- discovers `genai_config.json` from nested package layouts;
- resolves model-family prompt defaults and structured-content templates;
- supports per-model processor argument shapes, including Phi-4-MM and Whisper;
- preserves official lmms-eval scorers for OCRBench, DocVQA ANLS, ChartQA, TextVQA, AI2D, WER, BLEU, and other registered tasks;
- writes compact aggregate JSON and optional per-sample generations.

### 2. `CompositeToOnnxPackage`

`MobiusBuilder` emits a `CompositeModelHandler` containing decoder, vision, audio, and embedding components. Generic evaluation cannot open that handler as one inference session.

`CompositeToOnnxPackage`:

- hardlink-copies the complete composite package while preserving its nested graph layout;
- selects an entry component (default: `decoder`);
- returns a single `ONNXModelHandler`;
- preserves the complete package layout so ORT-GenAI can discover every component.

### 3. Component-aware mixed precision

`MultiModalMixedPrecision` classifies quantizable modules into:

- `vision`
- `audio`
- `text`
- `projector`
- `lm_head`
- `embeds`

It writes `mixed_precision_info` consumed by downstream PyTorch RTN, GPTQ, and K-Quant passes:

- 16/32-bit components are excluded from quantization;
- 2/4/8-bit components can receive per-module overrides;
- custom component-name rules can extend the built-in architecture mappings.

The pass plans component precision but does not quantize by itself.

`SelectiveMixedPrecision` provides the complementary layer-sensitivity path. FP16/FP32 promotions are emitted as quantization exclusions, and optional module globs are validated against fused QKV groups and tied embedding/LM-head pairs.

### 4. Multimodal calibration

- **`MultimodalActivationRangeCalibration`** records per-component activation ranges and derived qparams for vision/audio/text/projector analysis. It is a diagnostic/planning pass; it does not claim to implement MQuant's MSQ, AIFS, or runtime activation kernels.
- PyTorch RTN/GPTQ calibration replay now supports Gemma-4 per-layer inputs, Qwen2.5-VL visual inputs, and Qwen3-VL DeepStack feature injection.
- Quantizers preserve `MultiModalMixedPrecision` exclusions and avoid quantizing modules for which calibration statistics were not collected.

### 5. Quantizer and workflow integration

- ONNX blockwise RTN accepts `nodes_to_include` and `nodes_to_exclude`; ONNX K-Quant accepts `nodes_to_exclude` plus per-node `customized_weight_config` overrides. Both filters match exact node names and `*`/`?` glob patterns.
- K-Quant component scopes remain attached to composite component names.
- Native Olive GPTQ supports multimodal calibration inputs and component exclusions.
- ONNX HQQ supports exact and full fnmatch-style node selection, corrected output-channel-major `MatMulNBits` packing, and recursive orphan-initializer cleanup.
- `GptqModel` accepts the current GPTQModel API/configuration surface.
- Cache/resource handling preserves additional processor/tokenizer files across pass boundaries.
- New passes are registered in `olive/olive_config.json` and documented in the pass reference and quantization guide.

## Current validated recipes

The following examples use the exact current pass/evaluator field names and the tested quantization values. Machine-local snapshot, cache, model, and result paths from the validation workspace are replaced with portable paths.

### Recipe A: Gemma-4 E2B decoder-only ONNX K-Quant plus AI2D evaluation

```json
{
  "input_model": {
    "type": "HfModel",
    "model_path": "google/gemma-4-E2B-it"
  },
  "systems": {
    "local_system": {
      "type": "LocalSystem",
      "accelerators": [
        {
          "device": "gpu",
          "execution_providers": ["CUDAExecutionProvider"]
        }
      ]
    }
  },
  "passes": {
    "mobius_build": {
      "type": "MobiusBuilder",
      "precision": "fp16"
    },
    "int4_quantize": {
      "type": "OnnxKQuantQuantization",
      "bits": 4,
      "block_size": 32,
      "nodes_to_exclude": [
        "vision_encoder/*",
        "audio_encoder/*",
        "embedding/*",
        "decoder/lm_head/*"
      ]
    },
    "package": {
      "type": "CompositeToOnnxPackage"
    }
  },
  "evaluators": {
    "evaluator": {
      "type": "LMMSEvaluator",
      "tasks": ["ai2d_direct"],
      "batch_size": 1,
      "max_length": 32768,
      "ignore_stop_strings": ["\n\n"],
      "log_samples": true,
      "output_path": "results/gemma4_kquant_decoder_ai2d.json",
      "fail_on_error": true
    }
  },
  "evaluator": "evaluator",
  "evaluate_input_model": false,
  "target": "local_system",
  "output_dir": "models/gemma4_kquant_decoder",
  "cache_dir": "cache/gemma4_kquant_decoder",
  "no_artifacts": true,
  "clean_evaluation_cache": true
}
```

Run one benchmark per Olive process. To evaluate English audio instead, replace `ai2d_direct` and the output filename with one of `librispeech_test_clean`, `librispeech_test_other`, or `fleurs_en`. Keep custom tasks such as `ai2d_direct` in a versioned lmms-eval task bundle when they are not built into the pinned dependency.

### Recipe B: Qwen2.5-VL component-aware PyTorch RTN

```json
{
  "input_model": {
    "type": "HfModel",
    "model_path": "Qwen/Qwen2.5-VL-3B-Instruct",
    "task": "image-text-to-text",
    "load_kwargs": {
      "torch_dtype": "bfloat16",
      "attn_implementation": "sdpa"
    }
  },
  "systems": {
    "local_system": {
      "type": "LocalSystem",
      "accelerators": [
        {
          "device": "gpu",
          "execution_providers": ["CUDAExecutionProvider"]
        }
      ]
    }
  },
  "passes": {
    "component_plan": {
      "type": "MultiModalMixedPrecision",
      "component_precision": {
        "vision": 16,
        "projector": 16,
        "text": 4,
        "lm_head": 16,
        "embeds": 16
      },
      "bits": 4,
      "group_size": 32,
      "sym": false
    },
    "rtn": {
      "type": "Rtn",
      "bits": 4,
      "group_size": 32,
      "sym": false,
      "lm_head": false,
      "embeds": false
    },
    "mobius_build": {
      "type": "MobiusBuilder",
      "precision": "fp16"
    },
    "package": {
      "type": "CompositeToOnnxPackage"
    }
  },
  "evaluators": {
    "evaluator": {
      "type": "LMMSEvaluator",
      "tasks": ["ai2d_direct"],
      "batch_size": 1,
      "max_length": 32768,
      "ignore_stop_strings": ["\n\n"],
      "log_samples": true,
      "output_path": "results/qwen25vl_component_rtn_ai2d.json",
      "fail_on_error": true
    }
  },
  "evaluator": "evaluator",
  "evaluate_input_model": false,
  "target": "local_system",
  "output_dir": "models/qwen25vl_component_rtn",
  "cache_dir": "cache/qwen25vl_component_rtn",
  "no_artifacts": true,
  "clean_evaluation_cache": true
}
```

### Standalone Hugging Face evaluation

```json
{
  "input_model": {
    "type": "HfModel",
    "model_path": "Qwen/Qwen2.5-VL-3B-Instruct"
  },
  "evaluators": {
    "evaluator": {
      "type": "LMMSEvaluator",
      "model_class": "qwen2_5_vl",
      "trust_remote_code": true,
      "tasks": ["ai2d_direct"],
      "batch_size": 1,
      "log_samples": true,
      "output_path": "results/qwen25vl_hf_ai2d.json",
      "fail_on_error": true
    }
  },
  "evaluator": "evaluator",
  "evaluate_input_model": true,
  "output_dir": "models/qwen25vl_hf",
  "cache_dir": "cache/qwen25vl_hf",
  "no_artifacts": true,
  "clean_evaluation_cache": true
}
```

Run any recipe with:

```bash
olive run --config <recipe.json>
```

## `LMMSEvaluator` configuration

| Field | Default | Description |
|---|---:|---|
| `tasks` | required | lmms-eval task names |
| `model_class` | auto | Explicit lmms-eval wrapper for HF evaluation |
| `limit` | full split | Samples per task |
| `batch_size` | `1` | Only single-request generation is currently supported |
| `max_new_tokens` | `256` | Fallback completion budget when a task does not override it |
| `max_length` | `32768` | ORT-GenAI prompt + media + completion sequence budget |
| `system_prompt` | model-specific | Explicit system prompt override |
| `include_path` | none | One external lmms-eval task directory or a list of directories |
| `audio_target_sample_rate` | package-defined | Optional explicit host-resampling override; otherwise the package processor handles its declared/default rates |
| `image_serialization_profile` | `lossless` | ORT input serialization: lossless PNG or explicit `jpeg85` parity mode |
| `ignore_stop_strings` | none | Task stop strings to remove, such as `"\n\n"` |
| `trust_remote_code` | `false` | HF model/processor loading behavior |
| `log_samples` | `false` | Preserve per-sample generations |
| `output_path` | none | Aggregate result path |
| `fail_on_error` | `true` | Raise instead of recording a failed evaluator result |

## Tests

- `test/evaluator/test_lmms_ort.py`
- `test/evaluator/test_olive_evaluator.py`
- `test/test_cache.py`
- `test/passes/onnx/test_composite_to_onnx_package.py`
- `test/passes/onnx/test_hqq_quantization.py`
- `test/passes/onnx/test_kquant_quantization.py`
- `test/passes/onnx/test_rtn_quantization.py`
- `test/passes/pytorch/test_multimodal_activation_range.py`
- `test/passes/pytorch/test_multimodal_mixed_precision.py`
- `test/passes/pytorch/test_multimodal_quantization.py`
- `test/passes/pytorch/test_quant_utils.py`
- `test/passes/pytorch/test_selective_mixed_precision.py`

## Current limitations

- ORT-GenAI multimodal generation is single-request (`batch_size=1`).
- The adapter is invoked through Olive/Python; it is not registered as an lmms-eval v2 CLI entry point.
- `MultimodalActivationRangeCalibration` is diagnostic metadata only; it does not add MQuant runtime kernels.
- Multiround/interleaved lmms-eval generation is not implemented.
- Gemma-4 HF evaluation uses the corresponding wrapper from the pinned lmms-eval development revision above.
- Phi-4-MM packages whose tokenizer chat template accepts only flat string content cannot currently route image/audio requests through the structured-content adapter; those packages require a compatible exported template.

## Checklist

- [x] Add evaluator, adapter, packaging pass, multimodal planning/calibration passes, documentation, and tests.
- [x] Validate deployed ONNX/ORT-GenAI evaluation on Gemma-4, Qwen2.5-VL, and Qwen3-VL.
- [x] Publish final protocol-matched sample results in the PR comment.
- [x] Keep machine-local recipes, caches, models, and generated result artifacts out of the PR.

### Release note

Add multimodal vision/audio evaluation through lmms-eval, plus component-aware mixed precision, multimodal calibration and deployable ORT-GenAI package evaluation for Olive workflows.


DelwinKim:
## Outdated benchmark snapshot (historical only)

> [!WARNING]
> This July 28 snapshot is outdated and is **not** part of the current PR validation or merge claims. It is retained only for historical provenance. Current results are intentionally omitted from this draft.

<details>
<summary>Show outdated July 28 benchmark snapshot</summary>

## Outdated July 28 benchmark snapshot

> Updated 2026-07-28. This replaced the earlier July 1 `*_lite` screening table with the then-current protocol-matched results. Debug/smoke outputs and legacy flexible-extraction AI2D scores are excluded.

These runs exercise `LMMSEvaluator` end to end on deployed ONNX/ORT-GenAI packages. Vision, audio, translation, and fusion metrics use lmms-eval task scorers; AI2D uses a direct letter-only prompt with exact-match extraction.

### Protocols

- **AI2D direct:** all 3,088 test examples; direct letter-only prompt and exact-match extraction.
- **Full image:** complete ChartQA, DocVQA, TextVQA, OCRBench, and AI2D splits.
- **Full audio:** complete FLEURS, LibriSpeech, and CoVoST splits.
- **MathVision 500:** the same ordered first 500 examples for screening comparisons.
- All rows are evaluated ONNX packages, not PyTorch proxies. Higher is better except WER/CER.

### Exact variant definitions

| Label | Exact evaluated variant |
|---|---|
| FP16 | Original Hugging Face checkpoint exported by Mobius to FP16 ONNX; no weight quantization |
| PR515 | olive-recipes PR #515 CUDA mixed package: post-export ONNX K-Quant W4A16 group-32 asymmetric decoder, post-export ONNX RTN W8A16 group-32 asymmetric vision/audio encoders, FP16 embedding |
| PR558 | olive-recipes PR #558 CUDA five-recipe package built from pinned Gemma-4 E2B snapshot `9dbdf8a`: decoder K-Quant group-32 asymmetric with 135 W4A16 and 141 W8A16 MatMuls; vision/audio/embedding use ONNX RTN W8A16 group-32 asymmetric |
| KQ-all / KQ-dec | Post-export ONNX K-Quant W4A16 group-32 asymmetric on all eligible MatMuls / decoder projections only |
| RTN-all / RTN-dec | Post-export ONNX blockwise RTN W4A16 group-32 asymmetric on all eligible MatMuls / decoder projections only |
| Olive GPTQ | PyTorch GPTQ W4A16 group-32 asymmetric decoder-only; `desc_act=false`, 1% Hessian damping |
| QAT W4A16-CT | Google `gemma-4-E2B-it-qat-w4a16-ct` compressed-tensors checkpoint: symmetric W4A16 group-32 Linear weights, BF16 activations, BF16 vision/audio/projectors/embeddings/LM head; exported without another PTQ pass |
| Q4_0 GGUF | Google's QAT Q4_0 decoder imported as block-32 INT4 `MatMulNBits`; encoders/embeddings/LM head remain floating point |
| Q4_0 + INT8 embeds | Q4_0 decoder plus ONNX RTN W8A16 group-32 asymmetric token/per-layer embeddings and embedding projection |

Standalone W8 RTN and symmetric Google-policy RTN results are intentionally excluded. W8 source packages remain only in the matched INT8-embedding overlay experiment.

### Gemma-4 E2B - package size

| Variant | Bytes | Decimal GB |
|---|---:|---:|
| FP16 | 11,273,429,042 | 11.27 |
| PR515 | 7,611,546,812 | 7.61 |
| PR558 | 6,184,173,889 | 6.18 |
| KQ-all | 7,378,649,735 | 7.38 |
| KQ-dec | 8,624,417,851 | 8.62 |
| Olive GPTQ | 8,624,596,305 | 8.62 |
| Q4_0 GGUF | 8,595,362,354 | 8.60 |
| Q4_0 + INT8 embeds | 6,089,331,756 | 6.09 |
| RTN-all | 7,378,649,735 | 7.38 |
| RTN-dec | 8,624,417,851 | 8.62 |
| QAT W4A16-CT | 8,575,604,690 | 8.58 |

### Gemma-4 E2B - full image and fusion benchmarks

Higher is better. Values are percentages. The top four scores in each row
carry their rank as a subscript (`1` = best); ties share the same rank.
Image/fusion rows use complete benchmark splits; MathVision uses the same
ordered first 500 examples.

| Metric | FP16 | PR515 | PR558 | KQ-all | KQ-dec | RTN-all | RTN-dec | QAT W4A16-CT | Q4_0 GGUF | Q4_0 + INT8 embeds | Olive GPTQ |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| AI2D direct | 63.50<sub>1</sub> | 61.04 | 62.82<sub>2</sub> | 61.08 | 61.14 | 60.78 | 61.08 | 62.05 | 62.60<sub>4</sub> | 62.63<sub>3</sub> | 61.85 |
| ChartQA | 51.56<sub>4</sub> | 50.76 | 50.68 | 51.16 | 50.80 | 51.48 | 51.68<sub>3</sub> | 51.44 | 53.24<sub>2</sub> | 53.40<sub>1</sub> | 50.88 |
| DocVQA ANLS | 74.94<sub>1</sub> | 71.26 | 72.71<sub>4</sub> | 70.53 | 71.34 | 70.17 | 70.88 | 71.51 | 72.74<sub>3</sub> | 72.71<sub>4</sub> | 73.49<sub>2</sub> |
| TextVQA | 61.77<sub>2</sub> | 58.83 | 60.05 | 58.74 | 59.05 | 60.75 | 61.21 | 59.81 | 61.29<sub>4</sub> | 61.36<sub>3</sub> | 62.60<sub>1</sub> |
| OCRBench | 71.40<sub>1</sub> | 69.90<sub>4</sub> | 69.80 | 69.20 | 69.80 | 69.20 | 70.30<sub>2</sub> | 68.30 | 70.10<sub>3</sub> | 69.80 | 68.70 |
| OmniBench | 40.27 | 40.88<sub>4</sub> | 40.00 | 40.53 | 40.97<sub>3</sub> | 41.06<sub>2</sub> | 41.42<sub>1</sub> | 39.73 | 40.62 | 40.53 | 36.55 |
| MathVision standard (first 500) | 35.8<sub>1</sub> | 26.0 | 34.2<sub>2</sub> | 28.6 | 27.2 | 29.4 | 28.4 | 29.6 | 30.2<sub>4</sub> | 31.2<sub>3</sub> | 28.4 |

### Gemma-4 E2B - full audio benchmarks

WER/CER is lower-is-better; BLEU is higher-is-better. The top four scores in
each row carry their rank as a subscript (`1` = best); ties share the same
rank. The original combined FP16 run OOMed, so its replacement values below
are assembled from independent full-split task runs, all of which completed
successfully.

| Metric | FP16 | PR515 | PR558 | KQ-all | KQ-dec | RTN-all | RTN-dec | QAT W4A16-CT | Q4_0 + INT8 embeds | Olive GPTQ |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| FLEURS English WER | 9.81<sub>3</sub> | 10.00 | 9.51<sub>1</sub> | 11.48 | 10.16 | 10.30 | 10.91 | 9.97<sub>4</sub> | 10.46 | 9.75<sub>2</sub> |
| FLEURS Mandarin CER | 30.92<sub>2</sub> | 43.51 | 29.31<sub>1</sub> | 39.83 | 38.85 | 41.15 | 41.77 | 33.15<sub>3</sub> | 38.33 | 37.81<sub>4</sub> |
| FLEURS Cantonese CER | 48.85<sub>1</sub> | 66.10 | 53.09<sub>2</sub> | 75.55 | 64.29<sub>4</sub> | 70.69 | 64.51 | 67.85 | 68.79 | 59.80<sub>3</sub> |
| LibriSpeech clean WER | 5.52<sub>1</sub> | 6.85 | 6.35 | 6.84 | 6.79 | 5.90<sub>3</sub> | 5.54<sub>2</sub> | 6.40 | 6.06 | 5.98<sub>4</sub> |
| LibriSpeech other WER | 12.16<sub>1</sub> | 13.97 | 13.25 | 15.32 | 13.84 | 12.54<sub>3</sub> | 12.49<sub>2</sub> | 14.17 | 13.24 | 12.88<sub>4</sub> |
| CoVoST EN-to-ZH BLEU | 32.54<sub>1</sub> | 30.85 | 32.13<sub>2</sub> | 30.00 | 30.98 | 30.27 | 30.65 | 31.36<sub>4</sub> | 31.53<sub>3</sub> | 29.99 |
| CoVoST ZH-to-EN BLEU | 5.90<sub>1</sub> | 5.19<sub>3</sub> | 5.59<sub>2</sub> | 4.81 | 5.17<sub>4</sub> | 5.06 | 5.15 | 4.81 | 5.07 | 5.01 |

All 12 newly added PR558 runs completed with status 0 and zero empty
generations.

### PR558 recipe reproduction and harness compatibility

You built PR558's original CUDA package by running its five component
recipes (`export`, `text`, `vision`, `audio`, and `embedding`) and then
evaluated the resulting ORT-GenAI package through both the recipe-native
Olive evaluators and `LMMSEvaluator`.

| Full-split measurement | Result |
|---|---:|
| Package size | 6.18 GB |
| Recipe-native AI2D exact match (3,088 examples) | **62.40%** (1,927/3,088) |
| Recipe-native FLEURS English, Olive normalization (647 examples) | **8.92 WER** |
| `LMMSEvaluator` FLEURS English, lmms-eval normalization (647 examples) | **9.51 WER** |

The FLEURS numbers use different prompts and normalizers and therefore are
not expected to be identical. On the full set, applying both normalizers to
both prediction streams gives:

| Prediction source | Olive normalization | lmms-eval normalization |
|---|---:|---:|
| Recipe strict-ASR prompt | **8.92%** | 7.84% |
| Harness generic prompt | 10.21% | **9.51%** |

On matched 64-example compatibility runs, the recipe-native score was 10.33
WER and the harness score was 10.26 WER. The original five-recipe package and
the later unified package produced byte-identical prediction logs on both
AI2D-100 and FLEURS-64, confirming that the unified workflow reproduced the
evaluated PR558 package behavior.

### Gemma-4 E2B Q4_0 GGUF import

Both Q4_0 variants now have complete full-image results and are included in
the primary image/fusion table above. The Q4_0 + INT8-embedding package also
completed the full-audio suite and is included in the primary audio table.
The corresponding FP16-embedding full-audio run reached 28,068/28,399
generations (98.8%) before a 391 MB CUDA BFCArena allocation failed, so it has
no full-audio aggregate.

The deployed package imports Google's official QAT Q4_0 decoder as 275
block-32 INT4 `MatMulNBits` nodes. The LM head, vision/audio encoders, and
embedding component remain floating point.

| Benchmark | Q4_0 GGUF | Q4_0 + INT8 embeds |
|---|---:|---:|
| AI2D direct, full (3,088) | 62.60% | **62.63%** |
| MathVision standard, first 500 | 30.2% | **31.2%** |

Verified image/fusion results on the first 500 examples:

| Benchmark | Q4_0 GGUF | Q4_0 + INT8 embeds | QAT W4A16-CT | FP16 |
|---|---:|---:|---:|---:|
| AI2D direct | 63.4% | 63.4% | **64.6%** | 62.2% |
| ChartQA | 42.6% | **42.8%** | 41.4% | 41.4% |
| DocVQA ANLS | 71.65% | 71.59% | 73.12% | **74.22%** |
| TextVQA | **62.02%** | 61.94% | 59.42% | 61.68% |
| OCRBench¹ | 84.0% | 83.6% | 83.0% | **84.4%** |
| OmniBench | 38.8% | 38.8% | 38.6% | **42.8%** |

> ¹ Renormalized over the evaluated 500 samples; the saved aggregate uses
> OCRBench's full 1,000-example denominator.

Corrected audio results on the first 500 examples:

| Benchmark | Q4_0 GGUF | Q4_0 + INT8 embeds | QAT W4A16-CT | FP16 |
|---|---:|---:|---:|---:|
| LibriSpeech clean WER ↓ | 4.66 | 4.65 | 4.91 | **4.03** |
| LibriSpeech other WER ↓ | 20.62 | 20.67 | 21.78 | **19.50** |
| FLEURS English WER ↓ | 10.46 | 10.42 | 9.76 | **9.74** |
| FLEURS Mandarin WER ↓ | 37.51 | 38.16 | 35.02 | **28.17** |
| FLEURS Cantonese WER ↓ | 66.50 | 67.09 | 73.75 | **48.29** |
| CoVoST EN-to-ZH BLEU ↑ | 27.67 | 27.81 | 28.00 | **29.19** |
| CoVoST ZH-to-EN BLEU ↑ | 4.70 | 4.62 | 3.96 | **5.27** |

These are the corrected v13 audio packages. The original GGUF audio import
was invalid due to swapped light-convolution norms and double application of
Softplus; those superseded values are not reported. Corrected runs had zero
empty generations except Q4_0 GGUF FLEURS Cantonese (7/500).

### Qwen2.5-VL-3B - full image benchmarks

Image rows use complete benchmark splits. MathVision uses the same ordered
first 500 examples.

| Metric | FP16 | KQ-all | KQ-dec | RTN-dec |
|---|---:|---:|---:|---:|
| AI2D direct (full) | **75.81** | 74.38 | 75.13 | 72.09 |
| ChartQA (full) | **80.68** | 80.12 | 79.92 | 80.40 |
| DocVQA (full) | **82.59** | 80.77 | 79.98 | 77.74 |
| TextVQA (full) | **72.69** | 72.63 | 71.67 | 71.94 |
| OCRBench (full) | **76.30** | 75.60 | 75.30 | 72.80 |
| MathVision standard (first 500) | 21.6 | 20.8 | 22.4 | **23.6** |

The post-export RTN-dec package produced empty generations on several tasks. K-Quant retained most FP16 quality without empty outputs. A separate matched INT8-token-embedding matrix completed 14/14 runs; each first-500 metric changed by at most 0.8 percentage points with no consistent direction.
RTN-dec produced 24 empty generations in the MathVision-500 run; the other
three MathVision rows produced none.

### Qwen2.5-VL-3B - INT8 token-embedding overlays

Each row is a matched first-500 source → overlay pair. The overlay changes
only the token-embedding table to INT8 block-32
`GatherBlockQuantized`; every other component is unchanged. Values are
percentages except DocVQA, which is ANLS × 100. OCRBench retains the saved
full-1,000 denominator despite `limit=500`.

| Source package | AI2D direct | ChartQA | DocVQA | TextVQA | OCRBench | MathVision |
|---|---:|---:|---:|---:|---:|---:|
| FP16 | 80.2→80.2 | 66.2→66.2 | 82.99→82.69 | 77.0→77.0 | 40.4→40.4 | 22.6→22.0 |
| GPTQ mixed 64+64, W4A16 G32 asym dec | 79.2→79.0 | 65.0→65.0 | 82.08→82.08 | 76.7→76.7 | 39.6→39.7 | 22.2→21.4 |
| KQ-all W4A16 G32 asym | 78.6→78.6 | 65.0→65.0 | 77.05→77.05 | 77.0→77.2 | 40.7→40.7 | 21.2→20.6 |
| KQ-dec W4A16 G32 asym | 78.8→79.0 | 65.8→66.0 | 79.01→78.81 | 75.1→75.0 | 40.2→40.2 | 22.6→22.2 |
| RTN8-all W8A16 G32 asym | 80.0→80.0 | 66.2→66.2 | 82.96→82.92 | 77.4→77.6 | 40.2→40.2 | 21.6→22.2 |
| RTN8-dec W8A16 G32 asym | 80.8→80.8 | 66.2→66.2 | 82.65→82.65 | 77.1→77.3 | 40.3→40.3 | 22.0→21.8 |
| RTN-dec W4A16 G32 asym | 76.8→76.6 | 65.6→65.4 | 76.63→76.42 | 74.8→74.7 | 39.5→39.4 | 22.8→23.0 |

All 14 source/overlay runs completed successfully. Across these bundles, each
metric changed by at most 0.8 percentage points and there is no consistent
degradation or improvement.

### Qwen3-VL-2B - full image benchmarks

Image rows use complete benchmark splits. MathVision uses the same ordered
first 500 examples.

| Metric | FP16 | KQ-all | KQ-dec | RTN-dec |
|---|---:|---:|---:|---:|
| AI2D direct (full) | **66.71** | 65.38 | 65.19 | 65.19 |
| ChartQA (full) | **61.76** | 59.16 | 59.64 | 58.60 |
| DocVQA ANLS (full) | **76.25** | 71.92 | 72.92 | 73.00 |
| TextVQA (full) | **65.79** | 62.81 | 64.59 | 64.22 |
| OCRBench (full) | **64.50** | 60.30 | 63.50 | 61.20 |
| MathVision standard (first 500) | **11.8** | 8.6 | 8.8 | 8.6 |

Decoder-only K-Quant preserves more quality than all-MatMul K-Quant on Qwen3-VL, particularly OCRBench and TextVQA.

### Main takeaways

- Quantization quality is model-, component-, and benchmark-dependent; no single scheme wins every task.
- Decoder-only/component-aware policies usually protect multimodal quality better than indiscriminate all-MatMul W4.
- Multimodal calibration matters: mixed 64+64 GPTQ calibration reaches 75.58% AI2D on Qwen2.5-VL versus its 75.81% FP16 baseline, while Qwen3-VL still retains a larger GPTQ gap.
- INT8 token embeddings were effectively neutral in the completed Qwen2.5 matched matrix.

These tables are evidence that the evaluator handles official vision, document/OCR, audio, translation, and cross-modal scorers on deployed quantized packages. They should not be read as a claim that every difference is statistically significant.


</details>

jiafatom:
## Assessment: does this overlap with Olive's existing multimodal evaluation?

Olive already evaluates vision/audio/text, so it's worth being explicit about what this PR's `LMMSEvaluator` adds on top of that. Short version: **the evaluator is meaningful and not truly redundant, but it overlaps heavily with existing code and the PR bundles far more than multimodal eval.**

### What Olive already has (home-grown, in `OnnxEvaluator`)
- **Vision**: `_inference_vision` / `_inference_vision_genai` with custom metrics `ExactMatch`, `RelaxedAccuracy` (±5%), `WordSortRatio` — mapped to AI2D/ScienceQA/TextVQA/MathVista/MMMU/ChartQA/OCR.
- **Audio/ASR**: `_inference_text_genai` (whisper), `_inference_text_genai_streaming` (nemotron), `_inference_text_genai_multimodal` (chat-style, e.g. gemma4) with `WER`/`RTFx`.
- **Text**: `LMEvaluator` already wraps the EleutherAI **lm-eval** harness.

### What `LMMSEvaluator` adds beyond that overlap
1. **Protocol-matched official scorers.** The existing metrics are simplified reimplementations (normalized exact-match, ±5%, word-sort). lmms-eval brings the canonical scorers — DocVQA **ANLS**, **OCRBench**, ChartQA relaxed, TextVQA VQA-accuracy, AI2D, WER, **BLEU** — plus tasks we don't have today (**omni_bench, covost2, fleurs, librispeech**). This yields numbers comparable to published leaderboards.
2. **Architectural consistency.** It's the multimodal analogue of the existing `LMEvaluator`→lm-eval pattern (wrap a community harness) rather than per-benchmark custom inference + custom metrics.
3. **Larger task catalog + more model wrappers** without adding a bespoke pre/post-processor per dataset.
4. An `ortgenai_mm` adapter to evaluate deployed ORT-GenAI multimodal packages directly.

### Concerns
- **Duplication / maintenance.** We'd have two multimodal eval stacks. If we adopt lmms-eval, the home-grown vision/audio inference paths and custom metrics become deprecation candidates — this PR doesn't consolidate them.
- **Scope creep.** Most of the ~5k lines are *quantization*, not eval — `Mbq`, `MultiModalMixedPrecision`, `MultimodalActivationRangeCalibration`, `CompositeToOnnxPackage`, K-Quant/RTN node filtering. These are independent of the evaluation question.
- **Dependency risk.** It pins a **fork** (`DelwinKim/lmms-eval` at a specific commit) rather than an upstream release — a reproducibility/maintenance liability.
- **Limitations.** `batch_size=1` single-request only, not registered as an lmms-eval CLI entry point, Qwen3-VL MBQ unsupported.

### Suggestion
The evaluator is worth pursuing — standardized, leaderboard-comparable multimodal scoring is something the current custom code can't reliably match, and it mirrors how text eval already delegates to lm-eval. But for the draft as-is:
1. **Split** the evaluator/adapter from the quantization passes (MBQ, mixed-precision, calibration, packaging).
2. **Resolve overlap** — state a plan to deprecate or clearly delineate the home-grown vision/audio paths vs. `LMMSEvaluator`.
3. **Pin an upstream lmms-eval release** before merge.

jiafatom:
## Follow-up: handling tasks (and metrics) not covered by `LMMSEvaluator`

Two related questions given the overlap with Olive's existing eval: (a) how do we add a benchmark that isn't available through `LMMSEvaluator`, and (b) how do we add a metric that isn't there? Both are valid; here's how each works.

### Missing tasks
Three paths depending on *why* it's missing.

**1. Dataset/scorer isn't in lmms-eval → add it as an lmms-eval task.**
lmms-eval tasks are a **YAML + optional `utils.py`** (`doc_to_text`, `doc_to_visual`, `doc_to_target`, `process_results`, metric/aggregation). Author it and either contribute upstream or keep it in a **local task dir** registered via `TaskManager(include_path=<dir>)`.

⚠️ **Code gap:** `LMMSEvaluator.evaluate` currently calls `lmms_eval.evaluator.simple_evaluate(model=lm, tasks=tasks, ...)` with **no `task_manager`/`include_path`**, so only lmms-eval's *built-in* task registry resolves — custom local tasks can't be used today. Suggest adding a config field (e.g. `include_path`/`task_dir`) and passing `task_manager=TaskManager(include_path=...)` into `simple_evaluate`, mirroring what the text `LMEvaluator` already does with `TaskManager()`.

**2. Bespoke/proprietary task → use Olive's existing custom evaluation path.**
Reuse `AccuracyMetric` + custom pre/post-process + the existing `_inference_vision*` / `_inference_text_genai*` paths, or register a fully custom evaluator via `user_script` (`OliveEvaluatorConfig` imports the user module to register it). A reason to keep the custom stack as an escape hatch rather than delete it.

**3. Capability gap, not a dataset → extend the `ortgenai_mm` adapter.**
If the task exists but fails due to an adapter limitation (`batch_size>1`, multi-round/interleaved generation — `generate_until_multi_round` is a stub, a new `model_type`/processor shape, or a new media type), fix it in `olive/evaluator/lmms_ort.py` (`generate_until` / `loglikelihood` / `_build_prompt_for_request`).

### Missing metrics
Valid question, and mostly a sub-case of the above — in lmms-eval, **metrics live with the task**, not globally:

- **Per-task metric (most common):** compute it directly in the task's `process_results` in `utils.py` and return it in the results dict; declare its `metric_list` + `aggregation` + `higher_is_better` in the task YAML. No global registration needed.
- **Reusable metric across tasks:** register it in lmms-eval's metric registry (`lmms_eval/api/metrics.py`: `register_metric` / `register_aggregation` / `register_higher_is_better`), then reference it from any task YAML.
- **Olive side needs no change:** `LMMSEvaluator` converts results by iterating the numeric entries in `results["results"][task]`, so **any metric lmms-eval reports numerically flows into Olive's `MetricResult` automatically**. (One caveat: Olive assumes `higher_is_better=True` in that conversion — error-rate style metrics like WER/CER would be surfaced with the wrong direction and should honor lmms-eval's `higher_is_better` map instead.)

**Recommendation:** for anything with a published protocol, prefer adding it as an lmms-eval task/metric and close the `include_path` gap so custom tasks/metrics are first-class; reserve Olive's custom evaluator for truly internal benchmarks.

jiafatom:
We also need to compare the data generated here (from hf torch model) and the numbers that are reported by the model card. Then we can make sure our evaluation is correct.

DelwinKim:
@jiafatom 
"We also need to compare the data generated here (from hf torch model) and the numbers that are reported by the model card. Then we can make sure our evaluation is correct."

Yes, I agree with that. I've done those eval with the hf torch model and the results are within a reasonable range (I will post these numbers). I would note that I found comparing numbers reported by the model card to be "difficult" and in some cases "pointless" since a lot of these labs don't publish their evaluation methodology and they do "additional" stuff that inflates the performance of the actual model. This is pretty evident when you run "more difficult" benchmarks like mathvision or mmmu pro. With the simpler benchmarks, the replicated results are in the expected range.

jiafatom:
> @jiafatom "We also need to compare the data generated here (from hf torch model) and the numbers that are reported by the model card. Then we can make sure our evaluation is correct."
> 
> Yes, I agree with that. I've done those eval with the hf torch model and the results are within a reasonable range (I will post these numbers). I would note that I found comparing numbers reported by the model card to be "difficult" and in some cases "pointless" since a lot of these labs don't publish their evaluation methodology and they do "additional" stuff that inflates the performance of the actual model. This is pretty evident when you run "more difficult" benchmarks like mathvision or mmmu pro. With the simpler benchmarks, the replicated results are in the expected range.

I understand this. Since we want to introduce this evaluation metrics, so we need ensure the evaluation metrics give the correct result. If one or two model card inflates their models, I believe most of the model card - the result can be reproducible.  There must be some methodology behind the metrics of those "difficult" benchmarks, and we need understand what the trick is and to see if this makes sense or not, then to apply those "additional" stuff if it is reasonable.

DelwinKim:
> ## Follow-up: handling tasks (and metrics) not covered by `LMMSEvaluator`
> Two related questions given the overlap with Olive's existing eval: (a) how do we add a benchmark that isn't available through `LMMSEvaluator`, and (b) how do we add a metric that isn't there? Both are valid; here's how each works.
> 
> ### Missing tasks
> Three paths depending on _why_ it's missing.
> 
> **1. Dataset/scorer isn't in lmms-eval → add it as an lmms-eval task.** lmms-eval tasks are a **YAML + optional `utils.py`** (`doc_to_text`, `doc_to_visual`, `doc_to_target`, `process_results`, metric/aggregation). Author it and either contribute upstream or keep it in a **local task dir** registered via `TaskManager(include_path=<dir>)`.
> 
> ⚠️ **Code gap:** `LMMSEvaluator.evaluate` currently calls `lmms_eval.evaluator.simple_evaluate(model=lm, tasks=tasks, ...)` with **no `task_manager`/`include_path`**, so only lmms-eval's _built-in_ task registry resolves — custom local tasks can't be used today. Suggest adding a config field (e.g. `include_path`/`task_dir`) and passing `task_manager=TaskManager(include_path=...)` into `simple_evaluate`, mirroring what the text `LMEvaluator` already does with `TaskManager()`.
> 
> **2. Bespoke/proprietary task → use Olive's existing custom evaluation path.** Reuse `AccuracyMetric` + custom pre/post-process + the existing `_inference_vision*` / `_inference_text_genai*` paths, or register a fully custom evaluator via `user_script` (`OliveEvaluatorConfig` imports the user module to register it). A reason to keep the custom stack as an escape hatch rather than delete it.
> 
> **3. Capability gap, not a dataset → extend the `ortgenai_mm` adapter.** If the task exists but fails due to an adapter limitation (`batch_size>1`, multi-round/interleaved generation — `generate_until_multi_round` is a stub, a new `model_type`/processor shape, or a new media type), fix it in `olive/evaluator/lmms_ort.py` (`generate_until` / `loglikelihood` / `_build_prompt_for_request`).
> 
> ### Missing metrics
> Valid question, and mostly a sub-case of the above — in lmms-eval, **metrics live with the task**, not globally:
> 
> * **Per-task metric (most common):** compute it directly in the task's `process_results` in `utils.py` and return it in the results dict; declare its `metric_list` + `aggregation` + `higher_is_better` in the task YAML. No global registration needed.
> * **Reusable metric across tasks:** register it in lmms-eval's metric registry (`lmms_eval/api/metrics.py`: `register_metric` / `register_aggregation` / `register_higher_is_better`), then reference it from any task YAML.
> * **Olive side needs no change:** `LMMSEvaluator` converts results by iterating the numeric entries in `results["results"][task]`, so **any metric lmms-eval reports numerically flows into Olive's `MetricResult` automatically**. (One caveat: Olive assumes `higher_is_better=True` in that conversion — error-rate style metrics like WER/CER would be surfaced with the wrong direction and should honor lmms-eval's `higher_is_better` map instead.)
> 
> **Recommendation:** for anything with a published protocol, prefer adding it as an lmms-eval task/metric and close the `include_path` gap so custom tasks/metrics are first-class; reserve Olive's custom evaluator for truly internal benchmarks.

Thanks for the suggestion, implemented these solutions to these 2 Olive "gaps"

include_path now accepts one task directory or a list, validates each path, and passes a model-aware TaskManager(include_path=..., model_name=lm) for both ONNX and HF dispatch. When include_path is absent, Olive does not construct a manager, preserving lmms-eval's normal default-manager behavior. End-to-end resolution was tested with an external task directory and a real ORT-GenAI package.

Metric direction now comes from lmms-eval's per-task higher_is_better map rather than defaulting every metric to True; the regression test verifies that WER is lower-is-better. Numeric metrics returned by custom task process_results continue to flow into Olive without a new Olive metric registration.

One limitation remains important: include_path can add collision-free task names, but it cannot override a built-in task with the same name. Corrections to existing lmms-eval tasks still need to go upstream rather than being hidden in a local task directory.

For a benchmark that is absent from lmms-eval, I would follow the three paths you outlined:

For a public benchmark with a published protocol, add a versioned lmms-eval YAML/utils.py task, use include_path while it is under development, and contribute it upstream where possible.
For a proprietary or highly application-specific benchmark, retain Olive's existing custom evaluator/metric path as the escape hatch rather than forcing it into lmms-eval.
If the task already exists and the failure is an inference capability issue, extend ortgenai_mm. This PR currently supports single-request image and audio generation plus compatible text loglikelihood; it does not currently support batched generation, multi-round/interleaved generation, every processor signature, or additional media types.
I also agree with the metric split. A task-specific metric belongs in that task's process_results and YAML aggregation declaration; a genuinely reusable metric belongs in lmms-eval's metric registry. In either case, numeric results and their direction now pass through LMMSEvaluator into Olive's MetricResult.

So the intended boundary is: use lmms-eval for public, protocol-defined benchmarks; use Olive's custom evaluation path for internal/proprietary workloads; and change the adapter only when the benchmark exists but deployed ORT-GenAI inference lacks a required capability.

PR #2532, opened 2026-06-19
[AMD] Update QuarkQuantization pass for Quark 0.12

thpereir:
## Describe your changes

Updates `QuarkQuantization` to work with AMD Quark 0.12, which introduced two breaking API changes:

**ONNX path — calibration reader fix:**
Quark 0.12 rejects calibration inputs that are not model inputs. Olive was forwarding the full dataset batch (including label columns like `class`) to the reader, causing `INVALID_ARGUMENT: Invalid input name` during calibration — the workflow would complete with exit code 0 but produce no output model. Fixed by passing `model_path`/`io_config` to `create_calibration_dataloader()` so non-input columns are filtered out, matching the pattern already used by `OnnxQuantization`.

**Torch path — API update:**
- `revert_model_patching` was removed in Quark 0.12 (patching is now reverted internally during export). Removed the import and call.
- `prepare_for_moe_quant` is deprecated and raises on `transformers>=5`. Replaced with `preprocess_for_quantization`.

**Version gating:**
Both ONNX and Torch paths now gate on `amd-quark>=0.12.0` with a clear `ValueError` so users get an actionable error message instead of a cryptic `ImportError`.

  The `QuarkQuantization` pass now requires `amd-quark>=0.12.0`. Users on 0.11.x will receive a clear `ValueError` with an upgrade prompt. The ONNX calibration path and Torch path are updated to work correctly with the 0.12 API.

## (Optional) Issue link

### DIFF
--- a/olive/olive_config.json
+++ b/olive/olive_config.json
@@ -605,7 +605,7 @@
             "supported_algorithms": [ "awq" ],
             "supported_quantization_encodings": [  ],
             "run_on_target": true,
-            "extra_dependencies": [ "amd-quark" ]
+            "extra_dependencies": [ "amd-quark>=0.12.0" ]
         },
         "QuarkQuantizationVitisAI": {
             "module_path": "olive.passes.quark_vitisai.quark_quantization_vitisai.QuarkQuantizationVitisAI",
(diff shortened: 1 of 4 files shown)


xiaoyu-work:
We can also add a version pin here: https://github.com/microsoft/Olive/blob/main/olive/olive_config.json#L601-L619

thpereir:
> We can also add a version pin here: https://github.com/microsoft/Olive/blob/main/olive/olive_config.json#L601-L619

Good call, updated

xiaoyu-work (approved):

PR #2534, opened 2026-06-22
expose latencies with the speedup in OnnxDiscrepancyCheck

xadupre:
## Describe your changes
Speedup is not enough as a metric, latency must be added to understand which side is changing.


xadupre:
@copilot extend the existing unit test to cover the changes

PR #2538, opened 2026-06-22
Preserve graph output names in weight-only quantization passes

jiafatom:
## Problem

The RTN and HQQ weight-only quantization passes unconditionally append a `_Q{bits}` suffix to a quantized MatMul/Gather node's output name. When that node produces a **graph output**, the graph output itself gets renamed, silently breaking external consumers that reference the original name.

Concretely, ONNX Runtime GenAI's `genai_config.json` references model output names (e.g. a speech encoder's `audio_features`). After RTN int8 quantization, the terminal projector MatMul's output became `audio_features_Q8`, so GenAI raised:

```
Model output was not found: audio_features
```

on every inference, and all predictions came back empty. (Observed on a Gemma-3n multimodal export whose audio encoder's terminal projector MatMul is a graph output.)

## Fix

Only append the suffix when the node's output is **not** a graph output. Internal tensors are still renamed (their consumers are rewired by `replace_nodes_and_values`). This matches the guard already present in the k-quant pass (`kquant_quantization.py`).

- `rtn_quantization.py`: add `_rename_output_unless_graph_output` helper, used by both `_quantize_matmul` and `_quantize_gather`.
- `hqq_quantization.py`: guard the `_Q4` rename the same way.

## Tests

Added regression tests to the RTN, HQQ, and k-quant suites. Each builds a 2-MatMul model where the terminal MatMul feeds a named graph output, quantizes, and asserts:
- the graph output name is preserved (no suffix), and
- the internal MatMul output still gets the quant suffix.

All three suites pass (21 tests).

### DIFF
--- a/olive/passes/onnx/hqq_quantization.py
+++ b/olive/passes/onnx/hqq_quantization.py
@@ -181,7 +181,10 @@ class OnnxHqqQuantization(Pass):
         if accuracy_level > 0:
             kwargs["accuracy_level"] = accuracy_level
 
-        node.outputs[0].name = node.outputs[0].name + "_Q4"
+        # Only rename intermediate outputs; preserve graph output names so
+        # downstream consumers (e.g. genai_config.json) keep working.
+        if node.outputs[0] not in node.graph.outputs:
+            node.outputs[0].name = node.outputs[0].name + "_Q4"
 
         return ir.node(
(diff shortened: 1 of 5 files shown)


jiafatom:
Superseded by #2539, which uses a branch on microsoft/Olive directly (for full CI) with identical content.

PR #2569, opened 2026-07-14
Allow overriding past_present_share_buffer in LMEvaluator ortgenai backend

jiafatom:
## Describe your changes

The `ortgenai` lm-eval backend (`LMEvalORTGenAIEvaluator`) always derived `past_present_share_buffer` from the exported `genai_config.json`, with **no way to override it** from a recipe/evaluator config. Some models (e.g. **Gemma 4**) export with shared past/present buffers *enabled* but require it *disabled* for correct KV-cache handling during evaluation. Additionally, `LMEvaluator` gave recipe authors no way to enable a model's chat template or few-shot prompting, so instruction-tuned models were evaluated 0-shot on raw prompts and scored near-random. Today these gaps force recipe authors to bypass `LMEvaluator` and write a standalone `eval.py` that calls `lm_eval.simple_evaluate` directly — the declarative Olive config path simply produces wrong results for these models.

This PR makes these settings overridable end-to-end and adds optional sample logging:

- **`LMEvalORTGenAIEvaluator`** now accepts an explicit `past_present_share_buffer` argument that takes precedence over the exported value. The resolution logic is extracted into a small `_resolve_past_present_share_buffer(override, genai_config)` helper (explicit override wins; otherwise fall back to the exported value, defaulting to `False`).
- **`LMEvaluator`** gains a `model_args` dict that is forwarded verbatim to the lm-eval model backend constructor, so genai-specific options can be set from the evaluator config. User-provided `model_args` override the Olive-derived defaults (`batch_size`, `max_length`, `ep`, ...). Default is `{}`, so existing configs are unaffected.
- **`LMEvaluator` prompting controls**: new `apply_chat_template`, `system_instruction`, `fewshot_as_multiturn` and `num_fewshot` options, forwarded to `lm_eval.simple_evaluate`. Instruction-tuned models score near-random on multiple-choice tasks when evaluated 0-shot on raw prompts; wrapping requests in the model's chat template and matching the card's few-shot protocol fixes this. Empirically, on a Gemma-4 E2B export a German MMMLU subset went from ~0.16–0.24 (0-shot, no template) to ~0.52–0.68 (5-shot + chat template). Defaults preserve lm-eval's legacy behavior (no chat template, no system prompt, task-default few-shot), so existing configs are unaffected.
- **`LMEvaluator` sample logging**: a new `sample_log_num` (and optional `sample_log_dir`) option. When `sample_log_num > 0`, lm-eval's `log_samples` is enabled and the first N per-task predictions — question, options, resolved prediction (argmax of choice loglikelihoods to letter + text, or the raw generation), target, and `acc` — are written to `<task>_samples.jsonl`. Rather than duplicating the writer, `OliveEvaluator.save_sample_log` is decoupled to take a `(name, sample_log_dir)` pair instead of a `Metric`, so `LMEvaluator` reuses the exact same JSONL writer as the accuracy evaluators; a small `_lmeval_samples_to_output` helper maps lm-eval sample records into an `OliveModelOutput` (carrying question/options/prediction_index/acc via `extras`). This makes it possible to inspect why a task scores low (e.g. distinguishing a genuinely weak/quantized model from a broken eval). Default is `0`, so existing configs are unaffected.

With this, such models can be evaluated with a standard `LMEvaluator` config instead of a bespoke script:

```json
{
    "evaluators": {
        "mmlu": {
            "type": "LMEvaluator",
            "tasks": ["mmlu"],
            "model_class": "ortgenai",
            "num_fewshot": 5,
            "apply_chat_template": true,
            "fewshot_as_multiturn": true,
            "sample_log_num": 100,
            "model_args": { "past_present_share_buffer": false }
        }
    }
}
```

**Release note:** `LMEvaluator` now accepts a `model_args` dict forwarded to the lm-eval model backend; `apply_chat_template`, `system_instruction`, `fewshot_as_multiturn` and `num_fewshot` prompting options forwarded to `lm_eval.simple_evaluate` (so instruction-tuned models can be evaluated with their chat template and the card's few-shot protocol); and a `sample_log_num`/`sample_log_dir` option to log the first N per-task predictions vs. targets to a JSONL file for debugging (reusing the accuracy evaluators' `save_sample_log` writer). The `ortgenai` backend accepts an explicit `past_present_share_buffer` override (previously only read from `genai_config.json`).

## (Optional) Issue link

### DIFF
--- a/olive/evaluator/lmeval_ort.py
+++ b/olive/evaluator/lmeval_ort.py
@@ -545,6 +545,17 @@ class LMEvalORTGenAIEvaluator(LMEvalOnnxBase):
         normalized_ep = str(ep).lower().replace("executionprovider", "")
         return normalized_ep, cls._ORT_GENAI_PROVIDER_NAMES.get(normalized_ep, normalized_ep)
 
+    @staticmethod
+    def _resolve_past_present_share_buffer(override: bool | None, genai_config: dict) -> bool:
+        """Resolve the ``past_present_share_buffer`` search option.
+
+        An explicit ``override`` takes precedence; otherwise fall back to the exported
+        ``genai_config.json`` value (defaulting to ``False`` when absent).
+        """
+        if override is not None:
+            return override
+        return genai_config.get("search", {}).get("past_present_share_buffer", False)
+
     def __init__(
         self,
         pretrained: str,
@@ -553,6 +564,7 @@ class LMEvalORTGenAIEvaluator(LMEvalOnnxBase):
         ep: str = "follow_config",
         ep_options: dict | None = None,
         device: str = "cpu",
+        past_present_share_buffer: bool | None = None,
         **kwargs,
     ):
         """Initialize the evaluator.
@@ -563,6 +575,9 @@ class LMEvalORTGenAIEvaluator(LMEvalOnnxBase):
         :param ep: The execution provider to use. "follow_config" will use the provider specified in the genai_config file
         :param ep_options: The options to use for the execution provider. Only applicable if ep is not "follow_config"
         :param device: The device to run log likelihood calculations on
+        :param past_present_share_buffer: Override the exported ``search.past_present_share_buffer`` setting. When
+            ``None`` (default) the value from ``genai_config.json`` is used. Some models (e.g. Gemma 4) export with
+            shared buffers enabled but require it disabled for correct KV-cache handling during evaluation.
         """
         if og is None:
             raise ImportError("onnxruntime-genai is not installed.")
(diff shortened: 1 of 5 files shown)


jiafatom:
Thanks for the review! Addressed all three in 02d78188:

1. **`simple_evaluate` batch size inconsistency** — now uses `effective_batch_size = model_init_args["batch_size"]` so lm-eval batching matches the backend after any `model_args` override.
2. **Non-dict `model_args`** — validated in `__init__` with an early `ValueError("model_args must be a dict, ...")`.
3. **Test coverage** — the forwarding test now also asserts `simple_evaluate` receives the overridden batch size, plus a new test that a non-dict `model_args` is rejected.

xiaoyu-work (approved):

PR #2577, opened 2026-07-20
Add mobius inspect components api

xiaoyu-work:
## Describe your changes
Add mobius inspect components api

### DIFF
--- a/olive/common/mobius_utils.py
+++ b/olive/common/mobius_utils.py
@@ -1,0 +1,120 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+# --------------------------------------------------------------------------
+"""Helpers for obtaining a model component plan from mobius.
+
+Mobius owns the per-architecture knowledge of which components a model exposes (e.g. a VLM's
+``decoder`` / ``vision_encoder`` / ``embedding``), how each maps back to a submodule, and the role
+of each component. This adapter lets Olive consume that plan without re-implementing architecture-specific logic.
+
+``mobius-ai`` is imported lazily so Olive keeps working when it is not installed; only the code paths
+that actually need a component plan for a Hugging Face model require it.
+"""
+
+import logging
+from dataclasses import dataclass, field
+from typing import Optional
+
+logger = logging.getLogger(__name__)
+
+
+def _as_path_list(value: object) -> list[str]:
+    """Normalize a source-path value (tuple/list/str/None) into a list of paths."""
+    if value is None:
+        return []
+    if isinstance(value, str):
+        return [value] if value else []
+    return [str(p) for p in value if p]
+
+
+@dataclass
+class ComponentInfo:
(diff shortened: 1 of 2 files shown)


olive/common/mobius_utils.py
```diff
@@ -0,0 +1,120 @@
...
+        mobius releases, the legacy ``kind``/``source_path`` names are accepted as fallbacks.
+        """
+        if isinstance(data, cls):
+            return data
+        if isinstance(data, dict):
+            source = data.get("source") or {}
+            source_paths = data.get("source_paths")
+            if source_paths is None:
+                source_paths = data.get("source_path") or source.get("path")
+            recognized = {"name", "role", "kind", "source", "source_path", "source_paths"}
```
shaahji, line 68: Missing `path` in the set. If it exists, it is already accounted for on the previous line.

olive/common/mobius_utils.py
```diff
@@ -0,0 +1,120 @@
...
+            return cls(
+                name=data["name"],
+                role=data.get("role") or data.get("kind"),
+                source_paths=_as_path_list(source_paths),
+                metadata={k: v for k, v in data.items() if k not in recognized},
+            )
+        source_paths = getattr(data, "source_paths", None)
+        if source_paths is None:
+            source_paths = getattr(data, "source_path", None)
```
shaahji, line 77: Not looking for `path` here?

olive/common/mobius_utils.py
```diff
@@ -0,0 +1,120 @@
...
+            name=data.name,
+            role=getattr(data, "role", None) or getattr(data, "kind", None),
+            source_paths=_as_path_list(source_paths),
+        )
+
+
+def inspect_components(
+    model_name_or_path: str,
+    task: Optional[str] = None,
+    trust_remote_code: bool = False,
```
shaahji: Default should be `None`. This was a reported S360 issue; it recommends not defaulting to either true or false and let the underlying library decide what the _default_ should be.

shaahji (changes requested):

olive/common/mobius_utils.py
xiaoyu-work: `path` is nested under `source`, while recognized only filters top level keys, since `source` is already included, `path` will not be added to metadata

olive/common/mobius_utils.py
xiaoyu-work: Mobius objects expose `source_paths` and `source_path` here https://github.com/onnxruntime/mobius/blob/b7062d05d6c0ab671a5560e724340bbe6d4562cc/src/mobius/_inspect.py#L25-L50

olive/common/mobius_utils.py
xiaoyu-work: Done!

olive/common/mobius_utils.py
shaahji: Ahh missed that detail.

shaahji (approved):

PR #2578, opened 2026-07-20
Olive QAIRT model provenance initial

qti-kromero:
## Describe your changes

Adds model provenance tracking to QAIRT Olive passes. After each pass completes, an `olive_run_metadata.json` sidecar is written to the output directory recording:

- **`recipe_metadata`** — version and validated environment (read from `info.yml` in the recipe directory)
- **`ran_with`** — actual runtime versions of Python, `qairt-sdk`, and `qairt-dev`
- **`validation_delta`** — per-field match/mismatch between `ran_with` and `validated_with`

The metadata is accumulated across the pass chain: `QairtPreparation` seeds `recipe_metadata` from `info.yml` on its first run; `QairtGenAIBuilder` and `QairtEncapsulation` carry it forward via `model_attributes.additional_files`.

**File structure changes:**
- `base.py` → `pass_.py` (`pass` is a Python keyword)
- `run_metadata.py` → `utils/metadata.py`; `utils.py` → `utils/common.py` (now a package)

**`info.yml` schema** (olive-recipes companion change): `version` and `validated_with` are top-level fields in each QAIRT backend's `info.yml`, not embedded in the RunConfig JSON. This keeps recipe configs compatible with all Olive versions.

  QAIRT passes now write `olive_run_metadata.json` to the output directory after each pass, capturing recipe version, validated environment, and actual runtime versions with a per-field match delta.


olive/passes/qairt/base.py
```diff
@@ -0,0 +1,55 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Qualcomm Technologies, Inc. and/or its subsidiaries.
```
qti-kromero, line 2: let's rename this to pass.py to align with the naming of the class inside - already namespaced by qairt package

olive/passes/qairt/utils/metadata.py
```diff
@@ -0,0 +1,133 @@
...
+
+def _get_ran_with() -> dict[str, str]:
+    """Capture runtime environment versions for the current pass invocation."""
+    versions: dict[str, str] = {"python": sys.version.split()[0]}
+
+    try:
+        from qairt import __sdk_version__ as sdk_version
+
+        versions["qairt_sdk"] = sdk_version
+    except (ImportError, AttributeError):
```
qti-kromero, line 37: can we just combine these two and list the full unioned tuple for the except? we take the same pass action regardless

olive/passes/qairt/run_metadata.py
```diff
@@ -0,0 +1,133 @@
...
+
+    return versions
+
+
+def load_metadata(model_attributes: Optional[dict[str, Any]]) -> dict[str, Any]:
+    """Load existing olive_run_metadata.json from a model's additional_files, if present."""
+    if not model_attributes:
+        return {}
+
+    for filepath in model_attributes.get("additional_files") or []:
```
qti-kromero: can we just use a default argument for .get() of []?

olive/workflows/run/config.py
```diff
@@ -146,6 +146,9 @@ class RunConfig(NestedConfig):
         ),
     )
     passes: dict[str, list[RunPassConfig]] = Field(default_factory=dict, description="Pass configurations.")
+    vendor: Optional[dict[str, Any]] = Field(
```
qti-kromero: I don't think we need this any more

olive/passes/qairt/utils/metadata.py
```diff
@@ -0,0 +1,133 @@
...
+import sys
+from pathlib import Path
+from typing import Any, Optional
+
+logger = logging.getLogger(__name__)
+
+METADATA_FILENAME = "olive_run_metadata.json"
+
+
+def _get_ran_with() -> dict[str, str]:
```
qti-kromero: let's put this in a new folder "utils" under a new file "metadata.py"

PR #2579, opened 2026-07-21
[AMD] Add algo-config option in Quark

poganesh:
- Adds an algo_configs parameter to the QuarkQuantization pass (torch path), letting a recipe pass a custom quantization algorithm config (e.g. AWQ scaling_layers / model_decoder_layers) through to LLMTemplate.get_config(algo_configs=...).

- This is needed for architectures that Quark 0.12.post1 does not ship a built-in algorithm config for, e.g. qwen3, which is absent from Quark's AWQ_MAP and otherwise raises NotImplementedError: No built-in awq configuration is available for the 'qwen3' architecture. With this change, the recipe can supply the config (including scaling_layers: [] for auto-detection), unblocking AWQ quantization for such models.

- The parameter is optional and defaults to None; existing recipes and architectures with built-in configs are unaffected.

### DIFF
--- a/olive/passes/quark_quantizer/quark_quantization.py
+++ b/olive/passes/quark_quantizer/quark_quantization.py
@@ -127,6 +127,11 @@ class QuarkQuantization(Pass):
                 default_value=True,
                 description="[Torch] Trust remote code from HuggingFace",
             ),
+            "algo_configs": PassConfigParam(
+                type_=Optional[dict],
+                default_value=None,
+                description="[Torch] Custom algorithm configs by name, e.g. {'awq': {'scaling_layers': [...], 'model_decoder_layers': 'model.layers'}}. Used for architectures Quark lacks a built-in config for.",
+            ),
             # ── ONNX-specific configs ────────────────────────────────────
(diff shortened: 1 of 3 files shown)


poganesh:
@thpereir, @thiagocrepaldi  - Please help review the quark changes.

thiagocrepaldi (approved):
Code itself looks OK. Are you planning to add tests? any extra documentation planned, with maybe a real usage example?

poganesh:
> Code itself looks OK. Are you planning to add tests? any extra documentation planned, with maybe a real usage example?

@thiagocrepaldi 

- The new Quark 0.12 torch path had no existing tests (only the ONNX path in test_quark_onnx_quantization.py), so I added a new test_quark_torch_quantization.py covering the algo_configs handling introduced in the PR. 

- The primary use case is olive-recipes (e.g. qwen3, which Quark has no built-in AWQ config for). detailed config semantics are deferred to the Quark docs, in the olive-recipes README.

xiaoyu-work (approved):

PR #2608, opened 2026-08-05
Use the full calibration split and surface sample shortfalls

titaiwangms:
## Describe your changes

`get_calibration_dataset` / `get_calibration_data_config` in `olive/passes/pytorch/train_utils.py` defaulted the dataset split to `train[:1000]`. That slice is far too small for the default `max_samples=128`, so calibration silently ran on a fraction of the requested data. This changes the default to the full `train` split and makes any remaining shortfall visible.

### The shortfall

Measured locally against the cached wikitext-2-raw-v1 (gpt2 tokenizer, `max_seq_len=2048`):

| split | tokens | blocks available | blocks requested |
|---|---|---|---|
| `train[:1000]` | 61,816 | **~30** | 128 |
| `train` | 2,391,884 | ~1,167 | 128 |

Of the 1000 sliced rows, 353 are blank or section headers and get filtered out, leaving 647 usable rows. The result was that a pass asking for 128 calibration samples received roughly 30, with no indication that anything was wrong.

128 blocks x 2048 tokens is ~10-11% of the full split, so the new default is not "read everything" — it is "have enough to satisfy `max_samples`".

### Affected passes

Four passes take the default and are therefore affected:

- `Gptq` (via `quant_utils.py`)
- `AutoGPTQ`
- `GPTQModel`
- `SelectiveMixedPrecision`

Explicitly **not** affected: `Rotate` already passes `split="train"`, `AutoClip` uses `mit-han-lab/pile-val-backup` `validation[:1000]` explicitly, and `Rtn` needs no calibration data at all.

`test/utils.py` keeps its own `train[:1000]` slice — that one is a deliberate test-speed choice with `max_samples=1` and is left alone (a comment now says so).

### Accuracy impact

Qwen2.5-0.5B / 1.5B / 3B, 4-bit GPTQ, wikitext-2 test perplexity and pile-val perplexity, at 32 / 128 / 512 calibration blocks. 11 of the 12 comparisons improve over the 32-block baseline.

Caveats that should be read alongside that number:

- The absolute deltas are small.
- It is **not** monotone in block count everywhere. 1.5B on pile-val goes 0% -> **-2.7%** at 128 blocks -> +6.9% at 512, i.e. 128 blocks is worse than 32 for that one series.
- All of the evidence is from **dense** models. No MoE model was measured.
- 128 blocks costs roughly 25% more calibration wall-clock than 32; 512 costs about 2.3x. The default `max_samples` is unchanged, so this PR does not change wall-clock by itself — it only changes whether `max_samples` is actually met.

### This is a reproducibility break

Any workflow that relied on the old default will now calibrate on more data and produce different (bit-wise non-identical) quantized weights. Pinning `split="train[:1000]"` in the data config restores the previous behaviour exactly.

### Network footprint is unchanged

`datasets` downloads the full parquet shard regardless of a `[:N]` slice — the slicing happens in memory after the download. Moving from `train[:1000]` to `train` therefore does not download more data.

## Making the shortfall observable

Every tokenization strategy in `text_generation.py` can stop early and deliver fewer than `max_samples` without saying anything. A single check after all strategy branches now warns when that happens, with a strategy-specific explanation, because the cause and the useful remedy differ per strategy:

- `join` / `join-sliding-window`: the corpus needs `max_seq_len + (max_samples - 1) * step` tokens, where `step` is `max_seq_len` for `join` and `stride` for `join-sliding-window`. Raising `stride` is an effective remedy for the latter and not for the former.
- `join-random`: start positions are random and may overlap, so the corpus only ever needs `max_seq_len` tokens and lowering `max_samples` does not help. The real cause is `random_retries` exhaustion.
- `line-by-line` / `line-by-line-random`: at most one sample per row.

## Two latent bugs found along the way

Both were pre-existing and are fixed here because the new warning interacts directly with them.

**1. Off-by-one in the join strategies.** The block loop was `range(0, len(joined_input_ids) - args.max_seq_len, step)`. `range` is exclusive at the end, so a block starting exactly at `len - max_seq_len` was never emitted even though it is a full-length block. A corpus that was an exact fit for `max_samples` always came up one block short:

| tokens | required | delivered (before) | delivered (after) |
|---|---|---|---|
| 8 | 8 | 0 | 1 |
| 16 | 16 | 1 | 2 |
| 32 | 32 | 3 | 4 |
| 72 | 64 | 7 | 8 |

The new maximum `begin_loc` is `len - max_seq_len`, so `end_loc <= len` and every emitted sequence is still exactly `max_seq_len` long. Without the fix the new warning would have fired as a false alarm on exact-fit corpora.

**2. `drop_short_sequences` had no effect in `line-by-line-random`.** The resampling loop reassigns `encodings` on every attempt and was guarded by `if not encodings: continue`. Once the retries are exhausted, `encodings` still holds the last *rejected* (too-short) row, and a non-empty `BatchEncoding` is truthy, so the rejected row was appended anyway. `drop_short_sequences=True` therefore emitted short sequences instead of dropping them — silently producing wrong-length calibration data rather than less of it. The length is now re-checked explicitly, matching what `line-by-line` already did via an explicit `filter`.

Blast radius of bug 2 is essentially zero: `drop_short_sequences` defaults to `False` and is mutually exclusive with `pad_to_max_len` (which defaults to `True`), so triggering it required explicitly setting both. No recipe in the repository does.

### Known follow-up, not fixed here

In `line-by-line-random`, `cache[i] = encodings` means resampling an already-seen index returns the cached (short) row while still incrementing `resamples`. On a small dataset the retries burn on repeated rows, so far fewer distinct rows are explored than `random_retries` suggests. This is an efficiency issue rather than a correctness one now that bug 2 is fixed — the worst case is a shortfall, which is now reported.

**Release note:** Calibration data for the `Gptq`, `AutoGPTQ`, `GPTQModel` and `SelectiveMixedPrecision` passes now defaults to the full `train` split instead of `train[:1000]`, which was too small to satisfy the default `max_samples=128`. Quantized weights produced by these passes will differ from previous releases; set `split` explicitly in the data config to restore the old behaviour. Data preprocessing now warns when fewer samples than `max_samples` could be produced, and two pre-existing bugs were fixed: the `join` strategies dropped the last full block on exact-fit corpora, and `drop_short_sequences` was ignored by `line-by-line-random`.

### Tests

`test/data_container/test_text_generation.py` (14 tests) and `test/passes/pytorch/test_train_utils.py` (3 tests), both network-free. Includes a regression test for each of the two bugs above — the off-by-one test asserts an exact-fit corpus delivers all `max_samples`, and the `drop_short_sequences` test was verified to fail (4 short samples emitted instead of 0) with the fix reverted.

```
python -m pytest test/data_container test/passes/pytorch/test_train_utils.py -q
-> 1 failed, 94 passed, 1 skipped
```

The one failure is `test_dataloader.py::test_llm_augmented_dataloader[True]` (`TypeError: check_extra_options() takes 2 positional arguments but 7 were given`), which fails identically on `main` and is unrelated to this change.

### DIFF
--- a/olive/data/component/text_generation.py
+++ b/olive/data/component/text_generation.py
@@ -3,6 +3,7 @@
 # Licensed under the MIT License.
 # --------------------------------------------------------------------------
 
+import logging
 from pathlib import Path
 from random import Random
 from typing import Callable, Optional, Union
@@ -15,6 +16,8 @@ from olive.common.user_module_loader import UserModuleLoader
 from olive.common.utils import StrEnumBase
 from olive.data.component.dataset import ClassificationDataset
 from olive.data.constants import IGNORE_INDEX
+
+logger = logging.getLogger(__name__)
 
 
 class TextGenStrategy(StrEnumBase):
@@ -253,8 +256,10 @@ def text_gen_pre_process(dataset, tokenizer, all_kwargs):
                     joined_input_ids += input_ids + joiner_tokens
 
                 end_loc = 0  # position of unused token in joined_input_ids
-                # '- args.max_seq_len ' is used to make sure we don't get a sequence that is too short
-                for begin_loc in range(0, len(joined_input_ids) - args.max_seq_len, step):
+                # '- args.max_seq_len' is used to make sure we don't get a sequence that is too short
+                # '+ 1' since range is exclusive at the end: a block starting exactly at
+                # len(joined_input_ids) - max_seq_len is still a full length block
+                for begin_loc in range(0, len(joined_input_ids) - args.max_seq_len + 1, step):
                     # end_loc is the beginning of the next sequence
                     end_loc = begin_loc + args.max_seq_len
                     # get the input sequence
@@ -358,10 +363,23 @@ def text_gen_pre_process(dataset, tokenizer, all_kwargs):
                         # found a good sample
                         break
                     resamples += 1
-                if not encodings:
-                    # could not find a good sample after resampling
+                if encodings is None or (args.drop_short_sequences and encodings.input_ids.shape[1] < args.max_seq_len):
+                    # could not find a good sample after resampling. `encodings` holds the last rejected
+                    # row once the retries are exhausted, and a non-empty BatchEncoding is truthy, so the
+                    # length has to be re-checked explicitly to honor drop_short_sequences.
                     continue
                 append_text_gen_input_ids(tokenized_inputs, encodings.input_ids[0], encodings.attention_mask[0])
+
+    # every strategy can stop early and silently deliver fewer samples than requested, for reasons that
+    # depend on the strategy. Warn so that the shortfall is visible to the user.
+    num_delivered = len(tokenized_inputs["input_ids"])
+    if args.max_samples is not None and num_delivered < args.max_samples:
+        logger.warning(
+            "Only %d samples were generated but max_samples=%d was requested. %s",
+            num_delivered,
+            args.max_samples,
+            get_sample_shortfall_hint(args, total_examples),
+        )
 
     if not args.use_attention_mask:
         # remove attention_mask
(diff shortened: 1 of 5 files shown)


titaiwangms:
Superseded by #2609 (recreated from a branch on microsoft/Olive directly, since PRs from a fork don't get the `hf_token` secret needed for the Olive CI Hugging Face login step, which caused the Linux/Windows Olive CI checks to fail here). Also includes a fix for the stride-direction issue the automated review caught on this PR.

PR #2616, opened 2026-08-11
Guard RTN MoE quantization by expert layout

titaiwangms:
## Problem

PR #2584 merged RTN MoE quantization (`moe=True`) support without a layout check on fused-expert weights. RTN's `WeightQuantizer` groups unconditionally along a tensor's last dimension, which is only correct when the fused expert weight is stored `(num_experts, out_features, in_features)` -- K last. Architectures such as gpt-oss store the transposed `(num_experts, in, out)` layout instead, and are silently mis-quantized (wrong axis grouped, no error) today on `main` when run through `moe=True`.

## Fix

Adds `olive/passes/pytorch/moe_support.py` with `check_moe_layout_support`, gated into RTN's `_run_for_config` after `prepare_model()` and before `finalize()`. The check trusts `transformers`'s own `is_transposed` attribute (set by the `use_experts_implementation` decorator, not derived from config/checkpoint data) directly:

- Accept only experts modules that report `is_transposed is False`.
- Reject anything where `is_transposed` is missing or not a `bool` (covers older transformers releases, undecorated architectures such as llama4/aria, and unrecognized implementations).
- Reject `is_transposed=True` (e.g. gpt-oss).
- Exempt classic per-expert `nn.ModuleList` experts (e.g. Mixtral/PhiMoE on older transformers) that carry no direct 3D parameter.

A `trust_remote_code` custom experts implementation that misreports its own `is_transposed` is out of scope: that is treated as user-introduced misuse of an explicitly opted-in trust boundary, not a layout Olive can independently verify. No architecture allow-list is used.

Only affects the `moe=True` path -- gated behind `if qcfg.moe:` and only runs when experts modules are actually detected, so non-MoE quantization is unaffected.

## Testing

28/28 relevant tests pass (`test/passes/pytorch/test_moe_support.py`, `test/passes/pytorch/test_rtn.py`); lintrunner clean.

Note: PR #2610 (GPTQ MoE) will stack on top of this branch to reuse `check_moe_layout_support` and avoid duplicating the layout logic.

### DIFF
--- a/docs/source/features/quantization.md
+++ b/docs/source/features/quantization.md
@@ -87,15 +87,14 @@ Mixture-of-Experts (MoE) experts at full precision. Three independent category f
 | --- | --- | --- |
 | `lm_head` | `false` | Also quantize the language-model head. |
 | `embeds` | `false` | Also quantize the input embeddings. |
-| `moe` | `false` | Also quantize MoE expert weights (both 3D fused-parameter experts such as gpt-oss / newer Qwen3-MoE, and `nn.ModuleList`-style experts such as Mixtral). |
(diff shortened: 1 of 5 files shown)


jambayk (dismissed):

titaiwangms:
@jambayk Done! This is ready!

jambayk (approved):

PR #2628, opened 2026-08-17
Support Qwen3.5/3.6-MoE VL checkpoints in PyTorch-side quantization

titaiwangms:
## Summary

Enables Olive's PyTorch-side quantization passes (`Rtn`/`GPTQ`/`KQuant`, via `ModelWrapper`/`iter_quant_targets`) to work on a Qwen3.5/3.6-MoE checkpoint loaded as its **vision-language** class (`Qwen3_5MoeForConditionalGeneration`, `task="image-text-to-text"`), instead of only the text-only decoder class (`Qwen3_5MoeForCausalLM`) that was previously supported.

### Root cause

Two structural gaps, both purely data/mapping issues (no PyTorch quantization logic needed to change):

1. **Composite config resolution**: the VL checkpoint's top-level `config` has no flat `hidden_size`/`num_hidden_layers`/`num_attention_heads`/`num_key_value_heads`/`head_dim` — they live under `config.text_config`. `ModelWrapper.__init__` resolved these to `None` and crashed with a `TypeError`.
2. **Decoder module path**: the VL decoder is nested at `model.language_model.*` (alongside `model.visual`, the vision tower), not `model.*` directly, so `LAYERS`/`EMBEDDINGS`/`PRE_HEAD_LAYERNORM`/`ROTARY_EMBEDDING` couldn't resolve it.

### Fix

- `olive/assets/io_configs/defaults.yaml`: add `text_config.*` fallbacks for `num_layers`, `hidden_size`, `num_attention_heads`, `num_kv_heads` (mirrors the existing `num_experts` nested-fallback pattern), plus a new `head_dim` alias.
- `olive/common/hf/wrapper.py`: add a `"qwen3_5_moe"` entry to `LAYERS`/`EMBEDDINGS`/`PRE_HEAD_LAYERNORM`/`ROTARY_EMBEDDING` pointing at `model.language_model.*`. Scoped to the VL `model_type` only — the text-only checkpoint carries `model_type == "qwen3_5_moe_text"` and its flat config already matches `"default"`, so it is unaffected (see `test_hf_wrapper_text_only_config_unaffected_by_vl_aliases`).

### Two pre-existing (non-VL-specific) quantization gaps, found while validating the above

Both reproduce identically on the text-only checkpoint too, but affect Qwen3.5/3.6-MoE quantization quality generally:

- `MAMBA` mapping didn't recognize this architecture's `linear_attn` (GatedDeltaNet) attribute name, so its projections (`in_proj_qkv`/`in_proj_a`/`in_proj_b`/`in_proj_z`/`out_proj`) were swept into the generic 2D quantization walk instead of staying full precision like other Mamba/SSM blocks.
- Added a `SHARED_EXPERT_GATE` mapping + `LayerWrapper.get_shared_expert_gate()` accessor (also applicable to `qwen2_moe`/`qwen3_next`/`qwen3_omni_moe`, which use the same attribute name) and wired it into `iter_quant_targets`'s exclusion set — the single-row shared-expert sigmoid gate was being quantized like an ordinary `Linear`, despite being a routing-like signal (same reasoning as excluding the main router).

## Verification

- Confirmed against the real `Qwen/Qwen3.6-35B-A3B` checkpoint's config: alias/mapping resolution (`hidden=2048 heads=16 kv=2 head_dim=256 layers=40`, decoder path `model.language_model.layers`) now matches expectations exactly.
- Built a shape-faithful synthetic VL checkpoint (206M params, real vision tower + 4-layer MoE decoder) and ran the actual `Rtn` pass end-to-end (`bits=4, group_size=32, moe=True, modules_to_not_convert=["visual"]`):
  - `mlp.experts.*` (fused 3D MoE weights) — quantized ✅
  - `mlp.gate` (router), `mlp.shared_expert_gate`, `linear_attn.*` — full precision, untouched ✅
  - `model.visual.*` (vision tower) — 0 quant artifacts ✅
  - Reload of the quantized checkpoint via `HfModelHandler` succeeds.
- Added 5 new unit tests covering composite-config resolution, text-only-path non-regression, and `linear_attn`/`shared_expert_gate` exclusion.
- Full relevant suite: `test/common/`, `test/passes/pytorch/test_rtn.py`, `test/model/` — 622+ passed, 0 regressions (pre-existing unrelated `azure` module errors only). `lintrunner` clean (only the repo-wide `CPY001` false positive).

## Files

- `olive/assets/io_configs/defaults.yaml`
- `olive/common/hf/wrapper.py`
- `olive/common/quant/selection.py`
- `test/common/test_hf_wrapper.py`
- `test/common/quant/test_selection.py`

### DIFF
--- a/olive/assets/io_configs/defaults.yaml
+++ b/olive/assets/io_configs/defaults.yaml
@@ -3,12 +3,18 @@
 
 # Attribute aliases (same concept, different naming across models)
 aliases:
-  # Layer count
-  num_layers: [num_hidden_layers, n_layer, n_layers]
-  # Hidden dimensions
-  hidden_size: [dim, d_model, n_embd]
-  num_attention_heads: [num_heads, n_head, n_heads, encoder_attention_heads]
-  num_kv_heads: [num_key_value_heads]
+  # Layer count. ``text_config.num_hidden_layers`` covers composite VL configs (e.g.
+  # Qwen3.5/3.6-MoE-VL, GLM4V-MoE) whose top level only has ``vision_config``/``text_config``
+  # and no flat decoder attributes of its own.
+  num_layers: [num_hidden_layers, n_layer, n_layers, text_config.num_hidden_layers]
+  # Hidden dimensions. See ``num_layers`` above for why the ``text_config.*`` fallbacks exist.
+  hidden_size: [dim, d_model, n_embd, text_config.hidden_size]
+  num_attention_heads: [num_heads, n_head, n_heads, encoder_attention_heads, text_config.num_attention_heads]
+  num_kv_heads: [num_key_value_heads, text_config.num_key_value_heads]
+  # Explicit head_dim, so composite VL configs whose head_dim doesn't equal
+  # hidden_size // num_attention_heads (e.g. Qwen3.5/3.6-MoE-VL) resolve correctly instead of
+  # falling back to that derived (and here incorrect) value.
+  head_dim: [text_config.head_dim]
   # MoE expert count. Covers flat configs (Mixtral / gpt-oss / PhiMoE
   # ``num_local_experts``, DeepSeek ``n_routed_experts``, Ernie4.5 / Aria
   # ``moe_num_experts``) and nested sub-configs (DBRX ``ffn_config``, Llama4 /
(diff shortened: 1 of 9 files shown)


titaiwangms:
Superseded by #2630 — same branch/commit, pushed directly to microsoft/Olive instead of a fork so CI has access to the `hf_token` secret (Azure DevOps withholds secrets from fork PR builds, which is why `Hugging Face Login` failed here).

PR #2634, opened 2026-08-20
Support reproducible text-only MobiusBuilder exports

titaiwangms:
## Describe your changes

Extend `MobiusBuilder` so direct exports can reproducibly select the text backbone of supported multimodal checkpoints:

- forward the input `HfModelHandler` revision to `mobius.build`;
- expose an optional `text_only` pass setting, defaulting to `false`;
- pass `text_only=True` to Mobius only when explicitly requested, preserving compatibility for existing recipes;
- keep Mobius task auto-detection unchanged;
- forward the same revision and `trust_remote_code` policy to ORT GenAI config/tokenizer asset generation.

`text_only` selects the model architecture during Mobius build. It is intentionally distinct from `components_to_export`, which filters components after a package has already been built and therefore cannot produce a complete standalone runtime package by itself.

This enables VLM-capable checkpoints such as `Qwen/Qwen3.6-35B-A3B` to produce a single standalone text decoder with root-level `model.onnx` and decoder-only GenAI configuration while keeping weights and tokenizer assets on one immutable Hugging Face revision.

## Validation

- 35 MobiusBuilder tests passed
- coverage for default omission and explicit forwarding of `text_only`
- coverage that the HfModel task is not forwarded and Mobius continues auto-detection
- coverage for revision/trust forwarding to both build and GenAI config generation
- existing `components_to_export` regression coverage preserved
- targeted lint/format and diff checks passed; only the repository's pre-existing CPY001 header findings reproduce on the unchanged base files

- [x] Make sure all relevant tests can pass.
- [x] Update public pass configuration documentation in the implementation docstring.
- [x] Lint and apply fixes.
- [x] Include a user-facing release note.

Release note: `MobiusBuilder` can now export a supported multimodal checkpoint's standalone text backbone while honoring the configured Hugging Face revision for weights and runtime assets.

### DIFF
--- a/olive/passes/onnx/mobius_model_builder.py
+++ b/olive/passes/onnx/mobius_model_builder.py
@@ -157,8 +157,9 @@ class MobiusBuilder(Pass):
         dtype_str: str = _PRECISION_TO_DTYPE.get(config.precision, "f32")
         model_id: str = model.model_name_or_path
 
-        # Read trust_remote_code from the model's HuggingFace load kwargs.
-        trust_remote_code: bool = model.get_load_kwargs().get("trust_remote_code", False)
+        load_kwargs = model.get_load_kwargs()
+        revision: str | None = load_kwargs.get("revision")
+        trust_remote_code: bool = load_kwargs.get("trust_remote_code", False)
 
         logger.info(
             "MobiusBuilder: building '%s' (ep=%s, dtype=%s)",
@@ -182,6 +183,7 @@ class MobiusBuilder(Pass):
 
         pkg = build(
             model_id,
+            revision=revision,
             dtype=dtype_str,
             execution_provider=ep_str,
             load_weights=True,
--- a/olive/passes/onnx/mobius_model_builder.py
+++ b/olive/passes/onnx/mobius_model_builder.py
@@ -8,7 +8,7 @@ from __future__ import annotations
 
 import logging
 from pathlib import Path
-from typing import TYPE_CHECKING, ClassVar
+from typing import TYPE_CHECKING, Any, ClassVar
 
 from olive.common.utils import StrEnumBase
 from olive.constants import Precision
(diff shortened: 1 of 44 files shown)


xiaoyu-work:
do we still need this?

titaiwangms:
Yes, this is still needed for two reasons:

1. `components_to_export` selects which components are saved **after** Mobius has built a model. It does not change the model architecture. For example, filtering a VLM package to `decoder` leaves a decoder whose input is `inputs_embeds`; it does not produce the standalone text model whose root input is `input_ids`.
2. A direct FP16 export of a VLM-capable checkpoint such as `Qwen/Qwen3.6-35B-A3B` needs a **model-level** selection before the build so Mobius chooses the registered text sibling and generates a complete decoder-only package. Quantized flows can appear to work without it because an upstream quantization pass first materializes a local text model.

The revision forwarding is also still needed: `MobiusBuilder` otherwise drops the revision from the input `HfModelHandler`, and model weights plus GenAI/tokenizer assets may be resolved from different snapshots.

The distinction may be clearer if the Olive-facing setting is named `model_to_export: "text"` rather than exposing Mobius' boolean `text_only`; `components_to_export` would remain the separate package-component filter. I can make that naming change in this PR if preferred.

justinchuby:
LGTM. Would leave to @jambayk @xiaoyu-work for an approval

xiaoyu-work:

xiaoyu-work (approved):

olive/passes/onnx/mobius_model_builder.py
```diff
@@ -157,8 +173,9 @@ def _run_for_config(
         dtype_str: str = _PRECISION_TO_DTYPE.get(config.precision, "f32")
         model_id: str = model.model_name_or_path
 
-        # Read trust_remote_code from the model's HuggingFace load kwargs.
-        trust_remote_code: bool = model.get_load_kwargs().get("trust_remote_code", False)
+        load_kwargs = model.get_load_kwargs()
+        revision: str | None = load_kwargs.get("revision")
```
xiaoyu-work, line 177: what is revision used for?

justinchuby: revision allows the builder to select a specific commit from the hf repo to improve reproducibility.

justinchuby (approved):

PR #2636, opened 2026-08-25
Use onnx-shape-inference's infer_symbolic_shapes

ananyaa-ms:
- Replaced ORT shape inference with  onnx-shape-inference  and added a test for scalar  Constant  attributes.
- ORT crashes on valid  Constant(value_int=...)  nodes, while  onnx-shape-inference  handles them correctly.

### DIFF
--- a/olive/passes/onnx/common.py
+++ b/olive/passes/onnx/common.py
@@ -581,8 +581,8 @@ def model_has_adapters(model_path: Union[str, Path], adapter_type: AdapterType =
 
 def _fix_output_shapes(model_proto: onnx.ModelProto):
     """Run shape inference on the model and update the output shapes to make them fixed."""
-    from onnxruntime.tools.onnx_model_utils import is_fixed_size_tensor
     from onnx_shape_inference import infer_symbolic_shapes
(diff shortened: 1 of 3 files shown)


justinchuby (approved):

olive/passes/onnx/common.py
```diff
@@ -581,11 +581,12 @@ def model_has_adapters(model_path: Union[str, Path], adapter_type: AdapterType =
...
     from onnxruntime.tools.onnx_model_utils import is_fixed_size_tensor
-    from onnxruntime.tools.symbolic_shape_infer import SymbolicShapeInference
 
-    # use the onnxruntime shape inference tool since it can handle large models as well as contrib ops
-    inferred_proto = SymbolicShapeInference.infer_shapes(model_proto, auto_merge=True, guess_output_rank=True)
+    ir_model = ir.serde.deserialize_model(model_proto)
+    infer_symbolic_shapes(ir_model, warn_on_missing=False)
+    inferred_proto = ir.serde.serialize_model(ir_model)
 
     for idx, o in enumerate(model_proto.graph.output):
```
justinchuby, line 591: We will turn the full pass into onnx-ir. But that can be a next step later.

PR #2647, opened 2026-09-03
Fix boolean parsing for capture-onnx-graph options

sylvesterkaczmarek:
## Describe your changes

Fixes boolean argument parsing for `capture-onnx-graph`.

The `--exclude_embeds`, `--exclude_lm_head`, and `--enable_cuda_graph` options used `type=bool`. With argparse, values such as `false` are non-empty strings and therefore evaluate to `True`.

These options now use argparse's standard `store_true` flag semantics: absence leaves the option disabled (or `None` for `enable_cuda_graph`, preserving its existing tri-state default), while presence enables it.

Regression tests cover both default and enabled states.

- [x] Make sure focused validation passes (`ruff` and direct parser smoke test).
- [x] Update documents if necessary. No documentation change required.

### DIFF
--- a/olive/cli/capture_onnx.py
+++ b/olive/cli/capture_onnx.py
@@ -37,6 +37,19 @@ def parse_dim_dict(s):
         raise argparse.ArgumentTypeError("Format must be key=value,... with positive integers as values") from exc
 
 
+def parse_bool(value):
+    if isinstance(value, bool):
+        return value
+
+    normalized = value.lower()
+    if normalized in {"true", "1", "yes", "on"}:
+        return True
+    if normalized in {"false", "0", "no", "off"}:
+        return False
+
+    raise argparse.ArgumentTypeError(f"invalid boolean value: {value!r}")
+
+
 class CaptureOnnxGraphCommand(BaseOliveCLICommand):
     @staticmethod
     def register_subcommand(parser: ArgumentParser):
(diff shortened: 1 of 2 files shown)


olive/cli/capture_onnx.py
```diff
@@ -147,21 +160,21 @@ def register_subcommand(parser: ArgumentParser):
...
+            type=parse_bool,
             default=False,
             required=False,
             help="Remove language modeling head from your ONNX model.",
         )
         mb_group.add_argument(
             "--enable_cuda_graph",
-            type=bool,
+            type=parse_bool,
             default=None,  # Explicitly setting to None to differentiate between user intent and default.
```
shaahji, line 161: Rather than explicit parse, wouldn't use of `action="store_true"` be the correct pattern?

Using `type=bool` is wrong pattern but explicit parsing is also *non-pythonic*. The correct resolution to it is to use `action="store_true"`

shaahji (changes requested):

sylvesterkaczmarek:
Addressed the review: replaced the explicit parser with argparse `store_true` flags for all three options and updated the regression tests to cover absence/presence semantics. `enable_cuda_graph` keeps its existing `None` default so the current tri-state behavior is preserved. Ruff passes and a direct parser smoke test passes locally.

sylvesterkaczmarek:
@shaahji The current head uses `store_true` for all three flags and preserves the `None` default for `enable_cuda_graph`. The requested change and regression updates are in place. Please re-review when available.

olive/cli/capture_onnx.py
sylvesterkaczmarek: The current head uses argparse `store_true` for all three options, retaining the existing `None` default for `enable_cuda_graph`. This implements the requested parsing pattern; the re-review follow-up is posted above.

shaahji (approved):

sylvesterkaczmarek:
Thanks @shaahji for the re-review and approval. Ready from my side.

sylvesterkaczmarek:
The latest Olive CI run is red, but I checked it against current `main` (`b5a1cdcb`). `main` is failing the same Diffusers LoRA tests (`test_sd_lora_train_sd15`, `test_sd_lora_train_sdxl`, `test_sd_lora_train_flux`, and `test_sd_lora_dreambooth_sd15`) with the same PEFT `MagicMock` error. This PR only changes `olive/cli/capture_onnx.py` and its argument tests, so the Azure failure appears unrelated to this change.

PR #2666, opened 2026-09-15
Fix legacy past_key_values crash in the default ONNX export path

AmirF194:
## Describe your changes

xiaoyu-work's #2328 only patched the dynamo=True path. `_export_pytorch_model`'s dynamo=False branch (what `auto-opt` uses by default, without `--use_model_builder`) never called `_patch_model_if_necessary`, so the legacy list-format `past_key_values` Olive's own dummy-input generator still produces reaches the traced `forward()` unpatched.

This issue's own traceback was filed against 0.11.0, before #2328 landed, and shows the crash inside `patched_forward` itself (`DynamicCache.from_legacy_cache` didn't exist yet). On current main #2328 already removed that call for dynamo=False, so it now crashes one level earlier and differently: `AttributeError: 'list' object has no attribute 'get_seq_length'` in `Qwen2Model.forward`, still on transformers >= 5.0, still without `--use_model_builder`. Reproduced live against current main.

Fix: `_patch_model_if_necessary` now builds the `DynamicCache` from its layers instead of the removed `from_legacy_cache`/`to_legacy_cache` (the same conversion `_convert_past_key_values_to_dynamic_cache` already does for the dynamo=True path), and it's called from the dynamo=False branch too. Encoder-decoder (4-tuple) legacy caches on transformers >= 5.0 raise a clear `NotImplementedError` instead of crashing on the same removed method; that combination was already broken before this change and needs its own fix, out of scope here.

#2381 took a different approach, raising a clear error pointing at `--use_model_builder` instead of fixing the default path. This fixes the default path directly so it keeps working the way it did before #2328.

New test reproduces the crash on main and passes here, running `_export_pytorch_model` directly against `tiny-random-LlamaForCausalLM` with transformers 5.17 on Python 3.11. Rest of `test_conversion.py` passes unchanged. Haven't touched encoder-decoder caches or retested the transformers 4.45-4.x branch specifically, that code didn't change, just moved into a helper.

## (Optional) Issue link
Fixes #2335


AmirF194:
CodeQL, Docs, Lint and Test model fast are all waiting on approval to run for this fork PR. Could you approve the workflow runs so CI can execute?

